PR #5, opened 2020-12-16
Add support for link types selector

Mmduh-483:
Currently, the operator has no way to determine a specific link type
NICs for the deployed device plugin e.g. determining a node to be
only ethernet or InfiniBand NICs only
This patch adds support for "linkTypes" selector of SR-IOV network
device plugin, which provides a more flexible selection of the NICs

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -563,6 +563,9 @@ func (r *SriovNetworkNodePolicyReconciler) renderDevicePluginConfigData(pl *srio
 			if len(p.Spec.NicSelector.PfNames) > 0 {
 				netDeviceSelectors.PfNames = sriovnetworkv1.UniqueAppend(netDeviceSelectors.PfNames, p.Spec.NicSelector.PfNames...)
 			}
+			if len(p.Spec.NicSelector.LinkTypes) > 0 {
+				netDeviceSelectors.LinkTypes = sriovnetworkv1.UniqueAppend(netDeviceSelectors.LinkTypes, p.Spec.NicSelector.LinkTypes...)
+			}
 			if len(p.Spec.NicSelector.RootDevices) > 0 {
 				netDeviceSelectors.RootDevices = sriovnetworkv1.UniqueAppend(netDeviceSelectors.RootDevices, p.Spec.NicSelector.RootDevices...)
 			}
@@ -615,6 +618,9 @@ func (r *SriovNetworkNodePolicyReconciler) renderDevicePluginConfigData(pl *srio
 			}
 			if len(p.Spec.NicSelector.PfNames) > 0 {
 				netDeviceSelectors.PfNames = append(netDeviceSelectors.PfNames, p.Spec.NicSelector.PfNames...)
+			}
+			if len(p.Spec.NicSelector.LinkTypes) > 0 {
+				netDeviceSelectors.LinkTypes = append(netDeviceSelectors.LinkTypes, p.Spec.NicSelector.LinkTypes...)
 			}
 			if len(p.Spec.NicSelector.RootDevices) > 0 {
 				netDeviceSelectors.RootDevices = append(netDeviceSelectors.RootDevices, p.Spec.NicSelector.RootDevices...)
--- a/api/v1/sriovnetworknodepolicy_types.go
+++ b/api/v1/sriovnetworknodepolicy_types.go
@@ -52,6 +52,8 @@ type SriovNetworkNicSelector struct {
 	PfNames []string `json:"pfNames,omitempty"`
 	// Infrastructure Networking selection filter. Allowed value "openstack/NetworkID:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
 	NetFilter string `json:"netFilter,omitempty"`
+	// Link type of SR-IOV PF.
+	LinkTypes []string `json:"linkTypes,omitempty"`
 }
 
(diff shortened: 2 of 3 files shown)


controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -616,6 +619,9 @@ func (r *SriovNetworkNodePolicyReconciler) renderDevicePluginConfigData(pl *srio
 			if len(p.Spec.NicSelector.PfNames) > 0 {
 				netDeviceSelectors.PfNames = append(netDeviceSelectors.PfNames, p.Spec.NicSelector.PfNames...)
 			}
+			if l := len(p.Spec.NicSelector.LinkTypes); l > 0 {
```
moshe010, line 622: why you need the l. can you just write 
"if len(p.Spec.NicSelector.LinkTypes) > 0" like the others?

moshe010 (commented):
@pliurh @zshi-redhat I remember that you didn't want to add the link selector at the time. 
But our QA is complaining that there is no good what to create pools on host which has Mellanox IB card and Ethernet card so that can't re-use the setup to test both Ethernet  and IB. 
Can you reconsider adding this selector?

zshi-redhat:
> @pliurh @zshi-redhat I remember that you didn't want to add the link selector at the time.
> But our QA is complaining that there is no good what to create pools on host which has Mellanox IB card and Ethernet card so that can't re-use the setup to test both Ethernet and IB.
> Can you reconsider adding this selector?

@moshe010  is it an option to use rootDevices or pfNames selectors in such case.
I understand the request, the only concern I have is while using more selectors, there is an increasing chance of overlapping devices among resource pools.

pliurh:
> @pliurh @zshi-redhat I remember that you didn't want to add the link selector at the time.
> But our QA is complaining that there is no good what to create pools on host which has Mellanox IB card and Ethernet card so that can't re-use the setup to test both Ethernet and IB.
> Can you reconsider adding this selector?

Normally, we suggest users use `rootDevices` to specify the devices. Wouldn't it sufficient for the case you

pliurh:
`pfNames` shall not be used, if we want to change the device type from eth to ib or vice verse. As it may also be changed. But I think using the `rootDevices` should be sufficient.

moshe010 (commented):
So the overlapping can exist today. I don't see link type select going to be highly used but there are cases as state which it can be really helpful. (QA testing servers). 
using rootDevices  is problematic as it assume same pci address to all servers which is not our case a spicily that QA change the link type of cards.  It also more error prons. 

regarding pfNames  selectors I agree it problematic so maybe we can add validation which you can't use pfName selector when linkType is used.

moshe010:
As we talked in the meeting we will abandon this approach until we will have customers that we use it. 
instead of this PR we will create another approach to allow sriov-network-operator not  auto generate device plugin configmap and this will allow to advanced user to provide it own. sriov-device-plugin support link type selector so it work fine for our QA.

zshi-redhat:
Looking at this again, got a thought that we might be able to use the linkType from nodePolicy API to generate the device plugin configMap, this would avoid adding linkTypes selector but still be able to use it in device plugin config.

pliurh:
> Looking at this again, got a thought that we might be able to use the linkType from nodePolicy API to generate the device plugin configMap, this would avoid adding linkTypes selector but still be able to use it in device plugin config.

The flay`linkType` is to tell the config daemon to change the link type of the selected NICs. If we use this as a selector, then the config daemon cannot change a NIC's device type from eth to ib, as no eth NIC can be selected.

zshi-redhat:
> > Looking at this again, got a thought that we might be able to use the linkType from nodePolicy API to generate the device plugin configMap, this would avoid adding linkTypes selector but still be able to use it in device plugin config.
> 
> The flay`linkType` is to tell the config daemon to change the link type of the selected NICs. If we use this as a selector, then the config daemon cannot change a NIC's device type from eth to ib, as no eth NIC can be selected.

I mean use policy.LinkType to generate the device plugin selector, not the policy selector. Policy API remain unchanged, the way it selects which device to be configured and how it generates the sriov node state spec is not changed.

adrianchiris:
> I mean use policy.LinkType to generate the device plugin selector, not the policy selector. 

@zshi-redhat just to make sure i understand what you are proposing,

in this case `linkType` selector will always be present in sriov network device plugin config map ?

zshi-redhat:
> > I mean use policy.LinkType to generate the device plugin selector, not the policy selector.
> 
> @zshi-redhat just to make sure i understand what you are proposing,
> 
> in this case `linkType` selector will always be present in sriov network device plugin config map ?

It could be, or we only add linkType selector when it is explicitly configured to `IB`.

moshe010:
> > > I mean use policy.LinkType to generate the device plugin selector, not the policy selector.
> > 
> > 
> > @zshi-redhat just to make sure i understand what you are proposing,
> > in this case `linkType` selector will always be present in sriov network device plugin config map ?
> 
> It could be, or we only add linkType selector when it is explicitly configured to `IB`.

So just to add when it Ib seem good start

moshe010:
> So just to add when it Ib seem good start
Thinking about again we need to create the selector both for ib and ethernet networks. We can skip it in the case of virtual machines.

zshi-redhat:
> > So just to add when it Ib seem good start
> 
> Thinking about again we need to create the selector both for ib and ethernet networks. We can skip it in the case of virtual machines.

Since we only have two link types, it will be either IB or ETH. All devices except IB can be selected via linkType:eth.
Why skipped for virtual machines?  It might not be eth type?

moshe010:
Because we are not use in all cases you will have netdevice. You need netdevice to know the link type

PR #7, opened 2020-12-16
Support MetaPlugins for SriovIBNetwork

Mmduh-483:


moshe010 (commented):
@zshi-redhat @pliurh is there anything that block this PR? 
It LGTM to me

zshi-redhat:
> @zshi-redhat @pliurh is there anything that block this PR?
> It LGTM to me

The change looks good to me except that we need to update /manifests for this API change.
However, I'm not sure if we want to maintain the /manifests folder as a upstream project, as it's needed for openshift deployment only atm.

pliurh:
@zshi-redhat I prefer we maintain the manifests folder downstream only.

PR #8, opened 2020-12-17
Add deploy-setup-k8s-dev-mode Makefile target

oshoval:
`deploy-setup-k8s` disables webhooks, disregarding the
value of `ENABLE_ADMISSION_CONTROLLER`.

This PR adds `deploy-setup-k8s-dev-mode` make target which does
exactly the same as `deploy-setup-k8s` but respects previous
`ENABLE_ADMISSION_CONTROLLER` value.

This will allow to run k8s deploy with webhooks, without running the inner
make file targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,11 +188,13 @@ bundle-build:
 deploy-setup: skopeo install
 	hack/deploy-setup.sh $(NAMESPACE)
 
-deploy-setup-k8s: export NAMESPACE=sriov-network-operator
+deploy-setup-k8s-dev-mode: export NAMESPACE=sriov-network-operator
+deploy-setup-k8s-dev-mode: export CNI_BIN_PATH=/opt/cni/bin
+deploy-setup-k8s-dev-mode: export OPERATOR_EXEC=kubectl
+deploy-setup-k8s-dev-mode: deploy-setup
+
 deploy-setup-k8s: export ENABLE_ADMISSION_CONTROLLER=false
-deploy-setup-k8s: export CNI_BIN_PATH=/opt/cni/bin
-deploy-setup-k8s: export OPERATOR_EXEC=kubectl
-deploy-setup-k8s: deploy-setup
+deploy-setup-k8s: deploy-setup-k8s-dev-mode
 
 test-e2e-conformance:
 	./hack/run-e2e-conformance.sh


zshi-redhat:
> /cc @zshi-redhat

@oshoval does current operator webhook work in vanilla k8s environment? how the certificate is handled?

oshoval:
> > /cc @zshi-redhat
> 
> @oshoval does current operator webhook work in vanilla k8s environment? how the certificate is handled?

The default make target `deploy-setup-k8s` disables the webhook, so no need to use certificates,
we do want to use webhooks and when we use them we self sign the cert, and inject it manually (there is other issue on the reconcile of it).
This PR only introduce a new make target which will not disable the webhook in case the user set them to be enabled,
as the current make target  `deploy-setup-k8s` disregard the exported value of `ENABLE_ADMISSION_CONTROLLER` and sets it to false hard-coded.

zshi-redhat:
> > > /cc @zshi-redhat
> > 
> > 
> > @oshoval does current operator webhook work in vanilla k8s environment? how the certificate is handled?
> 
> The default make target `deploy-setup-k8s` disables the webhook, so no need to use certificates,
> we do want to use webhooks and when we use them we self sign the cert, and inject it manually (there is other issue on the reconcile of it #3).

ok, so once #3 is resolved, we would remove the hardcoded `ENABLE_ADMISSION_CONTROLLER=false` no longer need this dev-mode, correct?

oshoval:
> > > > /cc @zshi-redhat
> > > 
> > > 
> > > @oshoval does current operator webhook work in vanilla k8s environment? how the certificate is handled?
> > 
> > 
> > The default make target `deploy-setup-k8s` disables the webhook, so no need to use certificates,
> > we do want to use webhooks and when we use them we self sign the cert, and inject it manually (there is other issue on the reconcile of it #3).
> 
> ok, so once #3 is resolved, we would remove the hardcoded `ENABLE_ADMISSION_CONTROLLER=false` no longer need this dev-mode, correct?

I don't think so, the reason of `ENABLE_ADMISSION_CONTROLLER=false` is the unsupported nics.

zshi-redhat:
> > > > > /cc @zshi-redhat
> > > > 
> > > > 
> > > > @oshoval does current operator webhook work in vanilla k8s environment? how the certificate is handled?
> > > 
> > > 
> > > The default make target `deploy-setup-k8s` disables the webhook, so no need to use certificates,
> > > we do want to use webhooks and when we use them we self sign the cert, and inject it manually (there is other issue on the reconcile of it #3).
> > 
> > 
> > ok, so once #3 is resolved, we would remove the hardcoded `ENABLE_ADMISSION_CONTROLLER=false` no longer need this dev-mode, correct?
> 
> I don't think so, the reason of `ENABLE_ADMISSION_CONTROLLER=false` is the unsupported nics webhook validation.
> once #3 will be solved the new target will be able to use webhooks and inject self signed certs,
> while the previous target will stay with webhooks disabled i think.

I was thinking once #3 is resolved, then we can enable the admission controler webhook by default for vanilla k8s deployment. 
The admission controller webhook can validate both unsupported NICs ID and other validation/mutating functions that's already supported by existing webhooks.

User who wants to disable it can export the ENABLE_ADMISSION_CONTROLLER env var to false manually or in their automatic script, no need to maintain two different entries in Makefile for changing admission controller mode. wdyt?

oshoval:
> I was thinking once #3 is resolved, then we can enable the admission controler webhook by default for vanilla k8s deployment.
> The admission controller webhook can validate both unsupported NICs ID and other validation/mutating functions that's already supported by existing webhooks.
> 
> User who wants to disable it can export the ENABLE_ADMISSION_CONTROLLER env var to false manually or in their automatic script, no need to maintain two different entries in Makefile for changing admission controller mode. wdyt?

Yep, Its a possibility, but https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/10 should be fixed as well (i am on it),
and documentation should reflect `ENABLE_ADMISSION_CONTROLLER` and configmap of unsupported nics,
I will do that in case needed once we are done with all the changes we described.

Meanwhile should / can we merge this PR?

zshi-redhat:
> > I was thinking once #3 is resolved, then we can enable the admission controler webhook by default for vanilla k8s deployment.
> > The admission controller webhook can validate both unsupported NICs ID and other validation/mutating functions that's already supported by existing webhooks.
> > User who wants to disable it can export the ENABLE_ADMISSION_CONTROLLER env var to false manually or in their automatic script, no need to maintain two different entries in Makefile for changing admission controller mode. wdyt?
> 
> Yep, Its a possibility, but #10 should be fixed as well (i am on it),
> and documentation should emphasize imo `ENABLE_ADMISSION_CONTROLLER=false`in case using unsupported nics
> and how to use configmap in order to support them, as alternative.
> I will do that in case needed once we are done with all the changes we described.
> 
> Meanwhile should / can we merge this PR?
> The reason is that it allows to use k8s with webhooks with just one make target command, in respect to operator deployment.

@oshoval if this is just about having a way for running admission controllers on vanilla k8s, I think you probably can first deploy with `make deploy-setup-k8s` on vanilla k8s, then patch the default SriovOperatorConfig custom resource to enable both `EnableInjector` and `EnableOperatorWebhook`. Is it acceptable for your testing?

oshoval:
> > > I was thinking once #3 is resolved, then we can enable the admission controler webhook by default for vanilla k8s deployment.
> > > The admission controller webhook can validate both unsupported NICs ID and other validation/mutating functions that's already supported by existing webhooks.
> > > User who wants to disable it can export the ENABLE_ADMISSION_CONTROLLER env var to false manually or in their automatic script, no need to maintain two different entries in Makefile for changing admission controller mode. wdyt?
> > 
> > 
> > Yep, Its a possibility, but #10 should be fixed as well (i am on it),
> > and documentation should emphasize imo `ENABLE_ADMISSION_CONTROLLER=false`in case using unsupported nics
> > and how to use configmap in order to support them, as alternative.
> > I will do that in case needed once we are done with all the changes we described.
> > Meanwhile should / can we merge this PR?
> > The reason is that it allows to use k8s with webhooks with just one make target command, in respect to operator deployment.
> 
> @oshoval if this is just about having a way for running admission controllers on vanilla k8s, I think you probably can first deploy with `make deploy-setup-k8s` on vanilla k8s, then patch the default SriovOperatorConfig custom resource to enable both `EnableInjector` and `EnableOperatorWebhook`. Is it acceptable for your testing?


Yes, i am doing it little different,
https://github.com/kubevirt/kubevirtci/pull/500/files#diff-499cff045d8a875e8b39f4a9fd59354eb99e8f8d91a95ebc70fbf42d2d5b5beaR190
just thought its better in one liner, until we fix all the issues.

Closing thanks.

PR #15, opened 2020-12-20
Update DiscoverSriovDevices to list only pfs with networking devices

oshoval:
Currently `DiscoverSriovDevices` creates a list of interfaces,
by scanning `/sys/devices/pci`.
Some of them might not have a networking devices.
The interfaces without networking devices won't be available,
but would still be reset, as they won't be part of the node policy.
    
Updating `DiscoverSriovDevices` to list only pfs which have networking devices.
As a result sriov-operator will support namespace isolation,
because `/sys/class/net` entries are network-namespaced.

Fixes: https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/2

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -78,6 +78,18 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
 			glog.Warningf("DiscoverSriovDevices(): unable to parse device driver for device %+v %q", device, err)
 			continue
 		}
+
+		deviceNames, err := dputils.GetNetNames(device.Address)
+		if err != nil {
+			glog.Warningf("DiscoverSriovDevices(): unable to get device names for device %+v %q", device, err)
+			continue
+		}
+
+		if len(deviceNames) == 0 {
+			// no network devices found, skipping device
+			continue
+		}
+
 		iface := sriovnetworkv1.InterfaceExt{
 			PciAddress: device.Address,
 			Driver:     driver,


pkg/utils/utils.go
```diff
@@ -112,7 +124,16 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
 				}
 			}
 		}
-		pfList = append(pfList, iface)
+
+		for _, devName := range deviceNames {
```
oshoval, line 128: In case needed, this can be done in O(1) by checking if  `/sys/class/net/<devName>/device/uevent` exists
wasn't sure what is better, high level approach as the PR have, or optimized implementation.

pkg/utils/utils.go
```diff
@@ -112,7 +124,16 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
...
+			for _, nic := range net.NICs {
```
zshi-redhat, line 129: @SchSeba If we only loop against network devices here, will N3000 network function still be discovered?
I guess this is the same as asking if N3000 network function will be discovered as a separate PCI address.

zshi-redhat:
With this change, devices that're in non-default host namespace will not be discovered by config daemon, which in turn will not be published in node state, I don't see this is affecting current features in operator, but not sure if there are other cases that would need config daemon to have visibility on all host devices. @pliurh @adrianchiris wdyt?

pliurh:
I'm ok with this change.

pkg/utils/utils.go
```diff
@@ -112,7 +124,16 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
...
+				if devName == nic.Name {
+					glog.V(2).Infof("DiscoverSriovDevices() adding device %s, pci address %s\n", nic.Name, device.Address)
+					pfList = append(pfList, iface)
+					break
+				}
+			}
+		}
```
adrianchiris, line 136: Im trying to wrap my head around the changes in this PR.

looking at https://github.com/jaypipes/ghw/blob/14923eaaa4d8378051f741694e261747d16fc05e/net_linux.go#L43

the `NICs` attribute is essentially a list of all netdevices in the host built from `/sys/class/net`

`deviceNames` is yet another list of netdevices found for a specific pci device by scanning `/sys/bus/pci/devices/<pci addr>/net`

now, in my system, moving a netdev to a different namespace will remove it from both `/sys/class/net` and `/sys/bus/pci/devices/<pciAddr>/net`  (as they are both the same file under `/sys/devices/....`)

So it sifts down to the fact that the latter is a subset of the former, am i missing something ?

essentially the one case where you would not want `pfList` to contain a certain `iface` (device) is if it has no netdevs associated with it. so you could just check `len(deviceNames) == 0` at L92

adrianchiris:
>With this change, devices that're in non-default host namespace will not be discovered by config daemon, which in turn will not be published in node state, I don't see this is affecting current features in operator, but not sure if there are other cases that would need config daemon to have visibility on all host devices. @pliurh @adrianchiris wdyt?

This change mandates a netdevice to exist for a PCI device in DiscoverDevices. i see explicit logic that handles the fact that a PCI device might **_not_** have a network device, so i wonder if originally there was a reason for it.

pliurh:
> This change mandates a netdevice to exist for a PCI device in DiscoverDevices. i see explicit logic that handles the fact that a PCI device might not have a network device, so i wonder if originally there was a reason for it.

The original thought is that we might also want to support attaching PF to pod with vfio-pci driver. But as we decided to not use this operator for such a use case, it is fine to only discover network devices.

pkg/utils/utils.go
oshoval: You totally right, testing and updating according what you suggested.

adrianchiris:
> The original thought is that we might also want to support attaching PF to pod with vfio-pci driver. But as we decided to not use this operator for such a use case, it is fine to only discover network devices

Ack Thanks for the clarification @pliurh !

pliurh:
/hold
Suddenly thought of this change may affect Openstack support? Where the VF is a host device in VM and could be loaded with vfio-pci driver.

oshoval:
> /hold
> Suddenly thought of this change may affect Openstack support? Where the VF is a host device in VM and could be loaded with vfio-pci driver.

Would DiscoverSriovDevicesVirtual triggered instead ? 
if so we are safe, because this PR change only the non virtual discover
Ref: https://github.com/openshift/sriov-network-operator/pull/430#discussion_r534079710

pliurh:
/hold cancel
You're right

oshoval:
> /hold cancel
> You're right

Thanks, btw the hold label wasn't removed

pkg/utils/utils.go
oshoval: Done

oshoval:
Addressed comments and rebased.

martinkennelly:
Asking people within Intel to see if this blocks any of our use cases.

martinkennelly:
I do not see this affecting us. Thank you for your patience.

oshoval:
Thanks for verifying
@SchSeba would you please take a look on https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/15#discussion_r547868500?

pkg/utils/utils.go
SchSeba: Sorry for the late response

I check it and it will work also on the N3000 card

```
88:00.0 Ethernet controller: Intel Corporation Ethernet Controller XXV710 Intel(R) FPGA Programmable Acceleration Card N3000 for Networking (rev 02)
88:00.1 Ethernet controller: Intel Corporation Ethernet Controller XXV710 Intel(R) FPGA Programmable Acceleration Card N3000 for Networking (rev 02)
89:00.0 Processing accelerators: Intel Corporation Device 0b30
8a:00.0 Ethernet controller: Intel Corporation Ethernet Controller XXV710 Intel(R) FPGA Programmable Acceleration Card N3000 for Networking (rev 02)
8a:00.1 Ethernet controller: Intel Corporation Ethernet Controller XXV710 Intel(R) FPGA Programmable Acceleration Card N3000 for Networking (rev 02)

[root@cnfdb5 core]# ls /sys/bus/pci/devices/0000\:88\:00.0/net -la
total 0
drwxr-xr-x. 3 root root 0 Jan 17 15:53 .
drwxr-xr-x. 6 root root 0 Jan 17 15:53 ..
drwxr-xr-x. 5 root root 0 Jan 17 15:53 ens5f0
[root@cnfdb5 core]# ls /sys/bus/pci/devices/0000\:88\:00.1/net -la
total 0
drwxr-xr-x. 3 root root 0 Jan 17 15:53 .
drwxr-xr-x. 6 root root 0 Jan 17 15:53 ..
drwxr-xr-x. 5 root root 0 Jan 17 15:53 eth5
[root@cnfdb5 core]# ls /sys/bus/pci/devices/0000\:8a\:00.0/net -la
total 0
drwxr-xr-x. 3 root root 0 Jan 17 15:53 .
drwxr-xr-x. 6 root root 0 Jan 17 15:53 ..
drwxr-xr-x. 5 root root 0 Jan 17 15:53 eth6
[root@cnfdb5 core]# ls /sys/bus/pci/devices/0000\:8a\:00.1/net -la
total 0
drwxr-xr-x. 3 root root 0 Jan 17 15:53 .
drwxr-xr-x. 6 root root 0 Jan 17 15:53 ..
drwxr-xr-x. 5 root root 0 Jan 17 15:53 ens5f1
```

oshoval:
Are we safe to proceed please?

adrianchiris (approved):

zshi-redhat:
> Are we safe to proceed please? (no open comments atm IIUC)

@oshoval I think so, we will merge this PR next week (Feb 8, to align with release schedule).

zshi-redhat (approved):

Issue #10, opened 2020-12-17
Issues using unsupported NIC for Intel X710/X557-AT 10GBASE-T


oshoval:
Following https://github.com/openshift/sriov-network-operator/pull/416
Tried to use the following configMap for `Intel Corporation Ethernet Controller X710/X557-AT 10GBASE-T`
```
apiVersion: v1
data:
  X557: 8086 1589 154c
kind: ConfigMap
metadata:
  name: unsupported-nic-ids
  namespace: sriov-network-operator
```
The configMap doesn't get annotated, hence wont be accepted to be used as whitelist.

The 2 problems that caused it:

1. Using this configMap on a node where the `/host/etc/udev/rules.d/` didn't exists failed (using Kind nodes)
because WriteFile doesn't create the directory hierarchy
https://github.com/golang/go/issues/32916
```
I1216 13:45:10.325575    6814 daemon.go:299] Run(): period refresh
I1216 13:45:10.327794    6814 daemon.go:813] tryCreateUdevRule()
I1216 13:45:10.327869    6814 daemon.go:824] Old udev content '' and new content 'ACTION=="add|change|move", ATTRS{device}=="0x1014|0x1016|0x1018|0x101c|0x154c", ENV{NM_UNMANAGED}="1"' differ. Writing to file /host/etc/udev/rules.d/10-nm-unmanaged.rules.
E1216 13:45:10.327884    6814 daemon.go:833] tryCreateUdevRule(): fail to write file: open /host/etc/udev/rules.d/10-nm-unmanaged.rules: no such file or directory
I1216 13:45:10.327895    6814 daemon.go:301] Could not create udev rule: open /host/etc/udev/rules.d/10-nm-unmanaged.rules: no such file or directory
I1216 13:45:10.784003    6814 writer.go:70] Run(): period refresh
I1216 13:45:10.784035    6814 writer.go:80] pollNicStatus()
I1216 13:45:10.784042    6814 utils.go:46] DiscoverSriovDevices
```
Possible solution is to create the directory in case it doesn't exists.

Done the above solution and then saw the following problem with @akaris-redhat:
2. Since the X557 has VF Device ID 154c, which also exists in the `NicIdMap` https://github.com/openshift/sriov-network-operator/blob/c58d32a8079f5d6f5aa0fd73b83678263f1dbca5/api/v1/helper.go#L39
[...]

PR #23, opened 2020-12-23
daemon, Fix unsupported-nics logic in case VF device ID already exists

oshoval:
In case unsupported nic has a VF device id which already exists
in the udev rule, the rule won't be updated, as it depends only on the
VF device ids, and not on the tuple (Vendor, PF_ID, VF_ID).

Therefore:
* Fix logic to update the annotation even if the file contents are the same,
but the annotation is different or doesn't exist at all.
    
Note: since the annotation has the configMap data itself, both on the creation
and validation, it will be robust to configMap updates,
because if the annotation exists but the data in the annotation is different
from the configMap data, it will be rejected, until a new rule will be created,
if needed, and the updated annotation will be written.

Additional problems that were addressed:

* tryCreateUdevRule creates a udev file by using WriteFile
WriteFile will not create the directory hierarchy in case
it doesn't exist [1].
Therefore create the folder in case it doesn't exist.

* In case the daemon updates the unsupported-nic configmap
annotations, before k8s adds `last-applied-configuration` annotation,
a panic will occur at `annotateUnsupportedNicIdConfigMap`:
"Observed a panic: assignment to entry in nil map (assignment to entry in nil map)"
Fix it by allocation the map in case it is still nil.

[1] https://github.com/golang/go/issues/32916

Fixes: https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/10

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -815,7 +815,8 @@ func tryEnableRdma() (bool, error) {
 
 func tryCreateUdevRule(unsupportedNicIdMap map[string]string) (bool, error) {
 	glog.V(2).Infof("tryCreateUdevRule()")
-	filePath := "/host/etc/udev/rules.d/10-nm-unmanaged.rules"
+	dirPath := "/host/etc/udev/rules.d/"
+	filePath := dirPath + "10-nm-unmanaged.rules"
 
 	new_content := fmt.Sprintf("ACTION==\"add|change|move\", ATTRS{device}==\"%s\", ENV{NM_UNMANAGED}=\"1\"\n", strings.Join(sriovnetworkv1.GetMergedVfIds(unsupportedNicIdMap), "|"))
 
@@ -829,6 +830,12 @@ func tryCreateUdevRule(unsupportedNicIdMap map[string]string) (bool, error) {
 		strings.TrimSuffix(string(old_content), "\n"),
 		strings.TrimSuffix(new_content, "\n"),
 		filePath)
+
+	err = os.MkdirAll(dirPath, os.ModePerm)
+	if err != nil && !os.IsExist(err) {
+		glog.Errorf("tryCreateUdevRule(): failed to create dir %s: %v", dirPath, err)
+		return err
+	}
 
 	// if the file does not exist or if old_content != new_content
 	// write to file and create it if it doesn't exist
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -184,11 +184,18 @@ func (dn *Daemon) annotateUnsupportedNicIdConfigMap(cm *v1.ConfigMap, nodeName s
 	if err != nil {
 		return nil, err
 	}
+	cmData := string(jsonData)
+
+	annotationKey := "openshift.io/" + nodeName
+	annotationData, ok := cm.ObjectMeta.Annotations[annotationKey]
+	if ok && cmData == annotationData {
+		return cm, nil
+	}
 
 	if cm.ObjectMeta.Annotations == nil {
 		cm.ObjectMeta.Annotations = make(map[string]string)
 	}
-	cm.ObjectMeta.Annotations["openshift.io/"+nodeName] = string(jsonData)
+	cm.ObjectMeta.Annotations[annotationKey] = cmData
 	return dn.kubeClient.CoreV1().ConfigMaps(namespace).Update(context.Background(), cm, metav1.UpdateOptions{})
 }
 
@@ -204,17 +211,19 @@ func (dn *Daemon) tryCreateUdevRuleWrapper() error {
 	}
 
 	// update udev rule and if update succeeds
-	ruleWritten, err := tryCreateUdevRule(unsupportedNicIdMap)
+	err := tryCreateUdevRule(unsupportedNicIdMap)
 	if err != nil {
 		return err
 	}
 
-	// annotage ConfigMap if it exists
-	if configMapErr == nil && ruleWritten {
-		if _, err := dn.annotateUnsupportedNicIdConfigMap(cm, dn.name); err != nil {
-			glog.V(2).Info("Could not update Nic Id ConfigMap", "err", err)
-		}
+	if configMapErr != nil {
+		return nil
 	}
+
+	if _, err := dn.annotateUnsupportedNicIdConfigMap(cm, dn.name); err != nil {
+		glog.V(2).Info("Could not update Nic Id ConfigMap", "err", err)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -180,6 +180,15 @@ func (dn *Daemon) getUnsupportedNicIdConfigMap() (*v1.ConfigMap, error) {
 }
 
 func (dn *Daemon) annotateUnsupportedNicIdConfigMap(cm *v1.ConfigMap, nodeName string) (*v1.ConfigMap, error) {
+	annotationKey := "openshift.io/" + nodeName
```
andreaskaris, line 183: Can you reuse "annotationKey" in line 200:
    cm.ObjectMeta.Annotations["openshift.io/"+nodeName] = string(jsonData)

andreaskaris: So, change line 200 to:
    cm.ObjectMeta.Annotations["openshift.io/"+nodeName] = string(jsonData)

oshoval: Done thanks.
Did more little refactors of annotateUnsupportedNicIdConfigMap

oshoval:
```
[Fail] SriovNetwork Controller with SriovNetwork When a derived net-att-def CR is removed [It] should regenerate the net-att-def CR 
/home/runner/work/sriov-network-operator/sriov-network-operator/controllers/sriovnetwork_controller_test.go:204
```
I don't think its due to this PR.
```
2020-12-23T14:15:58.481Z	ERROR	controllers.SriovNetwork	Couldn't update NetworkAttachmentDefinition CR	{"sriovnetwork": "openshift-sriov-network-operator/test-5", "Namespace": "default", "Name": "test-5", "error": "Operation cannot be fulfilled on network-attachment-definitions.k8s.cni.cncf.io \"test-5\": the object has been modified; please apply your changes to the latest version and try again"}
github.com/go-logr/zapr.(*zapLogger).Error
	/home/runner/work/sriov-network-operator/sriov-network-operator/vendor/github.com/go-logr/zapr/zapr.go:132
github.com/k8snetworkplumbingwg/sriov-network-operator/controllers.(*SriovNetworkReconciler).Reconcile
	/home/runner/work/sriov-network-operator/sriov-network-operator/controllers/sriovnetwork_controller.go:149
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
```
RetryOnConflict would fix it i guess.
Should i retest ?

pliurh:
Retest could work. Additionally, I address this issue by https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/24

oshoval:
Thanks, all tested passed now,
and addressed the comments that were so far.

zshi-redhat:
> In case unsupported nic has a VF device id which already exists
> in the udev rule, the rule won't be updated, as it depends only on the
> VF device ids, and not on the tuple (Vendor, PF_ID, VF_ID).
> 
> Therefore:
> 
>     * Fix logic to update the annotation even if the file contents are the same,
>       but the annotation is different or doesn't exist at all.
> 
> 
> Note: since the annotation has the configMap data itself, both on the creation
> and validation, it will be robust to configMap updates,
> because if the annotation exists but the data in the annotation is different
> from the configMap data, it will be rejected, 

With current PR, if annotation exists but different from the configMap data, it will be updated to configMap data, isn't it?
why it will be rejected?

> until a new rule will be created,
> if needed, and the updated annotation will be written.
> 
> Additional problems that were addressed:
> 
>     * tryCreateUdevRule creates a udev file by using WriteFile
>       WriteFile will not create the directory hierarchy in case
>       it doesn't exist [1].
>       Therefore create the folder in case it doesn't exist.

I guess this issue is not on openshift?

> 
>     * In case the daemon updates the unsupported-nic configmap
>       annotations, before k8s adds `last-applied-configuration` annotation,
>       a panic will occur at `annotateUnsupportedNicIdConfigMap`:
>       "Observed a panic: assignment to entry in nil map (assignment to entry in nil map)"
>       Fix it by allocation the map in case it is still nil.
> 

good catch!

oshoval:
> > Note: since the annotation has the configMap data itself, both on the creation
> > and validation, it will be robust to configMap updates,
> > because if the annotation exists but the data in the annotation is different
> > from the configMap data, it will be rejected,
> 
> With current PR, if annotation exists but different from the configMap data, it will be updated to configMap data, isn't it?

Yes, and until it will be updated it will be rejected, therefore we are safe from races,
because the race can happen if the configMap is updated, and used before we gave the change to update the udev rule.
but since we annotate only after successful udev rule update we are safe.

> why it will be rejected?

Because a preliminary condition in validate.go is that the annotation data equals to the configmap data:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/webhook/validate.go#L318

> 
> > until a new rule will be created,
> > if needed, and the updated annotation will be written.
> > Additional problems that were addressed:
> > ```
> > * tryCreateUdevRule creates a udev file by using WriteFile
> >   WriteFile will not create the directory hierarchy in case
> >   it doesn't exist [1].
> >   Therefore create the folder in case it doesn't exist.
> > ```
> 
> I guess this issue is not on openshift?

I believe you are right, it depends if the folder exists or not there,
we can ask @andreaskaris if he tried that on openshift.
I tried it with nodes that didn't have this folder. 
Anyhow the creation doesn't harm in case the folder exists.
Issue was just that WriteFile is not creating the pull path.

zshi-redhat:
> > > Note: since the annotation has the configMap data itself, both on the creation
> > > and validation, it will be robust to configMap updates,
> > > because if the annotation exists but the data in the annotation is different
> > > from the configMap data, it will be rejected,
> > 
> > 
> > With current PR, if annotation exists but different from the configMap data, it will be updated to configMap data, isn't it?
> 
> Yes, and until it will be updated it will be rejected, therefore we are safe from races,
> because the race can happen if the configMap is updated, and used before we gave the change to update the udev rule.
> but since we annotate only after successful udev rule update we are safe.
> 
> > why it will be rejected?
> 
> Because a preliminary condition in validate.go is that the annotation data equals to the configmap data:
> https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/webhook/validate.go#L318

Ok. I didn't realize it was about webhook rejection, all good here.
> 
> > > until a new rule will be created,
> > > if needed, and the updated annotation will be written.
> > > Additional problems that were addressed:
> > > ```
> > > * tryCreateUdevRule creates a udev file by using WriteFile
> > >   WriteFile will not create the directory hierarchy in case
> > >   it doesn't exist [1].
> > >   Therefore create the folder in case it doesn't exist.
> > > ```
> > 
> > 
> > I guess this issue is not on openshift?
> 
> I believe you are right, it depends if the folder exists or not on the nodes.
> We can ask @andreaskaris if he tried that on openshift.
> I tried it with nodes that didn't have this folder.
> Anyhow the creation doesn't harm in case the folder exists.
> Issue was just that WriteFile is not creating the pull path.

The change is good, I just to make sure we don't have this issue in openshift.

zshi-redhat:
/lgtm

@pliurh do you want to take a second look?

oshoval:
Thanks @zshi-redhat

PR #26, opened 2020-12-24
Allow to use DOWN interface for tests, skip tests if no NIC available

kononovn:
Changed function findUnusedSriovDevices. If it returns error then skip tests instead raise Error.
Allow to use DOWN interfaces for tests.


SchSeba:
@kononovn please fix the tittle and commit message

PR #27, opened 2020-12-24
Fix sriov test

SchSeba:
This commit fix one sriov test by enable the linkState to be enable,
this will allow the pod to have a nic state up even if the PF is down.

This commit also extend the time we wait for the webhook pod to be remove


pliurh:
@SchSeba, Unlike openshift repo, prow commands don't work here. You may have to trigger the GitHub action rerun manually.

zshi-redhat:
@SchSeba could you rebase this PR against master, I think there were fixes to unit tests.

SchSeba:
@zshi-redhat done

PR #29, opened 2020-12-25
Rename vf representor with udev rules

zshi-redhat:
VF representor is named in the format of ethX, where X is an incremental
number, it doesn't tell which PF it belongs to and which VF is the peer device.
By renaming it with udev rules, VF rep name will be in the format of PF_[vf-index].
This commit add udev rules file and relevant script for machine config plugin,
which work only with openshift clusterType.

### DIFF
--- a/bindata/manifests/machine-config/files/switchdev-vf-link-name.sh.yaml
+++ b/bindata/manifests/machine-config/files/switchdev-vf-link-name.sh.yaml
@@ -1,0 +1,9 @@
+mode: 0755
+overwrite: true
+path: "/etc/udev/switchdev-vf-link-name.sh"
+contents:
+  inline: |
+    #!/bin/bash
+    set -x
+    PORT="$1"
+    echo "NUMBER=${PORT##pf*vf}"
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -105,10 +105,11 @@ type workItem struct {
 }
 
 const (
-	scriptsPath  = "/bindata/scripts/enable-rdma.sh"
-	annoKey      = "sriovnetwork.openshift.io/state"
-	annoIdle     = "Idle"
-	annoDraining = "Draining"
+	rdmaScriptsPath = "/bindata/scripts/enable-rdma.sh"
+	udevScriptsPath = "/bindata/scripts/load-udev.sh"
+	annoKey         = "sriovnetwork.openshift.io/state"
+	annoIdle        = "Idle"
+	annoDraining    = "Draining"
 )
 
 var namespace = os.Getenv("NAMESPACE")
@@ -202,6 +203,20 @@ func (dn *Daemon) annotateUnsupportedNicIdConfigMap(cm *v1.ConfigMap, nodeName s
 func (dn *Daemon) tryCreateUdevRuleWrapper() error {
 	var unsupportedNicIdMap map[string]string
 
+	ns, nodeStateErr := dn.client.SriovnetworkV1().SriovNetworkNodeStates(namespace).Get(
+		context.Background(),
+		dn.name,
+		metav1.GetOptions{},
+	)
+	if nodeStateErr != nil {
+		glog.Warningf("Could not fetch node state %s: %v, skip updating switchdev udev rules", dn.name, nodeStateErr)
+	} else {
+		err := tryCreateSwitchdevUdevRule(ns)
+		if err != nil {
+			glog.Warningf("Failed to create switchdev udev rules: %v", err)
+		}
+	}
+
 	// retrieve config map
 	cm, configMapErr := dn.getUnsupportedNicIdConfigMap()
 	if configMapErr != nil {
@@ -211,7 +226,7 @@ func (dn *Daemon) tryCreateUdevRuleWrapper() error {
 	}
 
 	// update udev rule and if update succeeds
-	err := tryCreateUdevRule(unsupportedNicIdMap)
+	err := tryCreateNMUdevRule(unsupportedNicIdMap)
 	if err != nil {
 		return err
 	}
@@ -800,7 +815,7 @@ func tryEnableRdma() (bool, error) {
 	glog.V(2).Infof("tryEnableRdma()")
 	var stdout, stderr bytes.Buffer
 
-	cmd := exec.Command("/bin/bash", scriptsPath)
+	cmd := exec.Command("/bin/bash", rdmaScriptsPath)
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	if err := cmd.Run(); err != nil {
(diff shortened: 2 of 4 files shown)


pliurh:
LGTM 
@adrianchiris PTAL

adrianchiris:
Overall looks OK, could you add a couple of sentences in the commit message describing why the change is needed and what is done in this commit

zshi-redhat:
> Overall looks OK, could you add a couple of sentences in the commit message describing why the change is needed and what is done in this commit

Done

bindata/manifests/machine-config/files/switchdev-vf-link-name.sh.yaml
```diff
@@ -0,0 +1,59 @@
+mode: 0755
+overwrite: true
+path: "/etc/udev/switchdev-vf-link-name.sh"
+contents:
+  inline: |
+    #!/bin/bash
```
moshe010, line 6: the switchdev-vf-link-name.sh is doing lot of logic which can case races. We saw it in openstack. It is better to do it like it done in os-net-config [1]. The script is just running simple echo command. We also mark the VF rep to be unmanaged by the NM with 'ENV{NM_UNMANAGED}="1" 


[1] https://github.com/openstack/os-net-config/blob/3144a8fc4fae95e3cfdc680ccdcbd5144a9a17a8/os_net_config/sriov_config.py#L344-L373
[2] - https://github.com/openstack/os-net-config/blob/3144a8fc4fae95e3cfdc680ccdcbd5144a9a17a8/os_net_config/sriov_config.py#L52-L58

moshe010 (changes requested):

bindata/manifests/machine-config/files/switchdev-vf-link-name.sh.yaml
wmousa: > the switchdev-vf-link-name.sh is doing lot of logic which can case races. We saw it in openstack. It is better to do it like it done in os-net-config [1]. The script is just running simple echo command. We also mark the VF rep to be unmanaged by the NM with 'ENV{NM_UNMANAGED}="1"
> 
> [1] https://github.com/openstack/os-net-config/blob/3144a8fc4fae95e3cfdc680ccdcbd5144a9a17a8/os_net_config/sriov_config.py#L344-L373
> [2] - https://github.com/openstack/os-net-config/blob/3144a8fc4fae95e3cfdc680ccdcbd5144a9a17a8/os_net_config/sriov_config.py#L52-L58

But Moshe in os-net-config, we know the name of the sriov_pf  and the switch_id and we add the udev rule according to that, Do we have such info here ? @moshe010

wmousa: > > the switchdev-vf-link-name.sh is doing lot of logic which can case races. We saw it in openstack. It is better to do it like it done in os-net-config [1]. The script is just running simple echo command. We also mark the VF rep to be unmanaged by the NM with 'ENV{NM_UNMANAGED}="1"
> > [1] https://github.com/openstack/os-net-config/blob/3144a8fc4fae95e3cfdc680ccdcbd5144a9a17a8/os_net_config/sriov_config.py#L344-L373
> > [2] - https://github.com/openstack/os-net-config/blob/3144a8fc4fae95e3cfdc680ccdcbd5144a9a17a8/os_net_config/sriov_config.py#L52-L58
> 
> But Moshe in os-net-config, we know the name of the sriov_pf and the switch_id and we add the udev rule according to that, Do we have such info here ? @moshe010

I think it's better to add a udev rule per pf here [1], you can see this one [2]
[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/bindata/manifests/machine-config/files/configure-switchdev.sh.yaml
[2] https://github.com/Mellanox/switchdev-utils/blob/master/switchdev_utils/switchdev.sh#L12-L74

zshi-redhat: switch_id is only available when switchdev mode is configured. If we were to add switch_id in udev rules, we also need to add logic in config daemon to reload udev rules which I intended to avoid. because appliance of udev rules will be done automatically along with rebooting (which is required to configure switchdev device) and config daemon doesn't have to track whether node is rebooted or not (to apply the udev rules). The current change looks simple than updating config daemon code since we handle all the switchdev config via plugins.

zshi-redhat: Removed the long bash script for renaming vf rep

pkg/daemon/daemon.go
```diff
@@ -822,6 +837,64 @@ func tryEnableRdma() (bool, error) {
...
+			switchID, err := utils.GetPhysSwitchID(ifaceStatus.Name)
+			if err != nil {
+				return err
+			}
+			portName, err := utils.GetPhysPortName(ifaceStatus.Name)
+			if err != nil {
+				return err
+			}
+			new_content = new_content + fmt.Sprintf("SUBSYSTEM==\"net\", ACTION==\"add|move\", ATTRS{phys_switch_id}==\"%s\", ATTR{phys_port_name}==\"pf%svf*\", IMPORT{program}=\"/etc/udev/switchdev-vf-link-name.sh $attr{phys_port_name}\", NAME=\"%s_$env{NUMBER}\"\n", switchID, strings.TrimPrefix(portName, "p"), ifaceStatus.Name)
+		}
```
moshe010, line 856: in os-net-config we have udev rule unmanage vf_representors by NM. I think you should added it as well.

https://github.com/openstack/os-net-config/blob/3144a8fc4fae95e3cfdc680ccdcbd5144a9a17a8/os_net_config/sriov_config.py#L369-L373

pkg/daemon/daemon.go
```diff
@@ -822,6 +837,64 @@ func tryEnableRdma() (bool, error) {
...
+	}
+
+	glog.V(2).Infof("Old udev content '%v' and new content '%v' differ. Writing to file %v.",
+		strings.TrimSuffix(string(old_content), "\n"),
+		strings.TrimSuffix(new_content, "\n"),
+		filePath)
+
+	// if the file does not exist or if old_content != new_content
+	// write to file and create it if it doesn't exist
+	err = ioutil.WriteFile(filePath, []byte(new_content), 0666)
```
moshe010, line 872: why rw to other? isn't 0664 is better?

pkg/daemon/daemon.go
```diff
@@ -822,6 +837,64 @@ func tryEnableRdma() (bool, error) {
...
+			if err != nil {
+				return err
+			}
+			new_content = new_content + fmt.Sprintf("SUBSYSTEM==\"net\", ACTION==\"add|move\", ATTRS{phys_switch_id}==\"%s\", ATTR{phys_port_name}==\"pf%svf*\", IMPORT{program}=\"/etc/udev/switchdev-vf-link-name.sh $attr{phys_port_name}\", NAME=\"%s_$env{NUMBER}\"\n", switchID, strings.TrimPrefix(portName, "p"), ifaceStatus.Name)
+		}
+	}
+
+	old_content, err := ioutil.ReadFile(filePath)
+	// if old_content = new_content, don't do anything
+	if err == nil && new_content == string(old_content) {
```
moshe010, line 861: you can string(old_content) once. currently you doing in L861 and L866

pkg/daemon/daemon.go
zshi-redhat: Done

pkg/daemon/daemon.go
zshi-redhat: Done

pkg/daemon/daemon.go
zshi-redhat: I don't see how I can do it once w/o adding a new string var, any idea?

bindata/manifests/machine-config/files/switchdev-vf-link-name.sh.yaml
zshi-redhat: @moshe010 @wmousa  have you seen any issue in openstack that udev rules cannot be applied due to vf rep is not link down?

moshe010: we did encounter  it. Are you saying that if VF rep is up udev rule will failed?

zshi-redhat: yes

moshe010: but why the VF reps are up? They are not moved to the container namespace right?

zshi-redhat: When the VF reps first get created, won't they be linked up auto?

moshe010: no they should be down. lined up auto if PF to VF.

moshe010: can you please check again? maybe NM starting them up?

zshi-redhat: If it's NM, then I probably need to set the UM udev rules before any switchdev device is configured, otherwise if NM udev rules is set after rebooting, then it takes another reboot to take effect.
I'll split the rules to separate functions and call them differently.

moshe010: Did it solve the issue?

zshi-redhat: I'm still testing it, what I observed was:
On first reboot, VF rep was not renamed correctly, still has the name ethx
On subsequential reboot, VF rep was renamed correctly.

zshi-redhat: ok, the issue that vf was not renamed on first reboot is because udev rules are added after reboot so they are not re-loaded automatically. Fixed it by explicitly reload the udev rules before triggering.

zshi-redhat:
@moshe010 any other suggestions on the renaming PR?

moshe010 (approved):
LGTM

moshe010:
@zshi-redhat  I think we good to go with this one. Can I merge it?

zshi-redhat:
> @zshi-redhat I think we good to go with this one. Can I merge it?

Yes, please. I just don't want to merge my own PR.

moshe010 (approved):

PR #39, opened 2021-01-07
Support changing Eswitch mode of SR-IOV NICs for kubernetes deployment

Mmduh-483:
This PR introduces new changes to support changing E-switch mode for vanilla Kubernetes
* Introduce service package to maintain services on vanilla Kubernetes, this is needed as an alternative for Openshift MCO
* Move bindata/manifests/machine-config to bindata/manifests/switchdev-config, to avoid confuion of Openshift MCO
* Introduce Kubernetes plugin to maintain services for switchdev
*  Add documention for hardware offload


bindata/manifests/k8s-services/configure-switchdev.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+set -eux
```
moshe010, line 2: why do you need this script if you take it from MCO?

pkg/service/utils.go
```diff
@@ -0,0 +1,110 @@
...
+
+	return &Service{
+		Name:    service.Name,
+		Path:    service.Path,
+		Content: string(data),
+	}, nil
+}
+
+// ServiceAppend appends missing fields of serviceB into serviceA
+func ServiceAppend(serviceA, serviceB *Service) (*Service, error) {
```
moshe010, line 73: I think you can change to  ServiceAppend(serviceA *Service, contentToAppend)((*Service, error)
Why do you need the serviceB?

pkg/service/utils.go
```diff
@@ -0,0 +1,110 @@
...
+
+		if !found {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// CompareServices compare 2 service and return true if serviceA has all the fields of serviceB
```
moshe010, line 38: update comment

pkg/service/utils.go
moshe010: Also maybe change to AppendToService to be same as RemoveFromService

pkg/service/utils.go
```diff
@@ -0,0 +1,110 @@
...
+	for _, optB := range optsB {
+		found := false
+		for _, optA := range optsA {
+			if optA.Match(optB) {
+				found = true
+				break
+			}
+		}
+
+		if !found {
```
moshe010, line 30: so after reading about the break label I think what you did first with the outer was better to break from both loops.

pkg/service/service_manager.go
```diff
@@ -0,0 +1,128 @@
...
+		return err
+	}
+	defer exit()
+
+	// Remove service file
+	cmd := exec.Command("systemctl", "stop", service)
+	return cmd.Run()
+}
+
+func (sm *serviceManager) isServiceLoaded(servicePath string) (bool, error) {
```
moshe010, line 118: What do you mean by loaded? 
Do you mean if the Service is activate? start? and why checking the file is good enough ?

pkg/service/service_manager.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	defer exit()
+
+	// Enable service
+	cmd := exec.Command("systemctl", "enable", service.Name)
+	return cmd.Run()
+}
+
+// DisableService disables service with systemctl disable and removes service file
+func (sm *serviceManager) DisableService(service *Service) error {
```
moshe010, line 68: are you using all the serviceManager apis? if not just keep the ones that you are currently use.

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,287 @@
...
+		err := ioutil.WriteFile(path.Join(chroot, switchdevRunScript.Path),
+			[]byte(switchdevRunScript.Contents.Inline), 0755)
+		if err != nil {
+			return err
+		}
+	}
+
+	if updateTarget.networkManager {
+		nmService, err := serviceManager.ReadService(networkManagerService.Path)
+		if err != nil {
```
moshe010, line 163: were is the ovs service?

moshe010 (changes requested):

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,336 @@
...
+	networkManager   bool
+	openVSwitch      bool
+}
+
+func (u *k8sUpdateTarget) needUpdate() bool {
+	return u.switchdevService || u.switchdevScript || u.networkManager || u.openVSwitch
+}
+
+const (
+	mcoManifestPath           = "bindata/manifests/machine-config/"
```
moshe010, line 55: let rename the folder to be bindata/manifests/switchdev-config/
and MCO andK8s both of them will use it.

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+type McoFile struct {
+	Path     string
+	Contents struct {
+		Inline string
+	}
+}
+
+type McoServiceInjection struct {
```
moshe010, line 36: remove all the Mco prefix from all the struct as they are not releated to MCO

moshe010 (changes requested):

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -50,6 +50,10 @@ spec:
         volumeMounts:
         - name: host
           mountPath: /host
+        lifecycle:
+          preStop:
+            exec:
+              command: ["/bin/sh","-c","rm -f /host/etc/systemd/system/switchdev-configuration.service"]
```
moshe010, line 56: this is not good enough. We should undo all the other changes

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,336 @@
...
+	// Check switchdev config
+	var update, remove bool
+	if update, remove, err = utils.WriteSwitchdevConfFile(new); err != nil {
+		glog.Errorf("k8s-plugin OnNodeStateChange():fail to update switchdev.conf file: %v", err)
+		return
+	}
+	if remove {
+		glog.Info("k8s-plugin OnNodeStateChange(): need reboot node to clean switchdev VFs")
+		needDrain = true
+		needReboot = true
```
moshe010, line 130: I don't think we need to reboot in any of this cases.

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,336 @@
...
+		return
+	}
+	if update {
+		glog.Info("k8s-plugin OnNodeStateChange(): need reboot node to use the up-to-date switchdev.conf")
```
moshe010, line 134: is not systemctl reload-deamon will load the new switchdev.conf

moshe010 (changes requested):

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,336 @@
...
+			return false, err
+		}
+		return needChange, nil
+	}
+
+	return false, nil
+}
+
+func updateService(serviceObj *service.Service) error {
+	systemService, err := serviceManager.ReadService(serviceObj.Path)
```
moshe010, line 318: if you will move the compare here you won't need to check if that service need update.
Just do it here.

moshe010 (changes requested):

pkg/plugins/k8s/k8s_plugin.go
Mmduh-483: `updateService` is called from `Apply` when there is change needed, the compare is running from `OnNodeChange` which check if there is change needed to be applied

pkg/plugins/k8s/k8s_plugin.go
Mmduh-483: switchdev.conf is updated by `utils.WriteSwitchdevConfFile` and the switchdev-service script reads it
Updating the service won't affect the file

pkg/plugins/k8s/k8s_plugin.go
Mmduh-483: The update VFs in the generic plugin ignores the switchdev vfs, if there is 8 switchdev vfs we want to change them in legacy, it will be ignored because the requested num of vfs == existing num of vfs

bindata/scripts/clean-k8s-services.sh
```diff
@@ -0,0 +1,26 @@
...
+  fi
+
+  if [ -f $ovs_service ]; then
+    sed -i.bak '/hw-offload/d' $ovs_service
+  fi
+}
+
+clean_services
+# Reload host services
+chroot $chroot_path /bin/bash -c systemctl daemon-reload > /dev/null
```
moshe010, line 26: shouldn't you restart the ovs and NM services?

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,386 @@
...
+	u.systemServices = []*service.Service{}
+}
+
+func (u *k8sUpdateTarget) String() string {
+	var updateList []string
+	if u.switchdevService {
+		updateList = append(updateList, "SwitchdevService")
+	}
+	if u.switchdevScript {
+		updateList = append(updateList, "SwitchdevService")
```
moshe010, line 71: should be "switchdevScript"

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,386 @@
...
+			services = append(services, systemService)
+		}
+	}
+
+	p.updateTarget.systemServices = services
+	return nil
+}
+
+func (p *K8sPlugin) isServicesNeedUpdate() error {
+	// Check switchdev
```
moshe010, line 333: it confusing it should return bool

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,386 @@
...
+
+// OnNodeStateChange Invoked when SriovNetworkNodeState CR is updated, return if need dain and/or reboot node
+func (p *K8sPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
+	glog.Info("k8s-plugin OnNodeStateChange()")
+	needDrain = false
+	needReboot = false
+
+	p.updateTarget.reset()
+	// Check services
+	err = p.isServicesNeedUpdate()
```
moshe010, line 135: isServicesNeedUpdate should return bool not err

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,386 @@
...
+		}
+		if err := yaml.Unmarshal(data, &p.switchdevRunScript); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (p *K8sPlugin) readSystemServiceManifestFile(path string) (*service.Service, error) {
```
moshe010, line 224: This should move to be in the function ServiceManager file like the compare services

moshe010 (changes requested):

pkg/plugins/k8s/k8s_plugin.go
Mmduh-483: This is reading from `bindata/manifests/machine-config` and parsing the yaml file, I don't think this is related to ServiceManager 
WDYT?

moshe010: Why not? because it end with yaml? I think in the future when we will add services if any it will be in the same approach. Don't you think?

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	return nil
+}
+
+func (p *K8sPlugin) getSystemServices() []*service.Service {
+	return []*service.Service{p.networkManagerService, p.openVSwitchService}
+}
+
+func (p *K8sPlugin) isSystemServiceNeedUpdate(serviceObj *service.Service) (bool, error) {
```
moshe010, line 251: let split the logic. 
You should have isSystemServiceExist() to validate at start 
and then isSystemServiceNeedUpdate can return just bool

moshe010 (changes requested):

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,350 @@
...
+	return u.switchdevService || u.switchdevScript || len(u.systemServices) > 0
+}
+
+func (u *k8sUpdateTarget) reset() {
+	u.switchdevService = false
+	u.switchdevScript = false
+	u.systemServices = []*service.Service{}
+}
+
+func (u *k8sUpdateTarget) String() string {
```
moshe010, line 45: where are you using it? and why is with capital latter?

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	if update {
+		glog.Info("k8s-plugin OnNodeStateChange(): need reboot node to use the up-to-date switchdev.conf")
+		needDrain = true
+		needReboot = true
+		return
+	}
+	if p.updateTarget.needUpdate() {
+		glog.Infof("k8s-plugin OnNodeStateChange(): needDrain to update %q", p.updateTarget)
+		needDrain = true
```
moshe010, line 140: don't you need reboot?

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+func (p *K8sPlugin) getSystemServices() []*service.Service {
+	return []*service.Service{p.networkManagerService, p.openVSwitchService}
+}
+
+func (p *K8sPlugin) isSystemServiceNeedUpdate(serviceObj *service.Service) bool {
+	systemService, err := p.serviceManager.ReadService(serviceObj.Path)
+	if err != nil {
+		return false
```
moshe010, line 264: you shoud log warn/err msg here

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	if systemService != nil {
+		needChange, err := service.CompareServices(systemService, serviceObj)
+		if err != nil {
+			return false
```
moshe010, line 269: same as above

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,350 @@
...
+
+func (p *K8sPlugin) systemServicesStateUpdate() error {
+	var services []*service.Service
+	for _, systemService := range p.getSystemServices() {
+		exist, err := p.serviceManager.IsServiceExist(systemService.Path)
+		if err != nil {
+			return err
+		}
+		if !exist {
+			continue
```
moshe010, line 285: if it not exist it a problem we should log error

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -0,0 +1,350 @@
...
+	networkManagerService, err := service.ReadServiceInjectionManifestFile(networkManagerUnitFile)
+	if err != nil {
+		return err
+	}
+
+	p.networkManagerService = networkManagerService
+	return nil
+}
+
+func (p *K8sPlugin) readOpenVSwichdManifest() error {
```
zshi-redhat, line 200: `readOpenVSwichdManifest` -> `readOpenVSwitchManifest`; lack of `t`

zshi-redhat (commented):
So k8s_plugin config is applied regardless of switchdev configuration or SriovOperatorConfig.EnableOvsOffload ？

pkg/plugins/k8s/k8s_plugin.go
Mmduh-483: On line 139 in print the log message, function `String` is used to represent any object as a string in go

pkg/plugins/k8s/k8s_plugin.go
Mmduh-483: Adding the services doesn't require reboot

Mmduh-483:
> So k8s_plugin config is applied regardless of switchdev configuration or SriovOperatorConfig.EnableOvsOffload ？

@zshi-redhat yes, I see the daemon can read the oeratorConfig but it is not passing them to plugins, is there a plan to pass them in the future? or use a 3rd party component daemon and plugins to read the operator config?

If k8s_plugin would use the config of enableOvsOfflad it should contact the k8s api @moshe010 WDYT?

pkg/plugins/k8s/k8s_plugin.go
Mmduh-483: if it not exist it means the NetworkManager or ovs is not installed on the host so a warning would serve better unless you consider them a pre-request, WDYT?

moshe010: The ovs is must not sure about the NM

pkg/plugins/k8s/k8s_plugin.go
moshe010: but what about the VF rep udev rename? it won't be triggered ...

Mmduh-483: It is possible to add an additional condition to the udev renaming to do reboot, but the PR of renaming is not merged yet

pkg/plugins/k8s/k8s_plugin.go
Mmduh-483: So it would be better to split the logic to check if the service needs an update because there are mandatory and optional services

moshe010: Let assume all of are mandatory ovs and NM

moshe010:
> > So k8s_plugin config is applied regardless of switchdev configuration or SriovOperatorConfig.EnableOvsOffload ？
> 
> @zshi-redhat yes, I see the daemon can read the oeratorConfig but it is not passing them to plugins, is there a plan to pass them in the future? or use a 3rd party component daemon and plugins to read the operator config?
> 
> If k8s_plugin would use the config of enableOvsOfflad it should contact the k8s api @moshe010 WDYT?

How it done with mco? it should be the same

zshi-redhat:
> > > So k8s_plugin config is applied regardless of switchdev configuration or SriovOperatorConfig.EnableOvsOffload ？
> > 
> > 
> > @zshi-redhat yes, I see the daemon can read the oeratorConfig but it is not passing them to plugins, is there a plan to pass them in the future? or use a 3rd party component daemon and plugins to read the operator config?
> > If k8s_plugin would use the config of enableOvsOfflad it should contact the k8s api @moshe010 WDYT?
> 
> How it done with mco? it should be the same

There is a conditional check in sriovoperatorconfig controller: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/controllers/sriovoperatorconfig_controller.go#L443

Basically only render and reconcile the MCO manifests when sriovOperatorConfig.EnableOvsOffload is true.
Since we are not using MCO in vanilla k8s, hense @Mmduh-483 has the question how to pass the flag to plugins.

zshi-redhat:
> > So k8s_plugin config is applied regardless of switchdev configuration or SriovOperatorConfig.EnableOvsOffload ？
> 
> @zshi-redhat yes, I see the daemon can read the oeratorConfig but it is not passing them to plugins, 

correct

> is there a plan to pass them in the future? 

we don't have yet.

> or use a 3rd party component daemon and plugins to read the operator config?

I prefer to not have plugin accessing specific API fields rather passing configurations as arguments to plugin functions(although there are plugins already accessing API directly via kubeclient)
> 
> If k8s_plugin would use the config of enableOvsOfflad it should contact the k8s api @moshe010 WDYT?

A rough idea, I was thinking to lift the service management to daemon with another goroutine (just like the writer goroutine in daemon), and manage the service injection/deletion in that separate goroutine. With that, SriovOperatorConfig will be able to be received in the service management goroutine and future injection can utilize the same. but I didn't think that through yet.

Let's discuss this and come up with a better idea, for current PR, I don't think we need to change it immediately as it's in good shape and we can do it in separate PRs if needed.

pkg/plugins/k8s/k8s_plugin.go
moshe010: Sorry I don't follow why it better to do it in the udev PR and not here?

pkg/utils/switchdev.go
```diff
@@ -0,0 +1,58 @@
...
+		glog.V(2).Info("WriteSwitchdevConfFile(): no update")
+		return
+	}
+	if newContent == "" {
+		remove = true
+		glog.V(2).Info("WriteSwitchdevConfFile(): remove content in switchdev.conf")
+	}
+	update = true
+	glog.V(2).Infof("WriteSwitchdevConfFile(): write %s to switchdev.conf", newContent)
+	err = ioutil.WriteFile(switchDevConfPath, []byte(newContent), 0666)
```
moshe010, line 52: I assume this what the permission in the MCO commit. I think it better to change it to 644. 
@zshi-redhat  what do you thing?

zshi-redhat: Yes, we only need this file be readable for users. mco/k8s plugins write the file.

pkg/plugins/k8s/k8s_plugin.go
Mmduh-483: I assume would  be better because the udev is not merged yet, and when udev is merged we need to do another PR to support it for k8s_plugin

pkg/utils/switchdev.go
Mmduh-483: updated

moshe010 (approved):

martinkennelly:
/LGTM - Can you document the feature to advertise it? I am not sure where, features in readme? - Maybe we need a switchdev doc.

zshi-redhat:
> /LGTM - Can you document the feature to advertise it? I am not sure where, features in readme? - Maybe we need a switchdev doc.

@Mmduh-483 would you mind write a section in doc/ovs-hw-offload.md for using switchdev device on k8s? Thanks!

Mmduh-483:
@moshe010 @zshi-redhat @martinkennelly added docs under `doc/ovs-hw-offload.md`

doc/ovs-hw-offload.md
```diff
@@ -0,0 +1,131 @@
...
+
+The OVS software based solution is CPU intensive, affecting system performance
+and preventing fully utilizing available bandwidth. OVS 2.8 and above support
+a feature called OVS Hardware Offload which improves performance significantly.
+This feature allows offloading the OVS data-plane to the NIC while maintaining
+OVS control-plane unmodified. It is using SR-IOV technology with VF representor
+host net-device. The VF representor plays the same role as TAP devices
+in Para-Virtual (PV) setup. A packet sent through the VF representor on the host
+arrives to the VF, and a packet sent through the VF is received by its representor.
+
```
moshe010, line 11: we should have prerequisites section:
1. ovs installed 
2. Network Manager installed

doc/ovs-hw-offload.md
```diff
@@ -0,0 +1,135 @@
...
+The following manufacturers are known to work:
+
+- Mellanox ConnectX-5 and above
+
+## Instructions for Mellanox ConnectX-5
+
+## Prerequisites
+
+- OpenVswitch installed
+- Network Manager installer
```
moshe010, line 23: s/installer/installed

moshe010 (approved):
@martinkennelly can you please take a look again. documentation was added

moshe010:
@zshi-redhat Can we merge this ?

pliurh (commented):
If I read the code correctly, the k8s plugin will always try to update the systemd services even when switchdev is not configured. Did I miss anything?

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -50,6 +50,10 @@ spec:
         volumeMounts:
         - name: host
           mountPath: /host
+        lifecycle:
+          preStop:
+            exec:
+              command: ["/bindata/scripts/clean-k8s-services.sh"]
```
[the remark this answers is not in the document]
Mmduh-483: Update the script to exist if CLUSTER_TYPE==openshift

Mmduh-483:
> If I read the code correctly, the k8s plugin will always try to update the systemd services even when switchdev is not configured. Did I miss anything?

Updated to update services and swithdev config if policy is switchdev

pliurh:
@Mmduh-483 Could you add more information to the commit description? This patch does more than just changing Eswitch mode. Also, I suggest adding a 'TODO' in the code, saying we will address the `enableOvsOffload` flag issue in the future.

Mmduh-483:
> @Mmduh-483 Could you add more information to the commit description? This patch does more than just changing E-switch mode. Also, I suggest adding a 'TODO' in the code, saying we will address the `enableOvsOffload` flag issue in the future.

Updated the PR with separated commits with detailed commit description for each commit and the PR description

Mmduh-483:
> @Mmduh-483 Could you add more information to the commit description? This patch does more than just changing Eswitch mode. Also, I suggest adding a 'TODO' in the code, saying we will address the `enableOvsOffload` flag issue in the future.

Updated the PR with separated commits with detailed commit description for each commit and the PR description

PR #46, opened 2021-01-13
Generate device plugin LinkType selector from NodePolicy LinkType

Mmduh-483:


controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -553,6 +554,13 @@ func (r *SriovNetworkNodePolicyReconciler) renderDevicePluginConfigData(pl *srio
 			if len(p.Spec.NicSelector.PfNames) > 0 {
 				netDeviceSelectors.PfNames = sriovnetworkv1.UniqueAppend(netDeviceSelectors.PfNames, p.Spec.NicSelector.PfNames...)
 			}
+			if p.Spec.LinkType != "" && !sriovnetworkv1.StringInArray(p.Spec.LinkType, netDeviceSelectors.LinkTypes) {
```
moshe010, line 557: in what case with p.Spec.LinkType == ""?

moshe010, line 557: !sriovnetworkv1.StringInArray(p.Spec.LinkType, netDeviceSelectors.LinkTypes) this is to check if that linktype selector already exist?

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -553,6 +554,13 @@ func (r *SriovNetworkNodePolicyReconciler) renderDevicePluginConfigData(pl *srio
...
+				linkType := linkTypeEthernet
+				if strings.ToLower(p.Spec.LinkType) == "ib" {
+					linkType = linkTypeInfiniband
+				}
+				netDeviceSelectors.LinkTypes = append(netDeviceSelectors.LinkTypes, linkType)
```
zshi-redhat, line 562: Might be good to us sriovnetworkv1.UniqueAppend?

controllers/sriovnetworknodepolicy_controller.go
zshi-redhat, line 557: netDeviceSelectors.LinkTypes use different value than p.Spec.LinkType, for example:
the former may includes `infiniband`, the latter may be `ib`. so the conditional check here would always true if p.Spec.LinkType != "" ?

zshi-redhat: This API filed is omitempty.

PR #47, opened 2021-01-18
nodeAffinity is not correctly updated with DeepDerivative

zshi-redhat:
https://bugzilla.redhat.com/show_bug.cgi?id=1914066

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -444,8 +444,13 @@ func (r *SriovNetworkNodePolicyReconciler) syncDaemonSet(cr *sriovnetworkv1.Srio
 		// This skips default values added by the api server.
 		// References in https://github.com/kubernetes-sigs/kubebuilder/issues/592#issuecomment-625738183
 		if equality.Semantic.DeepDerivative(in.Spec, ds.Spec) {
-			logger.Info("Daemonset spec did not change, not updating")
-			return nil
+			// DeepDerivative has issue detecting nodeAffinity change
+			// https://bugzilla.redhat.com/show_bug.cgi?id=1914066
+			if equality.Semantic.DeepEqual(in.Spec.Template.Spec.Affinity.NodeAffinity,
+				ds.Spec.Template.Spec.Affinity.NodeAffinity) {
+				logger.Info("Daemonset spec did not change, not updating")
+				return nil
+			}
 		}
 		err = r.Update(context.TODO(), in)
 		if err != nil {


fedepaol:
Ouch! Couple of questions: 

- Is this because the nodeAffinity is reset when deleting the policy (and as a 0 field is not compared)?
- Should we replace the other occurrences of `DeepDerivative` too?
- Did you check if this brings back the unwanted extra reconciliation loops?

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -443,7 +443,10 @@ func (r *SriovNetworkNodePolicyReconciler) syncDaemonSet(cr *sriovnetworkv1.Srio
 		// DeepDerivative checks for changes only comparing non zero fields in the source struct.
 		// This skips default values added by the api server.
 		// References in https://github.com/kubernetes-sigs/kubebuilder/issues/592#issuecomment-625738183
-		if equality.Semantic.DeepDerivative(in.Spec, ds.Spec) {
+
+		// Note: with DeepDerivative, controller has issue not updating nodeAffinity correctly:
+		// https://bugzilla.redhat.com/show_bug.cgi?id=1914066, Chnange it to DeepEqual()
```
fedepaol, line 448: Typo: "chnange".
I also wonder if we should delete the comment abovee

zshi-redhat:
> Ouch! Couple of questions:
> 
> * Is this because the nodeAffinity is reset when deleting the policy (and as a 0 field is not compared)?

NodeAffinity is not updated when deleting policy, the legacy nodeAffinity remains in the daemonset spec.

> * Should we replace the other occurrences of `DeepDerivative` too?

I didn't know any issue with the other one, I think we can keep it as it is for now.

> * Did you check if this brings back the unwanted extra reconciliation loops?

Yes, there is no endless loop in sriov operator log with DeepEqual.

controllers/sriovnetworknodepolicy_controller.go
zshi-redhat: I can delete previous comment if you prefer, just thought it might be useful for reference if later we hit the endless loop issue again.

fedepaol: I don't have a strong opinion, I guess it's fine since the comment below explains it pretty clearly.

fedepaol:
LGTM (apart from the typo)

controllers/sriovnetworknodepolicy_controller.go
zshi-redhat: fixed typo

fedepaol:
LGTM

pliurh:
With `DeepEqual`, I suppose we will hit the issue which @fedepaol tried to fix since it will be always false. Can we add a condition to check whether the nodeAffinity should be removed from the spec of DS, instead of using `DeepEqual`?

zshi-redhat:
> With `DeepEqual`, I suppose we will hit the issue which @fedepaol tried to fix since it will be always false. Can we add a condition to check whether the nodeAffinity should be removed from the spec of DS, instead of using `DeepEqual`?

Done

PR #49, opened 2021-01-21
Create ovn net-attach-def for switchdev policy (ovs hw offload)

zshi-redhat:


api/v1/helper.go
```diff
@@ -30,7 +30,8 @@ const (
 
 const invalidVfIndex = -1
 
-var MANIFESTS_PATH = "./bindata/manifests/cni-config"
+var SRIOV_NET_ATTACH_DEF_PATH = "./bindata/manifests/cni-config/sriov"
+var OVN_NET_ATTACH_DEF_PATH = "./bindata/manifests/cni-config/ovn"
```
moshe010, line 34: can you rename this to ovnkube? there are other cni that used ovn (e.g https://github.com/alauda/kube-ovn)  which we may want to add in the future

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -152,6 +157,47 @@ func (r *SriovNetworkNodePolicyReconciler) SetupWithManager(mgr ctrl.Manager) er
...
+func (r *SriovNetworkNodePolicyReconciler) syncOvnNetAttachDefs(pl *sriovnetworkv1.SriovNetworkNodePolicyList) error {
+	logger := r.Log.WithName("syncOvnNetAttachDefs")
+	logger.Info("Start to sync default ovn net-attach-defs")
+
+	for _, p := range pl.Items {
+		if p.Name == "default" {
+			continue
+		}
+		if p.Spec.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
+			raw, err := p.RenderNetAttDef()
```
moshe010, line 169: we don't have to address it now, but I think there is going to be several cni's that can leverage the switchdev config. 
option 1 - is that we have a flag that skip the  RenderNetAttDef
option 2 - is that we had network policy to each cni
or maybe we need both options. 
What do you think?

api/v1/helper.go
zshi-redhat: Done

controllers/sriovnetworknodepolicy_controller.go
zshi-redhat: I agree that we need distinguish switchdev for different CNI cases. I only add switchdev conditional check for ovs hw offload net-attach-def is because it is the only enabled feature atm for switchdev.

Regarding the options, I'd like to see if it's possible to do auto detection of whether a PF is enslaved as OVS port (ovs hw offload) or standalone/enslaved in linux bridge (accelerated bridge). If it's not something detectable, having a flag (API change) would be desired in my opinion.

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -152,6 +157,47 @@ func (r *SriovNetworkNodePolicyReconciler) SetupWithManager(mgr ctrl.Manager) er
...
 
+func (r *SriovNetworkNodePolicyReconciler) syncOvnNetAttachDefs(pl *sriovnetworkv1.SriovNetworkNodePolicyList) error {
+	logger := r.Log.WithName("syncOvnNetAttachDefs")
+	logger.Info("Start to sync default ovn net-attach-defs")
+
+	for _, p := range pl.Items {
+		if p.Name == "default" {
+			continue
+		}
+		if p.Spec.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
```
moshe010, line 168: I think we should have a flag to disable the render RenderNetAttDef. We see  other use-case for other cni that we can leverage as well.  In the future I think we should create several network types 
1. sriovnetwork
2. ibsriovnetwork
3. ovnkubnetwork
4. ovsnetwork
5. acceleratebridgenetwork

And that should implies the  SR-IOV mode (legacy/switchdev) 
What do you think?

moshe010 (changes requested):

controllers/sriovnetworknodepolicy_controller.go
zshi-redhat: @moshe010 sorry for late response.
By disabling RenderNetAttDef, we are expecting user to create their own net-attach-def based on the networktypes they will be using, am I understand it correctly ?

moshe010: yes

zshi-redhat: Could the flag be a env var that passed in to operator.yml or do you think it's prefered to be an API option in SriovOperatorConfig?

zshi-redhat: > I think we should have a flag to disable the render RenderNetAttDef. We see other use-case for other cni that we can leverage as well. In the future I think we should create several network types
> 
>     1. sriovnetwork
> 
>     2. ibsriovnetwork
> 
>     3. ovnkubnetwork
> 
>     4. ovsnetwork
> 
>     5. acceleratebridgenetwork
> 
> 
> And that should implies the SR-IOV mode (legacy/switchdev)
> What do you think?

Would there be cases that one network type may use both SR-IOV modes?

zshi-redhat:
The approach in this PR doesn't seems scale when multiple CNI network types are supported which may use different sriov modes. In which case we will have a mixed combinations of which device mode works with which CNI network type, and Operator cannot easily decide what net-attach-def should be created.

The intention of this PR was to ease user's configuration and create the net-attach-def automatically, but it turns out we cannot easily infer the relations between CNI and device Mode in Operator w/o adding additional configurations.

PR #50, opened 2021-01-21
Run generic plugin Apply when no vender plugin loaded

pliurh:
Fix the issue that generic plugin Apply won't be invoked if no supported
vendor NICs in the host.

https://bugzilla.redhat.com/show_bug.cgi?id=1914414

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -514,7 +514,7 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 		}
 	}
 
-	if len(dn.LoadedPlugins) > 1 && !reqReboot {
+	if !reqReboot {
 		// Apply generic_plugin last
 		err = dn.LoadedPlugins[GenericPlugin].Apply()
 		if err != nil {


adrianchiris:
with this change the intel plugin is no longer needed ?

zshi-redhat:
> with this change the intel plugin is no longer needed ?

With this change, generic_plugin is able to be applied no matter vendor plugins are loaded or not. This change is not specific to intel or mellanox. Previously, generic_plugin cannot be applied to other vendor NICs if none of intel or mellanox plugins are loaded.

moshe010:
I think adrian point was that you can also delete intel plugis as it does nothing today.

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

adrianchiris:
In generally LGTM on this one, I just wonder if intel plugin should be removed as it basically does nothing and generic plugin will always be invoked by the config daemon

pliurh:
Yes, the intel plugin does nothing today. 
I think we can remove it for now, and add it back if there are any intel devices that require additional configuration. We can address this in a separate PR.

moshe010 (approved):

zshi-redhat (approved):

PR #51, opened 2021-01-23
add CLUSTER_TYPE=kubernetes to deploy-setup-k8s

moshe010:
Signed-off-by: Moshe Levi <moshele@nvidia.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,6 +193,7 @@ deploy-setup-k8s: export NAMESPACE=sriov-network-operator
 deploy-setup-k8s: export ENABLE_ADMISSION_CONTROLLER=false
 deploy-setup-k8s: export CNI_BIN_PATH=/opt/cni/bin
 deploy-setup-k8s: export OPERATOR_EXEC=kubectl
+deploy-setup-k8s: export CLUSTER_TYPE=kubernetes
 deploy-setup-k8s: deploy-setup
 
 test-e2e-conformance:


adrianchiris (approved):

zshi-redhat:
Does this work w/o passing environment var to container/pod?

moshe010:
@zshi-redhat I don't understand the question we need this so that the SR-IOV operator will run in k8s mode and not openshift. 
In openshift it doing sync MCO because of the switchdev support. see [1] 

[1] - https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L639

zshi-redhat:
> @zshi-redhat I don't understand the question we need this so that the SR-IOV operator will run in k8s mode and not openshift.
> In openshift it doing sync MCO because of the switchdev support. see [1]
> 
> [1] - https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L639

I understood why we need the k8s cluster type. 
I was just wondering if the change in this PR is sufficient.

For openshift, the CLUSTER_TYPE environment variable is set in the [dockerfile](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/Dockerfile.rhel7#L16) and retrieved from [here](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L42), I didn't see how `export CLUSTER_TYPE=kubernetes` in Makefile can change the environment variable inside operator/config-daemon pods.

moshe010:
@zshi-redhat I understand you point. It seem like I didn't fix the problem :(. I will take close look on my setup to see what happened. 
Also can you explain why the Dockerfile.rhel7 are openshift? can we rename to  Dockerfile.openshift?

zshi-redhat:
> @zshi-redhat I understand you point. It seem like I didn't fix the problem :(. I will take close look on my setup to see what happened.
> Also can you explain why the Dockerfile.rhel7 are openshift? can we rename to Dockerfile.openshift?

It (.rhel7 suffix) is the naming convention used in openshift image build system, to differentiate from other Dockerfiles. If we change the name, will also need to sync the change to downstream openshift image build repo. (but it can be renamed to Dockerfile.openshift, it just needs extra step in openshift downstream to fix the image build repo)

In my opinion, we probably can remove Dockerfile.rhel7 from npwg repo.
/cc @pliurh

pliurh:
The .rhel7 is a historical naming convention when we used RHEL7 as the base image for downstream openshift builds, and CentOS for origin ones.

Personally, I prefer we keep both Dockerfiles. As there is also a community version openshift, for which users may want to have the .rhel7 Dockerfile to build the images.

Issue #48, opened 2021-01-18
sriov-network-config-daemon deployment failed with error: "failed to load kubelet kubeconfig: open /host/etc/kubernetes/kubeconfig: no such file or directory"


vtlrazin:
Deployment operator failed on sriov-network-config-daemon deployment stage with error:

`failed to load kubelet kubeconfig: open /host/etc/kubernetes/kubeconfig: no such file or directory`

Setup info:

OS - Ubuntu 18.04
Vanilla K8s - 1.17.5
SRIOV operator - github master

K8s cluster pods:

```
# kubectl -n sriov-network-operator get pod -o wide
NAME                                      READY   STATUS             RESTARTS   AGE     IP                NODE    NOMINATED NODE   READINESS GATES
sriov-network-config-daemon-5n9rv         0/1     CrashLoopBackOff   6          9m10s   192.168.222.102   node3   <none>           <none>
sriov-network-config-daemon-twvsm         0/1     CrashLoopBackOff   6          9m10s   192.168.222.103   node4   <none>           <none>
sriov-network-operator-56cb9ccd4d-wljvs   1/1     Running            0          9m27s   10.233.90.9       node1   <none>           <none>

```

Error log:

```
# kubectl -n sriov-network-operator describe pod sriov-network-config-daemon-twvsm
...
Events:
  Type     Reason     Age               From               Message
  ----     ------     ----              ----               -------
  Normal   Scheduled  82s               default-scheduler  Successfully assigned sriov-network-operator/sriov-network-config-daemon-twvsm to node4
  Normal   Pulling    81s               kubelet, node4     Pulling image "quay.io/openshift/origin-sriov-network-config-daemon@sha256:49a69724cd04d0264e8be94ce12893b1e4ecaf091bf4eaea02879c0e559aabf0"
  Normal   Pulled     49s               kubelet, node4     Successfully pulled image "quay.io/openshift/origin-sriov-network-config-daemon@sha256:49a69724cd04d0264e8be94ce12893b1e4ecaf091bf4eaea02879c0e559aabf0"
  Normal   Created    7s (x4 over 48s)  kubelet, node4     Created container sriov-network-config-daemon
  Normal   Started    7s (x4 over 48s)  kubelet, node4     Started container sriov-network-config-daemon
[...]

PR #52, opened 2021-01-27
Add missing CLUSTER_TYPE=kubernetes to vanilla k8s images

moshe010:
PR 51 try to fix it by adding the CLUSTER_TYPE=kubernetes in Makefile
which is wrong. The CLUSTER_TYPE=kubernetes should be exposed in the
sriov-operator and sriov-network-config-daemon Pods so that they will
behave according to the cluster type

This commit fix it by removing the CLUSTER_TYPE=kubernetes from the Makefile
and add it to sriov-operator and .sriov-network-config-daemon Dockerfile.

Fixes: #48
Signed-off-by: Moshe Levi <moshele@nvidia.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,4 +7,5 @@ FROM centos:7
 COPY --from=builder /go/src/github.com/k8snetworkplumbingwg/sriov-network-operator/build/_output/cmd/manager /usr/bin/sriov-network-operator
 COPY bindata /bindata
 ENV OPERATOR_NAME=sriov-network-operator
+ENV CLUSTER_TYPE=kubernetes
 CMD ["/usr/bin/sriov-network-operator"]
--- a/Dockerfile.sriov-network-config-daemon
+++ b/Dockerfile.sriov-network-config-daemon
@@ -12,4 +12,5 @@ COPY --from=builder /go/src/github.com/k8snetworkplumbingwg/sriov-network-operat
 COPY --from=builder /go/src/github.com/k8snetworkplumbingwg/sriov-network-operator/build/_output/pkg/plugins /plugins
 COPY bindata /bindata
 ENV PLUGINSPATH=/plugins
+ENV CLUSTER_TYPE=kubernetes
 CMD ["/usr/bin/sriov-network-config-daemon"]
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -7,4 +7,5 @@ FROM centos:7
 COPY --from=builder /go/src/github.com/k8snetworkplumbingwg/sriov-network-operator/build/_output/cmd/manager /usr/bin/sriov-network-operator
 COPY bindata /bindata
 ENV OPERATOR_NAME=sriov-network-operator
+CLUSTER_TYPE=kubernetes
```
pliurh, line 10: ENV CLUSTER_TYPE=kubernetes

Dockerfile.sriov-network-config-daemon
```diff
@@ -12,4 +12,5 @@ COPY --from=builder /go/src/github.com/k8snetworkplumbingwg/sriov-network-operat
 COPY --from=builder /go/src/github.com/k8snetworkplumbingwg/sriov-network-operator/build/_output/pkg/plugins /plugins
 COPY bindata /bindata
 ENV PLUGINSPATH=/plugins
+CLUSTER_TYPE=kubernetes
```
pliurh, line 15: ditto

moshe010:
sorry about that :(

zshi-redhat:
nit: is it because operator webhook is not used in vanilla k8s so Dockerfile.webhook is not updated?

btw, is there a need to change CLUSTER_TYPE via env variable? for example, current CLUSTER_TYPE is hard coded in Dockerfile/image, so an image is exclusively used for k8s or openshift. With env variable support, we can change the CLUSTER_TYPE on the fly and use openshift images on k8s or vice verse.

moshe010:
Is the webhook using this flag? I can try on my setup and check.
Maybe we should remove this ENV CLUSTER_TYPE and detect the cluster is self when we start to run the daemon.
Is there a way to detect openshift cluster?  if you have a specific label in openshift that sriov-operator can detect  and pass it to sriov-demon config.

zshi-redhat:
> Is the webhook using this flag? I can try on my setup and check.

Not yet, but I'm expecting there needs to be some check in the future. e.g. whether a device is switchdev mode capable, IB linkType should not be configured together with switchdev mode(?) etc

> Maybe we should remove this ENV CLUSTER_TYPE and detect the cluster is self when we start to run the daemon.
> Is there a way to detect openshift cluster? if you have a specific label in openshift that sriov-operator can detect and pass it to sriov-demon config.

A quick check on the node annotations/labels, looks like the existence of `node.openshift.io/os_id` label is unique on openshift.
I'm not sure if there are other reliable way to check the cluster is openshift.

moshe010:
the set Set CLUSTER_TYPE env variable in manifests it better approach. I just wonder if we should do autodetect the cluster ...

PR #58, opened 2021-02-08
Skip adding linkType selector for vfio-pci devices

zshi-redhat:
LinkType is not detectable for eth vfio-pci device using
netlink library which results in failure when exposing
vfio-pci resources.


moshe010 (commented):
@zshi-redhat can we just skip the vfio case instead the all Ethernet? Selector just for ib network will not solve the original issue we wanted to address

zshi-redhat:
> @zshi-redhat can we just skip the vfio case instead the all Ethernet? Selector just for ib network will not solve the original issue we wanted to address

Done

moshe010 (commented):
Can you update the commit msg to reflect the recent change?

zshi-redhat:
> Can you update the commit msg to reflect the recent change?

Good catch! Done.

moshe010 (approved):

moshe010 (approved):

PR #63, opened 2021-02-11
Fix bindata/scripts/load-udev.sh to run on ubuntu

Mmduh-483:
RHEL/centos has udevadm in /usr/sbin/udevadm but ubuntu udevadm is located in /usr/bin/udevadm which fails the script

### DIFF
--- a/bindata/scripts/load-udev.sh
+++ b/bindata/scripts/load-udev.sh
@@ -1,7 +1,25 @@
 #!/bin/bash
 
-echo "Reload udev rules: /usr/sbin/udevadm control --reload-rules"
-chroot /host/ /usr/sbin/udevadm control --reload-rules
+REDHAT_RELEASE_FILE="/host/etc/redhat-release"
+udevadm_bin=""
 
-echo "Trigger udev event: /usr/sbin/udevadm trigger --action add --attr-match subsystem=net"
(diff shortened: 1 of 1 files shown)


bindata/scripts/load-udev.sh
```diff
@@ -1,7 +1,7 @@
...
 
-echo "Reload udev rules: /usr/sbin/udevadm control --reload-rules"
-chroot /host/ /usr/sbin/udevadm control --reload-rules
+echo "Reload udev rules: udevadm control --reload-rules"
+chroot /host/ udevadm control --reload-rules
 
-echo "Trigger udev event: /usr/sbin/udevadm trigger --action add --attr-match subsystem=net"
-chroot /host/ /usr/sbin/udevadm trigger --action add --attr-match subsystem=net
+echo "Trigger udev event: udevadm trigger --action add --attr-match subsystem=net"
+chroot /host/ udevadm trigger --action add --attr-match subsystem=net
```
moshe010, line 7: it bad practice to remove the full path. We don't know how the PATH environment variable is setup. It is better to define the full path per OS

moshe010 (changes requested):

bindata/scripts/load-udev.sh
Mmduh-483: updated

moshe010 (approved):

zshi-redhat (approved):

PR #65, opened 2021-02-12
Allow modifying RESOURCE_PREFIX

manuelbuil:
RESOURCE_PREFIX is hardcoded to openshift.io, it would be nice to
change that if user wants

Signed-off-by: Manuel Buil <mbuil@suse.com>


martinkennelly (approved):
LGTM. Thanks for the change.

moshe010 (approved):

PR #67, opened 2021-02-16
Allow to modify CGO_ENABLED

manuelbuil:
Signed-off-by: Manuel Buil <mbuil@suse.com>

### DIFF
--- a/hack/build-go.sh
+++ b/hack/build-go.sh
@@ -32,7 +32,7 @@ fi
 
 mkdir -p ${BIN_PATH}
 
-CGO_ENABLED=1
+CGO_ENABLED=${CGO_ENABLED:-1}
 
 
 if [[ ${WHAT} == "manager" ]]; then


moshe010:
why do you want to do that? as far as I remember CGO_ENABLED=1 is required to plugin shared object.

manuelbuil:
> why do you want to do that? as far as I remember CGO_ENABLED=1 is required to plugin shared object.

Because if that is enabled, the resulting binary is not static anymore and depends on the following libraries:
```
	linux-vdso.so.1 (0x00007fff9e2b0000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fa5f5abd000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fa5f5702000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fa5f5cdc000)
```
As a result, the binary can't be run on slim images such as alpine or UBI.

As far as I understand, the C libraries are used for the net package but if CGO_ENABLED=0, then Go client is used for that

manuelbuil:
Another example: _build-webhook does not require CGO and without this PR, it is forced to use that and results in a dynamically linked binary

zshi-redhat (approved):

PR #76, opened 2021-02-25
Add support for helm deployment

Mmduh-483:


Mmduh-483:
need to add ReadMe

deployment/sriov-network-operator/values.yaml
```diff
@@ -0,0 +1,23 @@
...
+  nameOverride: ""
+  fullnameOverride: ""
+  resourcePrefix: "openshift.io"
+  enableAdmissionController: false
+  cniBinPath: "/opt/cni/bin"
+  clusterType: "kubernetes"
+
+# Image URIs for sriov-network-operator components
+images:
+  operator: quay.io/openshift/origin-sriov-network-operator
```
pliurh, line 17: It looks good to me. But we shall use the images build for vanilla k8s here. So I suggest we wait until #83 merged.

PR #71, opened 2021-02-19
Updating the hardware compatibility list

ptrnull:
Updating the quickstart document to have the compatible device list aligned with what is in the code and https://docs.openshift.com/container-platform/4.6/networking/hardware_networks/about-sriov.html#supported-devices_about-sriov

PR #85, opened 2021-03-08
Add additional supported MLNX device IDs

adrianchiris:
This commit adds the following NIC modes to the
supported devices map:

 1. ConnectX5 Ex
 2. ConnectX6 Dx

Signed-off-by: Adrian Chiris <adrianc@nvidia.com>

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -43,10 +43,12 @@ var NicIdMap = []string{
 	"8086 1592 1889", // Columbiaville E810-CQDA2/2CQDA2
 	"8086 1593 1889", // Columbiaville E810-XXVDA4
 	"8086 159b 1889", // Columbiaville E810-XXVDA2
+	"15b3 1013 1014", // ConnectX-4
 	"15b3 1015 1016", // ConnectX-4LX
 	"15b3 1017 1018", // ConnectX-5, PCIe 3.0
+	"15b3 1019 101a", // ConnectX-5 Ex
 	"15b3 101b 101c", // ConnectX-6
-	"15b3 1013 1014", // ConnectX-4
+	"15b3 101d 101e", // ConnectX-6 Dx
 	"15b3 a2d6 101e", // MT42822 BlueField-2 integrated ConnectX-6 Dx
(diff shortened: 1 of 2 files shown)


adrianchiris:
I will update Quickstart guide once #71 is merged.
Or if this lands first then add these two devices to the PR.

These NICs are tested internally within nvidia.

In general i dont think we should keep an explicit list of devices as this list might get unbearingly long.
for Openshift we can allow to filter supported device according to configMap.

zshi-redhat:
> I will update Quickstart guide once #71 is merged.
> Or if this lands first then add these two devices to the PR.
> 
> These NICs are tested internally within nvidia.
> 
> In general i dont think we should keep an explicit list of devices as this list might get unbearingly long.

@adrianchiris just to understand, are you suggesting to remove the NicIdMap?

adrianchiris:
> @adrianchiris just to understand, are you suggesting to remove the NicIdMap?

Yes, This is something i'd like to discuss, however this does not relate to this PR.

zshi-redhat:
> > @adrianchiris just to understand, are you suggesting to remove the NicIdMap?
> 
> Yes, This is something i'd like to discuss, however this does not relate to this PR.

Ok, the NicIdMap is used for generating device plugin configmap and/or perhaps in other code path (operator-webhook).
If we can find a way to remove all those direct uses of NicIdMap, I agree that we could delete it and allows more NICs to be used with SR-IOV Operator directly.

adrianchiris:
> Ok, the NicIdMap is used for generating device plugin configmap and/or perhaps in other code path (operator-webhook).
If we can find a way to remove all those direct uses of NicIdMap, I agree that we could delete it and allows more NICs to be used with SR-IOV Operator directly.

Will open an issue to track !

zshi-redhat:
@adrianchiris would you like to update the doc in this PR, this one https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/71 is merged.

adrianchiris:
@zshi-redhat  Done.

Issue #82, opened 2021-03-07
Add github Actions to create and push project images to container registry


adrianchiris:
- Generate GH token and store it as a secret to be used by the workflows
- Create github actions for:
  - Build and push images on PR merge.
  - Build and push images on tag creation.

**Artifacts to be created:**
  - sriov-network-operator
  - sriov-network-operator-config-daemon
  - sriov-network-operator-webhook

adrianchiris:
@pliurh i think the PAT you generated does not have enough permissions (write permissions) in order to push images.
i will close this issue once we got our first set of images available.

pliurh:
Yes. I'll ask Doug to see how we can solve it. It looks like we need the PAT from the owner of the org.

pliurh:
With #88, the images can be built as expected.

adrianchiris:
according to github docs a PAT is needed for ghcr.

https://docs.github.com/en/packages/guides/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry

anyway, i see we got our first images !

this can now be closed

PR #88, opened 2021-03-10
[AMEND] Use GITHUB_TOKEN to access github container registry

pliurh:


pliurh:
fix #82

adrianchiris:
Hmm, the reason i used a PAT is because of :

https://docs.github.com/en/packages/guides/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry

```
If you want to authenticate to GitHub Container Registry in a GitHub Actions workflow, then you must use a personal access token (PAT). The GITHUB_TOKEN does not currently have the required permissions. During the GitHub Container Registry beta, the only supported form of authentication is the PAT.
```

if it works, then docs must be out of date :)

PR #91, opened 2021-03-16
Replace nodeSelector with nodeAffinity master role

jcaamano:
Label node-role.kubernetes.io/master has traditionally no value
associated inherited from kubeadm deplyments but some platforms use
"true" as value. Use node affinity instead of node selector ignoring the
value of the label.

### DIFF
--- a/bindata/manifests/operator-webhook/server.yaml
+++ b/bindata/manifests/operator-webhook/server.yaml
@@ -24,7 +24,13 @@ spec:
       serviceAccountName: operator-webhook-sa
       nodeSelector:
         beta.kubernetes.io/os: linux
-        node-role.kubernetes.io/master:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: Exists
       tolerations:
       - key: "node-role.kubernetes.io/master"
         operator: Exists
--- a/bindata/manifests/webhook/server.yaml
+++ b/bindata/manifests/webhook/server.yaml
@@ -27,7 +27,13 @@ spec:
       serviceAccountName: network-resources-injector-sa
       nodeSelector:
         beta.kubernetes.io/os: linux
-        node-role.kubernetes.io/master:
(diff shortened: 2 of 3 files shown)


martinkennelly (approved):

zshi-redhat:
do we have the same issue for node-role.kubernetes.io/worker ?

jcaamano:
> do we have the same issue for node-role.kubernetes.io/worker ?

No to the same extent. `node-role.kubernetes.io/worker` label is not standard, kubeadm does not set it for example. I guess openshift does. But it is being replaced by nodeselector from node policies and/or operator config so not really effective.

zshi-redhat:
It might be good to do the same change for webhooks (webhook and operator-webhook) which run on master nodes (although they are not supported to run on k8s yet).

zshi-redhat:
> > do we have the same issue for node-role.kubernetes.io/worker ?
> 
> No to the same extent. `node-role.kubernetes.io/worker` label is not standard, kubeadm does not set it for example. I guess openshift does. But it is being replaced by nodeselector from node policies and/or operator config so not really effective.

ack+

manifests/4.7/sriov-network-operator.v4.7.0.clusterserviceversion.yaml
```diff
@@ -298,8 +298,13 @@ spec:
                 name: sriov-network-operator
             spec:
               serviceAccountName: sriov-network-operator
-              nodeSelector:
-                node-role.kubernetes.io/master: ""
+              affinity:
```
pliurh, line 301: This is for openshift, we don't maintain this file upstream.

SchSeba: +1 please drop this file change

jcaamano: Done!

jcaamano:
> It might be good to do the same change for webhooks (webhook and operator-webhook) which run on master nodes (although they are not supported to run on k8s yet).

This is already included in this PR or am I missing something else?

e0ne (approved):
LGTM

PR #93, opened 2021-03-23
Pause MCP before draining/rebooting node

pliurh:
The config daemon instance on each node need to compete for the 'Draining' lock before
it can draining/rebooting a node.

After a config daemon instance gets the lock, it checks its MachineConfigPool conditions.
If the MCP is in ready state, it will pause the MCP then process, otherwise, it wait for
the MCP getting ready. The MCP will be resumed after the config daemon and release the
lock after the config dameon finishes its work.

### DIFF
--- a/deploy/clusterrole.yaml
+++ b/deploy/clusterrole.yaml
@@ -48,3 +48,6 @@ rules:
 - apiGroups: [""]
   resources: ["pods/eviction"]
   verbs: ["create"]
+- apiGroups: ["machineconfiguration.openshift.io"]
+  resources: ["*"]
+  verbs: ["*"]
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -40,6 +40,9 @@ import (
 	snclientset "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/clientset/versioned"
 	sninformer "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/informers/externalversions"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	mcclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
+	mcfginformers "github.com/openshift/machine-config-operator/pkg/generated/informers/externalversions"
 )
 
 const (
@@ -68,6 +71,8 @@ type Daemon struct {
 	client snclientset.Interface
 	// kubeClient allows interaction with Kubernetes, including the node we are running on.
 	kubeClient *kubernetes.Clientset
+
+	mcClient *mcclientset.Clientset
 
 	nodeState *sriovnetworkv1.SriovNetworkNodeState
 
@@ -98,6 +103,8 @@ type Daemon struct {
 	nodeLister listerv1.NodeLister
 
 	workqueue workqueue.RateLimitingInterface
+
+	mcpName string
 }
 
 type workItem struct {
@@ -110,6 +117,7 @@ const (
 	annoKey         = "sriovnetwork.openshift.io/state"
 	annoIdle        = "Idle"
 	annoDraining    = "Draining"
+	annoMcpPaused   = "Draining_MCP_Paused"
 )
 
 var namespace = os.Getenv("NAMESPACE")
@@ -130,6 +138,7 @@ func New(
 	nodeName string,
 	client snclientset.Interface,
 	kubeClient *kubernetes.Clientset,
+	mcClient *mcclientset.Clientset,
 	exitCh chan<- error,
 	stopCh <-chan struct{},
 	syncCh <-chan struct{},
@@ -141,11 +150,11 @@ func New(
 		platform:   platformType,
 		client:     client,
 		kubeClient: kubeClient,
+		mcClient:   mcClient,
 		exitCh:     exitCh,
 		stopCh:     stopCh,
 		syncCh:     syncCh,
 		refreshCh:  refreshCh,
-		drainable:  true,
 		nodeState:  &sriovnetworkv1.SriovNetworkNodeState{},
 		drainer: &drain.Helper{
 			Client:              kubeClient,
@@ -413,13 +422,11 @@ func (dn *Daemon) nodeUpdateHandler(old, new interface{}) {
 		return
 	}
 	for _, node := range nodes {
-		if node.GetName() != dn.name && node.Annotations[annoKey] == annoDraining {
-			glog.V(2).Infof("nodeUpdateHandler(): node %s is draining", node.Name)
+		if node.GetName() != dn.name && (node.Annotations[annoKey] == annoDraining || node.Annotations[annoKey] == annoMcpPaused) {
 			dn.drainable = false
 			return
 		}
 	}
-	glog.V(2).Infof("nodeUpdateHandler(): no other node is draining")
 	dn.drainable = true
 }
 
@@ -522,6 +529,10 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 		}
 		return nil
 	}
+	if err = dn.getNodeMachinePool(); err != nil {
+		return err
+	}
+
 	if reqDrain && !dn.disableDrain {
 		glog.Info("nodeStateSyncHandler(): drain node")
 		if err := dn.drainNode(dn.name); err != nil {
@@ -553,7 +564,7 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 		}
 	}
 
-	if anno, ok := dn.node.Annotations[annoKey]; ok && anno == annoDraining {
+	if anno, ok := dn.node.Annotations[annoKey]; ok && (anno == annoDraining || anno == annoMcpPaused) {
 		if err := dn.completeDrain(); err != nil {
 			glog.Errorf("nodeStateSyncHandler(): failed to complete draining: %v", err)
 			return err
@@ -580,8 +591,17 @@ func (dn *Daemon) completeDrain() error {
 		return err
 	}
 
+	if utils.ClusterType == utils.ClusterTypeOpenshift {
+		glog.Infof("completeDrain(): resume MCP %s", dn.mcpName)
+		pausePatch := []byte("{\"spec\":{\"paused\":false}}")
+		if _, err := dn.mcClient.MachineconfigurationV1().MachineConfigPools().Patch(context.Background(), dn.mcpName, types.MergePatchType, pausePatch, metav1.PatchOptions{}); err != nil {
+			glog.Errorf("completeDrain(): failed to resume MCP %s: %v", dn.mcpName, err)
+			return err
+		}
+	}
+
 	if err := dn.annotateNode(dn.name, annoIdle); err != nil {
-		glog.Errorf("drainNode(): failed to annotate node: %v", err)
+		glog.Errorf("completeDrain(): failed to annotate node: %v", err)
 		return err
 	}
 	return nil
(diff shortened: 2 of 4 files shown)


pkg/daemon/daemon.go
```diff
@@ -750,18 +764,84 @@ func (dn *Daemon) annotateNode(node, value string) error {
...
+
+		ctx, cancel := context.WithCancel(context.TODO())
+		defer cancel()
+
+		mcpEventHanedler := func(obj interface{}) {
+			newMcp := obj.(*mcfgv1.MachineConfigPool)
+			if newMcp.GetName() != dn.mcpName {
+				return
+			}
+			if mcfgv1.IsMachineConfigPoolConditionFalse(newMcp.Status.Conditions, mcfgv1.MachineConfigPoolDegraded) &&
```
zshi-redhat, line 820: When node gets rebooted by sriov config daemon, will the `drainNode` be called again (in which case MCP is degraded due to pause:true)?

pkg/daemon/daemon.go
```diff
@@ -750,18 +764,84 @@ func (dn *Daemon) annotateNode(node, value string) error {
...
+				mcfgv1.IsMachineConfigPoolConditionTrue(newMcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdated) &&
+				mcfgv1.IsMachineConfigPoolConditionFalse(newMcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdating) {
+				glog.Infof("drainNode(): MCP %s is ready", newMcp.GetName())
```
zshi-redhat, line 823: Check `readyMachineCount` == `machineCount` as well ?

pkg/daemon/daemon.go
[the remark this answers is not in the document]
zshi-redhat: @yuqi-zhang @sinnykumari, do you think the checks here can 100% guarantee there is no race condition between MCO and SNO trying to start an update simultaneously?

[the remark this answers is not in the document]
pliurh: > By checking poolDegraded, it makes sure pool is not paused by user, right?

No, this is to check MCP is not in 'Degraded' state.

pliurh: Pausing an MCP will not make it 'Degraded'.

pkg/daemon/daemon.go
pliurh: I suppose the 'Updated' condition of MCP is sufficient.

pkg/daemon/daemon.go
```diff
@@ -580,8 +590,17 @@ func (dn *Daemon) completeDrain() error {
 		return err
 	}
 
+	if utils.ClusterType == utils.ClusterTypeOpenshift {
+		glog.Infof("completeDrain(): resume MCP %s", dn.mcpName)
```
yuqi-zhang, line 594: Making sure I understand this properly, the completeDrain() function gets called after the reboot correct? To uncordon the node?

pkg/daemon/daemon.go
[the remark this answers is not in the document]
yuqi-zhang: Pausing and degraded are separate concepts. I think the SRIOV operator shouldn't do anything if a pool is degraded (which means the pool is in an error state), as you have here. I think the checks (not degraded, not updating, updated) should be good.

pkg/daemon/daemon.go
```diff
@@ -747,21 +766,93 @@ func (dn *Daemon) annotateNode(node, value string) error {
...
+			UpdateFunc: func(old, new interface{}) {
+				mcpEventHandler(new)
+			},
+		})
+		mcpInformerFactory.Start(ctx.Done())
+		mcpInformerFactory.WaitForCacheSync(ctx.Done())
+		<-ctx.Done()
+
+		glog.Infof("drainNode(): pause MCP %s", dn.mcpName)
+		pausePatch := []byte("{\"spec\":{\"paused\":true}}")
```
yuqi-zhang, line 850: if we want to ensure maximum safety and elimate races, after the pause goes through here, we should add another check like above to see if the pool is progressing, since there is a tiny chance that the MCP started some kind of update in between.

If we check here and the pool is still good, then we can progress with drain. If the pool somehow is updating now, we should consider unpausing and then waiting in a loop for the MCP to go back to not updating.

yuqi-zhang (commented):
Overall this approach I think is more safe with the MCO and race conditions. Some more specific comments below

pkg/daemon/daemon.go
pliurh: Yes. It will be executed after reboot or after draining without a reboot.

pkg/daemon/daemon.go
zshi-redhat: Do we also want to check that pause is not already set by user?

pliurh: I updated the patch, PTAL.

pkg/daemon/daemon.go
```diff
@@ -747,21 +766,116 @@ func (dn *Daemon) annotateNode(node, value string) error {
...
+				if err != nil {
+					glog.V(2).Infof("drainNode(): fail to pause MCP %s: %v", dn.mcpName, err)
+					return
+				}
+				paused = true
+				return
+			}
+			if paused {
+				glog.Infof("drainNode(): MCP is processing, resume MCP %s", dn.mcpName)
+				pausePatch := []byte("{\"spec\":{\"paused\":false}}")
```
sinnykumari, line 859: For my understanding, can you please explain the condition in which we are unpausing the pool here.

sinnykumari (commented):
Thanks for implementing it in this way! overall this approach looks good.

pkg/daemon/daemon.go
pliurh: The MCP informer will not be stoped right after we pause the MCP. So the mcpEventHandler will be invoked once more as we updated the MCD. In this cycle, we check if the MCP is processing, if yes and the MCP was paused during the last cycle (indicated by the variable 'paused' here). We resume the MCP and wait until it finishes.

pliurh: This is to address https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/93#discussion_r599989656

sinnykumari: got it, makes sense.

pkg/daemon/daemon.go
```diff
@@ -747,21 +766,116 @@ func (dn *Daemon) annotateNode(node, value string) error {
...
+			if mcfgv1.IsMachineConfigPoolConditionFalse(newMcp.Status.Conditions, mcfgv1.MachineConfigPoolDegraded) &&
+				mcfgv1.IsMachineConfigPoolConditionTrue(newMcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdated) &&
+				mcfgv1.IsMachineConfigPoolConditionFalse(newMcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdating) {
+				glog.V(2).Infof("drainNode(): MCP %s is ready", dn.mcpName)
+				if paused {
+					glog.V(2).Info("drainNode(): stop MCP informer", dn.mcpName)
+					cancel()
+					return
+				}
+				if newMcp.Spec.Paused {
```
SchSeba, line 843: will this be the case if the sriov operator drains the node and then reboot it to change for example the mlx max number of VFs?

SchSeba (commented):
looks good I just have small comments thanks!

pkg/daemon/daemon.go
```diff
@@ -747,21 +766,116 @@ func (dn *Daemon) annotateNode(node, value string) error {
 	return nil
 }
 
+func (dn *Daemon) getNodeMachinePool() error {
```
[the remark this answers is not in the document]
pliurh: hmm, I think the current approach is ok. @sinnykumari @yuqi-zhang?

deploy/clusterrole.yaml
```diff
@@ -48,3 +48,6 @@ rules:
 - apiGroups: [""]
   resources: ["pods/eviction"]
   verbs: ["create"]
+- apiGroups: ["machineconfiguration.openshift.io"]
```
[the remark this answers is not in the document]
pliurh: As that is our downstream bundle, I will update it in the openshift repo only.

pkg/daemon/daemon.go
pliurh: After reboot, the config daemon will not drain the node, so we will not hit this logic.

pkg/daemon/daemon.go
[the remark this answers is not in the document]
sinnykumari: yes, current approach is fine to determine corresponding pool for the node.

pkg/daemon/daemon.go
SchSeba: Hi @pliurh did you try this on an intel nic?

I am waiting to get an env but I think in general for intel

1. add the iommu in the kargs
2. drainNode code https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/f552cd8eb8990787a24a0504f633405adfd2af76/pkg/daemon/daemon.go#L535
3. reboot https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/f552cd8eb8990787a24a0504f633405adfd2af76/pkg/daemon/daemon.go#L551
4. when the node is up again the default plugin will request to drain again https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/f552cd8eb8990787a24a0504f633405adfd2af76/pkg/plugins/generic/generic_plugin.go#L86
5. to configure the numOfVfs https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/f552cd8eb8990787a24a0504f633405adfd2af76/pkg/plugins/generic/generic_plugin.go#L193

it will stuck because the mcp is already pause do I miss something?

pkg/daemon/daemon.go
```diff
@@ -580,8 +590,17 @@ func (dn *Daemon) completeDrain() error {
 		return err
 	}
 
+	if utils.ClusterType == utils.ClusterTypeOpenshift {
+		glog.Infof("completeDrain(): resume MCP %s", dn.mcpName)
+		pausePatch := []byte("{\"spec\":{\"paused\":false}}")
+		if _, err := dn.mcClient.MachineconfigurationV1().MachineConfigPools().Patch(context.Background(), dn.mcpName, types.MergePatchType, pausePatch, metav1.PatchOptions{}); err != nil {
```
zshi-redhat, line 596: After rebooting the node, dn.mcpName is empty which results in failure updating the api.
dn.mcpName is assigned in drainNode function, but drainNode is not always called after rebooting.

zshi-redhat (changes requested):

pkg/daemon/daemon.go
pliurh: The `dn.mcpName` is assigned in `getNodeMachinePool`, which will be invoked in `nodeStateSyncHandler`. So it will always be assigned before we consume it.

zshi-redhat: I'm testing in a baremetal environment, what I observed was:
```
E0401 01:01:52.621664    5940 daemon.go:592] completeDrain(): failed to resume MCP : resource name may not be empty
E0401 01:01:52.621674    5940 daemon.go:564] nodeStateSyncHandler(): failed to complete draining: resource name may not be empty
```
The mcp never comes back.

zshi-redhat: ignore this, I might use wrong build.

pkg/daemon/daemon.go
pliurh: In step5, It will not require to drain a node, as the `ifaceStatus.NumVfs` is always 0 when a node boots up.

However, there is still a case that can trigger the problem, when a non-default MTU is specified. I'll fix that.

pliurh:
@zshi-redhat @SchSeba I just updated the patch. A new annotation value `Draining_MCP_Paused` was introduced to indicate whether the MCP is paused by the config daemon. Also, I changed the `needDrainNode`, so that changing MTU will not need a drain if `ifaceStatus.NumVfs` is 0.

pkg/daemon/daemon.go
```diff
@@ -747,21 +767,127 @@ func (dn *Daemon) annotateNode(node, value string) error {
...
+				}
+				if newMcp.Spec.Paused {
+					glog.V(2).Infof("drainNode(): MCP %s was paused by other, wait...", dn.mcpName)
+					return
+				}
+				glog.Infof("drainNode(): pause MCP %s", dn.mcpName)
+				pausePatch := []byte("{\"spec\":{\"paused\":true}}")
+				_, err = dn.mcClient.MachineconfigurationV1().MachineConfigPools().Patch(context.Background(), dn.mcpName, types.MergePatchType, pausePatch, metav1.PatchOptions{})
+				if err != nil {
+					glog.V(2).Infof("drainNode(): fail to pause MCP %s: %v", dn.mcpName, err)
```
SchSeba, line 853: nit: `fail` -> `Failed`

pkg/daemon/daemon.go
```diff
@@ -747,21 +767,127 @@ func (dn *Daemon) annotateNode(node, value string) error {
...
+			glog.Infof("drainNode():MCP %s is not ready: %v, wait...", newMcp.GetName(), newMcp.Status.Conditions)
 		}
-		return dn.drainable, nil
-	}, dn.stopCh)
 
-	err = dn.annotateNode(dn.name, annoDraining)
-	if err != nil {
-		glog.Errorf("drainNode(): Failed to annotate node: %v", err)
-		return err
+		mcpInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
SchSeba, line 882: nit: can we add a check if this is the MCP that the node is part of even before we send it to the `mcpEventHandler` function?

SchSeba (commented):
just two small nits and I think the PR is ready

pkg/daemon/daemon.go
pliurh: Hmm, I don't know how we can do that. Normally, the informer watches all the objects. I think it's sufficient that we check it in the `mcpEventHandler`.

pkg/daemon/daemon.go
pliurh: fixed.

PR #96, opened 2021-03-29
DRAFT: Workaround for mellanox errors in kernel lockdown mode

mmirecki:


pkg/utils/utils.go
```diff
@@ -693,3 +701,45 @@ func isSwitchdev(name string) bool {
...
+	return false
+}
+
+// IsKernelLockdownMode returns true when kernel lockdown mode is enabled
+func IsKernelLockdownMode() bool {
+	// Check 2 path to allow using with chroot
+	for _, path := range []string{"/host/sys/kernel/security/lockdown", "/sys/kernel/security/lockdown"} {
+		out, err := RunCommand("cat", path)
+		glog.V(2).Infof("IsKernelLockdownMode(): %s, %+v", out, err)
+		if err == nil && strings.Contains(out, "[integrity]") {
```
yuvalk, line 725: actually both `[integrity]` and `[confidentiality]` are considered lockdown
so better search for anything other than `[none]`

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -70,6 +69,10 @@ func (p *MellanoxPlugin) OnNodeStateAdd(state *sriovnetworkv1.SriovNetworkNodeSt
 // OnNodeStateChange Invoked when SriovNetworkNodeState CR is updated, return if need dain and/or reboot node
 func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
 	glog.Info("mellanox-Plugin OnNodeStateChange()")
+	if utils.IsKernelLockdownMode() {
```
SchSeba, line 72: we should return and error here so we don't continue to the apply

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -173,6 +176,10 @@ func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetwork
 
 // Apply config change
 func (p *MellanoxPlugin) Apply() error {
+	if utils.IsKernelLockdownMode() {
```
SchSeba, line 179: we don't need this lets stop in the `OnNodeStateChange`

pkg/utils/utils.go
```diff
@@ -136,7 +138,13 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
 // SyncNodeState Attempt to update the node state to match the desired state
 //
 func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
+	if IsKernelLockdownMode() && hasMellanoxInterfacesInSpec(newState) {
```
SchSeba, line 141: I don't think we need it here

pkg/utils/utils.go
```diff
@@ -693,3 +701,45 @@ func isSwitchdev(name string) bool {
...
+			}
+		}
+	}
+	return false
+}
+
+// IsKernelLockdownMode returns true when kernel lockdown mode is enabled
+func IsKernelLockdownMode() bool {
+	// Check 2 path to allow using with chroot
+	for _, path := range []string{"/host/sys/kernel/security/lockdown", "/sys/kernel/security/lockdown"} {
```
SchSeba, line 722: I think we can drop the two links as we only check now outside the chroot

SchSeba (commented):
Thanks for the PR I add a few comments

SchSeba:
closing this one in favor of https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/103

/close

PR #97, opened 2021-03-30
Support test-e2e-k8s rule

AbdYsn:
This patch adds the support for the `test-e2e-k8` make rule by
making the operator namespace in the e2e tests configurable then
setting the namespace value in the Makefile.

### DIFF
--- a/test/e2e/e2e_tests_suite_test.go
+++ b/test/e2e/e2e_tests_suite_test.go
@@ -31,11 +31,10 @@ import (
 var cfg *rest.Config
 var k8sClient client.Client
 var testEnv *envtest.Environment
+var testNamespace string
 
 // Define utility constants for object names and testing timeouts/durations and intervals.
 const (
-	testNamespace = "openshift-sriov-network-operator"
-
 	timeout  = time.Second * 30
 	duration = time.Second * 300
 	interval = time.Second * 1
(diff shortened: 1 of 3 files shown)


AbdYsn:
@adrianchiris can you have a look? thanks

moshe010 (approved):

martinkennelly (approved):

test/e2e/e2e_tests_suite_test.go
```diff
@@ -34,13 +34,13 @@ var testEnv *envtest.Environment
...
 // Define utility constants for object names and testing timeouts/durations and intervals.
 const (
-	testNamespace = "openshift-sriov-network-operator"
-
 	timeout  = time.Second * 30
 	duration = time.Second * 300
 	interval = time.Second * 1
 )
 
+var testNamespace string = os.Getenv("NAMESPACE")
```
adrianchiris, line 42: Should we panic in the test if NAMESPACE is somehow not defined (i.e os.Gentenv() returns an empty string) ?

adrianchiris:
@AbdYsn a small comment, otherwise i think we are good to merge this one !

test/e2e/e2e_tests_suite_test.go
AbdYsn: Done

PR #105, opened 2021-04-10
Don't run k8s-plugin when CLUSTER_TYPE is OpenShift

zshi-redhat:

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -491,6 +491,9 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 	mcoReboot := false
 	for k, p := range dn.LoadedPlugins {
 		d, r := false, false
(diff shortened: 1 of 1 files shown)


Mmduh-483:
@zshi-redhat k8s plugin is not supposed to be loaded when the cluster is openshift https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L674-L678

zshi-redhat:
> @zshi-redhat k8s plugin is not supposed to be loaded when the cluster is openshift https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L674-L678

@Mmduh-483  you're right, I hit the issue with my customized patch in this PR: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/106

The issue doesn't happen on master.

PR #106, opened 2021-04-11
OVS HWOL: use generic plugin to write switchdev device configuration

zshi-redhat:
Added SriovOperatorConfig.Spec.OvsHardwareOffload API field to define Labels that matches with MachineConfigPool NodeSelector.  Nodes that's selected by the Labels will be applied with "bindata/manifest/switchdev-config".

Removed MCO plugins and moved the writing of switchdev device config file into generic plugin (shared by OpenShift and Kubernetes deployment). Removed the writing of switchdev device config file from K8s plugin.

For configuring OVS HWOL on OpenShift, the steps are changed to:

- Configured default SriovOperatorConfig to apply switchdev-config services to MachineConfigPool
- Configure SriovNetworkNodePolicy with eSwitchMode: switchdev (This writes switchdev.conf file and trigger drain/reboot where needed)

For configuring OVS HWOL on Kubernetes, the steps are not changed.


pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -127,24 +127,6 @@ func (p *K8sPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetworkNodeS
 		}
 	}
 
-	// Check switchdev config
```
Mmduh-483, line 130: K8s plugin is only checking the services now, can you reformat the condition on line 121 to be 
if !utils.IsSwitchdevModeSpec(new.Spec) {
    continue
}

Can you also remove the word "not" in the comment in line 120

pkg/daemon/daemon.go
```diff
@@ -500,35 +499,25 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
...
 		glog.V(0).Infof("nodeStateSyncHandler(): plugin %s: reqDrain %v, reqReboot %v", k, d, r)
 		reqDrain = reqDrain || d
 		reqReboot = reqReboot || r
 	}
 	glog.V(0).Infof("nodeStateSyncHandler(): reqDrain %v, reqReboot %v disableDrain %v", reqDrain, reqReboot, dn.disableDrain)
 
 	for k, p := range dn.LoadedPlugins {
-		if k != GenericPlugin && k != McoPlugin {
+		if k != GenericPlugin {
+			if k == K8sPlugin && utils.ClusterType == utils.ClusterTypeOpenshift {
```
Mmduh-483, line 510: Is there a reason why this condition needed for the k8s plugin and openshift cluster type?
 on line 663 you ensure that plugin is not loaded

Mmduh-483 (commented):
Did you test the code on k8s cluster?

pkg/plugins/k8s/k8s_plugin.go
zshi-redhat: @Mmduh-483 I thought we still need the `p.servicesStateUpdate()` to set the right flags or content for K8s_plugin struct, isn't it? Otherwise, in line `p.updateTarget.needUpdate()` , it will always be false.

Mmduh-483: In the current implementation:
- p.servicesStateUpdate runs when if condition `utils.IsSwitchdevModeSpec(new.Spec) ` is met, I suggest a simple refactoring for the if condtion

- At the start of `OnNOdeChange` p.updateTarget is reset to be all false `p.updateTarget.needUpdate()` is always false if there is no switchdev spec

zshi-redhat: So your suggestion is to change:
```
        if utils.IsSwitchdevModeSpec(new.Spec) {
                // Check services
                err = p.servicesStateUpdate()
                if err != nil {
                        glog.Errorf("k8s-plugin OnNodeStateChange(): failed : %v", err)
                        return
                }
        }
```
To:
```
        if !utils.IsSwitchdevModeSpec(new.Spec) {
                continue
        }
        err = p.servicesStateUpdate()
        if err != nil {
                glog.Errorf("k8s-plugin OnNodeStateChange(): failed : %v", err)
                return
        }
```
Is it correct?

pkg/daemon/daemon.go
zshi-redhat: Will remove, there is no need for this conditional check.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -212,3 +196,25 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
 	}
 	return
 }
+
+func needRebootNode(state *sriovnetworkv1.SriovNetworkNodeState, loadVfioDriver *uint) (needReboot bool) {
```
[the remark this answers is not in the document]
zshi-redhat: Will add.

zshi-redhat:
> Did you test the code on k8s cluster?

I didn't, I only tried on OpenShift.

pkg/plugins/k8s/k8s_plugin.go
Mmduh-483: yes

pkg/daemon/daemon.go
zshi-redhat: Removed.

pkg/plugins/k8s/k8s_plugin.go
zshi-redhat: Done.

pkg/plugins/generic/generic_plugin.go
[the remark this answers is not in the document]
zshi-redhat: Added.

Mmduh-483:
@zshi-redhat lgtm for me, can you verify it is working on k8s cluster?

zshi-redhat:
> @zshi-redhat lgtm for me, can you verify it is working on k8s cluster?

@Mmduh-483 I'd like to, but I don't have an env that can be easily deployed with k8s.

controllers/sriovoperatorconfig_controller.go
```diff
@@ -409,101 +410,103 @@ func (r *SriovOperatorConfigReconciler) syncK8sResource(cr *sriovnetworkv1.Sriov
...
+
+	for _, mcp := range mcpList.Items {
+		selector, err := metav1.LabelSelectorAsSelector(mcp.Spec.NodeSelector)
+		if err != nil {
+			return fmt.Errorf("Invalid label selector in MachineConfigPool: %s, %v", mcp.GetName(), err)
+		}
+		// MachineConfigPool is selected when its NodeSelectors are included in MachineConfigPoolSelector
+		for _, pool := range dc.Spec.OvsHardwareOffload {
+			if selector.Matches(labels.Set(pool.MachineConfigPoolSelector)) {
+				mcpMap[mcp.GetName()] = true
```
pliurh, line 436: Shall we store this information in the `status`, so that users can know which MCP is matched?

api/v1/sriovoperatorconfig_types.go
```diff
@@ -24,6 +24,15 @@ type SriovOperatorConfigSpec struct {
 	DisableDrain bool `json:"disableDrain,omitempty"`
 	// Flag to enable OVS hardware offload. Set to 'true' to provision switchdev-configuration.service and enable OpenvSwitch hw-offload on nodes.
 	EnableOvsOffload bool `json:"enableOvsOffload,omitempty"`
+	// OvsHardwareOffload describes the OVS HWOL configuration for selected MachineConfigPool
+	OvsHardwareOffload []OvsHardwareOffloadConfig `json:"ovsHardwareOffload,omitempty"`
```
pliurh, line 28: I haven't seen such an API design in K8S API.  Normally, if users need to select objects, they shall use a selector which matches the label of the objects. Personally, I think this implementation is hard to understand.

Shall we ask users to label the MCP by themself, and we use a selector here?

controllers/sriovoperatorconfig_controller.go
```diff
@@ -409,101 +410,103 @@ func (r *SriovOperatorConfigReconciler) syncK8sResource(cr *sriovnetworkv1.Sriov
...
-				err = r.Create(context.TODO(), mc)
-				if err != nil {
-					return fmt.Errorf("Couldn't create MachineConfig: %v", err)
-				}
-				logger.Info("Created MachineConfig CR")
-			}
-		} else {
-			return fmt.Errorf("Failed to get MachineConfig: %v", err)
+	for mcpName, enable := range mcpMap {
+		mcName := "00-" + mcpName + "-" + OVS_HWOL_MACHINE_CONFIG_NAME_SUFFIX
```
pliurh, line 446: Do we have to create one MC for each selected pool? Have you considered if `spec.machineConfigSelector` in MCP can be used instead?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -212,3 +196,31 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
...
+				glog.Errorf("generic-plugin needRebootNode():fail to enable iommu in kernel args: %v", err)
+			}
+			if update {
+				glog.V(2).Infof("generic-plugin needRebootNode(): need reboot for enabling iommu kernel args")
+			}
+			needReboot = needReboot || update
+		}
+	}
+
+	update, remove, err := utils.WriteSwitchdevConfFile(state)
```
pliurh, line 216: Why we need to distinguish  'remove' from 'update' here?

api/v1/sriovoperatorconfig_types.go
zshi-redhat: > I haven't seen such an API design in K8S API. Normally, if users need to select objects, they shall use a selector which matches the label of the objects. Personally, I think this implementation is hard to understand.

My previous thinking was that we may want to re-use the same API for k8s case in the future.
Current `OvsHardwareOffload.MachineConfigPoolSelector` matches on node labels, where for OpenShift, it matches the (node) labels defined in MCP `nodeSelector`, for k8s, it could match the node labels directly.
> 
> Shall we ask users to label the MCP by themself, and we use a selector here?

One idea I had before was to label or annotate MCP (with ovs-hw-offload) directly, then not change the SR-IOV API, instead, we can have SR-IOV Operator watch on MCP update for creating/deleting HWOL MCs. wdyt?

controllers/sriovoperatorconfig_controller.go
zshi-redhat: +1

controllers/sriovoperatorconfig_controller.go
zshi-redhat: Yes, with current API change, we probably can use single MC be matched by all MCP.
I was thinking for advanced cases that we might have HWOL configurations that differ between each MCP in the future (e.g. the tc_policy config), in which case, having one MC per MCP would be needed.

pkg/plugins/generic/generic_plugin.go
zshi-redhat: We don't, will update.

api/v1/sriovoperatorconfig_types.go
zshi-redhat: > > I haven't seen such an API design in K8S API. Normally, if users need to select objects, they shall use a selector which matches the label of the objects. Personally, I think this implementation is hard to understand.
> 
There is a similar example in tuned Operator, which was suggested by MCO team: https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/apis/tuned/v1/tuned_types.go#L79
It defines the MachineConfigLabels in Tuned API to match on MCPs.

pliurh: > My previous thinking was that we may want to re-use the same API for k8s case in the future.

If we plan to reuse it for k8s, the name `MachineConfigPoolSelector` is kind of misleading. How about just use `NodeSelector` instead?

> One idea I had before was to label or annotate MCP (with ovs-hw-offload) directly, then not change the SR-IOV API, instead, we can have SR-IOV Operator watch on MCP update for creating/deleting HWOL MCs. wdyt?

I prefer we can have a common API for both OCP and K8S. Otherwise, we still need a way to tell when the ovs hw offload shall be applied in the k8s case. Currently, it's hard bonded with the switchdev configuration.

controllers/sriovoperatorconfig_controller.go
```diff
@@ -409,101 +410,103 @@ func (r *SriovOperatorConfigReconciler) syncK8sResource(cr *sriovnetworkv1.Sriov
...
+				mcpMap[mcp.GetName()] = true
+				break
+			}
+		}
 	}
 
-	foundMC := &mcfgv1.MachineConfig{}
-	foundMCP := &mcfgv1.MachineConfigPool{}
+	// Offload is not supported for master MachineConfigPool
+	delete(mcpMap, "master")
```
pliurh, line 443: Instead of deletion, can we check the MCP name before adding to `mcpMap`, and log an error if users select the master pool?

controllers/sriovoperatorconfig_controller.go
zshi-redhat: Added.

api/v1/sriovoperatorconfig_types.go
zshi-redhat: Changed to `NodeSelector`

pkg/plugins/generic/generic_plugin.go
zshi-redhat: Done

controllers/sriovoperatorconfig_controller.go
zshi-redhat: Done

PR #108, opened 2021-04-12
fix unbind driver with permission denied when setVfsadminMac

arikachen:
Using intel 82599. 
When node reboot, sriov-network-config-daemon start with error log and vf mac does't be configed.
```
Unbind(): fail to unbind driver for device 0000:01:10.0. open /sys/bus/pci/drivers/ixgbevf/unbind: permission denied
```

Signed-off-by: chenqijun <chenqijun@corp.netease.com>

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -574,8 +574,10 @@ func setVfsAdminMac(iface *sriovnetworkv1.InterfaceExt) error {
 		if err := netlink.LinkSetVfHardwareAddr(pfLink, vfID, vfLink.Attrs().HardwareAddr); err != nil {
 			return err
 		}
-		if err = Unbind(addr); err != nil {
-			return err
+		if vfLink.Attrs().EncapType == "infiniband" {
+			if err = Unbind(addr); err != nil {
+				return err
+			}
 		}
 		if err = BindDefaultDriver(addr); err != nil {
 			return err


zshi-redhat:
@arikachen  there is a wait (10 secs) workaround before unbinding the driver: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L277-L285, shall we just increase the timeout in case it's not sufficient?

Btw, how long do you see the VF becomes ready after creation?

arikachen:
@zshi-redhat thanks for you relpy.
> @arikachen there is a wait (10 secs) workaround before unbinding the driver: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L277-L285, shall we just increase the timeout in case it's not sufficient?
> 
I think it is too late, vf mac will not be set again. Maybe we could check and set vf mac every loop?
> Btw, how long do you see the VF becomes ready after creation?
3 secs with 32 vf

zshi-redhat:
> @zshi-redhat thanks for you relpy.
> 
> > @arikachen there is a wait (10 secs) workaround before unbinding the driver: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L277-L285, shall we just increase the timeout in case it's not sufficient?
> 
> I think it is too late, vf mac will not be set again. Maybe we could check and set vf mac every loop?

Okay, I thought you were referring to the unbind call here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L288
But there is another unbind call inside setVfsAdminMac.
I also noticed that there is another wait loop, right before calling the unbind in setVfsAdminMac: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L569, it has 5 sec timeout. which means in your case, you have to wait 5 + 3 sec for all the VFs become ready, correct?

The unbind here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L577 was added to fix an infiniband GUID issue, maybe there are two ways to fix the issue:
1. increase timeout in vfIsReady
2. add conditional check for unbind in setVfsAdminMac (only apply to infiniband VF).

> 
> > Btw, how long do you see the VF becomes ready after creation?
> > 3 secs with 32 vf

arikachen:
> > @zshi-redhat thanks for you relpy.
> > > @arikachen there is a wait (10 secs) workaround before unbinding the driver: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L277-L285, shall we just increase the timeout in case it's not sufficient?
> > 
> > 
> > I think it is too late, vf mac will not be set again. Maybe we could check and set vf mac every loop?
> 
> Okay, I thought you were referring to the unbind call here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L288
> But there is another unbind call inside setVfsAdminMac.
> I also noticed that there is another wait loop, right before calling the unbind in setVfsAdminMac: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L569, it has 5 sec timeout. which means in your case, you have to wait 5 + 3 sec for all the VFs become ready, correct?
> 
all VFs become ready within 3 secs total.
> The unbind here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L577 was added to fix an infiniband GUID issue, maybe there are two ways to fix this issue:
> 
> 1. increase timeout in vfIsReady
> 2. add conditional check for unbind in setVfsAdminMac (only apply to infiniband VF).

It is ok by testing with no unbind step in setVfsAdminMac. How to check with infiniband vf? 
> 
> > > Btw, how long do you see the VF becomes ready after creation?
> > > 3 secs with 32 vf

zshi-redhat:
> > I also noticed that there is another wait loop, right before calling the unbind in setVfsAdminMac: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L569, it has 5 sec timeout. which means in your case, you have to wait 5 + 3 sec for all the VFs become ready, correct?
> 
> all VFs become ready within 3 secs total.

If it is 3 secs then I'd expect there is no issue since vfIsReady already waits for 5 secs.
> 
> > The unbind here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L577 was added to fix an infiniband GUID issue, maybe there are two ways to fix this issue:
> > 
> > 1. increase timeout in vfIsReady
> > 2. add conditional check for unbind in setVfsAdminMac (only apply to infiniband VF).
> 
> It is ok by testing with no unbind step in setVfsAdminMac. How to check with infiniband vf?
> 
There is a getLinkType function in pkg/utils/utils.go: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L588

arikachen:
> > > I also noticed that there is another wait loop, right before calling the unbind in setVfsAdminMac: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L569, it has 5 sec timeout. which means in your case, you have to wait 5 + 3 sec for all the VFs become ready, correct?
> > 
> > 
> > all VFs become ready within 3 secs total.
> 
> If it is 3 secs then I'd expect there is no issue since vfIsReady already waits for 5 secs.

yes, 5 secs is enough in my environment.
> 
> > > The unbind here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L577 was added to fix an infiniband GUID issue, maybe there are two ways to fix this issue:
> > > 
> > > 1. increase timeout in vfIsReady
> > > 2. add conditional check for unbind in setVfsAdminMac (only apply to infiniband VF).
> > 
> > 
> > It is ok by testing with no unbind step in setVfsAdminMac. How to check with infiniband vf?
> 
> There is a getLinkType function in pkg/utils/utils.go: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L588

thanks.

pliurh:
Do we still need this PR? Please close it if not.

arikachen:
> Please close it if not.

Yes, we need this. PTAL.

zshi-redhat:
It seems that we can remove the unbind and bind (which was added for infiniband VF device) in the setVfsAdminMac, since infiniband VF device is not configured in `setVfsGuid`.
@Mmduh-483 can you please take a look if the above is true?

adrianchiris:
> infiniband VF device is not configured in setVfsGuid

What is configured in setVfsGuid then ?

for ETH case, in `setVfsAdminMac` there is an unbind/bind so that admin MAC is applied which in turn allocates a node GUID to be used for RDMA over Ethernet .

zshi-redhat:
> > infiniband VF device is not configured in setVfsGuid
> 
> What is configured in setVfsGuid then ?
> 
> for ETH case, in `setVfsAdminMac` there is an unbind/bind so that admin MAC is applied which in turn allocates a node GUID to be used for RDMA over Ethernet .

I missed this part, so the unbind and bind is for RoCE devices, not IB devices; and we still need that until the kernel/driver fix (I didn't remember what kernel/driver fix is needed), right?
> 
> @moshe010 please correct me if im worng here.

pkg/utils/utils.go
```diff
@@ -574,8 +574,10 @@ func setVfsAdminMac(iface *sriovnetworkv1.InterfaceExt) error {
 		if err := netlink.LinkSetVfHardwareAddr(pfLink, vfID, vfLink.Attrs().HardwareAddr); err != nil {
 			return err
 		}
-		if err = Unbind(addr); err != nil {
-			return err
+		if vfLink.Attrs().EncapType == "infiniband" {
```
adrianchiris, line 577: this method is only called for ethernet link type, please see https://github.com/openshift/sriov-network-operator/blob/b2b495af316f49434b9ae841b701923cf629246d/pkg/utils/utils.go#L236

adrianchiris:
> I missed this part, so the unbind and bind is for RoCE devices, not IB devices; and we still need that until the kernel/driver fix (I didn't remember what kernel/driver fix is needed), right?

Yes indeed.



`configureSriovDevice` calls `setVfsGuid` for IB and `setVfsAdminMac` for ETH

each is uses to set the GUID depending on the link type. for ETH this is set implicitly when setting and appying the admin mac

pkg/utils/utils.go
arikachen: ok， should I close this pr or just left as a issue?

adrianchiris: This change will break RDMA over ethernet as driver unbind/bind is needed to set the administrative MAC which in turn applies the GUID.

I dont fully understand the issue (why intel nic VF fails to unbind). more information is appreciated. in addition it would be great if you can open an issue for tracking.

PR #110, opened 2021-04-13
Add management workload annotations

lack:
In support of the workload partitioning feature
(https://github.com/openshift/enhancements/pull/703), we need to add
annotations to all management pods and namespaces so they can be
properly identified and assigned to segregated management cores on
clusters configured to do so.


lack:
After some discussion with @zshi-redhat , this change will go in downstream only :)

PR #114, opened 2021-04-14
Add webhook certificate handling for k8s

jcaamano:
Adds support for webhooks in k8s deployments through options to handle
webhoook certificate configuration, either via providing the CA bundle
through environment variable or cert-manager annotations.

### DIFF
--- a/bindata/manifests/operator-webhook/003-webhook.yaml
+++ b/bindata/manifests/operator-webhook/003-webhook.yaml
@@ -5,7 +5,11 @@ metadata:
   name: {{.SRIOVMutatingWebhookName}}
   namespace: {{.Namespace}}
   annotations:
+    {{- if eq .ClusterType "openshift" }}
     service.beta.openshift.io/inject-cabundle: "true"
+    {{- else if and (not .CaBundle) (eq .ClusterType "kubernetes") }}
+    cert-manager.io/inject-ca-from: {{.Namespace}}/operator-webhook-service
+    {{- end }}
 webhooks:
   - name: operator-webhook.sriovnetwork.openshift.io
     failurePolicy: Fail
@@ -14,6 +18,9 @@ webhooks:
         name: operator-webhook-service
         namespace: {{.Namespace}}
         path: "/mutating-custom-resource"
+      {{- if and (.CaBundle) (eq .ClusterType "kubernetes") }}
+      caBundle: "{{.CaBundle}}"
+      {{- end}}
     rules:
       - operations: [ "CREATE", "UPDATE" ]
         apiGroups: ["sriovnetwork.openshift.io"]
@@ -27,7 +34,11 @@ metadata:
   name: {{.SRIOVMutatingWebhookName}}
   namespace: {{.Namespace}}
   annotations:
+    {{- if eq .ClusterType "openshift" }}
     service.beta.openshift.io/inject-cabundle: "true"
+    {{- else if and (not .CaBundle) (eq .ClusterType "kubernetes") }}
+    cert-manager.io/inject-ca-from: {{.Namespace}}/operator-webhook-service
+    {{- end }}
 webhooks:
   - name: operator-webhook.sriovnetwork.openshift.io
     failurePolicy: Fail
@@ -36,6 +47,9 @@ webhooks:
         name: operator-webhook-service
         namespace: {{.Namespace}}
         path: "/validating-custom-resource"
+      {{- if and (.CaBundle) (eq .ClusterType "kubernetes") }}
+      caBundle: "{{.CaBundle}}"
+      {{- end }}
     rules:
       - operations: [ "CREATE", "UPDATE", "DELETE" ]
         apiGroups: ["sriovnetwork.openshift.io"]
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -54,10 +54,71 @@ make deploy-setup
 
 If you are running a Kubernetes cluster:
 ```bash
-export OPERATOR_EXEC=kubectl
-export ENABLE_ADMISSION_CONTROLLER=false
 make deploy-setup-k8s
 ```
+
+Webhooks are disabled when deploying on a Kubernetes cluster as per the instructions above. To enable webhooks on Kubernetes cluster, there are two options:
+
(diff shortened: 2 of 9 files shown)


doc/quickstart.md
```diff
@@ -54,11 +54,89 @@ make deploy-setup
...
 make deploy-setup-k8s
 ```
 
+Webhooks are disabled when deploying on a Kubernetes cluster as per the instructions above. To enable webhooks on Kubernetes cluster, there are two options:
+
+1. Create certificates for each of the two webhooks using a single CA whose cert you provide through an environment variable.
+
+   For example, given `cacert.pem`, `key.pem` and `cert.pem`:
+   ```bash
+   kubectl -n sriov-network-operator create secret tls operator-webhook-service --cert=cert.pem --key=key.pem
```
martinkennelly, line 66: This namespace wont exist if they have never launched the operator before.

doc/quickstart.md
```diff
@@ -54,11 +54,89 @@ make deploy-setup
...
+     dnsNames:
+     - network-resources-injector-service.sriov-network-operator.svc
+     issuerRef:
+       name: sriov-network-operator-ca
+   ```
+
+    And then deploy the operator:
+    ```bash
+    export ENABLE_ADMISSION_CONTROLLER=true
+    make deploy-setup-k8s
```
martinkennelly, line 137:
```
  Warning  FailedMount  40s                  kubelet            Unable to attach or mount volumes: unmounted volumes=[tls], unattached volumes=[tls network-resources-injector-sa-token-zwls5]: timed out waiting for the condition
  Warning  FailedMount  35s (x9 over 2m43s)  kubelet            MountVolume.SetUp failed for volume "tls" : secret "network-resources-injector-secret" not found
```
Got the following error when tried this method on a fresh k8 cluster. I made sure ```WEBHOOK_CA_BUNDLE```env var wasnt exported.

doc/quickstart.md
```diff
@@ -54,11 +54,89 @@ make deploy-setup
...
 ```
 
+Webhooks are disabled when deploying on a Kubernetes cluster as per the instructions above. To enable webhooks on Kubernetes cluster, there are two options:
+
+1. Create certificates for each of the two webhooks using a single CA whose cert you provide through an environment variable.
+
+   For example, given `cacert.pem`, `key.pem` and `cert.pem`:
+   ```bash
+   kubectl -n sriov-network-operator create secret tls operator-webhook-service --cert=cert.pem --key=key.pem
+   kubectl -n sriov-network-operator create secret tls network-resources-injector-secret --cert=cert.pem --key=key.pem
```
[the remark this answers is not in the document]
jcaamano: So this option is assuming the users are creating the certificates manually and cert-manager is not being used.

bindata/manifests/operator-webhook/003-webhook.yaml
```diff
@@ -5,7 +5,11 @@ metadata:
   name: {{.SRIOVMutatingWebhookName}}
   namespace: {{.Namespace}}
   annotations:
+    {{if and (not .CaBundle) (eq .ClusterType "openshift") }}
     service.beta.openshift.io/inject-cabundle: "true"
+    {{else if and (not .CaBundle) (eq .ClusterType "kubernetes") }}
+    cert-manager.io/inject-ca-from: {{.Namespace}}/operator-webhook-service
```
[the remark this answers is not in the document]
jcaamano: That requires creating the secrets manually with a specific annotation which would be an actual extra step on all the procedure. It's is primarily used when bootstrapping cert-manager itself. After that, cert-manager creates the secrets from certificates for you so all you can directly annotate the certificate and save that extra step.

doc/quickstart.md
jcaamano: Strange, this is working for me.

The secret `network-resources-injector-secret` should be created in namespace `sriov-network-operator` from the Certificate `network-resources-injector-service` if cert-manager is functioning correctly. Can you `describe`it a check what's its status? It is also strange that it happens for one of the webhooks and not the other. I wiull try it again on my side.

jcaamano: Tried again and working for me.

manuelbuil: It's working for me too on a fresh k8s install

manuelbuil (approved):

bindata/manifests/operator-webhook/003-webhook.yaml
```diff
@@ -14,6 +18,9 @@ webhooks:
         name: operator-webhook-service
         namespace: {{.Namespace}}
         path: "/mutating-custom-resource"
+      {{if .CaBundle}}
```
zshi-redhat, line 21: Could we check the "ClusterType" to be equals to "kubernetes" as well? Just in case `WEBHOOK_CA_BUNDLE` is accidentally set in "openshift" environment.

zshi-redhat: This would apply to other manifests in bindata.

jcaamano: The idea I had here is that passing the `CaBundle` would also be an option in Openshift. Thinking on situations where one would have a quick test or dev cluster where the ca operator is not deployed. This might not make sense though as I am not familiar with Openshift. If it does not, I can add the check for cluster type there as you mention. Let me know.

zshi-redhat: @jcaamano thanks for all the consideration!
We didn't do ca injection manually in openshift afaik, I think we could add it only for k8s at this moment.
If anything changes, we could remove the conditional check.

doc/quickstart.md
martinkennelly: worked on a fresh install. Thanks!

martinkennelly (approved):

bindata/manifests/operator-webhook/003-webhook.yaml
jacaru: @zshi-redhat I rewrote the conditions so that `CaBundle` is only considered for `ClusterType=kubernetes`. With `ClusterType=openshift`, `CaBundle` is ignored.

PR #121, opened 2021-04-29
Move RestoreNodeDrainState up in clean.All()

oribon:
Previously if the RestoreNodeDrainState was set to true but the test namespace didn't exist,
the restoration wouldn't happen because the func returns nil too early

### DIFF
--- a/test/util/clean/clean.go
+++ b/test/util/clean/clean.go
@@ -20,6 +20,12 @@ func All() error {
 		operatorNamespace = "openshift-sriov-network-operator"
 	}
 	clients := client.New("")
+	if RestoreNodeDrainState {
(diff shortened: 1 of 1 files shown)


test/util/clean/clean.go
```diff
@@ -15,15 +15,22 @@ var RestoreNodeDrainState bool
 // All cleans all the dangling resources created by conformance tests.
 // This includes pods, networks, policies and namespaces.
 func All() error {
+	var err error
```
SchSeba, line 18: please remove this just put

```
err := cluster.SetDisableNodeDrainState(clients, operatorNamespace, false)
```
and
```
err := namespaces.DeleteAndWait(clients, namespaces.Test, 5*time.Minute)
```

SchSeba (commented):
/lgtm
/cc @mmirecki

PR #123, opened 2021-05-06
Skip enabling hw-offload for ovs-vswitchd when containerized OpenVSwitch is used

Mmduh-483:
HW Offload needs to be enabled on OpenVSwitch which will fail if OpenVSwitch is deployed in a container.

This PR introduces a new flag `ContainerizedOvs` for `SriovOperatorConfig`, which makes the operator skip enabling hw-offload for ovs-vswitchd, in case OpenVSwitch is deployed in a container or maintained by 3rd party

### DIFF
--- a/api/v1/sriovoperatorconfig_types.go
+++ b/api/v1/sriovoperatorconfig_types.go
@@ -24,6 +24,8 @@ type SriovOperatorConfigSpec struct {
 	DisableDrain bool `json:"disableDrain,omitempty"`
 	// Flag to enable OVS hardware offload. Set to 'true' to provision switchdev-configuration.service and enable OpenvSwitch hw-offload on nodes.
 	EnableOvsOffload bool `json:"enableOvsOffload,omitempty"`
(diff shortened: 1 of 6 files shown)


Mmduh-483:
@zshi-redhat @pliurh Can you please provide review

zshi-redhat:
@Mmduh-483 Do you see the containerized-ovs be used in non-ovs-hwol cases ? I'm trying to create a separate CRD for sriov node specific configuration (e.g. https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/125/commits/9505309ca7e62809d1f7c6d60f5563641071ae18), wondering if it makes sense to put the containerized ovs flag in the SR-IOV node config CRD.

In containerized ovs case, who will configure the hw-offload flag? If I understand correctly, we just skip the ovs-vswitchd systemd configuration in SR-IOV Operator.

jjuele:

SchSeba:
I don't think this PR is still active
I will close it please reopen it if this is needed

PR #124, opened 2021-05-06
Modify downwardAPI tests to reflect changes in the sriov injector

mmirecki:
The sriov injector only injects the label/annotation fields into the downwardAPI volume if labels/annotations are actually present: 
[here](https://github.com/openshift/sriov-dp-admission-controller/blob/master/pkg/webhook/webhook.go#L411)
This PR modifies the tests to adapt to the new behaviour


SchSeba:
Hi @mmirecki can you please check the unit test that is failing?

mmirecki:
> Hi @mmirecki can you please check the unit test that is failing?

fixed

SchSeba (approved):
/lgtm

PR #130, opened 2021-05-08
Refactor node draining racing avoid condition

pliurh:
Utilize k8s leader election mechanism to prevent annotating nodes
at the same time.


pkg/daemon/daemon.go
```diff
@@ -791,27 +791,60 @@ func (dn *Daemon) getNodeMachinePool() error {
...
-		err = dn.annotateNode(dn.name, annoDraining)
-		if err != nil {
-			glog.Errorf("drainNode(): Failed to annotate node: %v", err)
-			return
-		}
-		cancel()
-	}, 3*time.Second, 3, true, ctx.Done())
+
+	done := make(chan bool)
+	go dn.getDrainLock(ctx, done)
```
zshi-redhat, line 846: Can we run `dn.getDrainLock` in foreground and not use `done` chan? 
I assume `leaderelection.RunOrDie` is a loop w/o timeout, is it true?

pkg/daemon/daemon.go
```diff
@@ -791,27 +791,60 @@ func (dn *Daemon) getNodeMachinePool() error {
...
+			Namespace: namespace,
+		},
+		Client: dn.kubeClient.CoordinationV1(),
+		LockConfig: resourcelock.ResourceLockConfig{
+			Identity: dn.name,
+		},
+	}
+
+	// start the leader election
+	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
```
adrianchiris, line 809: looking at leaderelection [docs](https://pkg.go.dev/k8s.io/client-go@v0.19.10/tools/leaderelection):

> Package leaderelection implements leader election of a set of endpoints. It uses an annotation in the endpoints object to store the record of the election state. This implementation does not guarantee that only one client is acting as a leader (a.k.a. fencing).

is it not an issue ?
how will the system behave when there is only one endpoint trying to take the lead on LeaseLock ?

pkg/daemon/daemon.go
```diff
@@ -791,27 +791,60 @@ func (dn *Daemon) getNodeMachinePool() error {
...
+		RenewDeadline:   3 * time.Second,
+		RetryPeriod:     1 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				glog.V(2).Info("drainNode(): started leading")
+				for {
+					time.Sleep(1 * time.Second)
+					if dn.drainable {
+						glog.V(2).Info("drainNode(): no other node is draining")
+						err = dn.annotateNode(dn.name, annoDraining)
```
adrianchiris, line 822: using this mechanism, do we still need to annotate the node with `annotDraining` ?

pliurh: That is the trick. The leader election mechanism requires the leader to keep updating the Lease object. But in our case, the node may reboot itself, then lose leadership. So I use a 2 layers lock here. The node can only start draining with 2 conditions: 1) it becomes the leader 2) no other node is draining which is indicated by the annotation.

pkg/daemon/daemon.go
pliurh: The `leaderelection.RunOrDie` runs forever unless the context is canceled. I didn't run it in the foreground, because I want it keeps leading until the node finishes draining. So the rest nodes can only start the election after the prior one finishes draining if there is no reboot required.

pkg/daemon/daemon.go
adrianchiris: OK thanks for explaining

pkg/daemon/daemon.go
```diff
@@ -791,27 +791,60 @@ func (dn *Daemon) getNodeMachinePool() error {
...
+		ReleaseOnCancel: true,
+		LeaseDuration:   5 * time.Second,
+		RenewDeadline:   3 * time.Second,
+		RetryPeriod:     1 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				glog.V(2).Info("drainNode(): started leading")
+				for {
+					time.Sleep(1 * time.Second)
+					if dn.drainable {
```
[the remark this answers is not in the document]
pliurh: That is intentional. We don't want a configuration mistake to break more nodes. If users encounter such problem, they'd better do some troubleshooting to find why the node cannot come back.

pkg/daemon/daemon.go
```diff
@@ -791,27 +791,60 @@ func (dn *Daemon) getNodeMachinePool() error {
...
+	// start the leader election
+	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
+		Lock:            lock,
+		ReleaseOnCancel: true,
+		LeaseDuration:   5 * time.Second,
+		RenewDeadline:   3 * time.Second,
+		RetryPeriod:     1 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				glog.V(2).Info("drainNode(): started leading")
```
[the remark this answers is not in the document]
pliurh: fixed

pkg/daemon/daemon.go
```diff
@@ -18,6 +18,14 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
```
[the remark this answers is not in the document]
pliurh: fixed

martinkennelly (approved):

adrianchiris (approved):
I would prefer the package updates be performed in a separate commit in the PR (easier to review this way). but would not block on it.

@SchSeba i see below you were requested as reviewer. once you approve this can be merged IMO

pkg/daemon/daemon.go
adrianchiris: @pliurh mind mentioning this in the commit message ? so its clear in the commit message how this mechanism is used to control node draining

PR #116, opened 2021-04-21
Move to admissionregistration.k8s.io/v1

jcaamano:
Switch to admission v1 from v1beta1. This PR drops support for k8s <= 1.15. 

Requires: https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/96

PR #131, opened 2021-05-09
Fix network injector webhook

SchSeba:
This commit fix the network injector webhook to stuck the all cluster.

https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/116 move the admission version to v1

In v1 the default `FailurePolicy` is `Fail` and not `Ignore` as v1Beta1


yuvalk:
/lgtm
:-)

zshi-redhat:
@SchSeba do we need this change for operator-webhook ?

SchSeba:
@zshi-redhat there is no need for the other webhook they should remain on Fail policy as they are only for the sriov CRDs so they will not stuck the system

PR #133, opened 2021-05-14
Add a sanity check test suite

fedepaol:
This adds a "validation" test suite that perform very basic tests. It checks that all the crds are deployed, and that the pods are running correctly. 

This serves two purposes:

- having a toggle to understand when a deployment finished (implemented   by the make deploy-wait rule)
- having a set of tests able to prevent DOA releases, verifying that the operator deploys correctly (even without needing a baremetal cluster)

### DIFF
--- a/hack/deploy-wait.sh
+++ b/hack/deploy-wait.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+ATTEMPTS=0
+MAX_ATTEMPTS=72
+ready=false
+sleep_time=10
+
+until $ready || [ $ATTEMPTS -eq $MAX_ATTEMPTS ]
+do
+    echo "running tests"
+    if SUITE=./test/validation ./hack/run-e2e-conformance.sh; then
+        echo "succeeded"
+        ready=true
+    else    
+        echo "failed, retrying"
+        sleep $sleep_time
+    fi
+    (( ATTEMPTS++ ))
+done
+
+if ! $ready; then 
+    echo "Timed out waiting for features to be ready"
+    oc get nodes
+    exit 1
+fi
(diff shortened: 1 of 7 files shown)


test/validation/tests/validation.go
```diff
@@ -0,0 +1,143 @@
...
+})
+
+func CheckStable() {
+	res, err := cluster.SriovStable(operatorNamespace, clients)
+	Expect(err).ToNot(HaveOccurred())
+	Expect(res).To(BeTrue(), "SR-IOV status is not stable")
+
+	isClusterReady, err := cluster.IsClusterStable(clients)
+	Expect(err).ToNot(HaveOccurred())
+	Expect(isClusterReady).To(BeFalse(), "Cluster is not stable")
```
pliurh, line 142: We shall expect `isClusterReady` to be true here, right?

hack/deploy-wait.sh
```diff
@@ -0,0 +1,25 @@
...
+        sleep $sleep_time
+    fi
+    (( ATTEMPTS++ ))
+done
+
+if ! $ready; then 
+    echo "Timed out waiting for features to be ready"
+    oc get nodes
+    exit 1
+fi
```
pliurh, line 25: nit: newline needed

test/validation/tests/validation.go
fedepaol: Oh dang, good catch! It's a leftover from me validating the script. Will change

fedepaol: Done

hack/deploy-wait.sh
fedepaol: Done

Issue #86, opened 2021-03-09
sriov-network-operator should not depend on hardcoded supported device list


adrianchiris:
The `NicIdMap` is used for generating device plugin configmap and/or perhaps in other code path (operator-webhook).
If we can find a way to remove all those direct uses of NicIdMap, we could delete it and allows more NICs to be used with SR-IOV Operator directly.

For platforms that wish to limit the devices which the sriov-network-operator operates on, a configMap may be used to provide the list of supported devices.

pliurh:
Agree. As OLM has supported ConfigMap https://github.com/operator-framework/operator-lifecycle-manager/issues/1120.

PR #145, opened 2021-06-09
Use ConfigMap to store supported NIC model list

pliurh:
Use a ConfigMap 'supported-nic-ids' to store the NIC model list
to replace the current hardcoded list. Deprecate the ConfigMap
'unsupported-nic-ids'.

Also deprecate the machanism which allow users can change the
content of ConfigMap on the fly. Now users have to restart the
webhook and config daemon pods after changing the ConfigMap.

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -12,8 +12,10 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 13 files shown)


api/v1/helper.go
```diff
@@ -73,6 +77,28 @@ func (e NetFilterType) String() string {
...
+		return err
+	}
+	for _, v := range cm.Data {
+		NicIdMap = append(NicIdMap, v)
+	}
+	// return a sorted slice so that udev rule is stable
+	sort.Slice(NicIdMap, func(i, j int) bool {
+		ip, _ := strconv.ParseInt(NicIdMap[i], 0, 32)
+		jp, _ := strconv.ParseInt(NicIdMap[j], 0, 32)
+		return ip < jp
```
martinkennelly, line 97: ip & jp always resolve to 0
https://play.golang.org/p/rqy0xpHulpG

api/v1/helper.go
```diff
@@ -73,6 +77,28 @@ func (e NetFilterType) String() string {
...
+		context.Background(),
+		SUPPORTED_NIC_ID_CONFIGMAP,
+		metav1.GetOptions{},
+	)
+	// if the configmap does not exist, return false
+	if err != nil {
+		return err
+	}
+	for _, v := range cm.Data {
+		NicIdMap = append(NicIdMap, v)
```
martinkennelly, line 91: Shouldn't we make the existing NicIdMap blank? Its currently populated on Ln 40.

zshi-redhat:
How will configmap be updated when new IDs are added in configmap during upgrade?

deploy/configmap.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: supported-nic-ids
+data:
+  I40e_XXV710: "8086 158a 154c"
```
zshi-redhat, line 6: Shall we define a consistent name format for the data keys?
For example: 
<vendor_driver_model_name>
Intel_ice_E810_Columbiaville
Nvidia_mlx5_MT42882_ConnectX-5

pliurh:
> How will configmap be updated when new IDs are added in configmap during upgrade?

During an upgrade, the ConfigMap shall be updated first, then the pod images. So that we can guarantee pods start with the right ConfigMap.

deploy/configmap.yaml
pliurh: I think we can do that. But keep in mind, the NIC name here is not used in the code logic, it is just for readability.

api/v1/helper.go
pliurh: oh, forget to remove the old initialization code.

zshi-redhat:
> > How will configmap be updated when new IDs are added in configmap during upgrade?
> 
> During an upgrade, the ConfigMap shall be updated first, then the pod images. So that we can guarantee pods start with the right ConfigMap.

Who updates the configMap? and How?
Does it need to be deleted and re-created manually? The name of configMap doesn't change across upgrade and I don't see a logic in the controller to update the configMap.

pliurh:
Users shall update the ConfigMap for the upstream version. For the downstream version, the OLM can update the ConfigMap. https://github.com/operator-framework/operator-lifecycle-manager/pull/1434

deploy/configmap.yaml
zshi-redhat: @pliurh are you going to update the keys here?

pliurh: done.

pliurh: done.

pliurh:
close #86

pliurh:
@adrianchiris PTAL

zshi-redhat:
> close #86

I think this PR address part of issue #86:  to move NIC id map to configMap.
There is still another part in the issue: `allows more NICs to be used with SR-IOV Operator **directly**`
I understand the second part is to not rely on the map for admitting use of new NICs.

zshi-redhat:
/cc @akaris-redhat 
This removes unsupported-nic-ids configmap.

PR #115, opened 2021-04-20
Add webhook reconciliation

jcaamano:
With managed certificate handling for k8s introduced with #114 , we can bring back webhook reconciliation 

Requires: #114 

Signed-off-by: Jaime Caamaño Ruiz <jcaamano@suse.com>

PR #146, opened 2021-06-15
Update Webhook Configuration objects when Reconciling

pliurh:
Sync content of MutatingWebhookConfiguration and ValidatingWebhookConfiguration
CRs when needed.

https://bugzilla.redhat.com/show_bug.cgi?id=1969761

### DIFF
--- a/bindata/manifests/operator-webhook/003-webhook.yaml
+++ b/bindata/manifests/operator-webhook/003-webhook.yaml
@@ -3,7 +3,6 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
   name: {{.SRIOVMutatingWebhookName}}
-  namespace: {{.Namespace}}
   annotations:
     {{- if eq .ClusterType "openshift" }}
     service.beta.openshift.io/inject-cabundle: "true"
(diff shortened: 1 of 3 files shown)


zshi-redhat:
@pliurh could you explain how this PR can fix the issue of admission-controller API upgrade from v1beta to v1?

pliurh:
> @pliurh could you explain how this PR can fix the issue of admission-controller API upgrade from v1beta to v1?

The root cause is not the mismatch of the API version itself, since kube-apiserver can handle the version conversion automatically. It is because, in functions `syncMutatingWebhook` and `syncValidatingWebhook`, there is no content sync logic for the webhook config CR if it already exists. The field `admissionReviewVersions` in the CRs are not updated. https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/fc360d956c237db6daab3c9c324af5bd6db07fb6/bindata/manifests/operator-webhook/003-webhook.yaml#L16

pliurh:
@jcaamano PTAL. 

I don't think we need to keep the caBundle when syncing the webhook config as you proposed in #115 since it can always be reinjected by either cert-manger or the openshift-service-ca-operator.

PR #149, opened 2021-06-25
Use hostNetwork for sriov-cni daemonset

zshi-redhat:
When ovn hardware offload is enabled with ovn-k8s as default
CNI plugin, pods using veth as default interface type can no
longer be created since OpenvSwitch is offloaded to the
SmartNIC system (with current design). sriov-cni pod is affected
in such case. This commit moves sriov-cni into config daemonset
which runs in host network and doesn't use veth interface.
    
This also remove the sriov-cni daemonset created by Operator.


pliurh:
Considering the job of the sriov-cni daemon is just to put the CNI binary to the node, it's a one-time job. Can we remove this DS, and put it as the init container of sriov network config daemon?

martinkennelly:
@zshi-redhat Could you explain why this is needed? I don't understand the need to give the DS this privilege since its only function is to place the binary in cni bin dir.

zshi-redhat:
> @zshi-redhat Could you explain why this is needed? I don't understand the need to give the DS this privilege since its only function is to place the binary in cni bin dir.

@martinkennelly This is related to the ovn hardware offload use case. When it is enable on the worker node with ovn-kubernetes CNI, we can no longer create regular veth pods on that particular node since openvswitch is offloaded to the SmartNIC system (vs on the host).

Discussed with @pliurh , we were thinking to delete sriov-cni daemonset and add it as a container in sriov-config-daemon pod.

adrianchiris:
/lgtm

@zshi-redhat mind updating commit message why we are moving sriov-cni container to sriov-config-daemon ?

this also has the benefit of reducing the resources this operator creates (one less daemonset that k8s needs to monitor)

zshi-redhat:
> /lgtm
> 
> @zshi-redhat mind updating commit message why we are moving sriov-cni container to sriov-config-daemon ?
> 
> this also has the benefit of reducing the resources this operator creates (one less daemonset that k8s needs to monitor)

Updated commit message.

adrianchiris (approved):
/lgtm

adrianchiris:
I have not tested this change (no e2e test ci yet :( ), so @zshi-redhat if you have done some testing internally this can be merged IMO

zshi-redhat:
> I have not tested this change (no e2e test ci yet :( ), so @zshi-redhat if you have done some testing internally this can be merged IMO

I built customized image and tried to create a policy, it worked as expected: no sriov-cni daemonset was created, sriov-cofig-daemon was created with several containers, including sriov-cni.

pliurh:
We also need to remove the legacy daemonsets of the CNI-plugins.

zshi-redhat:
> We also need to remove the legacy daemonsets of the CNI-plugins during an upgrade.

Added code to remove sriov-cni manifests

PR #151, opened 2021-06-25
Fixed typo in the operator helm chart

dpuzikox:
Fixing typo causing troubles

### DIFF
--- a/deployment/sriov-network-operator/templates/operator.yaml
+++ b/deployment/sriov-network-operator/templates/operator.yaml
@@ -9,7 +9,7 @@ spec:
   selector:
     matchLabels:
       name: sriov-network-operator
-  updateStrategy:
+  strategy:
     type: RollingUpdate
     rollingUpdate:
       maxUnavailable: 33%


zshi-redhat:
@dpuzikox thanks for the PR!
Could you explain why this is needed? I saw that we have other daemonsets (e.g. the [sriov-network-config-daemon](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/bindata/manifests/daemon/daemonset.yaml#L13)) which use updateStrategy as well.

dpuzikox:
> @dpuzikox thanks for the PR!
> Could you explain why this is needed? I saw that we have other daemonsets (e.g. the [sriov-network-config-daemon](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/bindata/manifests/daemon/daemonset.yaml#L13)) which use updateStrategy as well.

@zshi-redhat while `DaemonSet` has `updateStrategy` field :  https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#daemonset-v1-apps
the `Deployment` has differently named field for the same purpose: `strategy` : https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#deployment-v1-apps
[operator.yaml](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/deployment/sriov-network-operator/templates/operator.yaml) is a `Deployment` not a `DaemonSet` and currently helm deployments fails with error like: `apps.deployment has no field updateStrategy` as those fields are not interchangable.

e0ne (approved):

PR #155, opened 2021-07-01
Fix e2e tests on k8s environment

ykulazhenkov:
This PR fixes e2e tests on k8s enviroment

`DiscoverSriov()` uses `IsSupportedDevice()` function from `sriovv1`
package which relies on `sriovv1.NicIdMap`.
After recent changes, NicIdMap is not a static map anymore.
Need to update NicIdMap configmap with InitNicIdMap() function
call before usage.

Modify GetNodeSecureBootState to accept namespace argument and path sriov-network-operator
namespace to it. This required to keep all things in a single namespace during e2e.

FIX GetNodeSecureBootState function to handle
"/host/sys/kernel/security/lockdown" file absence correctly.
This path will be unavailable if kernel is compiled without lockdown
feature.

### DIFF
--- a/test/util/cluster/cluster.go
+++ b/test/util/cluster/cluster.go
@@ -10,11 +10,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	sriovv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
 	testclient "github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/client"
-	"github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/namespaces"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/nodes"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/pod"
 )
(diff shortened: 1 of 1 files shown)


ykulazhenkov:
@SchSeba  @pliurh PTAL, this change related to secure boot env fix and to changes for using ConfigMap for supported NICs

test/util/cluster/cluster.go
```diff
@@ -293,14 +299,14 @@ func GetNodeSecureBootState(clients *testclient.ClientSet, nodeName string) (boo
 		return false, err
 	}
 
-	stdout, stderr, err := pod.ExecCommand(clients, runningPod, "cat", "/host/sys/kernel/security/lockdown")
+	stdout, _, err := pod.ExecCommand(clients, runningPod, "cat", "/host/sys/kernel/security/lockdown")
```
moshe010, line 302: why did you remove the stderr check. I still can failed for other reason

moshe010 (changes requested):

test/util/cluster/cluster.go
ykulazhenkov: This check is useless. Stderr will be always empty. 

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/cb3b24b87a4de9e50f2b684b3a602dd6e657a8a1/test/util/pod/pod.go#L113

Check [Goclient src](https://github.com/kubernetes/client-go/blob/6711973af1c55d544f08252297ebf6dab2031975/tools/remotecommand/remotecommand.go#L46)

If command completes with non zero exit code, then err will be not nil.

adrianchiris (approved):

zshi-redhat (approved):

PR #125, opened 2021-05-07
OVS HWOL: support multiple MCPs

zshi-redhat:
Introduced a new CRD SriovNetworkPoolConfig for pool (not device)
related configurations such as OVS HWOL. This may be extended
to support future features that require pool specific configuration.
For example, SmartNIC offload, Accelerated bridge offload etc.

Removed MCO plugins and moved the writing of switchdev device config
file into generic plugin (shared by OpenShift and Kubernetes deployment).
Removed the writing of switchdev device config file from K8s plugin.

For configuring OVS HWOL on OpenShift, the steps are changed to:
- Create SriovNetworkPoolConfig to specify the name of MachineConfigPool
- Configure SriovNetworkNodePolicy with eSwitchMode: switchdev (This writes switchdev.conf file and trigger drain/reboot where needed)

For configuring OVS HWOL on Kubernetes, the steps are not changed.

PR #6, opened 2020-12-16
[WIP] Support RDMA namespace changing and generate CNI config with RDMA isolation support

Mmduh-483:

PR #156, opened 2021-07-02
WIP: add API to configure RDMA subsystem network namespace mode

ykulazhenkov:
This PR introduces API to configure RDMA mode on nodes.

RDMA subsystem network namespace mode is a node-level configuration. Reboot required to change this option.

This PR relies on  and extend SriovNetworkPoolConfig API proposed in https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/125

Replacement for https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/6

Status: **WIP**

API changes:

- Add  RDMAExclusiveMode to SriovNetworkPoolConfig

- Add NodeSettings to SriovNetworkNodeState.Spec

Proposed workflow to apply node level configuration on k8s clusters:
1. User create SriovNetworkPoolConfig CR
2. SriovNetworkPoolConfig controller (single instance, part of sriov-network-operator POD) reconcile CR, and update SriovNetworkNodeState.Spec.NodeSettings for nodes which match selectors.
3. sriov-network-config-daemon (runs on every node) read updated config from SriovNetworkNodeState CR and apply required node level configuration. We can use plugins to handle node level configuration.


zshi-redhat:
@ykulazhenkov Thanks for the PR!

I discussed with @pliurh on this PR and here are some suggestions:

1) The mechanism used for passing interfaces configuration via SriovNetworkNodeState can also be used to pass node level configuration (SriovNetworkPoolConfig) to sriov config daemonset. We can have another field (in the same level with Interfaces) in SriovNetworkNodeState.Spec (just like what you did) for node specific configuration. The advantages of this are we don't have to create another CRD and informer (in config daemon) for the same purpose, existing plugins can be used w/o adding additional plugin interfaces.

2) Regarding the plugin-rdma, we presume the rdma isolatoin is a generic feature that would apply to all vendor devices, would it make sense to add its function in generic_plugin (vs writing a separate plugin)?

ykulazhenkov:
@zshi-redhat Thanks for the feedback.

> Regarding the plugin-rdma, we presume the rdma isolatoin is a generic feature that would apply to all vendor devices, would it make sense to add its function in generic_plugin (vs writing a separate plugin)?

I agree, there is no need for the separate plugin for this. I will rework this.

jjuele:

SchSeba:
I don't think this PR is still active
I will close it please reopen if this is needed

PR #159, opened 2021-07-08
Change switchdev config to json format

pliurh:
1. use json format for config file
2. split VF creation and switchdev configuration in configure-switchdev.sh

### DIFF
--- a/pkg/utils/switchdev.go
+++ b/pkg/utils/switchdev.go
@@ -1,7 +1,8 @@
 package utils
 
 import (
-	"fmt"
+	"bytes"
+	"encoding/json"
 	"io/ioutil"
 	"os"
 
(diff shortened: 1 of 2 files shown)


pkg/utils/switchdev.go
```diff
@@ -11,9 +12,20 @@ import (
...
 const (
-	switchDevConfPath = "/host/etc/switchdev.conf"
+	switchDevConfPath = "/host/etc/sriov_interface_config.json"
 )
 
+type config struct {
+	Interfaces []Interface `json:"interfaces"`
+}
+
+type Interface struct {
```
zshi-redhat, line 22: @pliurh  can we use the interface structure in sriovnetworknodestate API?

pliurh: Good idea. Although we don't support configuring all the flags now, we might support them in the future.

PR #161, opened 2021-07-11
Remove sriov-cni check from E2E tests

AbdYsn:
Following k8snetworkplumbingwg/sriov-network-operator#149, sriov-cni
daemonset is no longer deployed separatly but as part of the config
daemon daemonset. This breaks the E2E tests, since the E2E
tests wait for the sriov-cni daemonsets to be ready.

This patch fix that by replacing the wait condition from the sriov-cni
to the config-daemon daemonset.


AbdYsn:
@adrianchiris @zshi-redhat Can you guys have a look at this?

moshe010 (approved):

PR #165, opened 2021-07-19
NRI: use control switch configmap to enable hugepage injection

zshi-redhat:
Network Resource Injector (NRI) has switches to config map to enable
optional features such as enableHugePageDownApi,
enableHonorExistingResources etc. This commit allows Operator to
create controlSwitch configmap for NRI and removes injectHugepageDownApi
from webhook-server container arguments.

### DIFF
--- a/bindata/manifests/webhook/004-configmap.yaml
+++ b/bindata/manifests/webhook/004-configmap.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nri-control-switches
+  namespace: {{.Namespace}}
+data:
+  config.json: |
+    {
+      "features": {
+        "enableHugePageDownApi": true
+      }
+    }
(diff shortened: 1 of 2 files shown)


bindata/manifests/webhook/004-configmap.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
adrianchiris, line 1: a comment on the name of the file.

should it be `004-configmaps.yaml` ?  or `004-nri-control-switches.yaml` ?

there may be other configmap that need to be deployed in the future from nri.

zshi-redhat: @adrianchiris I think it shall be `004-configmap.yaml`. we have other manifests that are named after the singular kubernetes resource such as rbac, service or webhook etc. Example like webhook contains both mutating and validating webhooks, but the name is singular.

adrianchiris: ack, makes sense. additional configmaps would be added to this file.

adrianchiris (approved):

PR #172, opened 2021-08-01
Make mstflint configurable for config daemon

e0ne:
mstflint in CentOS repository is too old and doesn't support latest
Nvidia Nics like BlueField 2.

This patch makes mstflint package location configurable so we can
build image with required version and location of it.


adrianchiris (approved):
/lgtm

to get bluefield-2 support in mstflint by default, base image should change to be centos8(ish)
since support was added inbox from mstflint version 4.14

However i can see benefits of having this configurable even if the base image is bumped (e.g if we want to pin this to a certain version or for testing)

moshe010 (approved):

zshi-redhat (approved):

PR #174, opened 2021-08-06
Remove SriovNetwork finalizers on controller shutdown

mmirecki:
The PR is added in response to customer complaints about SriovNetworks being stuck as terminating after uninstalling the sriov operator.
This PR removes the "netattdef.finalizers.sriovnetwork.openshift.io" finalizer from SriovNetworks when the sriovnetwork controller is deleted.  This finalizer is added to each SriovNetwork on reconciliation, but it is not removed when the operator is uninstalled, leaving SriovNetworks with finalizers and the SriovNetwork crd hanging in the terminating state forever.
The finalizers will be removed whenever the sriov operator pod is shut down, including pod restarts. The removed finalizer will be added when the pod restarts and the networks are reconciled.

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,6 +44,7 @@ import (
 
 	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/controllers"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
 	//+kubebuilder:scaffold:imports
 )
 
@@ -172,6 +173,9 @@ func main() {
 			os.Exit(1)
 		}
 	}()
+
+	// Remove all finalizers after controller is shut down
+	defer utils.Shutdown(mgr.GetClient())
 
 	setupLog.Info("starting manager")
 	if err := mgr.Start(stopCh); err != nil {
--- a/pkg/utils/shutdown.go
+++ b/pkg/utils/shutdown.go
@@ -1,0 +1,30 @@
+package utils
+
+import (
+	"context"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+)
+
+func Shutdown(c client.Client) {
+	shutdownLog := ctrl.Log.WithName("shutdown")
+	shutdownLog.Info("Clearing finalizers on exit")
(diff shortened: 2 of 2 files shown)


main.go
```diff
@@ -178,6 +179,8 @@ func main() {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
+	// Remove all finalizers after controller is shut down
+	utils.Shutdown(mgr.GetClient())
```
pliurh, line 183: Shall we move this function before the main loop and use `defer` call instead?

pliurh:
We also use a similar finalizer mechanism for the SriovNetworkPoolConfig. So maybe we should do the same for the SriovNetworkPoolConfig CRs.

main.go
mmirecki: done
Adding defer before we start the manager

SchSeba (commented):
Thanks for the contribution!

just a small comment

pkg/utils/shutdown.go
```diff
@@ -0,0 +1,25 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func Shutdown(c client.Client) {
+	networkList := &sriovnetworkv1.SriovNetworkList{}
+	err := c.List(context.TODO(), networkList)
+	if err != nil {
+		glog.Errorf("Failed to list SriovNetworks: %+v", err)
+	} else {
+		for _, instance := range networkList.Items {
```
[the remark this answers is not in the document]
mmirecki: done

pkg/utils/shutdown.go
```diff
@@ -0,0 +1,25 @@
+package utils
+
+import (
+	"context"
+
+	"github.com/golang/glog"
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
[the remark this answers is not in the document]
mmirecki: done

pkg/utils/shutdown.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+
+	"github.com/golang/glog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+)
+
+func Shutdown(c client.Client) {
+	networkList := &sriovnetworkv1.SriovNetworkList{}
```
pliurh, line 13: Can we add some log messages here?
This patch overall looks good to me.

mmirecki: Added logging.
Note that the log messages might be hard to see, as the pod will be removed right after this is executed.

PR #175, opened 2021-08-12
Use GlobalMgr for SriovNetwork controllers

oribon:
Using 2 managers for the same controllers causes inconsistency in
objects reconcilation. For example we saw a problem with some SriovNetwork
update events not reaching the controller (Reconcile was not triggered).
Using the same Mgr as the specified client for the controller solves these kind of problems.

Signed-off-by: Ori Braunshtein <obraunsh@redhat.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -108,14 +108,14 @@ func main() {
 	if err = (&controllers.SriovNetworkReconciler{
 		Client: mgrGlobal.GetClient(),
 		Scheme: mgrGlobal.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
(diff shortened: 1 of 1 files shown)


mgourin:
@oribon 
Could you elaborate on the differences "mgr" and "mgrGlobal" and how it caused the issue of not being able to delete the Sriov Networks?

pliurh:
This error was introduced by https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/bc40302df612fc62e0252977011fe7f95abe02b0/main.go

The `mgrGlobal` have access to all namespaces, which is useful to manager the net-att-def CRs in other namespaces.

PR #177, opened 2021-08-19
Set webhook failurePolicy to Ignore on controller pod shutdown

mmirecki:
This PR sets the failurePolicy of sriov webhooks to Ignore when an operator pod is exiting. This ensures that the policy does not cause incoming resources to fail when the sriov pods are down. The policy is changed to back to Fail on the next sriovconfig reconciliation loop (for example when the controller pod is started again).

The PR also changes the kubernetes API used in the shutdown function from the controller-runtime api to the basic client-go api.
This is due issues with the controller api during the controller shutdown. It was noticed that the api calls sometimes failed due to Informer not able to sync, probably caused by the manager being closed before the shutdown function was invoked.

### DIFF
--- a/main.go
+++ b/main.go
@@ -175,7 +175,7 @@ func main() {
 	}()
 
 	// Remove all finalizers after controller is shut down
-	defer utils.Shutdown(mgr.GetClient())
+	defer utils.Shutdown()
 
 	setupLog.Info("starting manager")
 	if err := mgr.Start(stopCh); err != nil {
--- a/pkg/utils/shutdown.go
+++ b/pkg/utils/shutdown.go
@@ -3,28 +3,101 @@ package utils
 import (
 	"context"
 
+	snclientset "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/clientset/versioned"
+	admv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	conf "sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
 )
 
-func Shutdown(c client.Client) {
-	shutdownLog := ctrl.Log.WithName("shutdown")
+var shutdownLog = ctrl.Log.WithName("shutdown")
+
+var sriovnetworksGVR = schema.GroupVersionResource{
+	Group:    "sriovnetwork.openshift.io",
+	Version:  "v1",
+	Resource: "sriovnetworks",
+}
+
+var failurePolicyIgnore = admv1.Ignore
+
+func Shutdown() {
+	updateFinalizers()
+	updateWebhooks()
+}
+
+func updateFinalizers() {
 	shutdownLog.Info("Clearing finalizers on exit")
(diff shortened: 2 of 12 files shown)


pkg/utils/shutdown.go
```diff
@@ -2,29 +2,113 @@ package utils
...
+	Group:    "sriovnetwork.openshift.io",
+	Version:  "v1",
+	Resource: "sriovnetworks",
+}
+
+var failurePolicyIgnore = admv1.Ignore
+
+func Shutdown() {
+	updateFinalizers()
+	//updateWebhooks()
```
pliurh, line 31: Shall it be commented out?

main.go
```diff
@@ -175,7 +175,7 @@ func main() {
 	}()
 
 	// Remove all finalizers after controller is shut down
-	defer utils.Shutdown(mgr.GetClient())
+	defer utils.Shutdown()
```
pliurh, line 178: Why not reuse the client anymore?

pkg/utils/shutdown.go
```diff
@@ -2,29 +2,113 @@ package utils
...
+		shutdownLog.Error(err, "Error getting cluster config")
+	}
+	return config
+}
+
+func updateFinalizers() {
 	shutdownLog.Info("Clearing finalizers on exit")
-	networkList := &sriovnetworkv1.SriovNetworkList{}
-	err := c.List(context.TODO(), networkList)
+	c, err := dynamic.NewForConfig(getConfig())
```
pliurh, line 44: Can we use the `GetConfigOrDie` from "sigs.k8s.io/controller-runtime/pkg/client/config" instead `getConfig()`? So that we can support running the operator in both in-cluster and local mode. The local mode is quite handy for development.

pkg/utils/shutdown.go
```diff
@@ -2,29 +2,113 @@ package utils
...
+	if err != nil {
+		shutdownLog.Error(err, "Error creating client")
+	}
+	sriovNetworkClient := c.Resource(sriovnetworksGVR)
```
pliurh, line 48: There is a go-client implementation under package/client for the SriovNetwork CRDs, I think we can use it here as well.

pkg/utils/shutdown.go
mmirecki: fixed

pkg/utils/shutdown.go
mmirecki: Done

pkg/utils/shutdown.go
mmirecki: Thanks for this tip
Done

pkg/utils/shutdown.go
```diff
@@ -3,28 +3,100 @@ package utils
...
+func Shutdown() {
+	updateFinalizers()
+	updateWebhooks()
+}
+
+func updateFinalizers() {
 	shutdownLog.Info("Clearing finalizers on exit")
-	networkList := &sriovnetworkv1.SriovNetworkList{}
-	err := c.List(context.TODO(), networkList)
+	//c2, err := dynamic.NewForConfig(conf.GetConfigOrDie())
```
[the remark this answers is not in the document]
mmirecki: removed

pkg/utils/shutdown.go
```diff
@@ -3,28 +3,100 @@ package utils
...
 	shutdownLog.Info("Done clearing finalizers on exit")
 }
+
+func updateWebhooks() {
+	shutdownLog.Info("Seting webhook failure policies to Ignore on exit")
+	clientset, err := kubernetes.NewForConfig(conf.GetConfigOrDie())
+	if err != nil {
+		shutdownLog.Error(err, "Error getting client")
+	}
+	validatingWebhookClient := clientset.AdmissionregistrationV1().ValidatingWebhookConfigurations()
```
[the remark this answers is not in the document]
mmirecki: Done

PR #179, opened 2021-09-06
Don't check MCP state when node in Draining_MCP_Paused state

pliurh:


pliurh:
Fix https://bugzilla.redhat.com/show_bug.cgi?id=1999079

zshi-redhat:
This fix looks good to me.

@pliurh would you mind adding some annotations in the code to explain why the conditional check is needed?

pliurh:
> This fix looks good to me.
> 
> @pliurh would you mind adding some annotations in the code to explain why the conditional check is needed?

Comments added.

zshi-redhat (approved):

PR #180, opened 2021-09-07
Restart device plugin pod after node policy applied

e0ne:
We need to restart device plugin pod after node policy applied
and all SR-IOV Network Config Daemon plugins finished. E.g.:

Generic plugin applies SR-IOV Network Node Policy and creates
VFs. That's mean we need to restart device plugin pod to found
newly created devices.

Related k8snetworkplumbingwg/sriov-network-device-plugin#276

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -518,15 +518,6 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 		return nil
 	}
 
-	// restart device plugin pod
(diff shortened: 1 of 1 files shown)


zshi-redhat:
@e0ne  Could you help me understand why we need to move the restart of device plugin after the completeDrain?

e0ne:
@zshi-redhat thanks for the review. the idea of my patch is to restart device plugin pod once after all sriov plugins applied. Digger deeper in the code, I agree that it should be done  before completeDrain. It means this PR is the same as #173 but with different commit message

e0ne:
it's duplicate of #173

PR #182, opened 2021-09-10
Update github.com/cespare/xxhash/v2 to v2.1.2

fedepaol:
github.com/cespare/xxhash/v2 v2.1.1 is incompatible with Go 1.17 (see https://github.com/cespare/xxhash/issues/54)
Here we bump up the dependency to solve the issue.


fedepaol:
cc @zshi-redhat @SchSeba

hack/tools.go
```diff
@@ -1,3 +1,4 @@
+//go:build tools
```
zshi-redhat, line 1: Is this change still needed?

fedepaol: Yes, that's why I pushed 3 times :-)

CI was complaining about it

zshi-redhat: wondering if we can remove the following line`+build tools`.

fedepaol: I think we can't. CI compares the generated file with what's in this PR and complains if they are different.

fedepaol:
```
diff --git a/hack/tools.go b/hack/tools.go
index 7a21725c..d1427a67 100644
--- a/hack/tools.go
+++ b/hack/tools.go
@@ -1,4 +1,4 @@
-// +build tools
+//go:build tools
```

This is if I remove the other line

zshi-redhat (approved):

PR #184, opened 2021-09-13
Do not set DpConfigVersion for node state by default

e0ne:
DpConfigVersion is needed only if node policy is created and node
state contains interfaces spec.

If we set DpConfigVersion for an empty node state it will lead to
unexpected behaviour after node reboot durin SR-IOV enabling:
device config daemon will ignore new state and won't apply it to
the node.

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -270,7 +270,6 @@ func (r *SriovNetworkNodePolicyReconciler) syncSriovNetworkNodeState(np *sriovne
 	if err != nil {
 		logger.Info("Fail to get SriovNetworkNodeState", "namespace", ns.Namespace, "name", ns.Name)
 		if errors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


pliurh:
Thanks for the PR. 
How about we only set `dpConfigVersion` when `spec.interfaces` is not null.

e0ne:
> Thanks for the PR.
> How about we only set `dpConfigVersion` when `spec.interfaces` is not null.

Thanks for the suggestion, I'll test this approach.  Apart of that, this patch introduces some unnecessary device plugin pod restarts:

`# kubectl -n network-operator logs sriov-device-plugin-q8pv2  --previous
I0920 13:10:04.036408       1 manager.go:52] Using Kubelet Plugin Registry Mode
I0920 13:10:04.036436       1 main.go:44] resource manager reading configs
E0920 13:10:04.036445       1 main.go:46] error getting resources from file error reading file /etc/pcidp/cloud-dev-13, open /etc/pcidp/worker-1: no such file or directory`

It requires additional investigation

PR #193, opened 2021-10-21
Use configurable image tags for major version branches

andreaskaris:
Ignore not found resources during uninstallation. 
Other dependencies such as make undeploy will be able to run even if
a specific resource was not found for uninstallation.

Signed-off-by: Andreas Karis <ak.karis@gmail.com>

### DIFF
--- a/hack/env.sh
+++ b/hack/env.sh
@@ -1,22 +1,26 @@
+export RELEASE_VERSION=4.7.0
+export IMAGE_TAG="" # set for master
+# export IMAGE_TAG=":${RELEASE_VERSION}" # set for branches
+
 if [ -z $SKIP_VAR_SET ]; then
         if ! skopeo -v &> /dev/null
         then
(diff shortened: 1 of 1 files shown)


andreaskaris:
RELEASE_VERSION has been set to 4.7.0 ever since that version. I grepped in the repo ... is that variable even used at all? If not, we might as well just remove it and only use the tag; I'd just like to have `make run` automatically pull the correct image.
E.g., I just cloned the operator, checked out the 4.7 branch, ran `make run` and all pods crashed because the CRs were deployed for 4.7 but the images were deployed for the latest release. Each branch's hack/env.sh should pin the images to the correct branch as well to avoid such mismatches.

e0ne (approved):

zeeke:
lgtm

SchSeba:
closing this PR if needed we can reopen but we also need to change the image location to the new ones

Issue #194, opened 2021-10-26
SriovNetworkNodeConfigPolicy priority inconsistent


mskrocki:
I have looked at the code and how it handles policies for the same root PF. I understand that the policy priorities are suppose to resolve competing configurations, and merge them if priorities are same.
Thing is I am unable to determine what is the "higher" priority in this project?
With: `policy1` with p=30 and `policy2` with p=40 which policy should overwrite which?

As a context will note that the polices are sorted by the priority in decreasing order: 100, 40, 30 and this is how they are handled thru the rest of the life-cycle.

Having defined that, the issue is either in: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/api/v1/helper.go#L264 where we select the last policy i.e. lower priority value (30 over 40) overwrites the final VFs number

or the issue is in https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L263 where we first apply the config of the first VfGroup to the VFs (here its 40 over 30).

PR #196, opened 2021-10-27
Fix priority handling for same-pf VFgroups.

mskrocki:
This change fixes VFGroups assigned to nodes based
on the policies priorities: highest priority (lowest
value of priority) should be the only one present in the
SriovNetworkNodeState.spec.

Additionally, for same priority policies we discard
overlapping VF ranges, only the highest priority is
present.

FIXES: #194

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -252,12 +252,12 @@ func UniqueAppend(inSlice []string, strings ...string) []string {
 }
 
 // Apply policy to SriovNetworkNodeState CR
-func (p *SriovNetworkNodePolicy) Apply(state *SriovNetworkNodeState, merge bool) {
+func (p *SriovNetworkNodePolicy) Apply(state *SriovNetworkNodeState, equalPriority bool) error {
 	s := p.Spec.NicSelector
 	if s.Vendor == "" && s.DeviceID == "" && len(s.RootDevices) == 0 && len(s.PfNames) == 0 &&
 		len(s.NetFilter) == 0 {
 		// Empty NicSelector match none
-		return
+		return nil
 	}
 	for _, iface := range state.Status.Interfaces {
 		if s.Selected(&iface) {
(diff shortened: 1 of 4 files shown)


api/v1/helper.go
```diff
@@ -258,55 +258,74 @@ func (p *SriovNetworkNodePolicy) Apply(state *SriovNetworkNodeState, merge bool)
...
-			input.Mtu = iface.Mtu
-		}
-		if input.NumVfs < iface.NumVfs {
-			input.NumVfs = iface.NumVfs
+// mergeConfigs merges configs from multiple polices where the last one has the
+// highest priority. This merge is dependent on: 1. SR-IOV partition is
+// configured with the #-notation in pfName and 2. SR-IOV policies have the same
+// priority.
+func (iface Interface) mergeConfigs(input *Interface, equalPriority bool) {
+	if !equalPriority {
```
zshi-redhat, line 292: I think we would still want to allow VF partition policies with different priority numbers, for example:

Policy 1:
```
spec:
  deviceType: netdevice
  resourceName: kernel-device
  numVfs: 8
  priority: 50
  nicSelector:
    pfNames:
    - "ens1f0#0-3"
```

Policy 2:
```
spec:
  deviceType: vfio-pci
  resourceName: dpdk-device
  numVfs: 8
  priority: 70
  nicSelector:
    pfNames:
    - "ens1f0#4-7"
```

With current logic, policy-2 won't be applied if I understand correctly, but we should allow the policy-2 vfGroup be created since it doesn't conflict with policy-1 configuration.

This means the conditional check `!equalPriority` shall not be applied to vfGroup merging in line 308.

mskrocki: From logical point of view I agree, these policies could be merged, but on the other hand wont that complicate the "priority's" field meaning? Now it is simple, merge only when same priority.
Why those policies cannot have same priority? If user explicitly defined the partitioning, the intention is to merge these, so in my opinion, they should have same priority.

zshi-redhat: My concern is it may break existing deployment if we change current logic. For example, if user already has VF partition + different priorities settings, upgrade of operator will result in unexpected result.

I agree with you on that priority field may not necessary in most cases.

mskrocki: Existing users can be instructed via documentation (or release notes), we are not changing any APIs. They would only have to adjust their priorities.
Additionally, this repo docs do not specify how exactly the priority field works, probably smth I can add to this PR.

zshi-redhat: > Existing users can be instructed via documentation (or release notes), we are not changing any APIs. 
User may not want to change their configuration in a production environment that is working, to me this is like a regression I'd like to avoid.

> They would only have to adjust their priorities.
In my opinion, priority shall apply when there is conflict between policies (only apply high priority one over the other), but not when there is no conflict.

adrianchiris: Discussed this in Today's meeting.

In general, the `priority` field is not well documented (except of the one-line in the sriovnetworknodepolicy CRD)
more so, the interaction between VF partitioning and priority when merging configurations.

so we should definitely document this feature :)

AFAIU, today, if you use VF partitioning with multiple policies, if there is an overlap in VF group you might end up with an un-expected configuration (lower priority takes precedence)

I can think of additional edge cases, where you apply configuration on a per device basis and not per vf group like switchdev. e.g two sriovnetworkpolicies one with switchdev enabled the other without, same PF with different vf groups. (not sure if it would fail or not)

IMHO, if one wants to partition a PF's VFs to different resources he should do that in the same priority.

mskrocki: I have updated the code to be backwards compatible. Added README changes.

api/v1/helper.go
```diff
@@ -268,55 +268,76 @@ func (p *SriovNetworkNodePolicy) Apply(state *SriovNetworkNodeState, merge bool)
...
+// highest priority. This merge is dependent on: 1. SR-IOV partition is
+// configured with the #-notation in pfName, 2. The VF groups are
+// non-overlapping or SR-IOV policies have the same priority.
+func (iface Interface) mergeConfigs(input *Interface, equalPriority bool) {
+	m := false
+	// merge VF groups (input.VfGroups already contains the highest priority):
+	// - skip group with same ResourceName,
+	// - skip overlapping groups (use only highest priority)
+	for _, gr := range iface.VfGroups {
+		if gr.ResourceName == input.VfGroups[0].ResourceName || gr.isVFRangeOverlapping(input.VfGroups[0]) {
```
zshi-redhat, line 307: This won't merge when a higher priority policy(input.vfGroup[0]) with same resourceName or overlapping VF group, but it should overwrite, right?
```
if gr.ResourceName == input.VfGroups[0].ResourceName ... {

    if !equalPriority {
        input.VfGroups[0] = gr
    }

}
```

api/v1/helper.go
```diff
@@ -268,55 +268,76 @@ func (p *SriovNetworkNodePolicy) Apply(state *SriovNetworkNodeState, merge bool)
...
+			continue
 		}
+		m = true
+		input.VfGroups = append(input.VfGroups, gr)
+	}
+
+	if !equalPriority && !m {
```
zshi-redhat, line 314: This may not needed as we would always want to update mtu and numvfs to highest number.
For example: 
```
if equalPriority {
      // update mtu and numVfs 
      // because higher value from same priority policies is applied
}
```

```
if !equalPriority {
      // update mtu and numVfs 
      // because value from higher priority policy gets applied
}
```

mskrocki: it is, we do not want to merge mtu and numvfs when we just overwrite i.e. m=false.

api/v1/helper.go
mskrocki: Mind what is used outside mergeConfigs, i.e. `*input` is the final object we use, not `iface`. So, input.VfGroups[0] already contains the highest priority group, and here we need to verify the other groups do not overlap or have same resourceName. If they dont we add them to the input.VfGroups list (line 311).

mskrocki:
ping, could we merge this?

api/v1/helper.go
zshi-redhat: input.VfGroups[0] already contains the highest priority group - right. 
when the highest priority group overlaps or has the same resourceName as lower priority one (`iface`), should it override (because highest priority wins)?

zshi-redhat (commented):
@mskrocki  sorry for late review.
got a follow-up questions in the above comment.

api/v1/helper.go
mskrocki: yes, because this is a conflicting situation: either we have duplicate resourceName, or there is VF range overlap. In such cases we will use the high priority one (lower value).

mskrocki: Will add that "override" in this "for" is done by just skipping given group and not adding it to the final "input" object.

zshi-redhat: > yes, because this is a conflicting situation: either we have duplicate resourceName, or there is VF range overlap. In such cases we will use the high priority one (lower value).

Yes, this is a conflicting situation.

mskrocki: Do you have any other doubts or questions about this?

zshi-redhat: I‘d expect line 314 will also be removed based on the above change, will it?
I don't have other comments beside this.

mskrocki: No, line 314 is required as per my reply below. We cannot update mtu and numvfs blindly for every call of mergeConfigs i.e. for case where we did NOT merge (enter "continue", line 308, every time), but just override we need to preserve the original values.

api/v1/helper.go
mskrocki: Will add that as mentioned above, *input is the final object we use, not iface and it already contains the highest priority policy values for mtu and NumVfs.

PR #201, opened 2021-11-12
Use systemd service to set numVFs for BF-2

pliurh:
To enable OVN hardware offloading with BlueField-2, VFs shall be
created before ovnkube-node running in dpu-host mode starts.

### DIFF
--- a/pkg/plugins/generic/generic_plugin.go
+++ b/pkg/plugins/generic/generic_plugin.go
@@ -170,10 +170,6 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
 		for _, iface := range desired {
 			if iface.PciAddress == ifaceStatus.PciAddress {
 				configured = true
-				if iface.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
-					// ignore swichdev device
-					break
-				}
 				if ifaceStatus.NumVfs != 0 {
 					if iface.NumVfs != ifaceStatus.NumVfs {
 						glog.V(2).Infof("generic-plugin needDrainNode(): need drain, expect NumVfs %v, current NumVfs %v", iface.NumVfs, ifaceStatus.NumVfs)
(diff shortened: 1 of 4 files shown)


adrianchiris:
Change looks OK to me. Could you add some more info in commit message on why this is needed ?

Dockerfile.sriov-network-config-daemon.rhel7
```diff
@@ -1,10 +1,10 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
```
zshi-redhat, line 1: we probably want to remove the rhel7 dockerfiles in k8s plumbing repo.

pliurh: fixed

adrianchiris (approved):

pkg/plugins/generic/generic_plugin.go
```diff
@@ -213,7 +213,7 @@ func needRebootNode(state *sriovnetworkv1.SriovNetworkNodeState, loadVfioDriver
 		}
 	}
 
-	if utils.IsSwitchdevModeSpec(state.Spec) {
+	if utils.IsSwitchdevModeSpec(state.Spec) || utils.IsDpuConfigured(state) {
```
zshi-redhat, line 216: Should we add a similar check in needDrainNode (line173)?

pliurh: I think we don't need the check at line 173 anymore. Since it was added with the MCO plugin and we have deprecated the MCO plugin. The MCO plugin is used to move a node to the MCP, so we don't need to drain/reboot the node by the config daemon. But now, as the config daemon is only responsible for updating the `sriov_config.json`, we shall use the generic plugin to drain/reboot the node.

zshi-redhat (approved):

pkg/utils/utils.go
```diff
@@ -148,8 +149,8 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 		for _, iface := range newState.Spec.Interfaces {
 			if iface.PciAddress == ifaceStatus.PciAddress {
 				configured = true
-				if iface.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV && ClusterType == ClusterTypeOpenshift {
-					// Skip sync for openshift, MCO will inject a systemd service to config switchdev devices.
+				if SkipConfigVf(iface, ifaceStatus) {
```
zshi-redhat, line 152: This removes `ClusterType == ClusterTypeOpenshift` conditional check.
@e0ne  Do you see this will be a problem for k8s switchdev configuration?

PR #204, opened 2021-11-18
Revert "Add a workaround for BZ 1875338"

zshi-redhat:
This reverts commit fa7011adc1d6a43210f6de95ae05090fcab57838.

Reverts since the aforementioned bz is fixed

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -276,35 +276,6 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 					glog.Warningf("configSriovDevice(): fail to bind default driver for device %s", addr)
 					return err
 				}
-
-				// validate the device exist only with the default driver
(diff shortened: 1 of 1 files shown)


pkg/utils/utils.go
```diff
@@ -276,35 +276,11 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
-
-					if err := BindDefaultDriver(addr); err != nil {
-						glog.Warningf("configSriovDevice(): fail to rebind default driver for device %s", addr)
-						return err
-					}
-
-					glog.Warningf("configSriovDevice(): W/A rebind for PCI %s", addr)
-
+				// keep VF MTU align with PF's
+				mtu := ifaceStatus.Mtu
```
pliurh, line 280: defined, but not used.

zshi-redhat: fixed

PR #205, opened 2021-11-21
Load tun kernel module

SchSeba:
The sriov-network-config-daemon will try to load the tun kernel driver when started

This PR is needed to continue to the PR merged in the sriov-device-plugin to allow the mount of tun device inside the container to create TAP devices.

The TAP devices are in use by dpdk applications that want to interact using a slow path with the Linux kernel networking.

### DIFF
--- a/test/conformance/tests/sriov_operator.go
+++ b/test/conformance/tests/sriov_operator.go
@@ -1136,7 +1136,7 @@ var _ = Describe("[sriov] operator", func() {
 						resourceName := "mainpfresource"
 						sriovDeviceList, err := sriovInfos.FindSriovDevices(testNode)
 						Expect(err).ToNot(HaveOccurred())
-						executorPod := createCustomTestPod(testNode, []string{}, true)
+						executorPod := createCustomTestPod(testNode, []string{}, true, nil)
 						mainDeviceForNode := findMainSriovDevice(executorPod, sriovDeviceList)
 						if mainDeviceForNode == nil {
 							Skip("Could not find pf used as gateway")
@@ -1572,7 +1572,7 @@ var _ = Describe("[sriov] operator", func() {
 
 		})
 
-		Context("vhost-net device Validation", func() {
+		Context("vhost-net and tun devices Validation", func() {
 			var node string
 			resourceName := "vhostresource"
 			vhostnetwork := "test-vhostnetwork"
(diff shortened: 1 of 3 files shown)


config/crd/bases/sriovnetwork.openshift.io_sriovnetworknodestates.yaml
```diff
@@ -50,6 +50,8 @@ spec:
                       type: integer
                     name:
                       type: string
+                    needVhostNet:
```
zshi-redhat, line 53: I prefer to not changing API and always loading the tun module like how rdma_core is installed, do you see any issue loading the tun module unconditionally?

SchSeba: sounds good to me :)

SchSeba:
@zshi-redhat done can you have another round of review?

BTW do you think we should do the same for the vfio-pci driver? just load it in the daemon initialization?

zshi-redhat:
> @zshi-redhat done can you have another round of review?
> 
> BTW do you think we should do the same for the vfio-pci driver? just load it in the daemon initialization?

It doesn't hurt to remove, but I think the current loading of vfio-pci module has more precise per-device control than tun or rdma modules. The main reason of not adding the same per-device control for tun module is that we don't want to change API in the backport or just for the module load. so I think we can leave the vfio-pci as it is.

zshi-redhat (approved):

zshi-redhat:
> Expose the Vhost field to the sriov-network-config-daemon, so it will be able to load the tun kernel driver when needed
> 
nit: this commit message need change to reflect current implemention.

> This PR is needed to continue to the PR merged in the sriov-device-plugin to allow the mount of tun device inside the container to create TAP devices.
> 
> The TAP devices are in use by dpdk applications that want to interact using a slow path with the Linux kernel networking.

SchSeba:
Hi @zshi-redhat I fixed the commits

zshi-redhat (approved):

test/conformance/tests/sriov_operator.go
```diff
@@ -2004,6 +2022,17 @@ func createCustomTestPod(node string, networks []string, hostNetwork bool) *k8sv
 			node,
 		)
 	}
+
+	if podCapabilies != nil && len(podCapabilies) != 0 {
```
Eoghan1232, line 2026: Should it be podCapabilies -> podCapabilities ?

test/conformance/tests/sriov_operator.go
```diff
@@ -1991,10 +2009,10 @@ func isPodConditionUnschedulable(pod *k8sv1.Pod, resourceName string) bool {
 }
 
 func createTestPod(node string, networks []string) *k8sv1.Pod {
-	return createCustomTestPod(node, networks, false)
+	return createCustomTestPod(node, networks, false, nil)
 }
 
-func createCustomTestPod(node string, networks []string, hostNetwork bool) *k8sv1.Pod {
+func createCustomTestPod(node string, networks []string, hostNetwork bool, podCapabilies []corev1.Capability) *k8sv1.Pod {
```
Eoghan1232, line 2015: podCapabilies -> podCapabilities ?

SchSeba:
@Eoghan1232 thanks I miss that!

Eoghan1232 (approved):
LGTM
Thanks for working on this!

SchSeba:
@pliurh can we merge this PR?

e0ne (approved):

PR #206, opened 2021-11-21
Remove the secure boot checker pod

SchSeba:
This commit removes the pod that is created to validate if the node has
secure boot enable or not

Before this PR after every test running there are test pods in the sriov
operator namespace that does not get clean


Eoghan1232 (approved):
lgtm

e0ne (approved):

test/util/cluster/cluster.go
```diff
@@ -308,5 +309,10 @@ func GetNodeSecureBootState(clients *testclient.ClientSet, nodeName, namespace s
 		return false, err
 	}
 
+	err = clients.Pods(namespace).Delete(context.Background(), runningPod.Name, metav1.DeleteOptions{GracePeriodSeconds: pointer.Int64Ptr(0)})
```
adrianchiris, line 312: should we use `defer` to delete the pod in case of early return ?

adrianchiris: @SchSeba  WDYT ? :)

SchSeba: sorry for the late response :)

adrianchiris (approved):

PR #207, opened 2021-11-22
Reload VF driver only when rdma is enabled

pliurh:
To set the GUID to VF, we need to set the admin MAC and reload the VF driver. It will take a long time if users set big 'numVFs'. This PR changes the logic to only reload the VF driver when `isRdma` is set. So that, users not using ROCE will no longer need to wait for the VF driver reload.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -233,13 +233,6 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 			glog.Errorf("configSriovDevice(): fail to set NumVfs for device %s", iface.PciAddress)
 			return err
 		}
-		if strings.EqualFold(iface.LinkType, "IB") {
-			if err = setVfsGuid(iface); err != nil {
-				return err
-			}
-		} else if err = setVfsAdminMac(ifaceStatus); err != nil {
-			return err
-		}
 	}
 	// set PF mtu
 	if iface.Mtu > 0 && iface.Mtu != ifaceStatus.Mtu {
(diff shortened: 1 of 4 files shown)


Mmduh-483:
@moshe010 @adrianchiris

zshi-redhat:
@pliurh Will this fix be backported to lower versions? If so, I'd prefer to not changing API. My understanding is we will eventually remove this once kernel fix is in place.

pliurh:
> @pliurh Will this fix be backported to lower versions? If so, I'd prefer to not changing API. My understanding is we will eventually remove this once kernel fix is in place.

If we want to fix that 4.6 bug, we need to backport it. Without modifying API, the config daemon cannot know if a VF will be used for ROCE. I don't know when it will be fixed in the kernel.

pkg/utils/utils.go
```diff
@@ -558,35 +566,25 @@ func vfIsReady(pciAddr string) (netlink.Link, error) {
 	return vfLink, nil
 }
 
-func setVfsAdminMac(iface *sriovnetworkv1.InterfaceExt) error {
-	glog.Infof("setVfsAdminMac(): device %s", iface.PciAddress)
-	pfLink, err := netlink.LinkByName(iface.Name)
+func setVfsAdminMac(vfAddr string, pfLink netlink.Link, isRdma bool) error {
```
adrianchiris, line 569: imo the name: setVFAdminMac or setVFAdminMAC is more appropriate now.

pkg/utils/utils.go
```diff
@@ -558,35 +566,25 @@ func vfIsReady(pciAddr string) (netlink.Link, error) {
...
-		}
-		if err = Unbind(addr); err != nil {
-			return err
-		}
-		if err = BindDefaultDriver(addr); err != nil {
+	if err := netlink.LinkSetVfHardwareAddr(pfLink, vfID, vfLink.Attrs().HardwareAddr); err != nil {
+		return err
+	}
+	if isRdma {
+		glog.Infof("setVfsAdminMac(): unbind driver for %s", vfAddr)
```
adrianchiris, line 586: should we add a method to handle the driver unbind and move the unbind logic out of SetVfAdminMac and SetVfGuid ?
so that these functions do exactly what their name say ? 

e.g add a function : `UnbindDriverIfNeeded()` and call it from `configureSriovDevice` ?

pkg/utils/utils.go
```diff
@@ -613,36 +611,22 @@ func getLinkType(ifaceStatus sriovnetworkv1.InterfaceExt) string {
 	return ""
 }
 
-func setVfsGuid(iface *sriovnetworkv1.Interface) error {
-	glog.Infof("setVfsGuid(): device %s", iface.PciAddress)
-	pfLink, err := netlink.LinkByName(iface.Name)
+func setVfsGuid(vfAddr string, pfLink netlink.Link) error {
```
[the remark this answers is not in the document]
pliurh: fixed

pkg/utils/utils.go
pliurh: fixed

pkg/utils/utils.go
pliurh: fixed

adrianchiris (approved):
LGTM, thanks Peng !

When merging please add the description of this PR to the merge commit so its clear why this change was introduced

PR #208, opened 2021-11-29
OVS HWOL: set flow steering mode for eswitch device

zshi-redhat:
Add a knob for steering mode configuration, defaults to dmfs

### DIFF
--- a/bindata/manifests/switchdev-config/files/configure-switchdev.sh.yaml
+++ b/bindata/manifests/switchdev-config/files/configure-switchdev.sh.yaml
@@ -54,14 +54,19 @@ contents:
         # Create udev rule to save PF name
         add_udev_rule_for_sriov_pf $name
 
-        echo "unload VF driver for $pci_addr"
+        echo "Unload VF driver for $pci_addr"
         VfDirs=$(ls /sys/bus/pci/devices/${pci_addr} | grep virtfn)
         for VfDir in $VfDirs
         do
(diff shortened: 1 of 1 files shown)


pliurh:
Can you add a description to the commit? IIUC, the `flow_steering_mode` flag is only valid for Mellanox NICs, shall we render this conditionally?

adrianchiris:
> Can you add a description to the commit? IIUC, the flow_steering_mode flag is only valid for Mellanox NICs, shall we render this conditionally?

ATM also the code above it (unbind mlx5_core) is only valid for mlnx NIC so not sure if this bit is required to be rendered conditionally.

LGTM from my side, A bit more context in commit message is appreciated :)

adrianchiris (approved):

zshi-redhat:
> > Can you add a description to the commit? IIUC, the flow_steering_mode flag is only valid for Mellanox NICs, shall we render this conditionally?

Updated the commit message.
> 
> ATM also the code above it (unbind mlx5_core) is only valid for mlnx NIC so not sure if this bit is required to be rendered conditionally.

Yeah, that's what I'm going to say, I agree that we need to remove the hardcoded driver. I think this can be done in a separate PR.
> 
> LGTM from my side, A bit more context in commit message is appreciated :)

pliurh:
> Updated the commit message.

I cannot see the commit message.

zshi-redhat:
> > Updated the commit message.
> 
> I cannot see the commit message.

I updated in this PR which would appear in the merge log.
Also updated in the commit message this time.

PR #209, opened 2021-12-02
Add pr-ci-triggers workflow

AbdYsn:
Added the pr-ci-triggers github workflow to print out available
vendors CIs trigger phrases.

### DIFF
--- a/.github/workflows/pr-ci-triggers.yml
+++ b/.github/workflows/pr-ci-triggers.yml
@@ -1,0 +1,31 @@
+on:
+  pull_request_target:
+
+jobs:
+  vendors-ci-triggers-list:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v5
+        env:
+          MESSAGE: >-
+            Thanks for your PR,\n
+            To run vendors CIs use one of:\n
+            * `/test-all`: To run all tests for all vendors.\n
+            * `/test-e2e-all`: To run all E2E tests for all vendors.\n
+            * `/test-e2e-nvidia-all`: To run all E2E tests for NVIDIA vendor.\n
+            \n
+            To skip the vendors CIs use one of:\n
+            * `/skip-all`: To skip all tests for all vendors.\n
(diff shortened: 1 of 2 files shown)


AbdYsn:
There seem to be a limitation in github where forks are not allowed to read `secrets` and so it is not possible to comment on the PR.

.github/workflows/pr-ci-triggers.yml
```diff
@@ -0,0 +1,24 @@
+on:
+  pull_request:
+
+jobs:
+  test:
```
adrianchiris, line 5: job name : `vendor-ci` ?

ci/examples/jenkins/sriov-network-operator-ci.yaml
```diff
@@ -59,7 +59,7 @@
           failure-status: "Build Failed, comment `/test-e2e-all`, /test-e2e-nvidia-all, or `/test-all` to retrigger"
           error-status: "Build Failed, comment `/test-e2e-all`, /test-e2e-nvidia-all, or `/test-all` to retrigger"
           status-context: '{project} CI'
-          trigger-phrase: ".*/test-(all|e2e-all|e2e-nvidia-all(,| |$)).*"
+          trigger-phrase: "^(?!Thanks for your PR).*/test-(all|e2e-all|e2e-nvidia-all(,| |$)).*"
```
adrianchiris, line 62: do you need this change ? the message does not contain the "full" trigger i.e `/<trigger-name>` but rather just `<trigger-name>`

AbdYsn: Actually, it should contain the full trigger, to make copy pasting it easier

.github/workflows/pr-ci-triggers.yml
AbdYsn: switched it to `vendors-ci-triggers-list`, since i think this would be more descriptive, WDYT?

adrianchiris (approved):

e0ne (approved):

PR #210, opened 2021-12-02
validate: allow netFilter to be set alone in NicSelector

EmilienM:
In `staticValidateSriovNetworkNodePolicy`, we now allow a NicSelector to
only have NetFilter defined.

This will be useful for the OpenStack platform, where we can request a
Network Node Policy only by its Network ID.

### DIFF
--- a/pkg/webhook/validate_test.go
+++ b/pkg/webhook/validate_test.go
@@ -2,9 +2,10 @@ package webhook
 
 import (
 	"fmt"
-	corev1 "k8s.io/api/core/v1"
 	"os"
 	"testing"
+
+	corev1 "k8s.io/api/core/v1"
 
 	. "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


pkg/webhook/validate_test.go
```diff
@@ -498,6 +499,52 @@ func TestStaticValidateSriovNetworkNodePolicyWithInvalidNicSelector(t *testing.T
...
+				"feature.node.kubernetes.io/network-sriov.capable": "true",
+			},
+			NumVfs:       63,
+			Priority:     99,
+			ResourceName: "p0",
+		},
+	}
+	g := NewGomegaWithT(t)
+	ok, err := validatePolicyForNodeState(policy, state, NewNode())
+	g.Expect(err).NotTo(HaveOccurred())
```
atyronesmith, line 544: Is this the correct logic?  Shouldn't you expect an error?

EmilienM: oops yes indeed :(
I'll check thanks.

atyronesmith (approved):
lgtm

zshi-redhat (approved):

zshi-redhat:
> @zshi-redhat @pliurh @atyronesmith could I backport it to older releases, I would like to have it for 4.8 if possible.
> 
> Thanks

@EmilienM I think it could, I don't see any obvious risk backporting this fix.

EmilienM:
@zshi-redhat thanks. Done via https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/212

PR #213, opened 2021-12-07
Better support for openshift single node

SchSeba:
This PR allows to still pause the MCP when running on OCP even if it's a single node with no drain option.

Another change is to the sriovOperatorConfig to mark the DisableDrain when there is only one node in the cluster

### DIFF
--- a/controllers/sriovoperatorconfig_controller.go
+++ b/controllers/sriovoperatorconfig_controller.go
@@ -78,6 +78,11 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 		Name: constants.DEFAULT_CONFIG_NAME, Namespace: namespace}, defaultConfig)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			singleNode, err := utils.IsSingleNodeCluster(r.Client)
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("Couldn't check the anount of nodes in the cluster")
+			}
+
 			// Default Config object not found, create it.
 			defaultConfig.SetNamespace(namespace)
 			defaultConfig.SetName(constants.DEFAULT_CONFIG_NAME)
@@ -86,7 +91,9 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 				EnableOperatorWebhook:    func() *bool { b := enableAdmissionController; return &b }(),
 				ConfigDaemonNodeSelector: map[string]string{},
 				LogLevel:                 2,
+				DisableDrain:             singleNode,
 			}
+
 			err = r.Create(context.TODO(), defaultConfig)
 			if err != nil {
 				logger.Error(err, "Failed to create default Operator Config", "Namespace",
--- a/main.go
+++ b/main.go
@@ -232,6 +232,12 @@ func createDefaultOperatorConfig(cfg *rest.Config) error {
 	if err != nil {
 		return fmt.Errorf("Couldn't create client: %v", err)
 	}
+
+	singleNode, err := utils.IsSingleNodeCluster(c)
+	if err != nil {
+		return fmt.Errorf("Couldn't check the anount of nodes in the cluster")
+	}
+
 	enableAdmissionController := os.Getenv("ENABLE_ADMISSION_CONTROLLER") == "true"
 	config := &sriovnetworkv1.SriovOperatorConfig{
 		Spec: sriovnetworkv1.SriovOperatorConfigSpec{
@@ -239,6 +245,7 @@ func createDefaultOperatorConfig(cfg *rest.Config) error {
 			EnableOperatorWebhook:    func() *bool { b := enableAdmissionController; return &b }(),
 			ConfigDaemonNodeSelector: map[string]string{},
 			LogLevel:                 2,
+			DisableDrain:             singleNode,
 		},
 	}
 	name := "default"
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -498,24 +498,26 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 	}
 
 	if reqDrain {
-		ctx, cancel := context.WithCancel(context.TODO())
-		defer cancel()
+		if !dn.disableDrain {
+			ctx, cancel := context.WithCancel(context.TODO())
+			defer cancel()
 
-		glog.Infof("nodeStateSyncHandler(): get drain lock for sriov daemon")
-		done := make(chan bool)
-		go dn.getDrainLock(ctx, done)
-		<-done
-
-		glog.Infof("nodeStateSyncHandler(): pause MCP")
-		if err := dn.pauseMCP(); err != nil {
-			return err
+			glog.Infof("nodeStateSyncHandler(): get drain lock for sriov daemon")
+			done := make(chan bool)
+			go dn.getDrainLock(ctx, done)
+			<-done
 		}
 
-		if !dn.disableDrain {
-			glog.Info("nodeStateSyncHandler(): drain node")
-			if err := dn.drainNode(); err != nil {
+		if utils.ClusterType == utils.ClusterTypeOpenshift {
+			glog.Infof("nodeStateSyncHandler(): pause MCP")
+			if err := dn.pauseMCP(); err != nil {
 				return err
 			}
+		}
+
+		glog.Info("nodeStateSyncHandler(): drain node")
+		if err := dn.drainNode(); err != nil {
+			return err
 		}
 	}
 
@@ -819,13 +821,8 @@ func (dn *Daemon) getDrainLock(ctx context.Context, done chan bool) {
 }
 
 func (dn *Daemon) pauseMCP() error {
-	glog.Info("pauseMCP(): check if pausing MCP is possible")
+	glog.Info("pauseMCP(): pausing MCP")
 	var err error
-
-	if utils.ClusterType != utils.ClusterTypeOpenshift {
-		glog.Infof("pauseMCP(): skipping MCP pause as the cluster is not an openshift cluster")
-		return nil
-	}
 
 	mcpInformerFactory := mcfginformers.NewSharedInformerFactory(dn.mcClient,
 		time.Second*30,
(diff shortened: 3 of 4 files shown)


controllers/sriovoperatorconfig_controller.go
```diff
@@ -78,6 +78,14 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
...
 	if err != nil {
 		if errors.IsNotFound(err) {
+			// Check if we only have one node
+			nodeList := &corev1.NodeList{}
+			err = r.List(context.TODO(), nodeList)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+
+			disableDrain := len(nodeList.Items) == 1
```
browsell, line 88: We should check that the the controlPlaneTopology = SingleReplica

controllers/sriovoperatorconfig_controller.go
```diff
@@ -78,6 +78,14 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 		Name: constants.DEFAULT_CONFIG_NAME, Namespace: namespace}, defaultConfig)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			// Check if we only have one node
```
zshi-redhat, line 81: This error won't be triggered unless user delete the default SriovOperatorConfig explicitly. Is it what we want? The default SriovOperatorConfig is created by OperatorConfig controller: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/main.go#L154

Do we need to consider the upgrade scenario? e.g. from lower version to 4.10 which contains this fix. If not, I think adding the disableDrain setting in the above 154 line shall be sufficient.

controllers/sriovoperatorconfig_controller.go
SchSeba: The `machineconfiguration.openshift.io/controlPlaneTopology: SingleReplica` is a MCO label. I want this feature to work also for plain k8s

controllers/sriovoperatorconfig_controller.go
SchSeba: @zshi-redhat thanks! I miss that file!

I will like to also work on updates so I put the change in both places.

The reason I put it on both places is a was able to make a race from when I install the operator + a policy and until there is a reconcile to add the skip

controllers/sriovoperatorconfig_controller.go
```diff
@@ -99,6 +100,25 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	nodeList := &corev1.NodeList{}
+	err = r.List(context.TODO(), nodeList)
+	if err != nil {
+		logger.Error(err, "Failed to list nodes")
+		return reconcile.Result{}, err
+	}
+
+	// Update the disableDrain field if needed
+	if len(nodeList.Items) == 1 && !defaultConfig.Spec.DisableDrain {
+		defaultConfig.Spec.DisableDrain = true
```
e0ne, line 113: Please, add logging here. We need to let users know that configuration is changed

main.go
```diff
@@ -232,13 +233,23 @@ func createDefaultOperatorConfig(cfg *rest.Config) error {
 	if err != nil {
 		return fmt.Errorf("Couldn't create client: %v", err)
 	}
+
+	// Check if we only have one node
```
e0ne, line 237: Actually this commit message doesn't align with a code below

controllers/sriovoperatorconfig_controller.go
```diff
@@ -99,6 +100,25 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 		return reconcile.Result{}, err
 	}
 
+	// Check if we only have one node
```
e0ne, line 103: The same comment a for main.go change. Do we want to move this check to pkg.utils module as 'GetNodesConut' function?

e0ne (changes requested):

controllers/sriovoperatorconfig_controller.go
adrianchiris: this is a one way switch, what if the cluster scales down to one then scales up to more that 1 node ?

pkg/daemon/daemon.go
```diff
@@ -497,11 +497,26 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
...
+		done := make(chan bool)
+		go dn.getDrainLock(ctx, done)
+		<-done
+
+		glog.Infof("nodeStateSyncHandler(): pause MCP")
+		if err := dn.pauseMCP(); err != nil {
 			return err
 		}
+
+		if !dn.disableDrain {
```
pliurh, line 514: It is better to keep this check consistent between function `drainNode` and `completeDrain`. Either put this check in both the functions or out of both.

controllers/sriovoperatorconfig_controller.go
SchSeba: done can you please have another look?

pkg/daemon/daemon.go
SchSeba: done can you please have another look?

SchSeba:
Hi @e0ne @pliurh @adrianchiris can you please give this another look?

controllers/sriovoperatorconfig_controller.go
adrianchiris: > This function is only if the config doesn't exist this means on install time or if the user for some reason removes this file.

This is not under the if block at L80, so it will be executed every time Reconcile is called.
maybe user put some configuration, then cluster scaled down (to 1) and operator restarted. this logic would then change the user configuration. later on cluster scaled up and we ended up with drain disabled.

So Here are my thoughts:
1. we document that in case of a single Node cluster, the user should explicitly set DisableDrain 
2. we do this logic only when creating the default object (within if statement block in L80)

pkg/daemon/daemon.go
```diff
@@ -801,99 +816,109 @@ func (dn *Daemon) getDrainLock(ctx context.Context, done chan bool) {
...
+func (dn *Daemon) pauseMCP() error {
+	glog.Info("pauseMCP(): check if pausing MCP is possible")
 	var err error
-	ctx, cancel := context.WithCancel(context.TODO())
-	defer cancel()
 
-	done := make(chan bool)
-	go dn.getDrainLock(ctx, done)
-	<-done
+	if utils.ClusterType != utils.ClusterTypeOpenshift {
```
adrianchiris, line 823: should we move (and inverse) this check to daemon.go L510 ? and only call pauseMCP in openshift cluster ?

pkg/utils/cluster.go
```diff
@@ -0,0 +1,26 @@
...
+func IsSingleNodeCluster(c client.Client) (bool, error) {
+	nodeList := &corev1.NodeList{}
+	err := c.List(context.TODO(), nodeList)
+	if err != nil {
+		glog.Errorf("Failed to list nodes: %v", err)
+		return false, err
+	}
+
+	if len(nodeList.Items) == 1 {
+		glog.Infof("only one node found in the cluster marking disableDrain as true")
```
adrianchiris, line 21: nit: i dont think this log is needed.

if you want to keep it, please update the message as disableDrain is now out of context.

controllers/sriovoperatorconfig_controller.go
adrianchiris: nit: i think reading L104 is pretty clear, you can drop this comment if you like.

main.go
adrianchiris: nit: i think reading L237 is pretty clear, you can drop this comment if you like.

adrianchiris:
@SchSeba might be a dumb question, but why do we want to disable drain on a single node cluster ? :)
critical services are daemonsets (usually) and these wont go away right ?

SchSeba:
Hi @adrianchiris there are cases when the user have a 1 node cluster and deploy some applications using `pod disruption budget` for example that is the case in openshift we have some platform pods using the PDB so the drain will never finish and the operator will just stuck trying to drain the node.

controllers/sriovoperatorconfig_controller.go
SchSeba: I also prefer 2 changing the code as requested

main.go
SchSeba: done

controllers/sriovoperatorconfig_controller.go
SchSeba: done

pkg/daemon/daemon.go
SchSeba: sure no problem

pkg/utils/cluster.go
SchSeba: I will like to leave the message makes debug easier :)

I just change the comment

pkg/daemon/daemon.go
```diff
@@ -497,9 +497,24 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
...
 	}
 
-	if reqDrain && !dn.disableDrain {
+	if reqDrain {
+		ctx, cancel := context.WithCancel(context.TODO())
+		defer cancel()
+
+		glog.Infof("nodeStateSyncHandler(): get drain lock for sriov daemon")
+		done := make(chan bool)
+		go dn.getDrainLock(ctx, done)
```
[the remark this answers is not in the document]
pliurh: I think it is a minor issue. If the cluster is a single node cluster, there is no other node that needs to wait for the lock. For multi-node clusters, this flag is mainly for testing. The drawback is that, as the `LeaseDuration` is 5s, we will have to waste 5s on each node when it requires draining.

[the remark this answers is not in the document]
adrianchiris: ack thx @pliurh  ! id prefer to condition on (disableDrain or Openshift ) to avoid api calls for the lease and the 5s delay per node.

main.go
adrianchiris: > done

done and not removed ? :)

pkg/daemon/daemon.go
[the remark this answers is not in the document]
SchSeba: @adrianchiris done please have a look I also test it internally

main.go
SchSeba: sorry about that

adrianchiris:
Failure related to #218

zshi-redhat:
> 

Fix proposed in #219

SchSeba:
Hi @adrianchiris anything we are missing here or we can merge this PR?

adrianchiris (approved):
Thanks for the reminder @SchSeba ! LGTM

PR #220, opened 2021-12-21
HA/SNO leader election config

eggfoobar:
Hi there, the purpose of this PR is to configure leader election to be driven by some conventions that allow for better behavior on SingleNode environments as well as normal HA clusters.

Currently leader election is defaulting to the Kubernetes defaults which are `leaseDuration: 15s` `renewDeadline: 15s` and a `retryPeriod: 2s`. On a typically well suited machine this is not much of a problem and cheap. However as resources become strapped this cadence starts to show some flaws and in situations where the `api-server` is down for longer than usual due to limited CPU bandwidth then the inability to get a response from the api-server consecutively will cause a restart. The new defaults allow for a more generous down time tolerance of (78s) and a more reasonable retry cadence of (26s) that keeps things well behaved in both respects of responsiveness during roll out and  stability.

The second path is for SingleNode situations, in those environments you are typically only running a single machine with much more strapped resources. Furthermore in stricter situations, like edge deployments, management workloads can be pinned down to a limited number of cores, and anything to lesson traffic against the `api-server` becomes a win. The config changes for those situations follow the same as HA environments but with a (270s) lease duration and a retry cadence of (60s) this gives us a much higher tolerance for `api-server` downtime at (180s).

Any and all feedback welcome, thanks!


pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,77 @@
...
+
+	if enabled {
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Second*3))
+		defer cancel()
+		infra, err := getClusterInfraStatus(ctx, restClient)
+		if err != nil {
+			glog.Warningf("unable to get cluster infrastructure status, using HA cluster values for leader election: %v", err)
+			return
+		}
+		if infra != nil && infra.ControlPlaneTopology == configv1.SingleReplicaTopologyMode {
```
zshi-redhat, line 45: `infra !=nil ` is checked in `getClusterInfraStatus`.

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,77 @@
...
+	return ret
+}
+
+// Retrieve the cluster status, used to determine if we should use different leader election.
+func getClusterInfraStatus(ctx context.Context, restClient *rest.Config) (*configv1.InfrastructureStatus, error) {
+	client, err := openshiftcorev1.NewForConfig(restClient)
+	if err != nil {
+		return nil, err
+	}
+	infra, err := client.Infrastructures().Get(ctx, infraResourceName, metav1.GetOptions{})
```
zshi-redhat, line 69: I guess this is specific to openshift, right?  does it work for native kubernetes?

qJkee: @zshi-redhat 
Yeah, that's true. Is there any way to check if we are in `Openshift` env?

SchSeba:
general question. we already support a single node in the operator for both k8s and ocp can you just add a function under https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/cluster.go and use the already existing function to understand if it's a single node cluster?

I don't think we need to import a package (ocp/api) just to use one function there.

Last point why the ocp/api doesn't have a function that returns the timer? this way if in the future there is a need to change the timer we just need to bump the package version and not change anything else on any project using this package.

pkg/leaderelection/leaderelection.go
zshi-redhat: Cluster type (openshift or kubernetes) is passed in as an environment variable when building the Dockerfile image: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/Dockerfile.rhel7#L16. Here is an example: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/cmd/sriov-network-config-daemon/start.go#L87

eggfoobar:
> general question. we already support a single node in the operator for both k8s and ocp can you just add a function under https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/cluster.go and use the already existing function to understand if it's a single node cluster?

That's a great idea, moved the logic there and used @zshi-redhat suggestion of checking against the `CLUSTER_TYPE` env for openshift and k8s clusters

> I don't think we need to import a package (ocp/api) just to use one function there.

I agree, I wasn't sure how best to approach this, to avoid any future mistakes figured it was best to pull in `ocp/api` and use the shared `Infrastructure` object.

> Last point why the ocp/api doesn't have a function that returns the timer? this way if in the future there is a need to change the timer we just need to bump the package version and not change anything else on any project using this package.

Correct me if I misunderstood your question @SchSeba , right now the functionality that returns default leader election timers for both regular and SNO clusters is part of `openshift/library-go`, we could pull in that pkg but that will force more dependency changes. The intent here was to keep this simple and copying those values seemed more manageable for this PR.

pkg/utils/cluster.go
```diff
@@ -21,6 +37,17 @@ func IsSingleNodeCluster(c client.Client) (bool, error) {
 		glog.Infof("IsSingleNodeCluster(): one node found in the cluster")
 		return true, nil
 	}
-
 	return false, nil
 }
+
+func openshiftSNOClusterStatus(c client.Client) (bool, error) {
```
SchSeba, line 43: question are we still need this function?

if the CLUSTER_TYPE is openshift and we have 1 node can we call it an SNO?

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,54 @@
+package leaderelection
+
+import (
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
```
SchSeba, line 7: small nit: can you move this as the last import?

eggfoobar: absolutely, updated

SchSeba:
@eggfoobar just a general question, is the access to the OCP API request any change in the operator RBAC's?

eggfoobar:
> @eggfoobar just a general question, is the access to the OCP API request any change in the operator RBAC's?

Thank you! Yes, I missed that. I updated the RBAC in the in the deployments where it seemed to be located.

pliurh:
@eggfoobar Could you squash the commits before we merge it?

eggfoobar:
@pliurh Sure thing, squashed it into two commits, is that okay or would you rather it be one?

pliurh:
two is ok.

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,53 @@
...
+			return
+		}
+		if isSno {
+			return leaderElectionSNOConfig(defaultConfig)
+		}
+	}
+	return
+}
+
+// Default leader election for SNO environments
```
adrianchiris, line 44: SNO also (might) stand for Sriov Network Operator
And i assume here it stands for Single Node Openshift

also IsSingleNodeCluster in L32 means if the cluster is a single node cluster, regardless of vanilla or openshift

can we rename this method to `leaderElectionSingleNodeConfig` ?  or similar ? as this configuration is platform agnostic (i.e same for openshift or vanilla k8s)

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,53 @@
...
+func GetLeaderElectionConfig(c client.Client, enabled bool) (defaultConfig configv1.LeaderElection) {
+	defaultConfig = configv1.LeaderElection{
+		Disable:       !enabled,
+		LeaseDuration: metav1.Duration{Duration: defaultLeaseDuration},
+		RenewDeadline: metav1.Duration{Duration: defaultRenewDeadline},
+		RetryPeriod:   metav1.Duration{Duration: defaultRetryPeriod},
+	}
+
+	if enabled {
+		isSno, err := utils.IsSingleNodeCluster(c)
```
adrianchiris, line 32: isSingleNode ?

if sno stands for single node openshift, thats not the case here as this bit will run on vanilla k8s as well

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,53 @@
+package leaderelection
+
+import (
+	"time"
+
+	"github.com/golang/glog"
+	configv1 "github.com/openshift/api/config/v1"
```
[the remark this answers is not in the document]
eggfoobar: Sure thing, updated to use the client-go leader election struct

pkg/leaderelection/leaderelection.go
eggfoobar: Thanks for the input, totally agree. Updated the wording to make it clear

pkg/leaderelection/leaderelection.go
eggfoobar: Updated as well

pkg/utils/cluster.go
```diff
@@ -2,14 +2,30 @@ package utils
...
+const (
+	// default Infrastructure resource name for Openshift
+	infraResourceName = "cluster"
+)
+
 func IsSingleNodeCluster(c client.Client) (bool, error) {
+	if os.Getenv("CLUSTER_TYPE") == ClusterTypeOpenshift {
+		return openshiftSNOClusterStatus(c)
+	}
+	return k8sSNOClusterStatus(c)
```
SchSeba, line 25: nit: can you change the name also here to not have `SNO`

eggfoobar: Sure thing, done!

e0ne (commented):
LGTM

e0ne (approved):

adrianchiris (approved):

PR #222, opened 2022-01-10
Remove SriovNetworkNodeConfigPolicy from docs

zeeke:
While going through the documentation I found that README.md was referring the type `SriovNetworkNodeConfigPolicy`, but the correct name of the CR is `SriovNetworkNodePolicy`.

Let me know if I got wrong while interpreting the docs.

Regards,


SchSeba:
Thanks for the contribution!

/lgtm

PR #223, opened 2022-01-13
Update base image version to use CentOS 8

e0ne:


adrianchiris (approved):

adrianchiris (approved):

zshi-redhat (approved):

SchSeba:
@e0ne question will it be possible to use golang:1.17 instead of 1.16?

e0ne:
@SchSeba I'll test with 1.17 and create PR

SchSeba:
@e0ne thanks!

PR #224, opened 2022-01-14
Ensure switchdev service dependencies

zshi-redhat:
It is found that VF netdev disappears after it was successfully
created (observed via ip link show). While investigating the root
cause, setting addtional dependency in NetworkManager.service
as temporary fix to the issue.

This commits also changes service and script names to lower case

### DIFF
--- a/bindata/manifests/switchdev-config/switchdev-units/NetworkManager.service.yaml
+++ b/bindata/manifests/switchdev-config/switchdev-units/NetworkManager.service.yaml
@@ -3,5 +3,7 @@ dropins:
 - name: 10-switchdev.conf
   contents: |
     [Unit]
-    Wants=switchdev-configuration-before-NM.service
-    After=switchdev-configuration-before-NM.service
+    Wants=switchdev-configuration-before-nm.service
+    Wants=switchdev-configuration-after-nm.service
+    After=switchdev-configuration-before-nm.service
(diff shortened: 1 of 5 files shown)


bindata/manifests/switchdev-config/switchdev-units/NetworkManager.service.yaml
```diff
@@ -3,5 +3,7 @@ dropins:
 - name: 10-switchdev.conf
   contents: |
     [Unit]
-    Wants=switchdev-configuration-before-NM.service
-    After=switchdev-configuration-before-NM.service
+    Wants=switchdev-configuration-before-nm.service
+    Wants=switchdev-configuration-after-nm.service
```
pliurh, line 7: Do we need this line?

zshi-redhat: From the testing, yes, this is required to ensure VF netdev driver is binded after reboot

e0ne (approved):

bindata/manifests/switchdev-config/switchdev-units/NetworkManager.service.yaml
zshi-redhat: @pliurh wdyt?

PR #227, opened 2022-01-14
Add Dockerfiles with Go 1.17 and CentOS Stream 9

fabiendupont:
The current version of Go is 1.17, so this allows validating that the
build chain works with this new version. It also introduces the CentOS
Stream 9 base image for the final image. This allows checking future
releases of Red Hat Enterprise Linux 9 ahead of time.

Signed-off-by: Fabien Dupont <fabiendupont@pm.me>

### DIFF
--- a/.github/workflows/pr-ci-build-container-images-cs9.yml
+++ b/.github/workflows/pr-ci-build-container-images-cs9.yml
@@ -1,0 +1,86 @@
+name: "Test build of CentOS Stream 9 based container images"
+on:
+  pull_request:
+    branches:
+      - master
+jobs:
+  test-build-image-operator:
+    runs-on: ubuntu-20.04
+    env:
+      IMAGE_NAME: ghcr.io/${{ github.repository }}
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Docker
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push sriov-network-operator
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: false
+          tags: |
+            ${{ env.IMAGE_NAME }}:latest
+          file: ./Dockerfile
+
+  test-build-image-config-daemon:
+    runs-on: ubuntu-20.04
+    env:
+      IMAGE_NAME: ghcr.io/${{ github.repository }}-config-daemon
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Docker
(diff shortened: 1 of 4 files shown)


e0ne:
Do we want to publish new images to GHCR? If yes, please create GitHub actions for it

adrianchiris:
> Do we want to publish new images to GHCR? If yes, please create GitHub actions for it

I dont think we want to publish an additional set of images we should rely on a "stable" image.

if we want to have an additional set of dockerfiles, maybe have a github action that test image build with these images ?

fabiendupont:
I agree with Adrian that it might not be desirable to publish the EL9 images on ghcr.io, just check that they properly build. To limit the number of builds, do you want to filter on some label on the PR?

adrianchiris:
> I agree with Adrian that it might not be desirable to publish the EL9 images on ghcr.io, just check that they properly build. To limit the number of builds, do you want to filter on some label on the PR?

Someone would need to put this label right ? im afraid we might forget.

maybe have a cron job for it ?
https://jasonet.co/posts/scheduled-actions/

personally i dont think its an issue to add a build test for PR

.github/workflows/pr-ci-build-container-images-cs9.yml
```diff
@@ -0,0 +1,87 @@
+name: "Test build of CentOS Stream 9 based container images"
+on:
+  pull_request:
+    branches:
+      - master
+      - main
```
e0ne, line 6: We don't have 'main' branch in this repository

fabiendupont: Indeed. I've removed it.

SchSeba (commented):
/lgtm

should we also have PRs for other sriov-operator components like the device plugin and the CNI (that is the one to be the most impacted one as it's the one running on the host itself)

fabiendupont:
> should we also have PRs for other sriov-operator components like the device plugin and the CNI (that is the one to be the most impacted one as it's the one running on the host itself)

Makes sense. Can you please provide links to the repositories and I can't take a look.

.github/workflows/pr-ci-build-container-images-cs9.yml
```diff
@@ -0,0 +1,86 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push sriov-network-operator
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: false
+          tags: |
+            ${{ env.IMAGE_NAME }}:latest
+          file: ./Dockerfile
```
zshi-redhat, line 32: Shall we use the `.el9` dockerfiles?
Or does it need to wait until this PR merged, otherwise CI would fail for this PR?

fabiendupont:
After discussing with @adrianchiris, it may become painful to add more distribution specific Dockerfiles. This repository maintains the code of the SRIOV Network Operator and building on top of other distributions can be done elsewhere.

I've created the [rh-ecosystem-edge/sriov-network-operator-build](https://github.com/rh-ecosystem-edge/sriov-network-operator-build) repository to build CentOS Stream 9 based images, and we'll surely add UBI based images as well. This separate repository has a GitHub Actions workflow that performs nightly builds and pushes them to Quay.io.

Closing this pull request. Thanks for the reviews and feedback.

Issue #230, opened 2022-01-19
Multiple SriovNetworkNodePolicy creation fails


e0ne:
Environment details:
- the latest SR-IOV Network Operator
- at least two worker nodes with 2 SR-IOV NICs on each of them: worker1 nic1, worker1 nic2, worker2 nic1, worker2 nic2
- SR-IOV is disabled on NICs

Steps to reproduce:
- create two yaml files with policies definition:
  - file1.yaml: it should contain policies for worker1 nic1 and worker2 nic1
  - file2.yaml: it should contain policies for worker1 nic2 and worker2 nic2
- apply file1.yaml 
- wait until worker1 starts reboot
- apply file2.yaml
- wait until worker1 started

Expected results:
All policies are applied

Actual results:
Part of policies applied. Leader elected for worker 2. Worker1 still has 'Draining' annotation, so config daemons do not proceed any configurations

SchSeba:
Hi @e0ne can you please check the sriov-network-config-daemon logs?

adrianchiris:
I believe we essentially hitting a deadlock in case worker1 after reboot needs to drain the node and worker 2 is holding the drain lock (is leader and waiting for draining annotation to be removed from node1).

Or generally speaking, any case where `nodeStateSyncHandler` is called in sriov-network-config-daemon while node has draining annot and during execution `reqDrain` is `true` and `disableDrain` is `false`.

zshi-redhat:
> I believe we essentially hitting a deadlock in case worker1, after reboot, needs to drain the node and worker 2 is holding the drain lock (is leader and waiting for draining annotation to be removed from node1).
> 
Wondering why worker-2 is able to get the drain lock while worker-1 still holds `Draining` annotation?

adrianchiris:
> Wondering why worker-2 is able to get the drain lock while worker-1 still holds Draining annotation?

as far as i saw there is no place in the daemon code that prevents it. as soon as `nodeStateSyncHandler()` config daemon releases leadership

adrianchiris:
> It gets the drainLock when dn.drainable is equal to true, but dn.drainable requires other nodes to not have Draining annotation.

[...]

PR #232, opened 2022-01-20
Continue node drain after reboot

e0ne:
Since drain operation started we don't need to requires
drain lock for this node because node already has required
annotation.

It's safe to continue node drain procedure without lock.

Closes: #230

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -498,20 +498,23 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 	}
 
 	if reqDrain {
-		if !dn.disableDrain {
-			ctx, cancel := context.WithCancel(context.TODO())
-			defer cancel()
+		if !dn.isNodeDraining() {
+			if !dn.disableDrain {
+				ctx, cancel := context.WithCancel(context.TODO())
+				defer cancel()
 
-			glog.Infof("nodeStateSyncHandler(): get drain lock for sriov daemon")
-			done := make(chan bool)
-			go dn.getDrainLock(ctx, done)
-			<-done
-		}
+				glog.Infof("nodeStateSyncHandler(): get drain lock for sriov daemon")
+				done := make(chan bool)
+				go dn.getDrainLock(ctx, done)
+				<-done
 
-		if utils.ClusterType == utils.ClusterTypeOpenshift {
-			glog.Infof("nodeStateSyncHandler(): pause MCP")
-			if err := dn.pauseMCP(); err != nil {
-				return err
+			}
+
+			if utils.ClusterType == utils.ClusterTypeOpenshift {
+				glog.Infof("nodeStateSyncHandler(): pause MCP")
+				if err := dn.pauseMCP(); err != nil {
+					return err
+				}
 			}
 		}
 
@@ -545,12 +548,12 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 		glog.Errorf("nodeStateSyncHandler(): fail to restart device plugin pod: %v", err)
 		return err
 	}
-	if anno, ok := dn.node.Annotations[annoKey]; ok && (anno == annoDraining || anno == annoMcpPaused) {
+	if dn.isNodeDraining() {
 		if err := dn.completeDrain(); err != nil {
 			glog.Errorf("nodeStateSyncHandler(): failed to complete draining: %v", err)
 			return err
 		}
-	} else if !ok {
+	} else {
 		if err := dn.annotateNode(dn.name, annoIdle); err != nil {
 			glog.Errorf("nodeStateSyncHandler(): failed to annotate node: %v", err)
 			return err
@@ -565,6 +568,13 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 	// wait for writer to refresh the status
 	<-dn.syncCh
 	return nil
+}
+
+func (dn *Daemon) isNodeDraining() bool {
+	if anno, ok := dn.node.Annotations[annoKey]; ok && (anno == annoDraining || anno == annoMcpPaused) {
+		return true
+	}
+	return false
 }
 
 func (dn *Daemon) completeDrain() error {
@@ -810,7 +820,7 @@ func (dn *Daemon) getDrainLock(ctx context.Context, done chan bool) {
 						done <- true
 						return
 					}
-					glog.V(3).Info("getDrainLock(): other node is draining, wait...")
+					glog.V(2).Info("getDrainLock(): other node is draining, wait...")
 				}
 			},
 			OnStoppedLeading: func() {
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -554,9 +554,11 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 			return err
 		}
 	} else {
-		if err := dn.annotateNode(dn.name, annoIdle); err != nil {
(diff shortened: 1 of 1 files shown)


e0ne:
Let's keep this PR in a WIP state because it requires additional testing, unit-tests implementation and most probably additional fixes

pkg/daemon/daemon.go
```diff
@@ -931,14 +941,14 @@ func (dn *Daemon) drainNode() error {
 		if err != nil {
 			lastErr = err
 			glog.Infof("Cordon failed with: %v, retrying", err)
-			return false, nil
+			return false, err
```
adrianchiris, line 944: this will cause the backoff loop to exit early since err != nil. I believe we want to keep retrying within the backoff loop in case of error like originally intended

pkg/daemon/daemon.go
```diff
@@ -931,14 +941,14 @@ func (dn *Daemon) drainNode() error {
...
 		}
 		err = drain.RunNodeDrain(dn.drainer, dn.name)
 		if err == nil {
-			return true, nil
+			glog.Infof("Draining failed with: %v, retrying", err)
```
adrianchiris, line 948: if err is nil it means drain succeeded so original code was good imo

e0ne: it makes sense, I'll revert this change

pliurh:
I tested this patch on OCP. It works.

pkg/daemon/daemon.go
```diff
@@ -545,12 +548,12 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
...
 		return err
 	}
-	if anno, ok := dn.node.Annotations[annoKey]; ok && (anno == annoDraining || anno == annoMcpPaused) {
+	if dn.isNodeDraining() {
 		if err := dn.completeDrain(); err != nil {
 			glog.Errorf("nodeStateSyncHandler(): failed to complete draining: %v", err)
 			return err
 		}
-	} else if !ok {
+	} else {
```
adrianchiris, line 556: we are now calling `dn.annotateNode()` in the else clause even if annotation is already set. this will invoke an extra GET to k8s API.

this IMO should be called like before only if annotation is not set WDYT?

pkg/daemon/daemon.go
```diff
@@ -498,20 +498,23 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
...
-		if !dn.disableDrain {
-			ctx, cancel := context.WithCancel(context.TODO())
-			defer cancel()
-
-			glog.Infof("nodeStateSyncHandler(): get drain lock for sriov daemon")
-			done := make(chan bool)
-			go dn.getDrainLock(ctx, done)
-			<-done
-		}
+		if !dn.isNodeDraining() {
```
zshi-redhat, line 501: This prevents daemon to go into `go dn.getDrainLock`, what if the other node is already in the go routine while the first node releases leader election lock?

I'm thinking the following scenario:
t1: node-1 gets leader election lock and node-2 is waiting for leader election to be released (node-2 daemon is already in the go routine loop for getting drain lock)
t2: node-1 daemon release leader election lock and reboot
t3: node-2 doesn't check line 501 since it is already in the go routine and gets the leader election lock.

pkg/daemon/daemon.go
e0ne: There is only two places where we add 'Idle' annotation: the line above and 'completeDrain' function, so IMO it doesn't make sense to check for it.

Probably we need to refactor the part of code where we do annotations to make it more clear

adrianchiris: it just to save a call to k8s API in annotateNode()

Think of a case where nodeStateSyncHandler runs but (for some reason) it does not require drain and the sriov-node-state annotation is already set to idle. in this case we dont really need to call annotateNode.

not sure how often we might hit this issue

pkg/daemon/daemon.go
zshi-redhat: you're right, I didn't notice that `isNodeDraining` checks for Draining annotations to avoid re-entering this clause after rebooting.

pkg/daemon/daemon.go
e0ne: got it, will fix it into 'annotateNode' function

zshi-redhat: Will there be a corner case where dn.node is not the latest if not get from annotateNode call?

pkg/daemon/daemon.go
```diff
@@ -711,6 +721,11 @@ func (a GlogLogger) Logf(format string, v ...interface{}) {
 func (dn *Daemon) annotateNode(node, value string) error {
 	glog.Infof("annotateNode(): Annotate node %s with: %s", node, value)
 
+	// Check if node already contains annotation
```
adrianchiris, line 724: IMO this common method should not change.

caller should determine (if relevant) wether or not to set the annotation based on `dn.node` state (line 553 above)

pkg/daemon/daemon.go
adrianchiris: dn.node will get updated via informer. it will eventually be consistent.

adrianchiris: I think this else clause aims to add idle annotation if annotation does not exist in `dn.node`. i prefer to keep this logic
(add a check : `hasSriovNodeStateAnnot()`) or similar.

pkg/daemon/daemon.go
```diff
@@ -545,15 +548,17 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
...
 		if err := dn.completeDrain(); err != nil {
 			glog.Errorf("nodeStateSyncHandler(): failed to complete draining: %v", err)
 			return err
 		}
-	} else if !ok {
-		if err := dn.annotateNode(dn.name, annoIdle); err != nil {
-			glog.Errorf("nodeStateSyncHandler(): failed to annotate node: %v", err)
-			return err
+	} else {
+		if !dn.nodeHasAnnotation(annoKey, annoIdle) {
```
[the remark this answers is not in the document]
adrianchiris: > but later on we might have additional state. thus, i think we should set annotation to idle only for case-1 only.
then if later on additional values are added this would not automatically reset annotation to idle.

i assume that if additional values are added it would need to be handled in other places as well. so its not an issue with this PR, im probably overthinking this :\

current approach is fine with me

adrianchiris (approved):

Eoghan1232 (approved):

pkg/daemon/daemon.go
zshi-redhat: > > Will there be a corner case where dn.node is not the latest if not get from annotateNode call?
> 
> dn.node will get updated via informer. 

I was thinking if there is delay between informer receive the node update event and dn.node be update
or a delay between node being update in api server and informer receive the node update event.

In the above case, dn.node is not the latest, which may result in inaccurate return from `nodeHasAnnotation`? (just thinking the possibility, I didn't test myself or it may be easy to test)

> it will eventually be consistent (thats how config daemon is designed).

If the `nodeHasAnnotation` returns incorrectly, I don't see how it can resolve by itself, guess the node will still maintain its old annotation forever unless new policy or state event?

adrianchiris: > If the nodeHasAnnotation returns incorrectly, I don't see how it can resolve by itself, guess the node will still maintain its old annotation forever unless new policy or state event?

i dont see how its different from what we had before, the change just moved the logic to a separate function.

your concern is valid theoretically, we have not encountered this. and its general for config daemon design.

zshi-redhat (approved):

Issue #81, opened 2021-03-04
[Enhancement] Support Generic SR-IOV configuration via system services


pliurh:
Adding this for further discussion and for this to be considered as an Enhancement for OpenShift 4.8

This Enhancement proposes to move generic_plugin SR-IOV configuration to boot time, to be performed by a system service,
Similarly to how switchdev is being configured today via SR-IOV network operator.

Doing so, When a node is rebooted, would not require a node to cordon, drain on boot as well as restart of device plugin as the machine will be properly configured on boot, making the node available to the k8s cluster faster and would keep pods running on the node after reboot in case they were created without a controller.

Proposal is to take a similar pattern to mco_plugin and create a sriov_config_service plugin to inject an sriov configuration service and its configuration to the node. This plugin would then be optionally enabled to replace the existing generic_plugin
via a configuration knob in SR-IOV config daemon.

mco_plugin can later be extended to support this functionality for OpenShift cluster, utilizing MCO to inject these services.

pliurh:
move here from https://github.com/openshift/sriov-network-operator/issues/438
@zshi-redhat @moshe010 @adrianchiris

pliurh:
There will some difficulty for Openshift. MCO uses MachineConfigPool to group the worker nodes with different MachineConfig. However, for one node, it can only belong to one MachineConfigPool. If we create a MachineConfigPool with the systemd service for SRIOV and put all SRIOV capable machines into it, users cannot create other custom pools for those workers.

This limitation also applies to the OVS HW offload case. But that's the only way for Openshift to config the switchdev devices, we don't have other choice.

zshi-redhat:
This might be desired in smart-nic case, where ovnkube-node running in host mode requires existence of VF0 to be the ovn mgmt port. Creating VFs via systemd service would allow VF to be ready before ovn pods start.

PR #235, opened 2022-01-21
Do not skip VFs configuration for BlueField 2 NICs in legacy mode

e0ne:
Let's configure VFs for BlueField 2 NICs in legacy mode if user
wants to use these NICs in a such way.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -36,7 +36,6 @@ const (
 	ClusterTypeOpenshift  = "openshift"
 	ClusterTypeKubernetes = "kubernetes"
 	VendorMellanox        = "15b3"
-	DeviceBF2             = "a2d6"
 )
 
 var InitialState sriovnetworkv1.SriovNetworkNodeState
@@ -173,15 +172,10 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 	return nil
 }
 
-// skip config VF for switchdev mode or BF-2 NICs
+// SkipConfigVf for switchdev mode or BF-2 NICs
 func SkipConfigVf(ifSpec sriovnetworkv1.Interface, ifStatus sriovnetworkv1.InterfaceExt) bool {
 	if ifSpec.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
 		glog.V(2).Infof("SkipConfigVf(): skip config VF for switchdev device")
-		return true
-	}
-	// Nvidia_mlx5_MT42822_BlueField-2_integrated_ConnectX-6_Dx
-	if ifStatus.Vendor == VendorMellanox && ifStatus.DeviceID == DeviceBF2 {
-		glog.V(2).Infof("SkipConfigVf(): skip config VF for BF2 device")
 		return true
 	}
 	return false


adrianchiris:
As background, the use-case is using DPU as a "regular NIC".

a mode which is called "separated" for Bluefield DPU

pkg/utils/utils.go
```diff
@@ -36,7 +36,6 @@ const (
 	ClusterTypeOpenshift  = "openshift"
 	ClusterTypeKubernetes = "kubernetes"
 	VendorMellanox        = "15b3"
```
adrianchiris, line 38: i think, with this change, `VendorMellanox` is also no longer used.

e0ne: It's still used here https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/235/files#diff-81ddbadfb415ccbb9c7af84f11668d1aa5e53c34025bf86d4702f16b4e42f045R708

pkg/utils/utils.go
```diff
@@ -36,7 +36,6 @@ const (
 	ClusterTypeOpenshift  = "openshift"
 	ClusterTypeKubernetes = "kubernetes"
```
e0ne, line 37: This const isn't used anywhere and should be cleaned up sometime

pkg/utils/utils.go
adrianchiris: ah, being used here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/c400742d910d731dfda9e81475d3b8047f051485/pkg/utils/utils.go#L714

search in github is not very good :(

Thanks for pointing this out

adrianchiris (approved):

zshi-redhat:
@e0ne @adrianchiris We may still want to skip BF2 for non-separated mode for the following reasons:
1) BF2 VF configuration is moved to use systemd service (similar to switchdev device configuration)
2) BF2 VF configuration doesn't require eswitchdev mode set to `switchdev`
3) Use systemd service for BF2 would allow VF with index 0 to become ready before any ovn-k8s pods are running on the host (ovnkube-node renames VF0 to ovn-k8s-mp0)

With this commit, BF2 configuration may be erased or reset once config-daemon starts.
IIUC, we can still configure BF2 w/o this commit, right?

e0ne:
@zshi-redhat, thanks for you feedback

> 1. BF2 VF configuration is moved to use systemd service (similar to switchdev device configuration)

Unfortunately, this assumption is not valid now. I'm going to fix it in scope of #81. It works for a switchdev only now

> 2. BF2 VF configuration doesn't require eswitchdev mode set to switchdev in BF2 embedded mode.

Could you please elaborate more on this case? It's not clear to me.

> 3. Use systemd service for BF2 would allow VF with index 0 to become ready before any ovn-k8s pods are running on the host (ovnkube-node renames VF0 to ovn-k8s-mp0)

How does it work/supposed to be working with SR-IOV Operator?

> With this commit, BF2 configuration may be erased or reset once config-daemon starts.

The current implementation resets all found devices. We need to fix it in a PR to not reset devices which aren't included in policy.

> IIUC, we can still configure BF2 in separated mode w/o this commit, right?

Without this commit we can't configure BF2 in a separated mode.

zshi-redhat:
> @zshi-redhat, thanks for you feedback
> 
> > 1. BF2 VF configuration is moved to use systemd service (similar to switchdev device configuration)
> 
> Unfortunately, this assumption is not valid now. I'm going to fix it in scope of #81. It works for a switchdev only now

It was done through this PR: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/201
Currently systemd service only applied to switchdev device and BF2.
The #81 will enable systemd service for all other devices.
> 
> > 2. BF2 VF configuration doesn't require eswitchdev mode set to switchdev in BF2 embedded mode.
> 
> Could you please elaborate more on this case? It's not clear to me.

Switchdev mode is required to enable OVS HWOL feature for CX- series cards, but in BF2 case (embedded cpu mode), offload happens in BF2 arm core and switchdev mode configuration is moved from host to BF2 system. On the corresponding x86 host,  switchdev mode is not needed.

> 
> > 3. Use systemd service for BF2 would allow VF with index 0 to become ready before any ovn-k8s pods are running on the host (ovnkube-node renames VF0 to ovn-k8s-mp0)
> 
> How does it work/supposed to be working with SR-IOV Operator?

There is no big difference on how sriov operator is used here, user just needs to create a sriovnetworknodepolicy (w/o specifying switchdev mode), all the rest configuration in sriovnetworknodepolicy are the same.

> 
> > With this commit, BF2 configuration may be erased or reset once config-daemon starts.
> 
> The current implementation resets all found devices. We need to fix it in a PR to not reset devices which aren't included in policy.
> 
> > IIUC, we can still configure BF2 in separated mode w/o this commit, right?
> 
> Without this commit we can't configure BF2 in a separated mode.

I think the config-daemon needs to know which mode BF2 is being configured, then use systemd for embedded cpu mode and config-daemon runtime config for separated mode.

e0ne:
@zshi-redhat PR #201  confused me too. Unfortunately it doesn't work as expected. In a current implementation BF2 configuration is ignored if we don't use switchdev mode.

e0ne:
> I think the config-daemon needs to know which mode BF2 is being configured, then use systemd for embedded cpu mode and config-daemon runtime config for separated mode.

It's what my PR does, isn't it?

zshi-redhat:
> @zshi-redhat PR #201 confused me too. Unfortunately it doesn't work as expected. In a current implementation BF2 configuration is ignored if we don't use switchdev mode.

In current implementation, bluefield-2 is ignored w/ or w/o switchdev config.
If it is switchdev, then it will be ignore at:
```
if ifSpec.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV
```
if it is not switchdev, then it will be ignore at:
```
if ifStatus.Vendor == VendorMellanox && ifStatus.DeviceID == DeviceBF2
```

Here ignore means not to run `configSriovDevice` in config-daemon.

zshi-redhat:
> > I think the config-daemon needs to know which mode BF2 is being configured, then use systemd for embedded cpu mode and config-daemon runtime config for separated mode.
> 
> It's what my PR does, isn't it?

For BlueFIeld in ConnectX mode, I think we may need :

```
if ifSpec.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV && ifStatus.Vendor == VendorMellanox && ifStatus.DeviceID == DeviceBF2
{
        return false
}


```

e0ne:
@zshi-redhat  I understood your point now. Will test your suggestion and update PR shortly

e0ne:
> For BlueFIeld in ConnectX mode, I think we may need the following as a temp solution:
> 
> ```
> if ifSpec.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV && ifStatus.Vendor == VendorMellanox && ifStatus.DeviceID == DeviceBF2
> {
>         return false
> }
> ```


It doesn't work to me. We it even breaks BF2 configuration for switchdev

zshi-redhat:
> > For BlueFIeld in ConnectX mode, I think we may need the following as a temp solution:
> > ```
> > if ifSpec.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV && ifStatus.Vendor == VendorMellanox && ifStatus.DeviceID == DeviceBF2
> > {
> >         return false
> > }
> > ```
> 
> It doesn't work to me. We it even breaks BF2 configuration for switchdev

Could you paste the change or issue ?

e0ne:
Will be fixed in #240

PR #237, opened 2022-01-26
fix: attach scheme to client for oc infra call

eggfoobar:
updated to attach scheme to client for infra call
updated error message output to bubble up error from single node call

Signed-off-by: ehila <ehila@redhat.com>

### DIFF
--- a/controllers/sriovoperatorconfig_controller.go
+++ b/controllers/sriovoperatorconfig_controller.go
@@ -80,7 +80,7 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 		if errors.IsNotFound(err) {
 			singleNode, err := utils.IsSingleNodeCluster(r.Client)
 			if err != nil {
-				return reconcile.Result{}, fmt.Errorf("Couldn't check the amount of nodes in the cluster")
+				return reconcile.Result{}, fmt.Errorf("Couldn't get cluster single node status: %s", err)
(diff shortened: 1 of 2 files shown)


pkg/utils/cluster.go
```diff
@@ -42,6 +42,7 @@ func k8sSingleNodeClusterStatus(c client.Client) (bool, error) {
 
 func openshiftSingleNodeClusterStatus(c client.Client) (bool, error) {
 	infra := &configv1.Infrastructure{}
+	configv1.AddToScheme(c.Scheme())
```
zshi-redhat, line 45: Wondering if this should be added here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/main.go#L57

eggfoobar: I wasn't sure either. I figured since this is a code path for Openshift clusters only it would be best to keep it close to the one spot it's being used, for now at least.

zshi-redhat: Maybe we can keep it here and move it later if needed?
@SchSeba wdyt?

SchSeba: I think is better to move this to the main.go file to have all the scheme changes/configuration in one please

eggfoobar: Sounds good, moved it to `main.go`

zshi-redhat (approved):

PR #238, opened 2022-01-28
webhook: relax NIC validation when using NetFilter selector

EmilienM:
When using NetFilter (for virtualized platforms), we do not provide a
DeviceID or a Vendor into the NIC selector. It's being figured out when
matching the network and the interface on the host.

So let's not fail early and let the validation process to continue,
later there is a function that tests if the VF is valid anyway.


EmilienM:
I tested the PR against OpenStack and this worked fine. I could even enable the webhook.

EmilienM:
This was wrong, I had a bad configmap. Nevermind.

PR #240, opened 2022-02-02
Fix BlueField2 SR-IOV configuration

e0ne:
This patch fixes current behaviour for BF2 NICs to configure
SR-IOV VFs in vanilla Kubernetes cluster.

PR https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/201 broke BF2 NICs configration in Kubernetes. It assumes
that BF2 will be configured via systemd service in OpenShift
using MachineConfigPool object.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -173,14 +173,14 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 	return nil
 }
 
-// skip config VF for switchdev mode or BF-2 NICs
+// SkipConfigVf Use systemd service to configure switchdev mode or BF-2 NICs in OpenShift
 func SkipConfigVf(ifSpec sriovnetworkv1.Interface, ifStatus sriovnetworkv1.InterfaceExt) bool {
 	if ifSpec.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
 		glog.V(2).Infof("SkipConfigVf(): skip config VF for switchdev device")
 		return true
 	}
-	// Nvidia_mlx5_MT42822_BlueField-2_integrated_ConnectX-6_Dx
-	if ifStatus.Vendor == VendorMellanox && ifStatus.DeviceID == DeviceBF2 {
+	// Nvidia_mlx5_MT42822_BlueField-2_integrated_ConnectX-6_Dx in OpenShift
+	if ClusterType == ClusterTypeOpenshift && ifStatus.Vendor == VendorMellanox && ifStatus.DeviceID == DeviceBF2 {
 		glog.V(2).Infof("SkipConfigVf(): skip config VF for BF2 device")
 		return true
 	}


e0ne:
@zshi-redhat @pliurh could you please review this PR?

zshi-redhat:
@e0ne Is this change to enable BF in ConnectX mode?

adrianchiris:
@zshi-redhat for now we dont have a use-case to deploy BF2 via systemd service with vanilla k8s. 
our use-case is to support DPU in connectX mode.

later on we will make sriov-network-operator DPU mode aware and support both use-cases.
(we may also need to support transitioning between the modes in this case).

once we move sriov configuration to systemd both flows (openshift and non-openshift) can be unified IMO. performing full sriov configuration via systemd.

zshi-redhat:
> @zshi-redhat for now we dont have a use-case to deploy BF2 via systemd service with vanilla k8s. our use-case is to support DPU in connectX mode (which requires "full" legacy sriov configuration).

So this patch is not supposed to enable switchdev mode with BF2 in connectX mode?
> 
> later on we will make sriov-network-operator DPU mode aware and support both use-cases. (we may also need to support transitioning between the modes in this case).

Ack+, both use-cases means supporting BF2 in different modes (connectX, embedded cpu)
> 
> once we move sriov configuration to systemd both flows (openshift and non-openshift) can be unified IMO. performing full sriov configuration via systemd.

Agree, I think k8s_plugin applies the systemd service to enable switchdev mode in vanilla k8s, I wonder if this approach will be applied to BF in connectX mode.

e0ne:
@zshi-redhat with this PR BF2 could be configured:
* by systemd service in switchdev mode both for vanilla k8s and OpenShift
* by config daemon to configure VFs like we do for the rest of NICs in vanilla k8s only


DPU mode aware logic will be implemented in a following PR

adrianchiris:
> So this patch is not supposed to enable switchdev mode with BF2 in connectX mode?

it is, as connectX, realized i wasnt clear on this :)

> Ack+, both use-cases means supporting BF2 in different modes (connectX, embedded cpu)

yes

>  I wonder if this approach will be applied to BF in connectX mode.

yes it will. the goal is for it to take the same code-path as connectx in both legacy/switchdev modes in vanilla k8s

pkg/utils/utils.go
```diff
@@ -173,14 +173,14 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
...
+// SkipConfigVf Use systemd service to configure switchdev mode or BF-2 NICs in OpenShift
 func SkipConfigVf(ifSpec sriovnetworkv1.Interface, ifStatus sriovnetworkv1.InterfaceExt) bool {
 	if ifSpec.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
 		glog.V(2).Infof("SkipConfigVf(): skip config VF for switchdev device")
 		return true
 	}
-	// Nvidia_mlx5_MT42822_BlueField-2_integrated_ConnectX-6_Dx
-	if ifStatus.Vendor == VendorMellanox && ifStatus.DeviceID == DeviceBF2 {
+	// Nvidia_mlx5_MT42822_BlueField-2_integrated_ConnectX-6_Dx in OpenShift
+	if ClusterType == ClusterTypeOpenshift && ifStatus.Vendor == VendorMellanox && ifStatus.DeviceID == DeviceBF2 {
```
zshi-redhat, line 183: We'd also like to support BF in connectx mode in openshift, this check makes the bf2 device in connectx mode to go through the path of systemd service configuration on openshift, which is not ideal since currently systemd service doesn't initialize the VF device as is done by config-daemon runtime.

Shall we take this opptunitiy to implement the VF initialization in systemd service?

adrianchiris: This PR is proposed as a temporary solution to enable the use-case for vanilla k8s. @e0ne is working of VF initialization in systemd.

for openshift i suspect the need to support BF as connectx will come at a later stage so it gives us some time to move to systemd., Driver changes might be needed as well as a new mstflint version that needs to be pulled to RHEL repo (for when we need to query the DPU mode and change it).

will this be acceptable ?

zshi-redhat: okay, the sysetmd can come later.

zshi-redhat (approved):

adrianchiris (approved):

PR #241, opened 2022-02-04
Automate network-resources-injector-secret deployment

EmilienM:
This patch is for when deploying on Kubernetes clusters, not OpenShift;
where it already has certificate management.

When deploying the Operator from source with the desire of using the
Webhook, a network-resources-injector-secret resource has to be created
and it depends on certificates.

Before, we were documented 2 options on how we can achieve it.
Now we automate the second option that is easy to use.

* Option 1: Using its own certs, then export ENABLE_CERT_MANAGER=false.

* Option 2 (default): it'll use jetstack/cert-manager to generate
  certificates and the secret for network-resources-injector.

Later, when the operator is created and webhook enabled, it'll not fail
because the `tls` volume can't be found.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,9 @@ export WATCH_NAMESPACE?=openshift-sriov-network-operator
 export GOFLAGS+=-mod=vendor
 export GO111MODULE=on
 PKGS=$(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
+# github.com/jetstack/cert-manager latest version
+export CERT_MANAGER_VERSION?=v1.7.0
+export ENABLE_CERT_MANAGER?=true
 
 # go source files, ignore vendor directory
 SRC = $(shell find . -type f -name '*.go' -not -path "./vendor/*")
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -58,54 +58,13 @@ Webhooks are disabled when deploying on a Kubernetes cluster as per the instruct
    kubectl -n sriov-network-operator create secret tls operator-webhook-service --cert=cert.pem --key=key.pem
    kubectl -n sriov-network-operator create secret tls network-resources-injector-secret --cert=cert.pem --key=key.pem
    export ENABLE_ADMISSION_CONTROLLER=true
+   export ENABLE_CERT_MANAGER=false
    export WEBHOOK_CA_BUNDLE=$(base64 -w 0 < cacert.pem)
    make deploy-setup-k8s
    ```
 
 2. Using https://cert-manager.io/, deploy it as:
-   ```bash
-   kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.3.0/cert-manager.yaml
-   ```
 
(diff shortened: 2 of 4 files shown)


EmilienM:
I manually tested this PR against `OpenShift`: https://paste.opendev.org/show/bns4K3mZsyQMJSEVsvXc/

e0ne (approved):
LGTM

doc/quickstart.md
```diff
@@ -58,54 +58,13 @@ Webhooks are disabled when deploying on a Kubernetes cluster as per the instruct
...
-   spec:
-     secretName: network-resources-injector-secret
-     dnsNames:
-     - network-resources-injector-service.sriov-network-operator.svc
-     issuerRef:
-       name: sriov-network-operator-selfsigned-issuer
-   EOF
-   ```
 
-    And then deploy the operator:
```
adrianchiris, line 108: no need to export CERT_MANAGER=true ?

doc/quickstart.md
```diff
@@ -58,54 +58,13 @@ Webhooks are disabled when deploying on a Kubernetes cluster as per the instruct
    kubectl -n sriov-network-operator create secret tls operator-webhook-service --cert=cert.pem --key=key.pem
    kubectl -n sriov-network-operator create secret tls network-resources-injector-secret --cert=cert.pem --key=key.pem
    export ENABLE_ADMISSION_CONTROLLER=true
+   export CERT_MANAGER=false
```
adrianchiris, line 61: we could set this to false by default in `hack/common` WDYT ?

EmilienM: The reason I set it to true by default is to make it easy for the user and have defaults that work ouf of the box.

If you prefer to disable it, I can change it but from a UX standpoint it's not great IMHO since you have to set it to true or install will fail...

adrianchiris: OK,
can you point me to where this is set to true by default ?

Makefile
```diff
@@ -24,6 +24,9 @@ export WATCH_NAMESPACE?=openshift-sriov-network-operator
 export GOFLAGS+=-mod=vendor
 export GO111MODULE=on
 PKGS=$(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
+# github.com/jetstack/cert-manager latest version
+export CERT_MANAGER_VERSION?=v1.7.0
+export CERT_MANAGER?=true
```
adrianchiris, line 29: nit: `ENABLE_CERT_MANAGER` ?  like we have `ENABLE_ADMISSION_CONTROLLER`

EmilienM: makes a lot of sense. Thanks

EmilienM:
I'm putting into draft now because I realized that we don't need it by default when OpenShift is used. I'll update the PR soon.

EmilienM:
@adrianchiris it's ready for review now.

EmilienM:
@SchSeba would you please looking at it please? Thanks

zeeke:
LGTM

e0ne (approved):

jjuele:

Issue #239, opened 2022-02-02
Can't build SR-IOV Network Operator images


e0ne:
CentOS 8 is EOL now [1]. It means we can't build docker images now because of this error:

```
CentOS Linux 8 - AppStream 110 B/s | 38 B 00:00
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
```

We need to update base images to fix our images build

[1] https://forums.centos.org/viewtopic.php?f=54&t=78708

e0ne:
I tested with `registry.access.redhat.com/ubi8/ubi:8.5-226` image:
```
Last metadata expiration check: 0:00:03 ago on Wed Feb  2 16:38:32 2022.
No match for argument: mstflint
Error: Unable to find a match: mstflint
The command '/bin/sh -c yum -y update && ARCH_DEP_PKGS=$(if [ "$(uname -m)" != "s390x" ]; then echo -n ${MSTFLINT} ; fi) && yum -y install hwdata $ARCH_DEP_PKGS && yum clean all' returned a non-zero code: 1
```

@zshi-redhat  @pliurh do you have any suggestions what base image we should use? I tested with `quay.io/centos/centos:stream9` and it works for me. Maybe it's better to use centos8 for now with updated repos?

PR #243, opened 2022-02-04
Update base images to CentOS 8 Stream

e0ne:
PR #223 updated base images to CentOS 8 which is EOL now so we can't
build new images using it.

Closes: #238


adrianchiris:
lgtm

@zshi-redhat @Eoghan1232 PTAL this addresses #239

adrianchiris (approved):

adrianchiris:
I think moving from centos8 to centos8-stream has the smallest delta then other alternatives (e.g rocky, centos9-stream, ubuntu base images)

Eoghan1232 (approved):
lgtm

Issue #244, opened 2022-02-05
Unable to switch vfio-pci to netdevice in policy


mskrocki:
Unable to change policy from `vfio-pci` to `netdevice`. When this is done the config daemon loops infinitely unable to recover.
Changing back to `vfio-pci`, removing the policy or changing the `numVFs` fixes this issue.

[setVfAdminMac](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L558) assumes that VFs are always using default kernel module, for the purpose of identifying the name of the VFs link. This is not the case when numVFs did not change and VFs are NOT reset, [here](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L245). The daemon hangs [here](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L568). Inside the [setSriovNumVfs](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L338) we are always resetting the VF number to 0.

This issue is introduced by https://github.com/k8snetworkplumbingwg/sriov-network-operator/commit/022c33a7b7c157a98e29ba6178583264babc6d29

Would like to know what should be the proper solution here:
- should we always reset VFs (set to zero), or
- change setVfAdminMac to use random MAC address if link is not found
- any other option???

PR #245, opened 2022-02-07
WIP: Set GUID and MAC for VF with default driver loaded

pliurh:


pliurh:
fix #244

e0ne (commented):
The change itself looks good to me but I prefer to have some more details in the commit message why do we need it

e0ne:
Do we still need #233  with this fix?

SchSeba:
I change my PR to contain also this change (it was implemented there in another method) so maybe we can just merge #233 ?

pkg/utils/utils.go
```diff
@@ -309,6 +303,15 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 						return err
 					}
 				}
+
+				// only set GUID and MAC for VF with default driver
+				if strings.EqualFold(iface.LinkType, "IB") {
+					if err = setVfGuid(addr, pfLink); err != nil {
+						return err
+					}
+				} else if err = setVfAdminMac(addr, pfLink); err != nil {
```
zshi-redhat, line 312: I think we can still set VF admin mac w/ vfio-pci driver, right?
Just like sriov-cni can set VF attributes(admin-mac, vlan, trust, spoofchk etc) in vfio-pci mode.

zshi-redhat (commented):
Could you remind me what issue does this PR fix?

pkg/utils/utils.go
pliurh: Does it make any sense to set the VF MAC when using vfio-pci driver? We invoke `vfIsReady` in `setVfAdminMac`, it returns error with vfio-pci driver.

pliurh: This is to resolve https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/244

mskrocki: I am afraid this will result in adapters with vfio-pci to use the ff:ff:ff:ff:ff:ff:ff address. Can we just skip the `setVfAdminMac` in case we did not change the vfNum? Considering this agent always creates the VFs, the mac will always be set, because it would happen when we created them.

zshi-redhat: > Does it make any sense to set the VF MAC when using vfio-pci driver? We invoke `vfIsReady` in `setVfAdminMac`, it returns error with vfio-pci driver.

I think yes. For some vendor NICs, the default admin MAC is all-zero, which means all packet (including packets with spoofed MAC) can be sent by app using the VF (when VF spoofChk is set to on).

pliurh: Fixed. PTAL.

pkg/utils/utils.go
```diff
@@ -287,12 +287,15 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
 					return err
 				}
-			} else if err = setVfAdminMac(addr, pfLink); err != nil {
-				return err
 			}
 			if err = unbindDriverIfNeeded(addr, isRdma); err != nil {
```
zshi-redhat, line 300: Put `unbindDriverIfNeeded` in the same conditional check (line 290)?

mskrocki: why do you want to pull `unbindDriverIfNeeded` into the new condition? I think this might lead to new unwanted cases, i.e. why you should only unbind when VF num changed? Since this function is not dependent on any module to be loaded I would keep it as is.

pkg/utils/utils.go
```diff
@@ -287,15 +287,18 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 					break
 				}
 			}
-			if strings.EqualFold(iface.LinkType, "IB") {
-				if err = setVfGuid(addr, pfLink); err != nil {
+			if iface.NumVfs != ifaceStatus.NumVfs {
```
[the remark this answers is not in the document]
pliurh: Agree. The change in this PR will be merged into #233, and this PR will be deprecated.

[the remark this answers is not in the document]
pliurh: @SchSeba ^^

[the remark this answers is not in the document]
SchSeba: Hi @pliurh @adrianchiris @mskrocki can you please take a look on https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/233

I think my implementation there is better I will explain on that PR

pkg/utils/utils.go
zshi-redhat: It is because `unbindDriverIfNeeded` is a workaround for infiniband (IPoIB) case (`isRdma` == true) to set GUID,
but when `isRdma` == true, user may just want to use dpdk with Mellanox device (in bifurcated mode), in which case, the unbind is not needed even `isRdma` equals to true.

zshi-redhat: In other words, `unbindDriverIfNeeded` is only needed when initializing(vf num changed) infiniband device.

PR #246, opened 2022-02-07
openstack: fallback to metadata server if config-drive is not deployed

EmilienM:
In most cases, OpenStack metadata service is available in the cloud,
which removes the requirement of config-drive.

This patch will change the logic to try reading Metadata and Network data from
the config-drive directory but if it's not present, it'll try to get
it by safely fetching the URL.

Note: it's adding a new dependency on hashicorp/go-retryablehttp which
is a library which provides a HTTP client interface with automatic retries
and exponential backoff.


EmilienM:
I'm currently testing it.

EmilienM:
After testing this, here are my early findings:

* The patch works as expected:
```
I0208 15:08:27.119809 1948520 utils_virtual.go:146] Unmarshal OpenStack meta_data
I0208 15:08:27.119894 1948520 utils_virtual.go:114] Getting body from http://169.254.169.254/openstack/latest/meta_data.json
2022/02/08 15:08:27 [DEBUG] GET http://169.254.169.254/openstack/latest/meta_data.json
I0208 15:08:27.129173 1948520 utils_virtual.go:160] Unmarshal OpenStack network_data
I0208 15:08:27.129214 1948520 utils_virtual.go:114] Getting body from http://169.254.169.254/openstack/latest/network_data.json
2022/02/08 15:08:27 [DEBUG] GET http://169.254.169.254/openstack/latest/network_data.json
```

* If metadata is disabled in a cloud or not reachable at all, the `sriov-network-config-daemon` will be really slower because it'll loop over trying to reach it anyway. We should have a parameter to disable it permanently and use config-drive all the times, so it stops trying to fetch the metadata server. See the loop:

```
I0208 15:06:04.680251 1948520 utils_virtual.go:114] Getting body from http://169.254.169.254/openstack/latest/network_data.json
2022/02/08 15:06:04 [DEBUG] GET http://169.254.169.254/openstack/latest/network_data.json
2022/02/08 15:06:04 [DEBUG] GET http://169.254.169.254/openstack/latest/network_data.json (status: 503): retrying in 1s (4 left)
2022/02/08 15:06:05 [DEBUG] GET http://169.254.169.254/openstack/latest/network_data.json (status: 503): retrying in 2s (3 left)
2022/02/08 15:06:07 [DEBUG] GET http://169.254.169.254/openstack/latest/network_data.json (status: 503): retrying in 4s (2 left)
2022/02/08 15:06:11 [DEBUG] GET http://169.254.169.254/openstack/latest/network_data.json (status: 503): retrying in 8s (1 left)
```

* Logging is missing when we fallback on reading config-drive file. This would help to confirm that this method was used instead of fetching metadata server (see logs from previous point).

EmilienM:
>     * If metadata is disabled in a cloud or not reachable at all, the `sriov-network-config-daemon` will be really slower because it'll loop over trying to reach it anyway. We should have a parameter to disable it permanently and use config-drive all the times, so it stops trying to fetch the metadata server. See the loop:

Well, actually if we invert the logic, where we first try config-drive directory and then fallback on metadata this will be better.

pkg/utils/utils_virtual.go
```diff
@@ -106,31 +110,65 @@ func metaData(platformType PlatformType, address string) (netFilter string, macA
...
-		return
+func readOpenstackMetaData() (metaData *OSPMetaData, networkData *OSPNetworkData) {
+	// We first try to read metadata from config-drive and fallback to reading from
+	// OpenStack metadata service.
+	if rawBytes, err := ioutil.ReadFile(ospMetaDataFile); err == nil {
+		glog.V(2).Infof("getting OpenStack meta_data from config-drive")
+		unmarshalOpenStackMetaData(ospMetaDataFile, rawBytes)
+	} else {
+		glog.V(2).Infof("getting OpenStack meta_data from metadata server")
+		rawBytes, err := getBodyFromUrl(ospMetaDataUrl)
```
EmilienM, line 153: I'm wondering if there would be a way to cache the data so it doesn't query the server every 10s when doing a reconcile. The metadata is not supposed to change unless the Machine had a change (e.g. via a new MachineConfig to add a new network interface).

What do you think and how could this be achieved?

EmilienM: BTW the reconcile loop logs: https://paste.opendev.org/show/bNfTzI7Q79Vkkb10JUK2/

EmilienM: ping @SchSeba ^

mdbooth: We're calling this code once *per network device* every 30 seconds. We should definitely not do this.

I think it's appropriate to cache here. The calling code clearly assumes this is cheap, and we're changing that assumption. However, we should also move the call to metadata() out of the device loop in the parent because that's unnecessarily inefficient anyway. Ideally we'd explicitly populate this metadata from an initialiser once and cache it somewhere appropriate. The writer object?

mdbooth (changes requested):

atyronesmith:
The metadata call could be moved up to the RunStartCmd I believe and a cache by pciAddress added.

EmilienM:
> The metadata call could be moved up to the RunStartCmd I believe and a cache by pciAddress added.

do you mean extracting out `pollNicStatus` (which calls `DiscoverSriovDevicesVirtual` which fetch metadata) from from daemon and add it to `cmd/sriov-network-config-daemon/start.go` (in `runStartCmd`)?
Or a way to fetch metadata (and cache it) in `runStartCmd` and pass the data down to `DiscoverSriovDevicesVirtual`?

pkg/utils/utils_virtual.go
```diff
@@ -42,9 +43,12 @@ var (
 )
 
 const (
-	ospMetaDataDir = "/host/var/config/openstack/latest/"
-	ospNetworkData = ospMetaDataDir + "/network_data.json"
-	ospMetaData    = ospMetaDataDir + "/meta_data.json"
+	ospMetaDataDir     = "/host/var/config/openstack/latest/"
+	ospMetaDataBaseUrl = "http://169.254.169.254/openstack/latest"
```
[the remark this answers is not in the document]
EmilienM: Yes, good idea. I'll do it in this PR via a separate commit.

pkg/utils/utils_virtual.go
```diff
@@ -106,31 +110,65 @@ func metaData(platformType PlatformType, address string) (netFilter string, macA
...
 	if err != nil {
-		glog.Errorf("error reading file %s, %v", ospNetworkData, err)
-		return
+		return nil, err
 	}
-
-	if err = json.Unmarshal(rawBytes, networkData); err != nil {
-		glog.Errorf("error unmarshalling raw bytes %v from %s", err, ospNetworkData)
-		return
+	rawBytes, err := ioutil.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
EmilienM: Thanks. I'll push it in the next iteration.

pkg/utils/utils_virtual.go
```diff
@@ -106,31 +110,65 @@ func metaData(platformType PlatformType, address string) (netFilter string, macA
...
+	if err != nil {
+		return nil, err
 	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
EmilienM: Thanks. I'll push it in the next iteration.

pkg/utils/utils_virtual.go
mdbooth: Do we expect this information to change? We certainly didn't expect it to change before as it was using config drive. If it will never change, there's no point in calling it every 30s. Even if it's just reading the config drive that's needlessly warming the planet.

EmilienM: The information could change if the worker MachineSet would be updated to e.g. have one more NIC attached?

mdbooth: Machines are explicitly immutable, so this would not happen via Machine updates. There could be some other mechanism to do it, though. Maybe CNI?

EmilienM: Oh right. Then I don't see anything that could change, TBH.
I'll investigate Pierre's solution for now, which seems a great improvement.

EmilienM:
I'll refactor this PR once https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/248 has landed.

EmilienM:
Being taken over here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/258

PR #247, opened 2022-02-08
Add openshift/api/config CRD to scheme

pliurh:
As we use configv1.Infrastructure in function configv1.Infrastructure.


zshi-redhat:
There is a PR fixing the same issue:https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/237

PR #248, opened 2022-02-10
openstack: move metadata fetch under daemon writer and reduce runs

EmilienM:
Machines are immutable, so there is no need to fetch for metadata every
30s.

This patch will move out the functions to read the metadata file and do
it in the daemon writer.

Old behavior:
Every 30 seconds, for each network device found, fetch, read and process
OpenStack metadata.

New behavior:
daemon process does a first run then runs a routine that executes every
30 seconds. So we'll only do fetch metadata two times in total.

Also in case the API contract is broken, we might encounter some issues so
let's switch to the latest stable, 2018-08-27, which has all the data
that we need and is currently in the same format as latest for our
needs.

### DIFF
--- a/pkg/daemon/writer.go
+++ b/pkg/daemon/writer.go
@@ -27,6 +27,8 @@ type NodeStateStatusWriter struct {
 	node               string
 	status             sriovnetworkv1.SriovNetworkNodeStateStatus
 	OnHeartbeatFailure func()
+	metaData           *utils.OSPMetaData
+	networkData        *utils.OSPNetworkData
 }
 
 // NewNodeStateStatusWriter Create a new NodeStateStatusWriter
@@ -43,6 +45,20 @@ func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func()) *Node
 func (writer *NodeStateStatusWriter) Run(stop <-chan struct{}, refresh <-chan Message, syncCh chan<- struct{}, destDir string, runonce bool, platformType utils.PlatformType) {
 	glog.V(0).Infof("Run(): start writer")
 	msg := Message{}
+
+	var err error
+
+	if platformType == utils.VirtualOpenStack {
+		writer.metaData, err = utils.ReadOpenstackMetaData()
+		if err != nil {
+			glog.Errorf("Run(): failed to read OpenStack meta_data: %v", err)
+		}
+		writer.networkData, err = utils.ReadOpenstackNetworkData()
+		if err != nil {
+			glog.Errorf("Run(): failed to read OpenStack network_data: %v", err)
+		}
+	}
+
 	if runonce {
 		glog.V(0).Info("Run(): once")
 		if err := writer.pollNicStatus(platformType); err != nil {
@@ -82,7 +98,7 @@ func (writer *NodeStateStatusWriter) pollNicStatus(platformType utils.PlatformTy
 	var err error
 
 	if platformType == utils.VirtualOpenStack {
-		iface, err = utils.DiscoverSriovDevicesVirtual(platformType)
+		iface, err = utils.DiscoverSriovDevicesVirtual(platformType, writer.metaData, writer.networkData)
 	} else {
 		iface, err = utils.DiscoverSriovDevices()
 	}
--- a/pkg/utils/utils_virtual.go
+++ b/pkg/utils/utils_virtual.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io/ioutil"
+	"os"
 	"strconv"
 
 	"github.com/golang/glog"
@@ -42,7 +42,7 @@ var (
 )
 
 const (
-	ospMetaDataDir = "/host/var/config/openstack/latest/"
+	ospMetaDataDir = "/host/var/config/openstack/2018-08-27"
 	ospNetworkData = ospMetaDataDir + "/network_data.json"
 	ospMetaData    = ospMetaDataDir + "/meta_data.json"
 )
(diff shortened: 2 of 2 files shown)


EmilienM:
Results can be seen here: https://paste.opendev.org/show/bHK9oRLZlbSc4jfUjH0y/

EmilienM:
Tested: https://paste.opendev.org/show/bpzQVRjAZc0XargXfsqP/

atyronesmith (approved):
lgtm

EmilienM:
@atyronesmith sorry, last small change: https://github.com/k8snetworkplumbingwg/sriov-network-operator/compare/37c550226cbc26849293ffd1eb20713f620b7dd2..494763265806686a1e9d52b1f5d309b3313d978b
it was something I missed... thanks for the lgtm

pierreprinetti:
oh sorry, I obviously have no review rights here! I hope I didn't mess with automation.

pkg/utils/utils_virtual.go
```diff
@@ -93,55 +93,44 @@ type OSPNetworkData struct {
...
-
-	return
-}
-
-func readOpenstackMetaData() (metaData *OSPMetaData, networkData *OSPNetworkData) {
-	networkData = &OSPNetworkData{}
-
-	rawBytes, err := ioutil.ReadFile(ospNetworkData)
+// ReadOpenstackMetaData reads the meta data from the openstack metadata file
+func ReadOpenstackMetaData() (metaData OSPMetaData, err error) {
```
SchSeba, line 97: can we add a glog.info with the name of the functions like we have in other functions

pkg/utils/utils_virtual.go
```diff
@@ -93,55 +93,44 @@ type OSPNetworkData struct {
...
+		err = fmt.Errorf("error unmarshalling metadata from file %s: %w", ospMetaData, err)
 	}
+	return metaData, err
+}
 
-	metaData = &OSPMetaData{}
-
-	rawBytes, err = ioutil.ReadFile(ospMetaData)
+// ReadOpenstackNetworkData reads the network data from the openstack metadata file
+func ReadOpenstackNetworkData() (networkData OSPNetworkData, err error) {
```
SchSeba, line 115: can we add a glog.info with the name of the functions like we have in other functions

pkg/utils/utils_virtual.go
```diff
@@ -93,55 +93,44 @@ type OSPNetworkData struct {
...
+	return networkData, err
 }
 
-func parseOpenstackMetaData(pciAddr string, metaData *OSPMetaData, networkData *OSPNetworkData) (networkID string, macAddress string) {
-
-	if metaData == nil || networkData == nil {
-		return
-	}
-
+func parseOpenstackMetaData(pciAddr string, metaData OSPMetaData, networkData OSPNetworkData) (networkID string, macAddress string) {
```
SchSeba, line 133: will the metaData.Devices not be nil if the `ReadOpenstackMetaData` function doesn't return anything or return error?

EmilienM: I added a check:

```golang
	if metaData.Devices == nil {
		return "", ""
	}
```

Tell me what you think.

SchSeba: maybe just better to use a pointer for the structs and leave the old validation?

```
if metaData == nil || networkData == nil {
		return
	}
```

pkg/utils/utils_virtual.go
```diff
@@ -42,7 +42,7 @@ var (
 )
 
 const (
-	ospMetaDataDir = "/host/var/config/openstack/latest/"
+	ospMetaDataDir = "/host/var/config/openstack/2018-08-27"
```
zshi-redhat, line 45: how often does this date change?

zshi-redhat:
> Machines are immutable, so there is no need to fetch for metadata every 30s.

Is there a use case of hot plugging?

pkg/utils/utils_virtual.go
EmilienM: > > Machines are immutable, so there is no need to fetch for metadata every 30s.
> 
> Is there a use case of hot plugging?

no, OpenStack doesn't support it. Nova will fail if we hotplug an SR-IOV device to a VM.

EmilienM:
> > Machines are immutable, so there is no need to fetch for metadata every 30s.
> 
> Is there a use case of hot plugging?

OpenStack Nova doesn't support hotplugging SRIOV devices. This is in the roadmap but not implemented yet, and won't be before a bit of time I think. It's safe to assume that hotplug is not a thing for now.

pkg/utils/utils_virtual.go
EmilienM: this version is the latest tag, so we can safely use it for now and new versions don't come often.... last one was in 2018, so...

pkg/daemon/writer.go
```diff
@@ -43,9 +43,27 @@ func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func()) *Node
...
+		networkData *utils.OSPNetworkData
+		err         error
+	)
+
+	if platformType == utils.VirtualOpenStack {
+		metaData, err = utils.ReadOpenstackMetaData()
+		if err != nil {
+			glog.Errorf("Run(): failed to read OpenStack meta_data: %v", err)
+		}
+		networkData, err = utils.ReadOpenstackNetworkData()
```
pliurh, line 58: Instead of adding extra parameters to `pollNicStatus`, shall we move `metaData` and `networkData` to struct `NodeStateStatusWriter`?

EmilienM: done

zshi-redhat (approved):

PR #250, opened 2022-02-11
Document how to build a custom image

EmilienM:
* Patch Makefile to allow overriding the APP_NAME, so any image can be
  built using `make image`.
* Add some docs about how to build & use an image.


EmilienM:
@SchSeba can you please look at this one?

PR #251, opened 2022-02-13
api: log unsupported models

EmilienM:
It can be really useful to provide logs when a device (PF or VF) is not
supported and give the vendorId + deviceId; so the deployer can check
what device is being selected and compare with the ConfigMap which
describes the supported nics.

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -103,6 +103,7 @@ func IsSupportedModel(vendorId, deviceId string) bool {
 			return true
 		}
 	}
+	log.Info("IsSupportedModel():", "Unsupported model:", "vendorId:", vendorId, "deviceId:", deviceId)
 	return false
 }
 
@@ -113,6 +114,7 @@ func IsVfSupportedModel(vendorId, deviceId string) bool {
 			return true
 		}
 	}
+	log.Info("IsVfSupportedModel():", "Unsupported VF model:", "vendorId:", vendorId, "deviceId:", deviceId)
 	return false
 }
 


api/v1/helper.go
```diff
@@ -103,6 +103,7 @@ func IsSupportedModel(vendorId, deviceId string) bool {
 			return true
 		}
 	}
+	log.Info("IsSupportedModel():", "Unsupported model:", "vendorId:", vendorId, "deviceId:", deviceId)
```
e0ne, line 106: IMO, it's better to use log.Warning here and below to get more attention into these log messages

e0ne (changes requested):

api/v1/helper.go
EmilienM: I agree but if you look at this file, all "errors" are reported via `log.Info`. The imported logger can do Info or Error. I could create another logger with glog and call `glog.Warning`, but... Up to you at this point.

EmilienM: @e0ne any thoughts?

Makefile
```diff
@@ -11,11 +11,11 @@ export OPERATOR_EXEC?=oc
 BUILD_GOPATH=$(TARGET_DIR):$(TARGET_DIR)/vendor:$(CURPATH)/cmd
 IMAGE_BUILDER?=docker
 IMAGE_BUILD_OPTS?=
-DOCKERFILE?=Dockerfile
+DOCKERFILE?=Dockerfile.webhook
```
adrianchiris, line 14: is this related to the PR ?

EmilienM: sorry it was a mistake on my side, I'll drop that

EmilienM:
@adrianchiris could I get this one approved please?

adrianchiris (approved):

adrianchiris:
merging as i see @pliurh LGTM'd it

PR #252, opened 2022-02-20
Skip unsupported NICs in DiscoverSriovDevices

rollandf:
Before creating the InterfaceExt in utils.DiscoverSriovDevices,
check that the discovered NIC is a supported model as configured
in the `supported-nic-ids` config map.

With this logic, the plugins will not get unsupported NIC to configure.

It is possible to get all the NICs discovered by using an env parameter:
`DEV_MODE` with value `TRUE` in the pod of the sriov-network-config-daemon.

Also, if DEV_MODE is set to TRUE, the webhook will allow unsupported NICS
in `SriovNetworkNodePolicy

### DIFF
--- a/bindata/manifests/daemon/daemonset.yaml
+++ b/bindata/manifests/daemon/daemonset.yaml
@@ -74,6 +74,8 @@ spec:
                 fieldPath: metadata.namespace
           - name: CLUSTER_TYPE
             value: "{{.ClusterType}}"
+          - name: DEV_MODE
+            value: "{{.DevMode}}"
         resources:
           requests:
             cpu: 100m
--- a/cmd/sriov-network-config-daemon/start.go
+++ b/cmd/sriov-network-config-daemon/start.go
@@ -130,8 +130,16 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 
 	config.Timeout = 5 * time.Second
 	writerclient := snclientset.NewForConfigOrDie(config)
+
+	mode := os.Getenv("DEV_MODE")
+	devMode := false
+	if mode == "TRUE" {
+		devMode = true
+		glog.V(0).Info("dev mode enabled")
+	}
+
 	glog.V(0).Info("starting node writer")
-	nodeWriter := daemon.NewNodeStateStatusWriter(writerclient, startOpts.nodeName, closeAllConns)
+	nodeWriter := daemon.NewNodeStateStatusWriter(writerclient, startOpts.nodeName, closeAllConns, devMode)
 
 	destdir := os.Getenv("DEST_DIR")
 	if destdir == "" {
--- a/hack/env.sh
+++ b/hack/env.sh
@@ -35,3 +35,4 @@ export ENABLE_ADMISSION_CONTROLLER=${ENABLE_ADMISSION_CONTROLLER:-"true"}
 export CLUSTER_TYPE=${CLUSTER_TYPE:-openshift}
 export NAMESPACE=${NAMESPACE:-"openshift-sriov-network-operator"}
 export WEBHOOK_CA_BUNDLE=${WEBHOOK_CA_BUNDLE:-""}
+export DEV_MODE=${DEV_MODE:-"FALSE"}
--- a/pkg/daemon/writer.go
+++ b/pkg/daemon/writer.go
@@ -23,20 +23,22 @@ const (
 )
 
 type NodeStateStatusWriter struct {
-	client             snclientset.Interface
-	node               string
-	status             sriovnetworkv1.SriovNetworkNodeStateStatus
-	OnHeartbeatFailure func()
-	metaData           *utils.OSPMetaData
-	networkData        *utils.OSPNetworkData
+	client                 snclientset.Interface
+	node                   string
+	status                 sriovnetworkv1.SriovNetworkNodeStateStatus
+	OnHeartbeatFailure     func()
+	metaData               *utils.OSPMetaData
+	networkData            *utils.OSPNetworkData
+	withUnsupportedDevices bool
 }
 
 // NewNodeStateStatusWriter Create a new NodeStateStatusWriter
-func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func()) *NodeStateStatusWriter {
+func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func(), devMode bool) *NodeStateStatusWriter {
 	return &NodeStateStatusWriter{
-		client:             c,
-		node:               n,
-		OnHeartbeatFailure: f,
+		client:                 c,
+		node:                   n,
+		OnHeartbeatFailure:     f,
+		withUnsupportedDevices: devMode,
 	}
 }
 
@@ -100,7 +102,7 @@ func (writer *NodeStateStatusWriter) pollNicStatus(platformType utils.PlatformTy
 	if platformType == utils.VirtualOpenStack {
 		iface, err = utils.DiscoverSriovDevicesVirtual(platformType, writer.metaData, writer.networkData)
 	} else {
-		iface, err = utils.DiscoverSriovDevices()
+		iface, err = utils.DiscoverSriovDevices(writer.withUnsupportedDevices)
 	}
 	if err != nil {
 		return err
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -48,7 +48,7 @@ func init() {
 	ClusterType = os.Getenv("CLUSTER_TYPE")
 }
 
-func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
+func DiscoverSriovDevices(withUnsupported bool) ([]sriovnetworkv1.InterfaceExt, error) {
 	glog.V(2).Info("DiscoverSriovDevices")
 	pfList := []sriovnetworkv1.InterfaceExt{}
 
@@ -94,6 +94,13 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
 		if len(deviceNames) == 0 {
 			// no network devices found, skipping device
 			continue
+		}
+
+		if !withUnsupported {
+			if !sriovnetworkv1.IsSupportedModel(device.Vendor.ID, device.Product.ID) {
+				glog.Infof("DiscoverSriovDevices(): unsupported device %+v", device)
+				continue
+			}
 		}
 
 		iface := sriovnetworkv1.InterfaceExt{
--- a/pkg/webhook/validate_test.go
+++ b/pkg/webhook/validate_test.go
@@ -339,6 +339,28 @@ func TestStaticValidateSriovNetworkNodePolicyWithInvalidVendor(t *testing.T) {
 	g.Expect(ok).To(Equal(false))
 }
 
+func TestStaticValidateSriovNetworkNodePolicyWithInvalidVendorDevMode(t *testing.T) {
+	t.Setenv("DEV_MODE", "TRUE")
+	policy := &SriovNetworkNodePolicy{
+		Spec: SriovNetworkNodePolicySpec{
+			DeviceType: "netdevice",
+			NicSelector: SriovNetworkNicSelector{
+				Vendor: "8087",
+			},
+			NodeSelector: map[string]string{
+				"feature.node.kubernetes.io/network-sriov.capable": "true",
+			},
+			NumVfs:       63,
+			Priority:     99,
+			ResourceName: "p0",
+		},
+	}
+	g := NewGomegaWithT(t)
+	ok, err := staticValidateSriovNetworkNodePolicy(policy)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(ok).To(Equal(true))
+}
+
 func TestStaticValidateSriovNetworkNodePolicyWithInvalidDevice(t *testing.T) {
 	policy := &SriovNetworkNodePolicy{
 		Spec: SriovNetworkNodePolicySpec{
--- a/bindata/manifests/operator-webhook/server.yaml
+++ b/bindata/manifests/operator-webhook/server.yaml
@@ -58,6 +58,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: DEV_MODE
+          value: "{{.DevMode}}"
         resources:
           requests:
             cpu: 10m
--- a/controllers/sriovoperatorconfig_controller.go
+++ b/controllers/sriovoperatorconfig_controller.go
@@ -179,6 +179,7 @@ func (r *SriovOperatorConfigReconciler) syncConfigDaemonSet(dc *sriovnetworkv1.S
 	data.Data["SRIOVInfiniBandCNIImage"] = os.Getenv("SRIOV_INFINIBAND_CNI_IMAGE")
 	data.Data["ReleaseVersion"] = os.Getenv("RELEASEVERSION")
 	data.Data["ClusterType"] = utils.ClusterType
+	data.Data["DevMode"] = os.Getenv("DEV_MODE")
 	envCniBinPath := os.Getenv("SRIOV_CNI_BIN_PATH")
 	if envCniBinPath == "" {
 		data.Data["CNIBinPath"] = "/var/lib/cni/bin"
@@ -231,6 +232,7 @@ func (r *SriovOperatorConfigReconciler) syncWebhookObjs(dc *sriovnetworkv1.Sriov
 		data.Data["ReleaseVersion"] = os.Getenv("RELEASEVERSION")
 		data.Data["ClusterType"] = utils.ClusterType
 		data.Data["CaBundle"] = os.Getenv("WEBHOOK_CA_BUNDLE")
+		data.Data["DevMode"] = os.Getenv("DEV_MODE")
 		objs, err := render.RenderDir(path, &data)
 		if err != nil {
 			logger.Error(err, "Fail to render webhook manifests")
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -63,6 +63,8 @@ spec:
               value: $RESOURCE_PREFIX
             - name: ENABLE_ADMISSION_CONTROLLER
               value: "$ENABLE_ADMISSION_CONTROLLER"
+            - name: DEV_MODE
+              value: "$DEV_MODE"
             - name: NAMESPACE
               valueFrom:
                 fieldRef:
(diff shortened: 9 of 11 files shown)


rollandf:
sriovnetworknodestates before:
```
status:
  interfaces:
  - deviceID: 165f
    driver: tg3
    linkSpeed: 1000 Mb/s
    linkType: ETH
    mac: 2c:ea:7f:ec:96:0c
    mtu: 1500
    name: eno1
    pciAddress: "0000:18:00.0"
    vendor: "14e4"
  - deviceID: 165f
    driver: tg3
    linkType: ETH
    mac: 2c:ea:7f:ec:96:0d
    mtu: 1500
    name: eno2
    pciAddress: "0000:18:00.1"
    vendor: "14e4"
  - deviceID: 165f
    driver: tg3
    linkType: ETH
    mac: 2c:ea:7f:ec:96:0e
    mtu: 1500
    name: eno3
    pciAddress: "0000:19:00.0"
    vendor: "14e4"
  - deviceID: 165f
    driver: tg3
    linkType: ETH
    mac: 2c:ea:7f:ec:96:0f
    mtu: 1500
    name: eno4
    pciAddress: "0000:19:00.1"
    vendor: "14e4"
  - deviceID: a2d6
    driver: mlx5_core
    linkSpeed: 25000 Mb/s
    linkType: ETH
    mac: b8:ce:f6:8d:c8:70
    mtu: 1500
    name: ens1f0
    pciAddress: 0000:3b:00.0
    totalvfs: 16
    vendor: 15b3
  - deviceID: a2d6
    driver: mlx5_core
    linkSpeed: 25000 Mb/s
    linkType: ETH
    mac: b8:ce:f6:8d:c8:71
    mtu: 1500
    name: ens1f1
    pciAddress: 0000:3b:00.1
    totalvfs: 16
    vendor: 15b3
  - deviceID: 101d
    driver: mlx5_core
    linkSpeed: 100000 Mb/s
    linkType: ETH
    mac: 04:3f:72:b2:c0:88
    mtu: 1500
    name: ens3f0
    pciAddress: 0000:d8:00.0
    totalvfs: 8
    vendor: 15b3
  - deviceID: 101d
    driver: mlx5_core
    linkSpeed: 100000 Mb/s
    linkType: ETH
    mac: 04:3f:72:b2:c0:89
    mtu: 1500
    name: ens3f1
    pciAddress: 0000:d8:00.1
    totalvfs: 8
    vendor: 15b3
  lastSyncError: exit status 3
  syncStatus: InProgress
```
============================
sriovnetworknodestates after:

```
status:
  interfaces:
  - deviceID: a2d6
    driver: mlx5_core
    linkSpeed: 25000 Mb/s
    linkType: ETH
    mac: b8:ce:f6:8d:c8:70
    mtu: 1500
    name: ens1f0
    pciAddress: 0000:3b:00.0
    totalvfs: 16
    vendor: 15b3
  - deviceID: a2d6
    driver: mlx5_core
    linkSpeed: 25000 Mb/s
    linkType: ETH
    mac: b8:ce:f6:8d:c8:71
    mtu: 1500
    name: ens1f1
    pciAddress: 0000:3b:00.1
    totalvfs: 16
    vendor: 15b3
  - deviceID: 101d
    driver: mlx5_core
    linkSpeed: 100000 Mb/s
    linkType: ETH
    mac: 04:3f:72:b2:c0:88
    mtu: 1500
    name: ens3f0
    pciAddress: 0000:d8:00.0
    totalvfs: 8
    vendor: 15b3
  - deviceID: 101d
    driver: mlx5_core
    linkSpeed: 100000 Mb/s
    linkType: ETH
    mac: 04:3f:72:b2:c0:89
    mtu: 1500
    name: ens3f1
    pciAddress: 0000:d8:00.1
    totalvfs: 8
    vendor: 15b3
```
==============================
Logs:
```
W0217 14:03:07.144808 1850566 utils.go:100] DiscoverSriovDevices(): unsupported device 0000:18:00.0 -> class: 'Network controller' vendor: 'Broadcom Inc. and subsidiaries' product: 'NetXtreme BCM5720 Gigabit Ethernet PCIe'
W0217 14:03:07.144857 1850566 utils.go:100] DiscoverSriovDevices(): unsupported device 0000:18:00.1 -> class: 'Network controller' vendor: 'Broadcom Inc. and subsidiaries' product: 'NetXtreme BCM5720 Gigabit Ethernet PCIe'
W0217 14:03:07.144885 1850566 utils.go:100] DiscoverSriovDevices(): unsupported device 0000:19:00.0 -> class: 'Network controller' vendor: 'Broadcom Inc. and subsidiaries' product: 'NetXtreme BCM5720 Gigabit Ethernet PCIe'
W0217 14:03:07.144911 1850566 utils.go:100] DiscoverSriovDevices(): unsupported device 0000:19:00.1 -> class: 'Network controller' vendor: 'Broadcom Inc. and subsidiaries' product: 'NetXtreme BCM5720 Gigabit Ethernet PCIe'
```

adrianchiris (approved):

e0ne (approved):
LGTM. Thank for this fix, Freddy!

pliurh:
@rollandf Thanks for your PR.
As you may have known, users can try unsupported NIC by adding the device ID to ConfigMap `supported-nic-ids`. By showing unsupported NICs in sriovnetworknodestates, it helps users to find all the NICs installed. So I prefer we keep that information.

adrianchiris:
@pliurh isnt it odd that we display nics which the operator doesnt support ? 
also to properly add a nic to the list you also need the VF device id right ? which is not present today.

pliurh:
@zshi-redhat @bn222 WDYT?

rollandf:
Added an env param to be able to see all the NICs

pkg/utils/utils.go
```diff
@@ -48,7 +48,7 @@ func init() {
 	ClusterType = os.Getenv("CLUSTER_TYPE")
 }
 
-func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
+func DiscoverSriovDevices(devMode bool) ([]sriovnetworkv1.InterfaceExt, error) {
```
adrianchiris, line 51: at this layer, maybe its worth naming it: "discoverAll" or "withUnsupported" ?

pkg/daemon/writer.go
```diff
@@ -29,14 +29,16 @@ type NodeStateStatusWriter struct {
 	OnHeartbeatFailure func()
 	metaData           *utils.OSPMetaData
 	networkData        *utils.OSPNetworkData
+	devMode            bool
 }
 
 // NewNodeStateStatusWriter Create a new NodeStateStatusWriter
-func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func()) *NodeStateStatusWriter {
+func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func(), mode bool) *NodeStateStatusWriter {
```
adrianchiris, line 36: nit : mode -> devMode

when reading this, its hard to understand from the function signature what is a "true" or "false" mode mean.
other arguments here also suffer from lack of clarity IMO (e.g `f`)

rollandf: Done

pkg/utils/utils.go
rollandf: changed to "withUnsupported"

pkg/daemon/writer.go
```diff
@@ -29,14 +29,16 @@ type NodeStateStatusWriter struct {
...
 }
 
 // NewNodeStateStatusWriter Create a new NodeStateStatusWriter
-func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func()) *NodeStateStatusWriter {
+func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func(), devMode bool) *NodeStateStatusWriter {
 	return &NodeStateStatusWriter{
 		client:             c,
 		node:               n,
 		OnHeartbeatFailure: f,
+		devMode:            devMode,
```
bn222, line 41: Since this is later passed to `utils.DiscoverSriovDevices`, I'm wondering if it makes any sense to keep the name "devMode" here. Wouldn't it make more sense to also rename to `withUnsupported` or even `withUnsupportedDevices`?

rollandf: Sounds good. Fixed.

hack/env.sh
```diff
@@ -35,3 +35,4 @@ export ENABLE_ADMISSION_CONTROLLER=${ENABLE_ADMISSION_CONTROLLER:-"true"}
 export CLUSTER_TYPE=${CLUSTER_TYPE:-openshift}
 export NAMESPACE=${NAMESPACE:-"openshift-sriov-network-operator"}
 export WEBHOOK_CA_BUNDLE=${WEBHOOK_CA_BUNDLE:-""}
+export DEV_MODE="FALSE"
```
adrianchiris, line 38: to allow to overwrite you can 👍 
export DEV_MODE=${DEV_MODE:-"FALSE"}

rollandf: Agreed

e0ne (approved):

adrianchiris (approved):

rollandf:
Anything missing in this PR?

adrianchiris:
@pliurh @zshi-redhat PTAL

pkg/daemon/writer.go
```diff
@@ -23,20 +23,22 @@ const (
...
+	status                 sriovnetworkv1.SriovNetworkNodeStateStatus
+	OnHeartbeatFailure     func()
+	metaData               *utils.OSPMetaData
+	networkData            *utils.OSPNetworkData
+	withUnsupportedDevices bool
 }
 
 // NewNodeStateStatusWriter Create a new NodeStateStatusWriter
-func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func()) *NodeStateStatusWriter {
+func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func(), devMode bool) *NodeStateStatusWriter {
```
bn222, line 36: Shouldn't we rename this `devMode` to `withUnsupportedDevices` as well?

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -74,6 +74,8 @@ spec:
                 fieldPath: metadata.namespace
           - name: CLUSTER_TYPE
             value: "{{.ClusterType}}"
+          - name: DEV_MODE
+            value: "{{.DevMode}}"
```
bn222, line 78: idem

adrianchiris:
> I feel that we should really rename devmode everywhere.

IMO we should not be explicit on the meaning of dev mode at upper layers to allow rolling in more functionality under the same config knob. 

For now, running config daemon in dev mode would expose unsupported devices in sriovnodestate status but later on it may have additional meaning.

bn222:
ok

cmd/sriov-network-config-daemon/start.go
```diff
@@ -130,8 +130,16 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
+	mode := os.Getenv("DEV_MODE")
+	devMode := false
+	if mode == "TRUE" {
+		devMode = true
+		glog.V(0).Info("dev mode enabled")
+	}
+
 	glog.V(0).Info("starting node writer")
-	nodeWriter := daemon.NewNodeStateStatusWriter(writerclient, startOpts.nodeName, closeAllConns)
+	nodeWriter := daemon.NewNodeStateStatusWriter(writerclient, startOpts.nodeName, closeAllConns, devMode)
```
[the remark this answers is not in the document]
adrianchiris: setting to resolved per:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/252#issuecomment-1066676597

cmd/sriov-network-config-daemon/start.go
```diff
@@ -130,8 +130,16 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 
 	config.Timeout = 5 * time.Second
 	writerclient := snclientset.NewForConfigOrDie(config)
+
+	mode := os.Getenv("DEV_MODE")
+	devMode := false
+	if mode == "TRUE" {
+		devMode = true
+		glog.V(0).Info("dev mode enabled")
```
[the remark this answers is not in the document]
adrianchiris: setting to resolved per:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/252#issuecomment-1066676597

bindata/manifests/daemon/daemonset.yaml
adrianchiris: setting to resolved per:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/252#issuecomment-1066676597

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -74,6 +74,8 @@ spec:
                 fieldPath: metadata.namespace
           - name: CLUSTER_TYPE
             value: "{{.ClusterType}}"
+          - name: DEV_MODE
```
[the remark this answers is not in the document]
adrianchiris: setting to resolved per:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/252#issuecomment-1066676597

pkg/daemon/writer.go
adrianchiris: setting to resolved per:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/252#issuecomment-1066676597

adrianchiris:
@pliurh @zshi-redhat @SchSeba PTAL. 

It would be great if we can merge this week

SchSeba:
Hi @rollandf @adrianchiris 

The code looks good from my point of view I just have one question.

Why do we add the flag as env variable in the operator instead of extending the sriovOperatorConfig?
if we do it this way this means that we will not be able to enable the dev mode in deployments with OLM (operator lifecycle management) is that ok?

adrianchiris:
> Why do we add the flag as env variable in the operator instead of extending the sriovOperatorConfig?

Running config daemon in dev mode is intended for development purposes so we didnt want to introduce changes to CRDs
in this case, you would not use OLM.

SchSeba:
@adrianchiris last question 

what about the validation webhook?
Today if the webhook is false we are able to test new nics without adding it to the supported nics for example if you don't know the vf deviceID.

after this change, the user will need to enable the dev mode + disable the webhook to be able and check new devices?

adrianchiris:
using an unsupported NIC is not something i envision will be used in production but rather in development/poc environments.

so we are talking about a development flow. 

admittedly i didnt think about the webhook :(  in this case i think of 3 ways:

1.disable webhook and run operator in dev mode which propagates the env variable to the config daemon
2. extend dev mode to webhook as well. in webhook if ran in dev mode it will not check sriovnetworkpolicy against supported dev ids
3. same as 2 but trigger dev mode from operatorconfig (API change)

I think 2 is a good happy middle WDYT ?
we should also document dev mode while we are at it.

so when you want to test a new hardware you deploy operator in dev mode (with or without webhook)

makes sense ?

SchSeba:
I think we can go with option 2
and yes please let's document the Dev mode

pkg/webhook/validate_test.go
```diff
@@ -339,6 +339,28 @@ func TestStaticValidateSriovNetworkNodePolicyWithInvalidVendor(t *testing.T) {
 	g.Expect(ok).To(Equal(false))
 }
 
+func TestStaticValidateSriovNetworkNodePolicyWithInvalidVendorDevMode(t *testing.T) {
+	os.Setenv("DEV_MODE", "TRUE")
```
zeeke, line 343: Please, be aware the `os.Setenv(...)` creates an order dependency in the test suite. If we can leverage go1.17 feature, you can use `t.Setenv(...)` to make sure the environment variable is cleaned after the test.

Otherwise, [here there are some solutions] for the env variable issue. The .3 should work well

zeeke:
Hi, to allow using unsupported NICs you can edit [configmap.yaml](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/deploy/configmap.yaml) file, that is deployed with `make deploy-setup` command.

I'm not sure what DEV_MODE feature adds other than what you already can do with the configmap and I see code becomes a little more complex with that variable.

pkg/webhook/validate_test.go
rollandf: Changed to `t.SetEnv`

rollandf:
> Hi, to allow using unsupported NICs you can edit [configmap.yaml](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/deploy/configmap.yaml) file, that is deployed with `make deploy-setup` command.
> 
> I'm not sure what DEV_MODE feature adds other than what you already can do with the configmap and I see code becomes a little more complex with that variable.

@zeeke The main purpose of this patch is to not expose the unsupported NICs in `SriovNetworkNodeState`.
However, community members wanted to keep the possibility to keep the previous behavior with a help of a env var.
You can read the discussion in the comments of this PR.

zeeke:
@rollandf I agree the PRs main purpose is really useful and I went over all the discussion, maybe I missed saying that.

I'm still missing the scenario where DEV_MODE env allows doing something the configmap can't do. If a dev is working on a new device, sooner or later he needs to update the configmap.yaml file.

Clusters that are currently running with unsupported NICs will require the operator's intervention after this PR in any case (adding a DEV_MODE env or updating the configmap).

Is there any other use case I'm missing?

rollandf:
> @rollandf I agree the PRs main purpose is really useful and I went over all the discussion, maybe I missed saying that.
> 
> I'm still missing the scenario where DEV_MODE env allows doing something the configmap can't do. If a dev is working on a new device, sooner or later he needs to update the configmap.yaml file.
> 
> Clusters that are currently running with unsupported NICs will require the (human) operator's intervention after this PR in any case (adding a DEV_MODE env or updating the configmap).
> 
> Is there any other use case I'm missing?

I think that @SchSeba mentioned following workflow:
- Be able to discover all NICS including unsupported in `sriovnetworknodestates`
- Be able to configure `SriovNetworkNodePolicy` with unsupported NICS without disabling the webhook

SchSeba:
Hi @rollandf you are right that is the workflow.

@zeeke this is fine we want the user to make changes if he wants to use unsupported nics.

This will help us in the troubleshooting face if we identify that the user is using DEV_MOD or change the configmap we can't support the operator.

Also when using OLM (production deployment) it will not be possible to enable the DEV_MOD.
Today we have users that just disable the webhook from the sriovOperatorConfig and start using unsupported nics this makes the troubleshooting harder

zeeke:
@rollandf @SchSeba thank you for the explanation, now it's more clear for me.

adrianchiris (approved):

PR #257, opened 2022-02-23
virtual/openstack: configure VFIO with NIOMMU by default

EmilienM:
As well known and well documented, in virtual deployments of Kubernetes such as running on top
of OpenStack where the underlying virtualization platform is KVM, it does not support a virtualized iommu,
the VFIO PCI driver needs to be loaded with a special flag.

This patch aims to do it by default in the virtual_plugin, since this
plugin is only used by OpenStack for now.

Note: it also changes `utils.LoadKernelModule` function to support
kernel module arguments. It also adds a note in `load-kmod.sh` about the
fact that a module won't be reloaded with potential new options if the
module is already loaded on the system.

### DIFF
--- a/bindata/scripts/load-kmod.sh
+++ b/bindata/scripts/load-kmod.sh
@@ -1,11 +1,16 @@
 #!/bin/sh
 # chroot /host/ modprobe $1
 kmod_name=$(tr "-" "_" <<< $1)
+kmod_args="${@:2}"
 chroot /host/ lsmod | grep "^$1" >& /dev/null
 
 if [ $? -eq 0 ]
 then
+        # NOTE: We do not check if the module is loaded with specific options
+        #       so a manual reload is required if the module is loaded with
+        #       new or different options.
+        echo "Module $kmod_name already loaded; no change will be applied..."
         exit 0 
 else
-        chroot /host/ modprobe $kmod_name
+        chroot /host/ modprobe $kmod_name $kmod_args
 fi
--- a/pkg/plugins/virtual/virtual_plugin.go
+++ b/pkg/plugins/virtual/virtual_plugin.go
@@ -84,6 +84,16 @@ func (p *VirtualPlugin) Apply() error {
 	glog.Infof("virtual-plugin Apply(): desiredState=%v", p.DesireState.Spec)
 
 	if p.LoadVfioDriver == loading {
+		// In virtual deployments of Kubernetes where the underlying virtualization platform does not support a virtualized iommu
+		// the VFIO PCI driver needs to be loaded with a special flag.
+		// This is the case for OpenStack deployments where the underlying virtualization platform is KVM.
+		// NOTE: if VFIO was already loaded for some reason, we will not try to load it again with the new options.
+		kernelArgs := "enable_unsafe_noiommu_mode=1"
+		if err := utils.LoadKernelModule("vfio", kernelArgs); err != nil {
+			glog.Errorf("virtual-plugin Apply(): fail to load vfio kmod: %v", err)
+			return err
+		}
+
 		if err := utils.LoadKernelModule("vfio_pci"); err != nil {
 			glog.Errorf("virtual-plugin Apply(): fail to load vfio_pci kmod: %v", err)
 			return err
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -534,12 +534,13 @@ func getVfInfo(pciAddr string, devices []*ghw.PCIDevice) sriovnetworkv1.VirtualF
 	return vf
 }
 
-func LoadKernelModule(name string) error {
-	glog.Infof("LoadKernelModule(): try to load kernel module %s", name)
-	cmd := exec.Command("/bin/sh", scriptsPath, name)
+func LoadKernelModule(name string, args ...string) error {
+	glog.Infof("LoadKernelModule(): try to load kernel module %s with arguments '%s'", name, args)
+	cmdArgs := strings.Join(args, " ")
+	cmd := exec.Command("/bin/sh", scriptsPath, name, cmdArgs)
 	err := cmd.Run()
 	if err != nil {
-		glog.Errorf("LoadKernelModule(): fail to load kernel module %s: %v", name, err)
+		glog.Errorf("LoadKernelModule(): fail to load kernel module %s with arguments '%s': %v", name, args, err)
 		return err
 	}
 	return nil


EmilienM:
testing it via our CI in OpenShift: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/257

EmilienM:
@atyronesmith @SchSeba would you be open to such a change?

atyronesmith:
Yes!

EmilienM:
This worked as expected:

```
I0223 22:18:55.973409   11002 utils.go:538] LoadKernelModule(): try to load kernel module vfio_pci with arguments enable_unsafe_noiommu_mode=1
```

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_sriov-network-operator/634/pull-ci-openshift-sriov-network-operator-release-4.10-e2e-openstack-nfv/1496590559802822656/artifacts/e2e-openstack-nfv/gather-extra/artifacts/pods/openshift-sriov-network-operator_sriov-network-config-daemon-clvpl_sriov-network-config-daemon.log

bindata/scripts/load-kmod.sh
```diff
@@ -1,11 +1,12 @@
 #!/bin/sh
 # chroot /host/ modprobe $1
 kmod_name=$(tr "-" "_" <<< $1)
+kmod_args="${@:2}"
```
SchSeba, line 4: will this also work if we only pass 1 arg?

pkg/utils/utils.go
```diff
@@ -534,12 +534,12 @@ func getVfInfo(pciAddr string, devices []*ghw.PCIDevice) sriovnetworkv1.VirtualF
 	return vf
 }
 
-func LoadKernelModule(name string) error {
-	glog.Infof("LoadKernelModule(): try to load kernel module %s", name)
-	cmd := exec.Command("/bin/sh", scriptsPath, name)
+func LoadKernelModule(name string, args string) error {
```
SchSeba, line 537: can you use here something like `args []string`

bindata/scripts/load-kmod.sh
EmilienM:
```bash
~ cat test.sh 
#!/bin/sh
# chroot /host/ modprobe $1
kmod_name=$(tr "-" "_" <<< $1)
kmod_args="${@:2}"
echo "Loading module $kmod_name with args $kmod_args"
```

```bash
~ ./test.sh foo bar
Loading module foo with args bar
```

``` bash
~ ./test.sh foo    
Loading module foo with args 
```

```bash
./test.sh foo arg1 arg2 arg3
Loading module foo with args arg1 arg2 arg3
```

pkg/utils/utils.go
EmilienM: done

bindata/scripts/load-kmod.sh
```diff
@@ -1,11 +1,13 @@
...
 chroot /host/ lsmod | grep "^$1" >& /dev/null
 
 if [ $? -eq 0 ]
 then
+        echo "Module $kmod_name already loaded; no change will be applied..."
```
bn222, line 9: I feel that the vagueness of the error might cause confusion while debugging if the module has already been loaded without enable_unsafe_noiommu_mode=1. I will not know easily that the missing option is the problem. Could it be possible to check that the module is loaded in the way we need it with modinfo and react accordingly (i.e. reload it with the correct options)?

pkg/plugins/virtual/virtual_plugin.go
```diff
@@ -84,7 +84,17 @@ func (p *VirtualPlugin) Apply() error {
 	glog.Infof("virtual-plugin Apply(): desiredState=%v", p.DesireState.Spec)
 
 	if p.LoadVfioDriver == loading {
-		if err := utils.LoadKernelModule("vfio_pci"); err != nil {
+		// In virtual deployments of Kubernetes where the underlying virtualization platform does not support a virtualized iommu
+		// the VFIO PCI driver needs to be loaded with a special flag.
+		// This is the case for OpenStack deployments where the underlying virtualization platform is KVM.
+		kernel_args := []string{"enable_unsafe_noiommu_mode=1"}
```
bn222, line 90: Is my understanding correct that vfio will from now on always be loaded with this arg? I'm unclear on the potential downsides of **always** loading it with this.

bindata/scripts/load-kmod.sh
EmilienM: yeah honestly I thought about it and I couldn't think of something easy to implement that wasn't too disruptive code-wise. I'll have a second look if you think it's really important.

EmilienM: `modinfo vfio` output:

```
filename:       /lib/modules/4.18.0-348.el8.x86_64/kernel/drivers/vfio/vfio.ko.xz
softdep:        post: vfio_iommu_type1 vfio_iommu_spapr_tce
alias:          devname:vfio/vfio
alias:          char-major-10-196
description:    VFIO - User Level meta-driver
author:         Alex Williamson <alex.williamson@redhat.com>
license:        GPL v2
version:        0.3
rhelversion:    8.5
srcversion:     D36061711A57E5FAC95068F
depends:        
intree:         Y
name:           vfio
vermagic:       4.18.0-348.el8.x86_64 SMP mod_unload modversions 
sig_id:         PKCS#7
signer:         Red Hat Enterprise Linux kernel signing key
sig_key:        77:A4:0E:A4:5E:89:11:3E:DC:EE:6E:4B:55:AA:F2:1A:6C:4A:91:21
sig_hashalgo:   sha256
signature:      (...)
parm:           enable_unsafe_noiommu_mode:Enable UNSAFE, no-IOMMU mode.  This mode provides no device isolation, no DMA translation, no host kernel protection, cannot be used for device assignment to virtual machines, requires RAWIO permissions, and will taint the kernel.  If you do not know what this is for, step away. (default: false) (bool)
```

Nothing tells which options are actually loaded.
The only command that I found was `modprobe -c` but it's for all the modules :-/

```
modprobe -c | grep enable_unsafe_noiommu_mode
options vfio enable_unsafe_noiommu_mode=1
```

I find a bit risky to play this game of grepping modprobe and reloading a module if an option wasn't found...

bn222: Hm, fair point, thank you for checking. Let's then just add a descriptive string that "(assuming module has enable_unsafe_noiommu_mode=1 set)" or something like that, to remind ourselves during the debug process about this.

EmilienM: @bn222 I added 2 comments:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/compare/856b088bdd2e4ebd4cb21ff97daac1f29517ad43..37809517fa0fa4ee04e532ef7d935ab3a41b500f

Let me know if it's enough.

Thanks again for your review!

bn222: ok thanks for taking this into account!

zshi-redhat (approved):

pkg/plugins/virtual/virtual_plugin.go
```diff
@@ -84,7 +84,18 @@ func (p *VirtualPlugin) Apply() error {
...
+		// the VFIO PCI driver needs to be loaded with a special flag.
+		// This is the case for OpenStack deployments where the underlying virtualization platform is KVM.
+		// NOTE: if VFIO was already loaded for some reason, we will not try to load it again with the new options.
+		kernel_args := []string{"enable_unsafe_noiommu_mode=1"}
+		if err := utils.LoadKernelModule("vfio", kernel_args); err != nil {
+			glog.Errorf("virtual-plugin Apply(): fail to load vfio kmod: %v", err)
+			return err
+		}
+
+		no_args := []string{}
```
adrianchiris, line 97: lets use camelCase

pkg/plugins/virtual/virtual_plugin.go
```diff
@@ -84,7 +84,18 @@ func (p *VirtualPlugin) Apply() error {
 	glog.Infof("virtual-plugin Apply(): desiredState=%v", p.DesireState.Spec)
 
 	if p.LoadVfioDriver == loading {
-		if err := utils.LoadKernelModule("vfio_pci"); err != nil {
+		// In virtual deployments of Kubernetes where the underlying virtualization platform does not support a virtualized iommu
+		// the VFIO PCI driver needs to be loaded with a special flag.
+		// This is the case for OpenStack deployments where the underlying virtualization platform is KVM.
+		// NOTE: if VFIO was already loaded for some reason, we will not try to load it again with the new options.
+		kernel_args := []string{"enable_unsafe_noiommu_mode=1"}
```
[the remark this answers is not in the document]
adrianchiris: >the virtual plugin is only used for OpenStack (now) and enable_unsafe_noiommu_mode is always needed on OpenStack because we don't support virtualized IOMMU. So I think it's safe to enable it by default for now.

i have limited expeirence with SNO on virtualized env with openstack where this plugin is used. 
we may need later on to expose the option to configure vfio kernel args policy.

for now if its always needed then i think its ok.

pkg/utils/utils.go
```diff
@@ -534,12 +534,13 @@ func getVfInfo(pciAddr string, devices []*ghw.PCIDevice) sriovnetworkv1.VirtualF
 	return vf
 }
 
-func LoadKernelModule(name string) error {
-	glog.Infof("LoadKernelModule(): try to load kernel module %s", name)
-	cmd := exec.Command("/bin/sh", scriptsPath, name)
+func LoadKernelModule(name string, args []string) error {
+	glog.Infof("LoadKernelModule(): try to load kernel module %s with arguments %s", name, args)
+	cmd_args := strings.Join(args, " ")
```
adrianchiris, line 539: can we use camel case ?

pkg/utils/utils.go
```diff
@@ -534,12 +534,13 @@ func getVfInfo(pciAddr string, devices []*ghw.PCIDevice) sriovnetworkv1.VirtualF
 	return vf
 }
 
-func LoadKernelModule(name string) error {
-	glog.Infof("LoadKernelModule(): try to load kernel module %s", name)
-	cmd := exec.Command("/bin/sh", scriptsPath, name)
+func LoadKernelModule(name string, args []string) error {
```
adrianchiris, line 537: you can do a variadic declaration of args , this way if not provided it will evaluate to empty list, else you will have the args in the list

e.g
`LoadKernelModule(name string, args ...string)`

then you can call it with or without args

bindata/scripts/load-kmod.sh
```diff
@@ -1,11 +1,16 @@
 #!/bin/sh
 # chroot /host/ modprobe $1
 kmod_name=$(tr "-" "_" <<< $1)
+kmod_args="${@:2}"
 chroot /host/ lsmod | grep "^$1" >& /dev/null
 
 if [ $? -eq 0 ]
 then
+        # NOTE: We do not check if the module is loaded with specific options
```
adrianchiris, line 9: unfortunately i dont have a setup to check this, 

but for a given kmod i think its parameters are under:
/sys/module/<kmod>/parameters/<param name>

what are you getting in this path for vfio ?
and what is the value you read from "enable_unsafe_noiommu_mode" file ?

from: https://man7.org/linux/man-pages/man5/sysfs.5.html

              parameters
                     This directory contains one file for each module
                     parameter, with each file containing the value of
                     the corresponding parameter.  Some of these files
                     are writable, allowing the

adrianchiris: moreover this can be used to validate that the provided parameters are supported.
(in addition to checking if module need to be reloaded)

EmilienM:
```
[root@test ~]# cat /sys/module/vfio/parameters/enable_unsafe_noiommu_mode 
Y
[root@test ~]# ls -al /sys/module/vfio/parameters/enable_unsafe_noiommu_mode 
-rw-r--r--. 1 root root 4096 Mar  8 14:50 /sys/module/vfio/parameters/enable_unsafe_noiommu_mode
```

Like I said in previous comments, I don't think we want to add complex logic to check whether or not a module is enabled with the right parameters, and should leave it to the user somehow.
Yes we could parse kernel arguments and check if the files are there but I'm sure we'll hit corner cases and cause more harm that we we think.
If you're ok, I'll not touch that code for now.

EmilienM: For reference, we had this conversation here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/257/files#r819412615

pkg/plugins/virtual/virtual_plugin.go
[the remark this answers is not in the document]
EmilienM: yes it's always needed and I don't foresee it changing since it'll require *huge* changes in OpenStack to support this.

pkg/utils/utils.go
EmilienM: good idea

bindata/scripts/load-kmod.sh
adrianchiris: if there is no strong use-case for it now, im ok with deferring it.

adrianchiris (commented):
minor nit otherwise lgtm !

EmilienM:
@adrianchiris thanks for the review, all done. Let me know if there is anything.

adrianchiris (approved):
LGTM

PR #258, opened 2022-02-24
openstack: include support to metadata service

MaysaMacedo:
In most cases, OpenStack metadata service is available in the cloud,
which removes the requirement of config-drive.

This patch will change the logic to try reading Metadata and Network data from
the config-drive directory but if it's not present, it'll try to get
it by safely fetching the URL.

Note: it's adding a new dependency on hashicorp/go-retryablehttp which
is a library which provides a HTTP client interface with automatic retries
and exponential backoff.


pkg/utils/utils_virtual.go
```diff
@@ -93,43 +98,79 @@ type OSPNetworkData struct {
 	// Omit Services
 }
 
-// ReadOpenstackMetaData reads the meta data from the openstack metadata file
-func ReadOpenstackMetaData() (metaData *OSPMetaData, err error) {
+// ReadOpenstackDataFiles reads the meta_data and network_data files
+func ReadOpenstackDataFiles() (metaData *OSPMetaData, networkData *OSPNetworkData, err error) {
+	metaData = &OSPMetaData{}
+	networkData = &OSPNetworkData{}
 	glog.Infof("ReadOpenstackMetaData(): read OpenStack meta_data")
```
EmilienM, line 105: maybe add "from config-drive"

pkg/utils/utils_virtual.go
```diff
@@ -93,43 +98,79 @@ type OSPNetworkData struct {
...
-		err = fmt.Errorf("error unmarshalling metadata from file %s: %w", ospMetaData, err)
+	if err = json.NewDecoder(metadataf).Decode(&metaData); err != nil {
+		return metaData, networkData, fmt.Errorf("error unmarshalling metadata from file %s: %w", ospMetaDataFile, err)
 	}
-	return metaData, err
-}
 
-// ReadOpenstackNetworkData reads the network data from the openstack metadata file
-func ReadOpenstackNetworkData() (networkData *OSPNetworkData, err error) {
 	glog.Infof("ReadOpenstackNetworkData(): read OpenStack network_data")
```
EmilienM, line 115: same here

pkg/daemon/writer.go
```diff
@@ -49,13 +49,13 @@ func (writer *NodeStateStatusWriter) Run(stop <-chan struct{}, refresh <-chan Me
...
-			glog.Errorf("Run(): failed to read OpenStack meta_data: %v", err)
-		}
-		writer.networkData, err = utils.ReadOpenstackNetworkData()
-		if err != nil {
-			glog.Errorf("Run(): failed to read OpenStack network_data: %v", err)
+			glog.Errorf("Run(): failed to read OpenStack data files: %v", err)
+			writer.metaData, writer.networkData, err = utils.FetchOpenstackData()
+			if err != nil {
+				glog.Errorf("Run(): failed to fetch OpenStack data: %v", err)
+			}
```
EmilienM, line 58: @pierreprinetti I think your feedback from https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/246/files#r803725901 is addressed here. Can you please confirm?

pkg/utils/utils_virtual.go
```diff
@@ -42,9 +44,12 @@ var (
 )
 
 const (
-	ospMetaDataDir = "/host/var/config/openstack/2018-08-27"
-	ospNetworkData = ospMetaDataDir + "/network_data.json"
-	ospMetaData    = ospMetaDataDir + "/meta_data.json"
+	ospMetaDataDir     = "/host/var/config/openstack/2018-08-27"
+	ospMetaDataBaseUrl = "http://169.254.169.254/openstack/latest"
```
EmilienM, line 48: please use `2018-08-27` instead of `latest` for consistency and safety against a stable API.

MaysaMacedo:
It worked fine to fetch the metadata service:
```
I0301 18:00:41.740862   49272 utils_virtual.go:105] ReadOpenstackMetaData(): read OpenStack meta_data from config-drive                                                                       
2022/03/01 18:00:41 [DEBUG] GET http://169.254.169.254/openstack/2018-08-27/meta_data.json                                                                                                    
2022/03/01 18:00:41 [DEBUG] GET http://169.254.169.254/openstack/2018-08-27/network_data.json 
```

EmilienM:
/lgtm
this looks good to me. Thanks!

MaysaMacedo:
@SchSeba can you take a look on this PR? Thanks

EmilienM:
LGTM

pkg/daemon/writer.go
```diff
@@ -49,13 +49,12 @@ func (writer *NodeStateStatusWriter) Run(stop <-chan struct{}, refresh <-chan Me
 	var err error
 
 	if platformType == utils.VirtualOpenStack {
-		writer.metaData, err = utils.ReadOpenstackMetaData()
+		writer.metaData, writer.networkData, err = utils.ReadOpenstackDataFiles()
```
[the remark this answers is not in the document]
adrianchiris: also maybe at this level (writer) we should have just a single call here to:

`utils.GetOpenstackData() ` which is where the logic of getting from either config drive or metadata service will reside. WDYT?

adrianchiris:
> In most cases, OpenStack metadata service is available in the cloud

should fetching from metadata service be the default ?

pkg/daemon/writer.go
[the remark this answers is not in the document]
MaysaMacedo: yeah, works for me.

MaysaMacedo:
> > In most cases, OpenStack metadata service is available in the cloud
> 
> should fetching from metadata service be the default ?

Using config-drive is considered more stable than using metadata, that's why we try to handle it first.
The order chosen was initially proposed by @EmilienM, maybe he has more information about it.

EmilienM:
> > > In most cases, OpenStack metadata service is available in the cloud
> > 
> > 
> > should fetching from metadata service be the default ?
> 
> Using config-drive is considered more stable than using metadata, that's why we try to handle it first. The order chosen was initially proposed by @EmilienM, maybe he has more information about it.

Right, let's try to read the files first (cheaper & faster than fetching URL), by default, and if it's not there we just fallback into fetching the metadata URL.

MaysaMacedo:
@adrianchiris Thanks for the review. Can you take another look on this PR?

adrianchiris (approved):

EmilienM:
LGTM

zshi-redhat (approved):

PR #268, opened 2022-03-28
Sync SriovNetworkNodeState CRD into helm chart

e0ne:
SriovNetworkNodeState is outdated. IsRdma field was introduced
in the PR #207.

This patch syncs CRDs between config/crd/bases and helm chart.


adrianchiris (approved):

adrianchiris:
@SchSeba @pliurh PTAL

SchSeba:
/lgtm

in general maybe we can have a `make` command that will copy all the files into the right folders :)

PR #269, opened 2022-03-28
Remove unused VfIds global variable

zeeke:
The `VfIds` seems to be unused in the repository. May we get rid of it?


SchSeba:
Thanks!

/lgtm

PR #262, opened 2022-03-02
Improve the virtual plugin support

SchSeba:
This commit add the support for virtio interfaces like vhostuser for openstack virtual workers.

implementation details:
* on first run (after a reboot) we get all the information we need when the devices are visible to the kernel
* we match the mac address to the openstack network ID
* if the sriov-network-config-daemon gets reboot it will use the initial file on the node so even if the nics are in vfio the node state will be right
* introduce a new system service to remove the initial state file after a reboot to support nic changes both for virtual and BM environments

Signed-off-by: Sebastian Sch <sebassch@gmail.com>

PR #271, opened 2022-03-29
OpenStack: Support generic virtio device for DPDK

MaysaMacedo:
This commit includes device with vendor 1af4, PF device 1000
and VF device 0000 to the supported NICs config-map to ensure
the webhook allows creation of SriovNetworkNodePolicy for DPDK
in the OpenStack CI.

### DIFF
--- a/deploy/configmap.yaml
+++ b/deploy/configmap.yaml
@@ -21,4 +21,4 @@ data:
   Broadcom_bnxt_BCM57414_2x25G: "14e4 16d7 16dc"
   Broadcom_bnxt_BCM75508_2x100G: "14e4 1750 1806"
   Qlogic_qede_QL45000_50G: "1077 1654 1664"
-  
+  Red_Hat_Virtio_network_device: "1af4" "1000" "0000"
--- a/deployment/sriov-network-operator/templates/configmap.yaml
+++ b/deployment/sriov-network-operator/templates/configmap.yaml
@@ -21,4 +21,4 @@ data:
   Broadcom_bnxt_BCM57414_2x25G: "14e4 16d7 16dc"
   Broadcom_bnxt_BCM75508_2x100G: "14e4 1750 1806"
   Qlogic_qede_QL45000_50G: "1077 1654 1664"
-  
+  Red_Hat_Virtio_network_device: "1af4" "1000" "0000"


deploy/configmap.yaml
```diff
@@ -21,4 +21,4 @@ data:
   Broadcom_bnxt_BCM57414_2x25G: "14e4 16d7 16dc"
   Broadcom_bnxt_BCM75508_2x100G: "14e4 1750 1806"
   Qlogic_qede_QL45000_50G: "1077 1654 1664"
-  
+  Red_Hat_Virtio_network_device: "1af4" "0001" "0000"
```
EmilienM, line 24: isn't it `1af4 1000 0000` ?

From my logs when I tested it:
```
I0323 11:12:34.454601       1 factory.go:106] device added: [pciAddr: 0000:00:04.0, vendor: 1af4, device: 1000, driver: vfio-pci]
I0323 11:12:34.454613       1 manager.go:139] New resource server is created for dpdk1 ResourcePool
```

MaysaMacedo: Ouch, yes! thanks

SchSeba:
when using the netFilter that is the filter we use for virtual environments we skip the nic verification so I don't think we need this PR.

deploy/configmap.yaml
```diff
@@ -21,4 +21,4 @@ data:
   Broadcom_bnxt_BCM57414_2x25G: "14e4 16d7 16dc"
   Broadcom_bnxt_BCM75508_2x100G: "14e4 1750 1806"
   Qlogic_qede_QL45000_50G: "1077 1654 1664"
-  
+  Red_Hat_Virtio_network_device: "1af4" "1000" "0000"
```
SchSeba, line 24: The vf and PF have the same deviceID it should be 1000 also for the VF

```
- Vfs:
      - deviceID: "1000"
        driver: vfio-pci
        pciAddress: "0000:00:04.0"
        vendor: 1af4
        vfID: 0
      deviceID: "1000"
      driver: vfio-pci
      netFilter: openstack/NetworkID:45da469e-6a3a-41f4-80e8-3c762dbd13b4
      numVfs: 1
      pciAddress: "0000:00:04.0"
      totalvfs: 1
      vendor: 1af4

```

MaysaMacedo: is it 0000 because of `vfID: 0`?

SchSeba:
Hi @MaysaMacedo can we close this one in favor of https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/262

deploy/configmap.yaml
SchSeba: the sriov-operator looks for the vendorID and  deviceID the vfID is just the number of the allocated VF by the kernel.
you can see that using the output

`ip link`

PR #273, opened 2022-03-31
Add unit tests for `daemon` package

zeeke:
To improve the overall quality of the `daemon` package, I think it's useful to have a unit test suite to make changes a little safer.

Besides a basic unit test, I added another to verify non latest NodeState generations are ignored, as it was the effective behavior
but it wasn't very clear from the `nodeStateSyncHandler()`'s parameter point of view.


rollandf:
@zeeke Can you please rebase this one?

zeeke:
Sure!

I think I can cover https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/333 too with these kind of tests.

rollandf:
LGTM

zeeke:
Just rebased and improved the test cases to something more meaningful. I'm not very happy with the changes I had to do to make the config-daemon testable, as they include the following hacks and tricks:

1. Wrappers for `ioutil.ReadFile`, `ioutil.WriteFile`, `exec.Command`, `syscall.Chroot`, `os.MkDirAll`
2. `//go:linkname sysBusPci ...` hack to mock of DevicePlugin invocation calls. This point can be solved in cleaner ways, but I want to be sure this testing strategy is worth the effort:
  a. By exporting `sysBusPci` variable in device-plugin repo, perhaps using a dedicated function
  b. Copy-paste device-plugin functions, as they are small utility functions and use `fswrap` in those methods.(Go proverb [“a little copying is better than a little dependency”](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=568s))
  c. Mocking the actual interaction with device-plugin utils package using wrapper like I did with filesystem.
3. [Dummy interfaces](https://developers.redhat.com/blog/2018/10/22/introduction-to-linux-interfaces-for-virtual-networking#dummy) to allow `netlink` package to find good IP links. This also involves running tests with `sudo`.

any suggestion on the above items?

pkg/daemon/daemon_test.go
```diff
@@ -0,0 +1,390 @@
...
+			mcClient,
+			exitCh,
+			stopCh,
+			syncCh,
+			refreshCh,
+			utils.Baremetal,
+		)
+
+		p, _ := generic.NewGenericPlugin()
+		sut.enabledPlugins = map[string]plugin.VendorPlugin{generic.PluginName: p}
```
adrianchiris, line 140: we should provide a mock interface for generic plugin and not invoke it here IMO.

UT for generic plugin should be added separately

pkg/daemon/daemon_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+		err = netlink.LinkAdd(&linkDummy)
```
adrianchiris, line 142: can we avoid adding dummy links ?
these remain in the system if not cleaned.

pkg/daemon/daemon_test.go
zeeke: I agree. I can work on a version of these tests with all plugins mocked out. 

Some testing hacks would probably go away we will also get less meaningful tests: the current test version has SriovNetworkNodeState as input and filesystem values/iplinks as output.

I'll use a different PR, so we can see and compare both strategies better. WDYT?

zeeke:
Here there is a version of `daemon_test.go` where all plugins are mocked (call it plugin-less):

"plugin-less" https://github.com/zeeke/sriov-network-operator-1/blob/simpler-daemon-tests/pkg/daemon/daemon_test.go

There are much fewer test doubles, therefore the overall test structure is simpler. BTW, the tests we can implement for the `Daemon` component have SriovNetworkNodeState as inputs, while on the output (assertion side) we have `refreshCh`, DevicePlugin pod status, and plugin callback invocation.

The other version (let's call it plugin-aware) maps SriovNetworkNodeStates to filesystem reads/writes and device configuration. 

"plugin-aware" represents better what the component specifications are, but at the cost of a higher complexity and maintenance cost.

Has anyone any improvement points for both versions?

@adrianchiris is "plugin-less" version similar to what you were expecting? 

"plugin-aware` https://github.com/zeeke/sriov-network-operator-1/blob/node-state-generation/pkg/daemon/daemon_test.go

zeeke:
As discussed in the last Network Plumbing WG meeting, we prefer the plugin-less version. Keeping the component under test small involves having much fewer test doubles and simpler test implementation.

Going to tidy up my branches and code

pkg/client/clientset/versioned/typed/sriovnetwork/v1/fake/fake_sriovnetwork.go
```diff
@@ -20,9 +20,9 @@ type FakeSriovNetworks struct {
 	ns   string
 }
 
-var sriovnetworksResource = schema.GroupVersionResource{Group: "sriovnetwork", Version: "v1", Resource: "sriovnetworks"}
+var sriovnetworksResource = schema.GroupVersionResource{Group: "sriovnetwork.openshift.io", Version: "v1", Resource: "sriovnetworks"}
```
SchSeba, line 23: how this change occurs?

pkg/daemon/daemon_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	snclientset "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/clientset/versioned"
+	fakesnclientset "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/clientset/versioned/fake"
+	plugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/generic"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
+
+	fakemcclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/fake"
+
+	fakek8s "k8s.io/client-go/kubernetes/fake"
```
SchSeba, line 23: can you please fix the imports to be something like this


```
	"context"
	"flag"
	"testing"

	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"
	
	corev1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
    fakek8s "k8s.io/client-go/kubernetes/fake"
    fakemcclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/fake"
    
	snclientset "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/clientset/versioned"
	fakesnclientset "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/clientset/versioned/fake"
	plugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins"
	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/generic"
	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
	"github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/fakefilesystem"
```

pkg/daemon/daemon_test.go
```diff
@@ -0,0 +1,265 @@
...
+	var exitCh chan error
+	var refreshCh chan Message
+
+	var cleanFakeFs func()
+
+	var sut *Daemon
+
+	BeforeEach(func() {
+		stopCh = make(chan struct{})
+		refreshCh = make(chan Message, 1024)
```
SchSeba, line 66: why the size here is not 1 like in the running case?
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/781c598256b7182671b3fc007cc95344f4a3289c/cmd/sriov-network-config-daemon/start.go#L81

pkg/daemon/daemon_test.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func updateSriovNetworkNodeState(c snclientset.Interface, nodeState *sriovnetworkv1.SriovNetworkNodeState) error {
+	_, err := c.SriovnetworkV1().
+		SriovNetworkNodeStates(namespace).
+		Update(context.Background(), nodeState, metav1.UpdateOptions{})
+	return err
+}
+
+type fakePlugin struct{}
```
SchSeba, line 245: can we move this plugin maybe to the plugin folder?

with a comment that is a plugin for the unit tests

pkg/daemon/daemon_test.go
```diff
@@ -0,0 +1,265 @@
...
+	})
+
+	Context("Should", func() {
+
+		It("restart sriov-device-plugin pod", func() {
+			go func() {
+				Expect(sut.Run(stopCh, exitCh)).To(BeNil())
+			}()
+
+			_, err := sut.kubeClient.CoreV1().Nodes().
```
SchSeba, line 126: can we move this one as part of the beforeEach

pkg/daemon/daemon_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	var cleanFakeFs func()
+
+	var sut *Daemon
+
+	BeforeEach(func() {
+		stopCh = make(chan struct{})
+		refreshCh = make(chan Message, 1024)
+		exitCh = make(chan error)
+		syncCh = make(chan struct{}, 64)
```
SchSeba, line 68: why we need to fill this channel will 64 messages?

pkg/daemon/daemon.go
```diff
@@ -273,11 +276,13 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 		case <-stopCh:
 			glog.V(0).Info("Run(): stop daemon")
 			return nil
-		case err := <-exitCh:
+		case err, more := <-exitCh:
```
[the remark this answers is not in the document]
zeeke: Because when closing `exitCh` (and it's actually the only use of it, no one is writing in it), `err.Error()` produces a "nil pointer dereference"

pkg/client/clientset/versioned/typed/sriovnetwork/v1/fake/fake_sriovnetwork.go
zeeke: I manually changed the file, as it seems [hack/update-codegen.sh](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/hack/update-codegen.sh) is not aligned with the generated clientset.

Maybe it is worth creating a separate issue to get it working again.

pkg/daemon/daemon_test.go
zeeke: good point. I don't remember if I got issues with a zero-length channel but now tests pass even with it.
Changing..

pkg/daemon/daemon_test.go
zeeke: sure

pkg/daemon/daemon_test.go
zeeke: It simulates `NodeStateStatusWriter` writing on the channel, it avoids the daemon waiting.

cgoncalves:
LGTM. This PR brings a lot of value to the project, boosting people's confidence in its stability and helps mitigate code regressions. Thanks, Andrea!

@rollandf @adrianchiris @SchSeba could you please review this PR once more? I believe @zeeke has addressed all your comments.

pkg/daemon/daemon.go
```diff
@@ -932,7 +939,7 @@ func tryEnableRdma() (bool, error) {
 	glog.V(2).Infof("tryEnableRdma()")
 	var stdout, stderr bytes.Buffer
 
-	cmd := exec.Command("/bin/bash", rdmaScriptsPath)
+	cmd := exec.Command("/bin/bash", filesystemRoot+rdmaScriptsPath)
```
SchSeba, line 942: can we use the file/path package and not do `+`

pkg/daemon/daemon.go
```diff
@@ -957,7 +964,7 @@ func tryEnableRdma() (bool, error) {
 func tryCreateSwitchdevUdevRule(nodeState *sriovnetworkv1.SriovNetworkNodeState) error {
 	glog.V(2).Infof("tryCreateSwitchdevUdevRule()")
 	var newContent string
-	filePath := "/host/etc/udev/rules.d/20-switchdev.rules"
+	filePath := filesystemRoot + "/host/etc/udev/rules.d/20-switchdev.rules"
```
SchSeba, line 967: same here please :)

pkg/daemon/daemon.go
```diff
@@ -993,7 +1000,7 @@ func tryCreateSwitchdevUdevRule(nodeState *sriovnetworkv1.SriovNetworkNodeState)
 	}
 
 	var stdout, stderr bytes.Buffer
-	cmd := exec.Command("/bin/bash", udevScriptsPath)
+	cmd := exec.Command("/bin/bash", filesystemRoot+udevScriptsPath)
```
SchSeba, line 1003: same

pkg/daemon/daemon.go
```diff
@@ -1014,7 +1021,7 @@ func tryCreateSwitchdevUdevRule(nodeState *sriovnetworkv1.SriovNetworkNodeState)
 
 func tryCreateNMUdevRule() error {
 	glog.V(2).Infof("tryCreateNMUdevRule()")
-	dirPath := "/host/etc/udev/rules.d/"
+	dirPath := filesystemRoot + "/host/etc/udev/rules.d/"
```
SchSeba, line 1024: same

SchSeba (commented):
just a small nit but I can already give the lgtm

/lgtm

pkg/daemon/daemon_test.go
```diff
@@ -0,0 +1,212 @@
...
+				}
+
+				return len(podList.Items), nil
+			}, "10s").Should(BeZero())
+
+		})
+	})
+})
+
+func TestConfigDaemon(t *testing.T) {
```
adrianchiris, line 223: can you move this and all the below before the Describe() call ?

test/util/fakefilesystem/fakefilesystem.go
```diff
@@ -0,0 +1,60 @@
...
+// Example usage:
+//
+// ```
+// rootDir, fakeFsClean, err := defer fs.Use()
+// if err != nil { ... }
+// defer fakeFsClean()
+// ````
+func (f *FS) Use() (string, func(), error) {
+	// create the new fake fs root dir in /tmp/sriov...
+	rootDir, err := ioutil.TempDir("", "sriov")
```
adrianchiris, line 27: nit: "sriov-operator" ?

adrianchiris (commented):
Hey @zeeke ! appologies for the delay.

a few nits and otherwise LGTM !

test/util/fakefilesystem/fakefilesystem.go
zeeke: sure, that's better!

adrianchiris (approved):
Thanks @zeeke  ! LGTM.

i see @SchSeba comments addressed and he LGTM it as well

merging

Issue #279, opened 2022-04-07
sriov-network-config-daemon fails to configure Mellanox NICs


e0ne:
Steps to reproduce:
- deploy the latest SR-IOV Network Operator
- apply policy to configure VFs on Mellanox NICs

Expected results: VFs are configured

Actual results:
Operator fails to load mellanox plugin and configure VFs:
```I0407 07:41:39.725566 1910191 start.go:141] starting node writer
I0407 07:41:39.749812 1910191 start.go:161] Running on platform: Baremetal
I0407 07:41:39.749841 1910191 writer.go:48] Run(): start writer
I0407 07:41:39.749852 1910191 writer.go:61] Run(): once
I0407 07:41:39.804107 1910191 utils.go:101] DiscoverSriovDevices(): unsupported device 0000:01:00.0 -> class: 'Network controller' vendor: 'Broadcom Inc. and subsidiaries' product: 'NetXtreme BCM5720 Gigabit Ethernet PCIe'
I0407 07:41:39.804178 1910191 utils.go:101] DiscoverSriovDevices(): unsupported device 0000:01:00.1 -> class: 'Network controller' vendor: 'Broadcom Inc. and subsidiaries' product: 'NetXtreme BCM5720 Gigabit Ethernet PCIe'
I0407 07:41:39.804211 1910191 utils.go:101] DiscoverSriovDevices(): unsupported device 0000:18:00.0 -> class: 'Network controller' vendor: 'Broadcom Inc. and subsidiaries' product: 'BCM57416 NetXtreme-E Dual-Media 10G RDMA Ethernet Controller'
I0407 07:41:39.804238 1910191 utils.go:101] DiscoverSriovDevices(): unsupported device 0000:18:00.1 -> class: 'Network controller' vendor: 'Broadcom Inc. and subsidiaries' product: 'BCM57416 NetXtreme-E Dual-Media 10G RDMA Ethernet Controller'
I0407 07:41:39.804287 1910191 utils.go:101] DiscoverSriovDevices(): unsupported device 0000:3b:00.0 -> class: 'Network controller' vendor: 'Mellanox Technologies' product: 'MT42822 BlueField-2 integrated ConnectX-6 Dx network controller'
[...]

PR #280, opened 2022-04-07
Init NicIdMap on the config daemon startup

e0ne:
After PR #252 is merged there is a race condition between NicIdMap
init and accec to it.
This patch moves InitNicIdMap function call before NicIdMap will be
used in node writer and config daemon run function.

Closes-Bug: #279


ykulazhenkov (approved):

adrianchiris (approved):

zeeke:
lgtm

Issue #249, opened 2022-02-10
sytemd services are not deleted


e0ne:
Steps to reproduce:
* deploy SriovNetworkNodePolicy CR to configure some NIC with `eSwitchMode: switchdev`
* wait until policy will be applied
* verify systemd services `switchdev-configuration-*-nm` are deployed
* delete SriovNetworkNodePolicy 
* wait until policy will be deleted

Expected results:
VFs are deleted, `switchdev-configuration-*-nm` services and their components are deleted

Actual results:
systemd services and their scripts are still available on the node

adrianchiris:
fix should be here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/7102b792c4fe5c30ab3ffb68f808f8a1a279dcc5/bindata/scripts/clean-k8s-services.sh#L10

rollandf:
I will take a look

PR #282, opened 2022-04-13
Remove switchdev scripts when uneeded

rollandf:
Fixes issue #249


zeeke:
LGTM

e0ne (approved):
LGTM as a partial fix of the issue. We need to do this cleanup as a part of sriov node policy deletion process

bindata/scripts/clean-k8s-services.sh
```diff
@@ -10,7 +10,11 @@ chroot_path="/host"
 function clean_services() {
   # Remove switchdev service files
   rm -f $chroot_path/etc/systemd/system/switchdev-configuration.service
```
adrianchiris, line 12: this and L15 are no longer created by config daemon. however we should keep them in case of leftovers from old SR-IOV network operator that did not do the clean of this file.

can you add a note explaining why we delete file that were not created by current version of network operator ?

rollandf: Added

adrianchiris (approved):

adrianchiris (approved):

SchSeba (approved):

Issue #283, opened 2022-04-14
SriovNetworkNodeState temporary state cause unneeded plugins actions


rollandf:
There is a stage when the SriovNetworkNodeState is initializing where the spec is empty because the SriovNetworkNodePolicyReconciler did not yet applied the policies.

It can cause a non required action by the plugins, that will try to apply the empty spec by resetting the NIC for example.

The flow currently is:
1 Creation of NodeState by NodePolicyController ( no policies applied at this state, as Interfaces are not present yet)
2 Status update with Interfaces by NodeStateStatusWriter 
3 Spec Update by NodePolicyController
4 Plugin action based on spec

Now, the plugins are also triggered between 2 and 3, causing unneeded configuration.

There is a need to be able to know if the empty spec is intentional or just a temporary state.

SchSeba:
The fix is in place closing this issue

PR #287, opened 2022-04-26
Skip config if policy not applied yet

rollandf:
There is a stage when the SriovNetworkNodeState is initializing
where the spec is empty because the SriovNetworkNodePolicyReconciler
did not yet applied the policies.

It can cause a non required action by the plugins,
that will try to apply the empty spec by resetting the NIC for example.

The config daemon will not run the plugins if the generation is 1
and the Spec.Interfaces is empty.

Solves issue #283

For e2e tests, the wait timeout to get to initial Sync state has been
increased.
This change is needed as now the config daemon will not apply on "empty"
spec until the SriovNetworkNodePolicyReconciler will iterate on the Interfaces.
The reconcile loop interval is 5 minutes, so the test timeout needed to be increased.

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -278,6 +278,12 @@ func (r *SriovNetworkNodePolicyReconciler) syncSriovNetworkNodeState(np *sriovne
 			return fmt.Errorf("failed to get SriovNetworkNodeState: %v", err)
 		}
 	} else {
+		if len(found.Status.Interfaces) == 0 {
+			logger.Info("SriovNetworkNodeState Status Interfaces are empty. Skip update of policies in spec",
+				"namespace", ns.Namespace, "name", ns.Name)
+			return nil
+		}
+
 		logger.Info("SriovNetworkNodeState already exists, updating")
 		newVersion := found.DeepCopy()
 		newVersion.Spec = ns.Spec
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -418,6 +418,19 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 		return nil
 	}
 
+	if latestState.GetGeneration() == 1 && len(latestState.Spec.Interfaces) == 0 {
+		glog.V(0).Infof("nodeStateSyncHandler(): Name: %s, Interface policy spec not yet set by controller", latestState.Name)
+		if latestState.Status.SyncStatus != "Succeeded" {
+			dn.refreshCh <- Message{
+				syncStatus:    "Succeeded",
+				lastSyncError: "",
+			}
+			// wait for writer to refresh status
+			<-dn.syncCh
+		}
+		return nil
+	}
+
 	dn.refreshCh <- Message{
 		syncStatus:    "InProgress",
 		lastSyncError: "",


config/crd/bases/sriovnetwork.openshift.io_sriovnetworknodestates.yaml
```diff
@@ -75,6 +75,8 @@ spec:
                   - pciAddress
                   type: object
                 type: array
+              isPolicySet:
```
e0ne, line 78: Please, add the similar change to https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/deployment/sriov-network-operator/crds/sriovnetwork.openshift.io_sriovnetworknodestates.yaml

e0ne (changes requested):

e0ne (approved):
LGTM

pkg/daemon/daemon.go
```diff
@@ -448,11 +448,15 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 		return nil
 	}
 
+	if !latestState.Spec.IsPolicySet {
+		glog.V(0).Infof("nodeStateSyncHandler(): Interface policy spec not yet set")
```
adrianchiris, line 452: nit: `SriovNetworkNodeState spec not yet set by controller`  ?

adrianchiris:
one minor nit, otherwise LGTM

AbdYsn:
@adrianchiris @rollandf i am not sure if it is related, but in last CI failure, the sriovnetworknodestate was missing the `syncStatus` field, see [sriovnetworknodestate](http://13.74.249.42/sriov-network-operator-ci/52/logs/worker-node-state.yaml.gz),
we also see
```

I0508 11:08:50.751581    1394 writer.go:163] setNodeStateStatus(): syncStatus: , lastSyncError: 
I0508 11:08:56.009989    1394 daemon.go:311] get item: 1
I0508 11:08:56.010049    1394 daemon.go:401] nodeStateSyncHandler(): new generation is 1
I0508 11:08:56.023458    1394 daemon.go:424] nodeStateSyncHandler(): Interface policy spec not yet set
I0508 11:08:56.023491    1394 daemon.go:346] Successfully synced
I0508 11:08:56.023509    1394 daemon.go:309] worker queue size: 0
I0508 11:08:56.116046    1394 daemon.go:284] Run(): period refresh
I0508 11:08:56.121430    1394 daemon.go:985] tryCreateSwitchdevUdevRule()
I0508 11:08:56.121583    1394 daemon.go:1043] tryCreateNMUdevRule()
I0508 11:09:11.010630    1394 daemon.go:311] get item: 1
I0508 11:09:11.010672    1394 daemon.go:401] nodeStateSyncHandler(): new generation is 1
I0508 11:09:11.019890    1394 daemon.go:424] nodeStateSyncHandler(): Interface policy spec not yet set
I0508 11:09:11.019917    1394 daemon.go:346] Successfully synced
I0508 11:09:11.019928    1394 daemon.go:309] worker queue size: 0
```
in the config daemon logs

test/e2e/sriovoperatornodepolicy_test.go
```diff
@@ -32,7 +32,7 @@ var _ = Describe("Operator", func() {
 		Eventually(func() *cluster.EnabledNodes {
 			sriovInfos, _ = cluster.DiscoverSriov(clients, testNamespace)
 			return sriovInfos
-		}, timeout, interval).ShouldNot(BeNil())
+		}, time.Minute*5, time.Second*20).ShouldNot(BeNil())
```
AbdYsn, line 35: can you use multiples of values defined at the [test suite](https://github.com/rollandf/sriov-network-operator/blob/nodestatepolicy/test/e2e/e2e_tests_suite_test.go#L42-L44) instead of hard coding it?

rollandf: Done

adrianchiris (approved):

test/e2e/sriovoperatornodepolicy_test.go
```diff
@@ -32,7 +32,7 @@ var _ = Describe("Operator", func() {
 		Eventually(func() *cluster.EnabledNodes {
 			sriovInfos, _ = cluster.DiscoverSriov(clients, testNamespace)
 			return sriovInfos
-		}, timeout, interval).ShouldNot(BeNil())
+		}, Timeout*15, RetryInterval*20).ShouldNot(BeNil())
```
adrianchiris, line 35: With this change, sriovnetworknodestate will be reconciled by config daemon only after the controller set the spec (empty or other)

reason for timeout increase is for controller to resync nodestate with the actual spec and allow config daemon to reconcile.

see
https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/305

adrianchiris (approved):
LGTM

adrianchiris:
@SchSeba @Eoghan1232  can we merge this ?

Eoghan1232 (approved):

Eoghan1232:
LGTM

SchSeba:
@rollandf @adrianchiris my only question here is if there is an option to not change the CRD with a new field.

rollandf:
@SchSeba Any action items on my side on this PR?

SchSeba:
Hi @rollandf @adrianchiris I review this PR and I have questions.

looks like we are not changing the `IsPolicySet` to false ever again. if that is the case to not run the plugins only on the first run after the installation I think we can just use
 
```
if len(found.Status.Interfaces) == 0 {
			logger.Info("SriovNetworkNodeState Status Interfaces are empty. Skip update of policies in spec")
			return nil
		}
```

if that is not the case then the PR will not work as expected in the following scenario
1. install the operator
2. after the daemon created the devices in the status area and the user creates the policy the `IsPolicySet` will be set to true
3. delete the sriov policy (the  `IsPolicySet` will still be true)
4. reboot the node or the config daemon (it will still apply the plugins)

Let me know if I miss something

SchSeba:
In general to fix this I think we just need to improve the config-daemon verification to see if we even need to apply the plugins or not

SchSeba:
for the daemon can we just do something like

if spec.interface is empty and latestState.GetGeneration() == 1: skip config

that should do the same no?

adrianchiris:
> for the daemon can we just do something like
>if spec.interface is empty and latestState.GetGeneration() == 1: skip config
> that should do the same no?

that is correct but only if there is going to be an update to the spec from the controller.
in case the controller calculated the desired state and no policy applies, then spec would be empty,
we would then want config daemon to reset interfaces. and this will not happen since generation will still be 1. (controller will not update)

rollandf:
With `if spec.interface is empty and latestState.GetGeneration() == 1: skip config` there can be still a problematic scenario (not sure if it is a corner case):

1. Policy with config + node state with spec
2. Delete policy and node state
3. Daemon will not reset the interfaces since the generation is 1 again after NS deletion

rollandf:
> Hi @rollandf @adrianchiris I review this PR and I have questions.
> 
> looks like we are not changing the `IsPolicySet` to false ever again. if that is the case to not run the plugins only on the first run after the installation I think we can just use
> 
> ```
> if len(found.Status.Interfaces) == 0 {
> 			logger.Info("SriovNetworkNodeState Status Interfaces are empty. Skip update of policies in spec")
> 			return nil
> 		}
> ```
> 
> if that is not the case then the PR will not work as expected in the following scenario
> 
>     1. install the operator
> 
>     2. after the daemon created the devices in the status area and the user creates the policy the `IsPolicySet` will be set to true
> 
>     3. delete the sriov policy (the  `IsPolicySet` will still be true)
> 
>     4. reboot the node or the config daemon (it will still apply the plugins)
> 
> 
> Let me know if I miss something

I don't understand step 4. The spec in NS will eventually be updated(cleared in this case) and then the interfaces will be reset.

SchSeba:
> With if spec.interface is empty and latestState.GetGeneration() == 1: skip config there can be still a problematic scenario (not sure if it is a corner case):

> Policy with config + node state with spec
Delete policy and node state
Daemon will not reset the interfaces since the generation is 1 again after NS deletion

This can be a corner case but in general we can just add here
```
latestState, err = dn.client.SriovnetworkV1().SriovNetworkNodeStates(namespace).Get(context.Background(), dn.name, metav1.GetOptions{})
	if err != nil {
		glog.Warningf("nodeStateSyncHandler(): Failed to fetch node state %s: %v", dn.name, err)
		return err
	}
```

```
if errors.IsNotFound(err): reset sriov devices
```

rollandf:
@SchSeba 
Another issue is that in case that there is no policy, the generation will stay 1 and any existing interface configuration will stay as is and not reset.

SchSeba:
@rollandf as we spoke offline the operator will not reset sriov devices if there is no policy. 

the sriov operator is the only operator that should change the configuration on a node so if something/someone that is not the operator changes it the operator will **not** reset them and that is expected if you ask me

adrianchiris:
> the sriov operator is the only operator that should change the configuration on a node so if something/someone that is not the operator changes it the operator will not reset them and that is expected if you ask me

ack, one caveat i can think of is if somehow we end up in a state where config daemon configured node with sriov,  it then went down (no reboot) then policy and node state were deleted for the node and finally config daemon came up.

it would then leave the node configured with sriov.

we did not encounter such case so its purely theoretical. Im fine with current changes as it provides improvement over current state.

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -278,6 +278,11 @@ func (r *SriovNetworkNodePolicyReconciler) syncSriovNetworkNodeState(np *sriovne
 			return fmt.Errorf("failed to get SriovNetworkNodeState: %v", err)
 		}
 	} else {
+		if len(found.Status.Interfaces) == 0 {
```
adrianchiris, line 281: i think e2e test need to be adjusted for this change

from CI log:

```
[91m[1m• Failure in Spec Setup (BeforeEach) [900.023 seconds][0m
Operator
[90m/jenkins/workspace/sriov-network-operator-ci/test/e2e/sriovoperatornodepolicy_test.go:24[0m
  [91m[1mwith single policy [BeforeEach][0m
  [90m/jenkins/workspace/sriov-network-operator-ci/test/e2e/sriovoperatornodepolicy_test.go:43[0m
    should config sriov
    [90m/jenkins/workspace/sriov-network-operator-ci/test/e2e/sriovoperatornodepolicy_test.go:94[0m
      Set MTU and vfio driver
      [90m/jenkins/workspace/sriov-network-operator-ci/test/e2e/sriovoperatornodepolicy_test.go:166[0m

      [91mTimed out after 900.001s.
      Expected
          <*cluster.EnabledNodes | 0x0>: nil
      not to be nil[0m

      /jenkins/workspace/sriov-network-operator-ci/test/e2e/sriovoperatornodepolicy_test.go:35
[90m------------------------------[0m
[91m[1m•! Panic in Spec Setup (JustBeforeEach) [0.000 seconds][0m
Operator
[90m/jenkins/workspace/sriov-network-operator-ci/test/e2e/sriovoperatornodepolicy_test.go:24[0m
  with single policy
  [90m/jenkins/workspace/sriov-network-operator-ci/test/e2e/sriovoperatornodepolicy_test.go:43[0m
    [91m[1mshould config sriov [JustBeforeEach][0m
    [90m/jenkins/workspace/sriov-network-operator-ci/test/e2e/sriovoperatornodepolicy_test.go:94[0m
      Set MTU and default driver
      [90m/jenkins/workspace/sriov-network-operator-ci/test/e2e/sriovoperatornodepolicy_test.go:167[0m

      [91m[1mTest Panicked[0m
      [91mruntime error: invalid memory address or nil pointer dereference[0m
      /usr/local/go/src/runtime/panic.go:221

      [91mFull Stack Trace[0m
      github.com/k8snetworkplumbingwg/sriov-network-operator/test/e2e.glob..func3.2.1()
      	/jenkins/workspace/sriov-network-operator-ci/test/e2e/sriovoperatornodepolicy_test.go:88 +0x5c
      github.com/k8snetworkplumbingwg/sriov-network-operator/test/e2e.TestSriovTests(0x0)
      	/jenkins/workspace/sriov-network-operator-ci/test/e2e/e2e_tests_suite_test.go:48 +0x16c
      testing.tRunner(0xc0006824e0, 0x1a315b0)
      	/usr/local/go/src/testing/testing.go:1259 +0x102
      created by testing.(*T).Run
      	/usr/local/go/src/testing/testing.go:1306 +0x35a
```

adrianchiris (approved):

SchSeba (commented):
just a small nit beside that LGTM

rollandf:
@SchSeba Can you please take another look?

SchSeba (approved):

test/e2e/sriovoperatornodepolicy_test.go
adrianchiris: @rollandf question : do we need this now that we are updating sync status immediately to success on generation 1 ?

adrianchiris (commented):
final question about latest changes and i think we can merge

adrianchiris (approved):
This has been a long one :) LGTM

Issue #288, opened 2022-04-26
Add golangci-lint target


rollandf:
For better lint, add golangci-lint tool to project

adrianchiris:
+1 on this

we can reference from sriov-network-device-plugin

GA:
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/58eb3660dcaeec068c8c8878825c6d724020966a/.github/workflows/build-test-lint.yml#L45

adding a .golangci.yaml for linter config
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/58eb3660dcaeec068c8c8878825c6d724020966a/.github/workflows/build-test-lint.yml#L45


and makefile target
https://github.com/Mellanox/rdma-cni/blob/b013003eae5344065fa47699aecde4a733234faa/Makefile#L81

SchSeba:
@rollandf can we close this one or we are still missing some features?

rollandf:
We can close it. There are still linters that are commented out, but we can handle them without this issue.

PR #293, opened 2022-04-28
Add golangci-lint

rollandf:
- Add golangci.yaml config file
- Add Makefile lint target to golangci-lint
- Add GitHub action

Related to issue #288

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -44,3 +44,19 @@ jobs:
 
     - name: test bindata/scripts
       run: make test-bindata-scripts
+
+  golangci:
+    name: Golangci-lint
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go 1.17
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17.8
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v3
+      with:
+        # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
+        version: v1.46
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,119 @@
+# Tested with golangci-lint ver. 1.37
+run:
+  timeout: 3m
+  skip-dirs:
+    - vendor/
+    - .github/
+    - deployment/
+    - doc/
+    - bindata/
+linters-settings:
+  depguard:
+    list-type: blacklist
+    packages:
+      # logging is allowed only by logutils.Log, logrus
+      # is allowed to use only in logutils package
+      - github.com/sirupsen/logrus
+    packages-with-error-message:
+      - github.com/sirupsen/logrus: "logging is allowed only by logutils.Log"
+  dupl:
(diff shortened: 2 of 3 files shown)


.golangci.yml
```diff
@@ -0,0 +1,120 @@
+# Tested with golangci-lint ver. 1.37
+run:
```
zeeke, line 2: `run` key is already present on [line 114](https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/293/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R114). This duplication inhibits the `timeout :3m` spec and force you to add `args: --timeout 3m0s`  to the action job definition.

rollandf: Fixed, thanks

rollandf:
Commented out currently failing linters.
PRs fixing the issue should uncomment the specific linter fixed.

.github/workflows/go.yml
```diff
@@ -44,3 +44,19 @@ jobs:
...
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17.8
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v3
+      with:
+        # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
+        version: v1.37
```
adrianchiris, line 62: while we are at it, lets use a newer golangci version WDYT ?
latest is 1.46.1

rollandf: done

rollandf:
@SchSeba @adrianchiris PTAL

adrianchiris (approved):
Thanks @rollandf ! LGTM

Makefile
```diff
@@ -53,6 +53,15 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+GOLANGCI_LINT = $(GOBIN)/golangci-lint
+# golangci-lint version should be updated periodically
+# we keep it fixed to avoid it from unexpectedly failing on the project
+# in case of a version bump
+GOLANGCI_LINT_VER = v1.46.1
+TIMEOUT = 15
```
zeeke, line 61: This variable is actually unused. Maybe you wanted to use it in `-timeout` parameters of `go test`?

Makefile
```diff
@@ -53,6 +53,15 @@ else
...
+Q = $(if $(filter 1,$V),,@)
```
zeeke, line 62: `$V` is uninitialized. Maybe it comes from [this Makefile](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/master/Makefile#L53). I don't know if it can cause problems, as I didn't well understand what that wrapper actually does.

rollandf: Removing it.
Thanks for the review

Makefile
rollandf: Removing it.

rollandf:
@zeeke Can you check if any more changes are required?

adrianchiris:
Straightforward change that will improve the projects quality ! thanks @rollandf !

PR #299, opened 2022-05-02
Generic Plugin Reset interface on error

adrianchiris:
If error occures during sriov configuration in
generic plugin, config daemon will log the error
and requeue the request to reconcile the node.

On the next run, when generic plugin Apply() method
is called, interface configuration may be skipped,
leaving the interface partially configured.

To ensure generic plugin re-configures the interface
on the next iteration, we reset the interface
before returning an error.


e0ne (approved):
LGTM

adrianchiris:
@SchSeba PTAL

Eoghan1232 (approved):

rollandf:
LGTM

PR #301, opened 2022-05-05
Improve `GetSriovNicIPs` error message

zeeke:
In case of error, it is useful to know the reason why it was not possible to get the IP address of SR-IOV interface.
This commit also adds some unit test for `GetSriovNicIPs(...)`


rollandf:
LGTM

adrianchiris (approved):

PR #309, opened 2022-05-26
external control plane

zshi-redhat:
Enable sriov operator in the cluster w/o master nodes (k8s apiserver and other control plane components are managed externally).

### DIFF
--- a/bindata/manifests/operator-webhook/server.yaml
+++ b/bindata/manifests/operator-webhook/server.yaml
@@ -28,6 +28,7 @@ spec:
       nodeSelector:
         beta.kubernetes.io/os: linux
       affinity:
+        {{ if not .ExternalControlPlane }}
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
@@ -37,7 +38,9 @@ spec:
             - matchExpressions:
                 - key: node-role.kubernetes.io/control-plane
                   operator: Exists
+        {{ end }}
       tolerations:
+      {{ if not .ExternalControlPlane }}
       - key: "node-role.kubernetes.io/master"
         operator: Exists
         effect: NoSchedule
@@ -47,6 +50,7 @@ spec:
       - key: "node.kubernetes.io/not-ready"
         operator: Exists
         effect: NoSchedule
+      {{ end }}
       {{- if .ImagePullSecrets }}
       imagePullSecrets:
       {{- range .ImagePullSecrets }}
--- a/bindata/manifests/webhook/server.yaml
+++ b/bindata/manifests/webhook/server.yaml
@@ -8,7 +8,7 @@ metadata:
   namespace: {{.Namespace}}
   annotations:
     kubernetes.io/description: |
-      This daemon set launches the network resource injector component on master nodes.
+      This daemon set launches the network resource injector component on master or worker nodes.
     release.openshift.io/version: "{{.ReleaseVersion}}"
 spec:
   selector:
@@ -31,6 +31,7 @@ spec:
       nodeSelector:
         beta.kubernetes.io/os: linux
       affinity:
+        {{ if not .ExternalControlPlane }}
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
@@ -40,7 +41,9 @@ spec:
             - matchExpressions:
                 - key: node-role.kubernetes.io/control-plane
                   operator: Exists
+        {{ end }}
       tolerations:
+      {{ if not .ExternalControlPlane }}
       - key: "node-role.kubernetes.io/master"
         operator: Exists
         effect: NoSchedule
@@ -50,6 +53,7 @@ spec:
       - key: "node.kubernetes.io/not-ready"
         operator: Exists
         effect: NoSchedule
+      {{ end }}
       {{- if .ImagePullSecrets }}
       imagePullSecrets:
       {{- range .ImagePullSecrets }}
--- a/deployment/sriov-network-operator/templates/operator.yaml
+++ b/deployment/sriov-network-operator/templates/operator.yaml
@@ -81,6 +81,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
             - name: RELEASE_VERSION
               value: {{ .Release.AppVersion }}
             - name: SRIOV_CNI_BIN_PATH
--- a/pkg/utils/cluster.go
+++ b/pkg/utils/cluster.go
@@ -15,39 +15,99 @@ import (
 
 const (
 	// default Infrastructure resource name for Openshift
-	infraResourceName = "cluster"
+	infraResourceName        = "cluster"
+	workerRoleName           = "worker"
+	masterRoleName           = "master"
+	workerNodeLabelKey       = "node-role.kubernetes.io/worker"
+	masterNodeLabelKey       = "node-role.kubernetes.io/master"
+	controlPlaneNodeLabelKey = "node-role.kubernetes.io/control-plane"
 )
+
+func getNodeRole(node corev1.Node) string {
+	for k := range node.Labels {
+		if k == workerNodeLabelKey {
+			return workerRoleName
+		} else if k == masterNodeLabelKey || k == controlPlaneNodeLabelKey {
+			return masterRoleName
+		}
+	}
+	return ""
+}
 
 func IsSingleNodeCluster(c client.Client) (bool, error) {
 	if os.Getenv("CLUSTER_TYPE") == ClusterTypeOpenshift {
-		return openshiftSingleNodeClusterStatus(c)
+		topo, err := openshiftControlPlaneTopologyStatus(c)
+		if err != nil {
+			return false, err
+		}
+		if topo == configv1.SingleReplicaTopologyMode {
+			return true, nil
+		}
+		return false, nil
 	}
 	return k8sSingleNodeClusterStatus(c)
+}
+
+// IsExternalControlPlaneCluster detects control plane location of the cluster.
+// On OpenShift, the control plane topology is configured in configv1.Infrastucture struct.
+// On kubernetes, it is determined by which node the sriov operator is scheduled on. If operator
+// pod is schedule on worker node, it is considered as external control plane.
+func IsExternalControlPlaneCluster(c client.Client) (bool, error) {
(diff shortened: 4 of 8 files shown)


pkg/utils/cluster.go
```diff
@@ -40,14 +63,14 @@ func k8sSingleNodeClusterStatus(c client.Client) (bool, error) {
 	return false, nil
 }
 
-func openshiftSingleNodeClusterStatus(c client.Client) (bool, error) {
+func openshiftControlPlaneTopologyStatus(c client.Client) (configv1.TopologyMode, error) {
 	infra := &configv1.Infrastructure{}
 	err := c.Get(context.TODO(), types.NamespacedName{Name: infraResourceName}, infra)
 	if err != nil {
-		return false, err
+		return "", err
```
zeeke, line 70: Please, add context information to the error:

```suggestion
return "", fmt.Errorf("can't get Infrastructure [%s]: %w, infraResourceName, err)
```

zshi-redhat: updated.

adrianchiris:
@zshi-redhat can you explain abit more about the use-case ?

pkg/utils/cluster.go
```diff
@@ -20,34 +20,57 @@ const (
...
+		case configv1.SingleReplicaTopologyMode:
+			return true, nil
+		}
+		return false, nil
 	}
 	return k8sSingleNodeClusterStatus(c)
 }
 
+func IsExternalControlPlaneCluster(c client.Client) (bool, error) {
+	if os.Getenv("CLUSTER_TYPE") == ClusterTypeOpenshift {
```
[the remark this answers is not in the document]
zshi-redhat: updated to also support k8s platform.
for k8s, external-control-plane is determined by checking which node is operator pod scheduled on.

bindata/manifests/operator-webhook/server.yaml
```diff
@@ -32,13 +32,21 @@ spec:
...
+              {{ else }}
               - key: node-role.kubernetes.io/master
+              {{ end }}
                 operator: Exists
             - matchExpressions:
                 - key: node-role.kubernetes.io/control-plane
                   operator: Exists
       tolerations:
+      {{ if .ExternalControlPlane }}
+      - key: "node-role.kubernetes.io/worker"
```
adrianchiris, line 46: worker taint ? that's a thing ?

bindata/manifests/operator-webhook/server.yaml
```diff
@@ -32,13 +32,21 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
             - matchExpressions:
+              {{ if .ExternalControlPlane }}
```
adrianchiris, line 35: in case of external control plane maybe drop the entire  'requiredDuringScheduling....' 

or alternatively just add another `- matchExpression` as these are ORed

bindata/manifests/webhook/server.yaml
```diff
@@ -35,13 +35,21 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
             - matchExpressions:
+              {{ if .ExternalControlPlane }}
```
adrianchiris, line 38: same comment as above

deployment/sriov-network-operator/templates/operator.yaml
```diff
@@ -47,6 +47,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: K8S_POD_NAME
```
adrianchiris, line 50: 1. you already have POD_NAME defined below
2. you can use `spec.nodeName` to expose Node name directly and avoid logic added to get node name from pod

pkg/utils/cluster.go
```diff
@@ -15,39 +15,119 @@ import (
...
+func operatorNodeRole(c client.Client) (string, error) {
+	pod := &corev1.Pod{}
+	err := c.Get(context.TODO(), types.NamespacedName{Namespace: os.Getenv("NAMESPACE"), Name: os.Getenv("K8S_POD_NAME")}, pod)
+	if err != nil {
+		glog.Errorf("k8sIsExternalTopologyMode(): Failed to get operator pod: %v", err)
+		return "", err
+	}
+
+	nodeList := &corev1.NodeList{}
+	err = c.List(context.TODO(), nodeList)
```
adrianchiris, line 109: you gen get node by name no ? no need to list and iterate here

pkg/utils/cluster.go
```diff
@@ -15,39 +15,119 @@ import (
 
 const (
 	// default Infrastructure resource name for Openshift
-	infraResourceName = "cluster"
+	infraResourceName  = "cluster"
+	workerNodeLabelKey = "node-role.kubernetes.io/worker"
+	masterNodeLabelKey = "node-role.kubernetes.io/master"
```
adrianchiris, line 20: what about `node-role.kubernetes.io/control-plane` ?
in k8s 1.24 kubeadm now replaced master with control-plane (master label deprecated)

zshi-redhat: added support for both.

pkg/utils/cluster.go
zshi-redhat: fixed.

pkg/utils/cluster.go
```diff
@@ -15,39 +15,119 @@ import (
...
+			return false, nil
+		case configv1.SingleReplicaTopologyMode:
+			return false, nil
+		}
+	} else {
+		role, err := operatorNodeRole(c)
+		if err != nil {
+			return false, err
+		}
+		switch role {
```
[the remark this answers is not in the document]
zshi-redhat: fixed.

bindata/manifests/webhook/server.yaml
```diff
@@ -35,13 +35,21 @@ spec:
...
+              {{ else }}
               - key: node-role.kubernetes.io/master
+              {{ end }}
                 operator: Exists
             - matchExpressions:
                 - key: node-role.kubernetes.io/control-plane
                   operator: Exists
       tolerations:
+      {{ if .ExternalControlPlane }}
+      - key: "node-role.kubernetes.io/worker"
```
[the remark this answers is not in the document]
zshi-redhat: removed "node-role.kubernetes.io/worker"

bindata/manifests/webhook/server.yaml
zshi-redhat: fixed.

bindata/manifests/operator-webhook/server.yaml
zshi-redhat: removed.

bindata/manifests/operator-webhook/server.yaml
zshi-redhat: dropped `requiredDuringScheduling` in External mode.

deployment/sriov-network-operator/templates/operator.yaml
zshi-redhat: removed `K8S_POD_NAME` and added `NODE_NAME`

pkg/utils/cluster.go
```diff
@@ -15,39 +15,105 @@ import (
...
+		}
+		switch topo {
+		case "External":
+			return true, nil
+		case configv1.HighlyAvailableTopologyMode:
+			return false, nil
+		case configv1.SingleReplicaTopologyMode:
+			return false, nil
+		}
+	} else {
```
bn222, line 70: Nit: Let's do:

`
else if ... == ClusterTypeKubernetes {
    ...
} else {
    // return error
}
`

bn222:
Can you put your first comment in the commit message? Or something more elaborate?

pkg/utils/cluster.go
zshi-redhat: Done

zshi-redhat:
> Can you put your first comment in the commit message? Or something more elaborate?

added the comment in the commit message.

SchSeba:
Hi @zshi-redhat will you please be able to rebase this PR?

zshi-redhat:
> Hi @zshi-redhat will you please be able to rebase this PR?

rebased, thanks for reminding!

bn222:
@zshi-redhat We could label the workers in guest cluster with label "master" (such that the role will be master,worker). In that case, this would not be needed. For the sake of my understanding, is that correct? I still think it's nice to allow to run on workers so we will still want to have this PR in.

zshi-redhat:
> @zshi-redhat We could label the workers in guest cluster with label "master" (such that the role will be master,worker). In that case, this would not be needed. For the sake of my understanding, is that correct? 

@bn222  This may not be desired from my understanding, since the node pool is supposed to only contain worker nodes, not sure if there will be any side effects to other pods.

btw, do we still need this PR? or is it going to be integreted to Tyler's?

zshi-redhat:
@adrianchiris  PTAL.

pkg/utils/cluster.go
```diff
@@ -15,39 +15,105 @@ import (
...
+// pod is schedule on worker node, it is considered as external control plane.
+func IsExternalControlPlaneCluster(c client.Client) (bool, error) {
+	if os.Getenv("CLUSTER_TYPE") == ClusterTypeOpenshift {
+		topo, err := openshiftControlPlaneTopologyStatus(c)
+		if err != nil {
+			return false, err
+		}
+		switch topo {
+		case "External":
+			return true, nil
```
adrianchiris, line 64: nit you can just check with an if statement that `topo == "External" the fallback at line 79 will take care of the rest

pkg/utils/cluster.go
```diff
@@ -15,39 +15,105 @@ import (
...
+
 func IsSingleNodeCluster(c client.Client) (bool, error) {
 	if os.Getenv("CLUSTER_TYPE") == ClusterTypeOpenshift {
-		return openshiftSingleNodeClusterStatus(c)
+		topo, err := openshiftControlPlaneTopologyStatus(c)
+		if err != nil {
+			return false, err
+		}
+		switch topo {
+		case configv1.SingleReplicaTopologyMode:
```
adrianchiris, line 42: nit: you can just use if statement here and compare `topo == configv1.SingleReplicaTopoligyMode`
and return `true, nil`

the fallback at L47 will take care of the rest

adrianchiris (commented):
a couple of nits otherwise LGTM

pkg/utils/cluster.go
zshi-redhat: done

pkg/utils/cluster.go
zshi-redhat: done

adrianchiris (approved):
LGTM, lets wait for e2e test to pass

Issue #90, opened 2021-03-15
Statically-linked plugins


manuelbuil:
Plugins are currently built using the buildmode=plugins and used by the network config daemon through the plugin library: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/daemon/plugin.go#L38. The consequence of this approach is that the xxx.so plugin libraries are dynamically linked with libc, as it requires dlopen() to load them. That makes it harder to move the binary across platforms and is at odds with the go best practices, that recommend statically linked builds.

I guess the reason to use the plugins library is to be able to update plugins without having to build the whole network config daemon again. However, as network config daemon is deployed in a container, the use case of replacing one xxx.so plugin while the container is running is, in my opinion, a bit obscure as we would break the immutability. Instead, the best practice would be to generate a new image that contains the replaced xxx.so plugin and thus, it does not matter if we had to rebuild network config daemon again, as it needs to be redeployed anyway.

I wanted to start a discussion about this to understand if there are more use cases that the plugins library fulfill and that I am missing. And also to see if we could agree on a different and statically-linked way to integrate plugins into the network config daemon and understand

pliurh:
Initially, the idea was that vendors may want to create/maintain their plugin code out of the SRIOV operator repo. And the go plugin is the simplest approach from the implementation perspective comparing to other alternatives (hashicorp go-plugin, side-car pods talk with gRPC). And this was an openshift project back to that time, so the dynamically linked with libc was not an issue.

@manuelbuil what's your suggestion if we want to drop the golang plugin mechanism.

manuelbuil:
Thanks for the reply @pliurh. Do you see that requirement still being valid? I mean, are you expecting vendors to develop their plugins out of the SRIOV repo? If that is the case, perhaps we could explore the alternative of doing it as Kubernetes does it with the CNI plugins, i.e. each plugin could be a binary with a simple CLI and sriov-operator executes the CLI passing parameters through STDIN. I think that would make the binary static

pliurh:
I suppose not anymore. @adrianchiris @martinkennelly any comments?

martinkennelly:
@manuelbuil My preference is to keep vendor work in-tree so the plugin design is of limited value to me right now.
If I wanted to add functionality there, I would push it to this repository.
I am fine with converting it but that is easy for me to say since we have no functionality in the Intel plugin right now.

@adrianchiris WDYT?

[...]

PR #312, opened 2022-06-01
Remove go plugins

SchSeba:
This commit removes the go plugins for vendors.

Fixes: https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/90


pkg/daemon/plugin.go
```diff
@@ -1,11 +1,12 @@
...
 
 import (
-	"fmt"
-	"plugin"
-
-	"github.com/golang/glog"
 	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	genericplugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/generic"
+	intelplugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/intel"
+	k8splugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/k8s"
```
zeeke, line 7: I have a little concern about this import, as the `k8s.init()` func contains a [panic call](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/ba5ad00dccdb2c5949a507af3ce9d1787c79e4ee/pkg/plugins/k8s/k8s_plugin.go#L104). I'm not sure about how likely it is going to happen in production, but prior to this changes the init func was called only 

```
if utils.ClusterType != utils.ClusterTypeOpenshift {
```

zeeke: That's good. thanks

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -127,6 +129,14 @@ func (p *K8sPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetworkNodeS
 	needDrain = false
 	needReboot = false
 
+	if p.loadedManifests != loaded {
```
adrianchiris, line 132: to avoid this, you could define a `New<plugin-name>Plugin()` method for the plugins and use it in `plugins.go`

in this method you can perform one-time initialization.

e0ne (approved):
/LGTM

zeeke:
LGTM

zeeke:
LGTM

pkg/plugins/generic/generic_plugin.go
```diff
@@ -29,15 +33,13 @@ const (
 	loaded
 )
 
-var Plugin GenericPlugin
-
 // Initialize our plugin and set up initial values
-func init() {
-	Plugin = GenericPlugin{
-		PluginName:     "generic_plugin",
+func NewPlugin() (plugin.VendorPlugin, error) {
```
adrianchiris, line 37: nit: NewGenericPlugin

pkg/daemon/daemon.go
```diff
@@ -637,31 +635,44 @@ func (dn *Daemon) restartDevicePluginPod() error {
 	return nil
 }
 
-func (dn *Daemon) loadVendorPlugins(ns *sriovnetworkv1.SriovNetworkNodeState) error {
-	var pl []string
+func (dn *Daemon) EnableVendorPlugins(ns *sriovnetworkv1.SriovNetworkNodeState) error {
```
adrianchiris, line 638: by no mean a must but you could move these two methods (EnableVendorPlugins and registerVendorPlugins) to daemon/plugins.go

pkg/daemon/daemon.go
```diff
@@ -74,7 +73,7 @@ type Daemon struct {
 
 	nodeState *sriovnetworkv1.SriovNetworkNodeState
 
-	LoadedPlugins map[string]VendorPlugin
+	enabledPlugins map[string]plugin.VendorPlugin
```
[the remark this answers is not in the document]
SchSeba: enabled means this are the plugins in use by the config daemon.

for example on a system with intel only it will only enabled that one or for hwoffload it will load both k8s and mlx so I think the node is fine :)

pkg/daemon/daemon.go
SchSeba: much cleaner agree, done

pkg/daemon/daemon.go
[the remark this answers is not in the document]
adrianchiris: ack

pkg/daemon/plugin.go
```diff
@@ -2,60 +2,86 @@ package daemon
...
 )
 
-// loadPlugin loads a single plugin from a file path
-func loadPlugin(path string) (VendorPlugin, error) {
-	glog.Infof("loadPlugin(): load plugin from %s", path)
-	plug, err := plugin.Open(path)
-	if err != nil {
-		return nil, err
-	}
+func EnablePlugins(platform utils.PlatformType, ns *sriovnetworkv1.SriovNetworkNodeState) (map[string]plugin.VendorPlugin, error) {
```
adrianchiris, line 30: nit : make this package internal ? (`enablePlugins`)

adrianchiris (commented):
a couple of nits otherwise LGTM ! thanks for addressing my comments

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -35,18 +38,18 @@ const (
...
 var mellanoxNicsStatus map[string]map[string]sriovnetworkv1.InterfaceExt
 var mellanoxNicsSpec map[string]sriovnetworkv1.Interface
 
 // Initialize our plugin and set up initial values
-func init() {
-	Plugin = MellanoxPlugin{
-		PluginName:  "mellanox_plugin",
-		SpecVersion: "1.0",
-	}
+func NewMellxanoxPlugin() (plugin.VendorPlugin, error) {
```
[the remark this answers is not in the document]
adrianchiris: ack i see. alternative is to define a factory method that creates plugin based on vendor ID, then use it.

pkg/daemon/plugin.go
SchSeba: done

adrianchiris (approved):
lets just make sure vendor CI passes and we can merge IMO

LGTM

PR #313, opened 2022-06-02
Update sysctl to a safe one in metapluging chaining test

mmirecki:
The test should use a network interface level sysctl.


SchSeba (approved):
/lgtm

adrianchiris (approved):

Issue #322, opened 2022-06-16
Drain node doesn't work after reboot


e0ne:
After node is rebooted operator doesn't drain node. It leads to non-working workloads on a rebooted node

adrianchiris:
@SchSeba have you encountered this in Openshift ?

adrianchiris:
So i assume the check is there since, when moving from zero to non-zero number of VFs it is assumed that there are no pods consuming VFs which makes drain operation not needed.

now, say we have a deployment with pods running on a node (these pods have sriov secondary network).

On node reboot. if node comes up "fast" enough the pods of the deployment will not get rescheduled.
then kubelet will spin them up when it comes up. at that point i expect CNI call to fail and pod to enter error state.

@e0ne is that the case ?

SchSeba:
@adrianchiris we see this in OCP also and there is an open issue on k8s (kubelet) for it https://github.com/kubernetes/kubernetes/issues/109595

can you try to remove the checkpoint file from kubelet on reboot that should stuck the pods on pending until the device plugin is really ready

adrianchiris:
thanks for linking the k8s issue.

So, until this is sorted, i think we should introduce a temporary fix in sriov network operator to avoid such cases.
for that i think drain is the correct way to go as this (node reboot) is not something is expected to happen often.
(with a TODO to remove it ofc)

id prefer to avoid deleting kubelet cache files in system service, this feels to me like adding un-needed complexity which also tinkers with a critical component that we don't really own (and might have un-intentional side effects ).

PR #323, opened 2022-06-16
Fix drain logic after node reboot

e0ne:
We don't need to check for ifaceStatus.NumVfs in a 'needDrainNode'
function because it blocks drain after node reboot.

Closes: #322

Signed-off-by: Ivan Kolodiazhnyi <ikolodiazhny@nvidia.com>


zeeke:
Is there a chance to add some unit tests for this scenario? Plugin's logic starts becoming a little complex to leave it untested.

adrianchiris (approved):

adrianchiris:
Note: additional discussion in #322

adrianchiris:
@SchSeba are you ok with current approach ?

Eoghan1232 (approved):
Overall, LGTM.

PR #325, opened 2022-06-16
Generate random mac address when virtual mac address is zero.

vrindle:


vrindle:
@zshi-redhat This generates a random mac address. Sending to you for review.

pkg/utils/utils.go
```diff
@@ -329,6 +331,13 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 							return err
 						}
 					}
+					if addr == zero_mac_address {
```
zshi-redhat, line 334: `addr` here is VF PCI address, you would need to check all zero mac against `vfLink.Attrs().HardwareAddr`.

SchSeba:
Hi @vrindle can you please give some info about this PR?

If I remember we copy the mac address from the interface attached to the kernel driver into the vf <id> mac address

zshi-redhat:
> Hi @vrindle can you please give some info about this PR?
> 
> If I remember we copy the mac address from the interface attached to the kernel driver into the vf mac address

I think the issue is the VF effective mac address is all zero for some vendor NICs.

SchSeba (commented):
thanks for the PR

/lgtm

vrindle:
@adrianchiris Could you review this PR as well? Thank you.

pkg/utils/utils.go
```diff
@@ -374,6 +383,17 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
 
+func generateRandomMacAddressVfs() (string, error) {
+	buf := make([]byte, 6)
+	_, err := rand.Read(buf)
+	if err != nil {
+		return nil, err
+	}
+	// Set the local bit
+	buf[0] |= 2
+	return fmt.Sprintf("%02x:%02x:%02x:%02x:%02x:%02x", buf[0], buf[1], buf[2], buf[3], buf[4], buf[5]), nil
```
bn222: Let's add a check and retry until we stop being unlucky (i.e. retry when we generate 00:00... or FF:FF... randomly).

pkg/utils/utils.go
```diff
@@ -383,15 +375,15 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 	return nil
 }
 
-func generateRandomMacAddressVfs() (string, error) {
+func generateRandomMacAddressVfs() ([]byte, error) {
 	buf := make([]byte, 6)
 	_, err := rand.Read(buf)
```
wizhaoredhat, line 380: Some MAC addresses are restricted. E.g. IEEE802.1 reserved MACs.
If we are unlucky and somehow land on restricted MAC addresses, we shouldn't use those MAC address for the interface.

wizhaoredhat: OR even multicast MAC addresses shouldn't be allowed.

bn222: Yes, we can go very far with this and also disallow MACs that have been generated for other VFs. I think if we just fix the first byte to some hardcoded value, then we are ok.

vrindle: [We or the first byte with ](https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/325/files#diff-81ddbadfb415ccbb9c7af84f11668d1aa5e53c34025bf86d4702f16b4e42f045R385)2 and guarantee the local bit is set so we know we have a unicast address.

Is there anything else we need to do here?

bn222: looks ok, thanks

adrianchiris:
Hi, Can you provide more context in commit message on why this is needed ?

vrindle:
> Hi, Can you provide more context in commit message on why this is needed ?

Hi, I will fix up the commit message, once the CI is passing. Essentially VF effective mac address is all zero for some vendor NICs, so we have to set the VF effective MAC address so that it passes our QE tests. I will change the commit message to make sure that it reflects why this change is necessary.

pkg/utils/utils.go
```diff
@@ -374,6 +375,17 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
 }
 
+func generateRandomMacAddressVfs() ([]byte, error) {
+	buf := make([]byte, 6)
+	_, err := rand.Read(buf)
+	if err != nil {
+		return nil, err
+	}
+	// Set the local bit
+	buf[0] |= 2
```
adrianchiris, line 385: as far as i see. 

need to:
1. make sure mcast bit is 0 
2. local bit is 1

this only takes care of the latter

`buf[0] = (buf[0] & 0xfe) |  2`  will do the trick.

since you have 1 bit which is 0 and one bit with is 1  then the generated address cannot be all zeroes or all ones

adrianchiris: was looking at :

adrianchiris:
> Hi, I will fix up the commit message, once the CI is passing. Essentially VF effective mac address is all zero for some vendor NICs, so we have to set the VF effective MAC address so that it passes our QE tests. I will change the commit message to make sure that it reflects why this change is necessary.

Which vendor ? why is the MAC address zero for VF netdev, this seems like a kernel (driver) bug.

pkg/utils/utils.go
```diff
@@ -600,6 +612,31 @@ func vfIsReady(pciAddr string) (netlink.Link, error) {
...
+	case
+		[6]byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00},
+		[6]byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF}:
+		return false
+	}
+	return true
+}
+
+func getEffectiveMacAddress(hwAddr HardwareAddr) (HardwareAddr, error) {
+	newAddr = [6]byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00}
```
adrianchiris, line 626: why not use: `make([]byte, 6)` ? it will result in a slice of zero bytes ?

vrindle: The reason was that by making it more explicit, it would be easier to read.

vrindle: @adrianchiris Is that alright? 

> > Hi, I will fix up the commit message, once the CI is passing. Essentially VF effective mac address is all zero for some vendor NICs, so we have to set the VF effective MAC address so that it passes our QE tests. I will change the commit message to make sure that it reflects why this change is necessary.
> 
> Which vendor ? why is the MAC address zero for VF netdev, this seems like a kernel (driver) bug.

This is for Pensando DSC cards that we need to enable to get SRIOV network support on them. While other vendors do provide this within the drivers, this vendor does not. Therefore we must set it in the SRIOV network operator.

bn222:
iI a way (besides checking the mac) that the card is **purposefully** not setting the MAC? We've had other cases where the mac was not set unintentionally. With your changes, the mac will be set either way complicating the debug process.

vrindle:
> iI a way (besides checking the mac) that the card is **purposefully** not setting the MAC? We've had other cases where the mac was not set unintentionally. With your changes, the mac will be set either way complicating the debug process.

Based on my dialogue with Pensando, the MAC address is not being set because it is not deemed to be necessary. The other vendors, however, do set the MAC address for the effective VF. In order to pass QE tests the VF effective MAC address must be set, hence this change.

bn222:
I understand that we need to set the MAC to something, but to determine that you just check "if mac is zero, then we set it". Can we have something better than "if mac is zero", something like "if the kernel reports that the mac will not be set purposefully for this device, then we set it"?

Suppose that we have such a nob (the question here is if such a thing exists?), then we use it in place of the "if mac is zero" predicate.

bn222:
I've discussed the idea to use something more clever for "if mac is zero" offline with @vrindle. At this point, we will go ahead but keep a note for future reference to come back to this

pkg/utils/utils.go
```diff
@@ -375,6 +376,17 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
 }
 
+func generateRandomMacAddressVfs() ([6]byte, error) {
+	buf := make([]byte, 6)
+	_, err := rand.Read(buf)
+	if err != nil {
+		return nil, err
+	}
+	// Set the local bit
+	buf[0] = (buf[0] & 0xfe) | 2
```
wizhaoredhat, line 386: From the Kernel:
/**
 * eth_random_addr - Generate software assigned random Ethernet address
 * @addr: Pointer to a six-byte array containing the Ethernet address
 *
 * Generate a random Ethernet address (MAC) that is not multicast
 * and has the local assigned bit set.
 */
static inline void eth_random_addr(u8 *addr)
{
	get_random_bytes(addr, ETH_ALEN);
	addr[0] &= 0xfe;	/* clear multicast bit */
	addr[0] |= 0x02;	/* set local assignment bit (IEEE802) */
}

Could you expand the commentary to explain your bit operations?

wizhaoredhat: Yes, I agree that it is the same. But your commentary "// Set the local bit" could be improved to explain the bit operations in more detail.

vrindle: I can 

> > iI a way (besides checking the mac) that the card is **purposefully** not setting the MAC? We've had other cases where the mac was not set unintentionally. With your changes, the mac will be set either way complicating the debug process.
> 
> This is a very good point. As per the further discussions, it seems that this is a Pensando specific issue. I think we have the following options:
> 
> 1. Create the random MAC addresses in the Kernel driver for Pensando cards. Similar to the IAVF drivers.
> 2. Create a PensandoPlugin for this specific issue to isolate the changes to Pensando cards only.
> 3. Keep this PR as is. However the con is what Balazs was quoted above.
> 
> Did we go over these options and decide that option 3 is the best approach?

Discussed with @bn222 offline. We will look into this as a seperate issue.

PR #327, opened 2022-06-23
Revert "Generate random mac address when virtual mac address is zero."

adrianchiris:
Reverts k8snetworkplumbingwg/sriov-network-operator#325

- There should be acceptance from more than one company according to organization policy
- There are still concerns for this PR which i have.
   - pensando is not officially supported so why are we adding WA for this ?
   - why isnt mac addresss set in driver ? 
   - if it is deemed un-nessecary for pensando kenrel folks then this perhaps should be handled in a vendor plugin and not in main code. this requires further discussion

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -330,7 +330,6 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 							return err
 						}
 					}
-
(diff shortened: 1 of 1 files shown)


adrianchiris:
@vrindle please re-propose the original PR and lets have further discussion on it

1. follow this doc to add support for pensando to begin with : https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/doc/supported-hardware.md
2.  re propose PR for the WA with pensando driver

PR #330, opened 2022-06-30
Make Operator Config more strict

zeeke:
Asserting the cluster does not contain "*operator-webhook*" cluster role
binding can lead to failed test if there are other operators installed.

Signed-off-by: Andrea Panattoni <apanatto@redhat.com>


SchSeba:
/lgtm

Thanks @zeeke !

zeeke:
Did some rework as discussed during the last community meeting. 

Also reverted the renaming of test files, as they need to be without `_test.go` suffix.

zeeke:
As discussed during the last WG meeting, I added comments to `test/conformance/tests` and `test/validation/tests` to pay attention on import mechanism.

About the linter exception, if not using `dot-import-whitelist`, the only other way is to add the two folders to the `run.skip-dirs` array, but it would turn off every linter, as it is a golangci-lint setting. IMO I prefer to keep some linter rules on them.

SchSeba (approved):
Thanks!

/lgtm

adrianchiris (approved):

PR #331, opened 2022-07-06
Create new api module with minimal dependencies

dprince:
This patch creates a new '/api' go module with the existing v1 API. This makes it possible for
external projects to import the API structs for the SRIOV Operator without the larger
set of dependencies in the projects top level go.mod

In order to minimize external deps 2 functions using the 'render' package which required machine-config-operator
needed to be moved to utils and converted to standard golang functions.

A similar API package pattern (if implemented) in MCO would be potentially helpful to isolote SRIOV Operator from its dependencies.

See notes on multiple go.mod files in the same repository here:
https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository
(See the "Two example scenarios where it can make sense to have more than one go.mod in a repository" section).


dprince:
For some extra context we are using the SRIOV Operator from the [OSP Director Operator](https://github.com/openstack-k8s-operators/osp-director-operator) and having the ability to import just the API go module would help keep our imports to a minimum

SchSeba:
Hi @dprince will this create duplicate stuff between the 2 vendor folders?

dprince:
> Hi @dprince will this create duplicate stuff between the 2 vendor folders?

Hi. It would create a duplicate of a subset of the items (only the ones in the api/go.mod). The alternative approach would be to move the API into its own repo which wouldn't work as nicely with operator SDK and would still have its own copy of a subset of the go.mod items.

We recently found that we needed to bump up to operator SDK 1.22 which requires controller-runtime 12.1 to do this as well as that fixes an issue with 'make bundle' with nested API modules. Other than that we (openstack-k8s-operators) intend to adopt this same patter for our operators.

dprince:
If it is agreed and helps I could push a patch to bump to SDK 1.22 first? Going from 19 to 22 isn't that difficult however it does recommend bumping to golang 1.18.

jjuele:

PR #333, opened 2022-07-10
Wait for writer response on Refresh

rollandf:
When sending a Refresh message to Writer's refresh channel,
if the state is "Succeeded" or "Failed", the Writer will update
the sync channel when finished.

Therefore there is a need to wait on sync channel on the Daemon side
when sending "Succeeded" or "Failed".

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -413,6 +413,8 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 				syncStatus:    "Succeeded",
 				lastSyncError: "",
 			}
+			// wait for writer to refresh the status
+			<-dn.syncCh
 		}
 
 		return nil


rollandf:
@fedepaol can you PTAL?

pkg/daemon/daemon.go
```diff
@@ -413,6 +413,8 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 				syncStatus:    "Succeeded",
 				lastSyncError: "",
 			}
+			// wait for writer to refresh the status
+			<-dn.syncCh
```
adrianchiris, line 417: IIUC as SyncCh is initialized with a zero capacity. then [1] will block until someone is trying to read from the channel.

nice catch here @rollandf 

[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/815fd134ba8000756791051fca60179ec66ddb46/pkg/daemon/writer.go#L98

adrianchiris (approved):

e0ne (approved):

PR #339, opened 2022-07-14
Makefile: copy generated CRDs to Helm dir

rollandf:
On `manifests` target, after generating the CRDS, copy them
to the Helm `crds` directory.


e0ne (approved):
/LGTM
It would be good to check this in CI but I don't want to block patch for it since it fixes real issue

adrianchiris (approved):

adrianchiris:
@SchSeba @Eoghan1232 PTAL, small change here

SchSeba (approved):
/lgtm

PR #341, opened 2022-07-15
Remove duplicated and unused code in plugins

e0ne:


zeeke:
LGTM . Thanks for cleaning

adrianchiris:
I am a bit on the fence here with this change.

although not directly used i think its useful to:

1. have a way to determine when node state is first added
2. have access to the "old" state previous to the change

which is more consistent with the way k8s client informer callback work.

An alternative is to keep only `OnNodeStateChange` like in this change (drop OnNodeStateAdd).
but keep old, new args. then on Add daemon will call : `OnNodeStateChange(nil, new)`

e0ne:
> I am a bit on the fence here with this change.
> 
> although not directly used i think its useful to:
> 
> 1. have a way to determine when node state is first added

I agree on this, so let's leave this https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/341/files#diff-a53b7b593d3d778e62eaeeafa40088656f9212bfa2c2b7991df15fa78e60b0f0L454 'if' statement and move loggin into a daemon.go


> 2. have access to the "old" state previous to the change
> 
> which is more consistent with the way k8s client informer callback work.
> 
> An alternative is to keep only `OnNodeStateChange` like in this change (drop OnNodeStateAdd). but keep old, new args. then on Add daemon will call : `OnNodeStateChange(nil, new)`

I don't understand why do we need to have a dead code just to have a consistency with k8s client. Once we need 'old' object  in OnNodeStateChange we can re-introduce it

pkg/plugins/intel/intel_plugin.go
```diff
@@ -33,13 +33,7 @@ func (p *IntelPlugin) Spec() string {
...
 }
 
-// OnNodeStateAdd Invoked when SriovNetworkNodeState CR is created, return if need dain and/or reboot node
-func (p *IntelPlugin) OnNodeStateAdd(state *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
-	glog.Info("intel-plugin OnNodeStateAdd()")
-	return false, false, nil
-}
-
-// OnNodeStateChange Invoked when SriovNetworkNodeState CR is updated, return if need dain and/or reboot node
+// OnNodeStateChange Invoked when SriovNetworkNodeState CR is created or updated, return if need dain and/or reboot node
```
adrianchiris, line 36: nit: technically this plugin does nothing, so we dont really need it.

anyway this can be another cleanup in a separate PR :)

adrianchiris (approved):

SchSeba (approved):
/lgtm

PR #344, opened 2022-07-25
load vhost-net kernel module

SchSeba:
This commit is needed when using DPDK with vhost-net in virtio mode.
We must load the vhost-net kernel module if we don't want the dpdk
application to run with privilege

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -206,6 +206,7 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 
 	tryEnableRdma()
 	tryEnableTun()
+	tryEnableVhostNet()
 
 	if err := dn.tryCreateUdevRuleWrapper(); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -905,6 +906,12 @@ func tryEnableTun() {
 	}
 }
 
+func tryEnableVhostNet() {
+	if err := utils.LoadKernelModule("vhost-net"); err != nil {
+		glog.Errorf("tryEnableTun(): VHOST-NET kernel module not loaded: %v", err)
```
zeeke, line 911: typo: please change to `tryEnableVhostNet()`

pkg/daemon/daemon.go
```diff
@@ -905,6 +906,12 @@ func tryEnableTun() {
 	}
 }
 
+func tryEnableVhostNet() {
+	if err := utils.LoadKernelModule("vhost-net"); err != nil {
```
zeeke, line 910: Maybe better to change to `vhost_net`, as `lsmod` print modules with underscore and we would not grep it in [load-kmod.sh](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/bindata/scripts/load-kmod.sh)


```
lsmod | grep "vhost_net"                                                                                                                                                                                             
vhost_net              28672  0
vhost                  49152  1 vhost_net
tap                    28672  1 vhost_net
tun                    49152  4 vhost_net
```

```suggestion
	if err := utils.LoadKernelModule("vhost_net"); err != nil {
```

pkg/daemon/daemon.go
SchSeba: nice catch!

pkg/daemon/daemon.go
SchSeba: both of them work I am fine chaining :)

zeeke:
LGTM

adrianchiris (approved):

PR #346, opened 2022-07-28
Fix unit test

SchSeba:


Makefile
```diff
@@ -203,21 +204,18 @@ test-e2e-conformance:
...
 
 test-bindata-scripts: fakechroot
 	fakechroot ./test/scripts/enable-kargs_test.sh
 
-test-%: generate vet manifests
-	mkdir -p ${ENVTEST_ASSETS_DIR}
-	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.8.3/hack/setup-envtest.sh
-	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test ./$*/... -coverprofile cover-$*.out -coverpkg ./... -v
+test-%: generate vet manifests envtest
+test-%: envtest
```
bn222, line 217: 2x test-%

Makefile
```diff
@@ -203,21 +204,18 @@ test-e2e-conformance:
...
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir=/tmp -p path)" HOME=$(shell pwd) go test ./$*/... -coverprofile cover-$*.out -coverpkg ./... -v
```
bn222, line 218: what if `pwd` returns a string with a space? I think we need some quotes around the call to the shell

Makefile
```diff
@@ -153,6 +150,10 @@ KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary.
 	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.5.5)
 
+ENVTEST = $(shell pwd)/bin/setup-envtest
```
bn222, line 153: Also, seems to need quotes here?

SchSeba: we don't have that in any of the other installations do you want me to add?

bn222: should be ok as it is now, but just something to keep in mind

Makefile
```diff
@@ -203,21 +204,17 @@ test-e2e-conformance:
 test-e2e-validation-only:
 	SUITE=./test/validation ./hack/run-e2e-conformance.sh	
 
-test-e2e: generate vet manifests skopeo
-	mkdir -p ${ENVTEST_ASSETS_DIR}
-	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.8.3/hack/setup-envtest.sh
-	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); source hack/env.sh; go test ./test/e2e/... -timeout 60m -coverprofile cover.out -v
+test-e2e: generate vet manifests skopeo envtest
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir=/tmp -p path)"; source hack/env.sh; HOME="$(shell pwd)" go test ./test/e2e/... -timeout 60m -coverprofile cover.out -v
```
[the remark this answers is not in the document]
SchSeba: I wasn't sure if it needs it or not. it was using it before so I did the same :)

e0ne (approved):
/LGTM

adrianchiris (approved):

PR #351, opened 2022-08-17
Set `mod=readonly` when installing binaries

zeeke:
If no `-mod` argument is specified, GOLFAGS environment
variable takes place. If it contains `mod=vendor` (like Makefile
actually does), goinstall can fail with:

```
go install: github.com/onsi/ginkgo/ginkgo@v1.16.5: cannot query module due to -mod=vendor
```

Signed-off-by: Andrea Panattoni <apanatto@redhat.com>


zeeke:
Build is failing due to go1.19, fixed in 
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/352

SchSeba (approved):

adrianchiris (approved):

PR #310, opened 2022-05-29
Bump to golang 1.18

SchSeba:

PR #352, opened 2022-08-17
Stick with go <1.19

zeeke:
Using a dynamic version like '>=1.18.2' may lead to unexpected
build results, with new go tools in place.

This PR should fix [errors like this](https://github.com/k8snetworkplumbingwg/sriov-network-operator/runs/7875123215?check_suite_focus=true), as they occur due to go 1.19

```
Run actions/setup-go@v3
Setup go version spec >=1.18.2
Found in cache @ /opt/hostedtoolcache/go/1.19.0/x64
Added go to the path
Successfully set up Go version >=1.1[8](https://github.com/k8snetworkplumbingwg/sriov-network-operator/runs/7839168685?check_suite_focus=true#step:2:9).2
go version go1.1[9](https://github.com/k8snetworkplumbingwg/sriov-network-operator/runs/7839168685?check_suite_focus=true#step:2:10) linux/amd64
```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -12,7 +12,7 @@ jobs:
     - name: Set up Go 1.18
       uses: actions/setup-go@v3
       with:
-        go-version: '>=1.18.2'
+        go-version: '>=1.18.2 <1.19'
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
@@ -31,7 +31,7 @@ jobs:
     - name: Set up Go 1.18
       uses: actions/setup-go@v2
       with:
-        go-version: 1.18.2
+        go-version: '>=1.18.2 <1.19'
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
@@ -52,7 +52,7 @@ jobs:
     - name: Set up Go 1.18
       uses: actions/setup-go@v2
       with:
-        go-version: 1.18.2
+        go-version: '>=1.18.2 <1.19'
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
     - name: golangci-lint
@@ -69,7 +69,7 @@ jobs:
     - name: Set up Go 1.18
       uses: actions/setup-go@v2
       with:
-        go-version: 1.18.2
+        go-version: '>=1.18.2 <1.19'
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,6 @@
 # Tested with golangci-lint ver. 1.37
 run:
-  timeout: 3m
+  timeout: 10m
   skip-dirs:
     - vendor/
     - .github/


SchSeba:
Thanks!

/lgtm

.github/workflows/go.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     - name: Set up Go 1.18
       uses: actions/setup-go@v3
       with:
-        go-version: '>=1.18.2'
+        go-version: 1.18.2
```
e0ne, line 15: It's better to specify '<1.19' until we support go v1.19+

e0ne (changes requested):

.github/workflows/go.yml
zeeke: We also need `>= 1.18.2` for what was discussed on #310. 
Any idea on a clean way to specify both conditions?

SchSeba: I think we can go with <1.19 there is already 1.18.5 so there is no risk here

zeeke: ok

zeeke:
Retitling as we set go version <1.19

SchSeba:
Thanks!

/lgtm

SchSeba (approved):

adrianchiris:
Before i LGTM. are we sure the `>1.19` will run latest go `1.18` and will not by some weird circumstances run older version of go ?

e0ne (approved):
Thanks for addressing my comments. CI use go 1.18.5 now https://github.com/k8snetworkplumbingwg/sriov-network-operator/runs/7895516472?check_suite_focus=true#step:2:12

SchSeba:
I will always run latest version that is < 1.19

zeeke:
> Before i LGTM. are we sure the `>1.19` will run latest go `1.18` and will not by some weird circumstances run older version of go ?

Good point. I tried understanding how `actions/setup-go` actually works:

- https://github.com/actions/setup-go/blob/main/src/installer.ts#L58
  - https://github.com/actions/toolkit/blob/master/packages/tool-cache/src/tool-cache.ts#L473
    - https://github.com/actions/toolkit/blob/master/packages/tool-cache/src/tool-cache.ts#L671

And it seems to look in a local, repository-shared, cache. So it seems to be possible to get a go1.17 build in some rare case.

Options:
a. stick with 1.8.2 (Pro: We know for sure which version we use in CI, Cons: manual bump of go version)
b. use '>=1.8.2 && < 1.9.0'
c. keep '<1.9.0' and see how often an old go version occurs

(personally, I'd go for a)

SchSeba:
I will prefer to go with `b` it should not be any change between Z stream of go 1.18

There was a specific issue in 1.18.1 but in general we should not care what Z stream it use

adrianchiris:
i would go with `b` as well

zeeke:
Ok, going for `b`

SchSeba (approved):
LGTM

Thanks @zeeke and sorry for the ping-pong :)

adrianchiris (approved):

PR #353, opened 2022-08-22
Add support for BF2 in connectX mode

SchSeba:
Fix the issue introduced in https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/240/files#r808624002

Allow to use BF in connectX mode also for OCP platform.
This is needed until we support the systemd configuration

### DIFF
--- a/pkg/plugins/generic/generic_plugin.go
+++ b/pkg/plugins/generic/generic_plugin.go
@@ -88,12 +88,21 @@ func (p *GenericPlugin) Apply() error {
 			return nil
 		}
 	}
+
+	// Create a map with all the PFs we will need to configure
+	// we need to create it here before we access the host file system using the chroot function
+	// because the skipConfigVf needs the mstconfig package that exist only inside the sriov-config-daemon file system
+	pfsToSkip, err := utils.GetPfsToSkip(p.DesireState)
+	if err != nil {
+		return err
+	}
+
 	exit, err := utils.Chroot("/host")
 	if err != nil {
 		return err
 	}
 	defer exit()
-	if err := utils.SyncNodeState(p.DesireState); err != nil {
+	if err := utils.SyncNodeState(p.DesireState, pfsToSkip); err != nil {
 		return err
 	}
 	p.LastState = &sriovnetworkv1.SriovNetworkNodeState{}
--- a/pkg/plugins/mellanox/mellanox_plugin.go
+++ b/pkg/plugins/mellanox/mellanox_plugin.go
@@ -2,7 +2,6 @@ package mellanox
 
 import (
 	"fmt"
-	"regexp"
 	"strconv"
 	"strings"
 
@@ -215,24 +214,17 @@ func configFW() error {
 	return nil
 }
 
-func mstConfigReadData(pciAddress string) (string, error) {
-	glog.Infof("mellanox-plugin mstConfigReadData(): device %s", pciAddress)
-	args := []string{"-e", "-d", pciAddress, "q"}
-	out, err := utils.RunCommand("mstconfig", args...)
-	return out, err
-}
-
 func getMlnxNicFwData(pciAddress string) (current, next *mlnxNic, err error) {
 	glog.Infof("mellanox-plugin getMlnxNicFwData(): device %s", pciAddress)
 	err = nil
 	attrs := []string{TotalVfs, EnableSriov, LinkTypeP1, LinkTypeP2}
 
-	out, err := mstConfigReadData(pciAddress)
+	out, err := utils.MstConfigReadData(pciAddress)
 	if err != nil {
 		glog.Errorf("mellanox-plugin getMlnxNicFwData(): failed %v", err)
 		return
 	}
-	mstCurrentData, mstNextData := parseMstconfigOutput(out, attrs)
+	mstCurrentData, mstNextData := utils.ParseMstconfigOutput(out, attrs)
 	current, err = mlnxNicFromMap(mstCurrentData)
 	if err != nil {
 		glog.Errorf("mellanox-plugin getMlnxNicFwData(): %v", err)
@@ -263,25 +255,6 @@ func mlnxNicFromMap(mstData map[string]string) (*mlnxNic, error) {
 	}
 
 	return fwData, nil
-}
-
-func parseMstconfigOutput(mstOutput string, attributes []string) (fwCurrent, fwNext map[string]string) {
-	glog.Infof("mellanox-plugin parseMstconfigOutput(): Attributes %v", attributes)
-	fwCurrent = map[string]string{}
-	fwNext = map[string]string{}
-	formatRegex := regexp.MustCompile(`(?P<Attribute>\w+)\s+(?P<Default>\S+)\s+(?P<Current>\S+)\s+(?P<Next>\S+)`)
-	mstOutputLines := strings.Split(mstOutput, "\n")
-	for _, attr := range attributes {
-		for _, line := range mstOutputLines {
-			if strings.Contains(line, attr) {
-				regexResult := formatRegex.FindStringSubmatch(line)
-				fwCurrent[attr] = regexResult[3]
-				fwNext[attr] = regexResult[4]
-				break
-			}
-		}
-	}
-	return
 }
 
 func getPciAddressPrefix(pciAddress string) string {
--- a/pkg/utils/switchdev.go
+++ b/pkg/utils/switchdev.go
@@ -29,15 +29,23 @@ func IsSwitchdevModeSpec(spec sriovnetworkv1.SriovNetworkNodeStateSpec) bool {
 }
 
 func WriteSwitchdevConfFile(newState *sriovnetworkv1.SriovNetworkNodeState) (update bool, err error) {
+	// Create a map with all the PFs we will need to SKIP for systemd configuration
+	pfsToSkip, err := GetPfsToSkip(newState)
+	if err != nil {
+		return false, err
+	}
+
 	cfg := config{}
 	for _, iface := range newState.Spec.Interfaces {
 		for _, ifaceStatus := range newState.Status.Interfaces {
 			if iface.PciAddress != ifaceStatus.PciAddress {
 				continue
 			}
-			if !SkipConfigVf(iface, ifaceStatus) {
+
+			if skip := pfsToSkip[iface.PciAddress]; !skip {
 				continue
 			}
+
 			i := sriovnetworkv1.Interface{}
 			if iface.NumVfs > 0 {
 				i = sriovnetworkv1.Interface{
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -147,7 +147,7 @@ func DiscoverSriovDevices(withUnsupported bool) ([]sriovnetworkv1.InterfaceExt, 
 
 // SyncNodeState Attempt to update the node state to match the desired state
 //
-func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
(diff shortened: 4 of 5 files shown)


SchSeba:
/cc @adrianchiris @e0ne @zshi-redhat @bn222 @zshi-redhat  @zeeke 

please have a look

pkg/plugins/generic/generic_plugin.go
```diff
@@ -87,12 +87,31 @@ func (p *GenericPlugin) Apply() error {
 			return nil
 		}
 	}
+
+	// Create a map with all the PFs we will need to configure
+	// we need to create it here before we access the host file system using the chroot function
+	// because the SkipConfigVf needs the mstconfig package that exist only inside the sriov-config-daemon file system
+	vfsToConfig := map[string]bool{}
```
zeeke, line 94: please, rename this variable (and parameter) to `vfsToSkip` or something like that, as its entries seem to be true when the Virtual Functions must not be configured.

Another solution may be to make `utils.SyncNodeState(...)` receive `ifSpecs []sriovnetworkv1.Interface, ifStatuses []sriovnetworkv1.InterfaceExt` and filter out the to-skip interfcaes before calling it.

pkg/utils/utils_mlx.go
```diff
@@ -0,0 +1,102 @@
...
+	internalCPUIbVportoStatus, exist := mstCurrentData[internalCPUIbVporto]
+	if !exist {
+		return 0, fmt.Errorf("failed to find %s in the mstconfig output command", internalCPUIbVporto)
+	}
+	internalCPUOffloadEngineStatus, exist := mstCurrentData[internalCPUOffloadEngine]
+	if !exist {
+		return 0, fmt.Errorf("failed to find %s in the mstconfig output command", internalCPUOffloadEngine)
+	}
+
+	// check for DPU
```
wizhaoredhat, line 86: Just a comment: These registers aren't available with some versions of mstflint.

INTERNAL_CPU_MODEL should also be 1 as well. This is true for both DPU mode and NIC mode. Maybe we should add a check here just to be safe.

SchSeba: @wizhaoredhat if for both it will be 1 why do we need to check it?

wizhaoredhat: It is possible that the NIC was incorrectly set to Separated Host Mode which will have a value of zero.

pkg/utils/utils.go
```diff
@@ -185,17 +190,29 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
...
+		mode, err := mellanoxBlueFieldMode(ifStatus.PciAddress)
+		if err != nil {
+			return false, fmt.Errorf("failed to read Mellanox Bluefield card mode for %s,%v", ifStatus.PciAddress, err)
+		}
+
+		if mode == bluefield2Conntexd {
+			return false, nil
+		}
+
+		glog.V(2).Infof("SkipConfigVf(): skip config VF for BF2 device on DPU mode")
```
e0ne, line 211: Neet to confirm that such use-case is not needed

pkg/plugins/generic/generic_plugin.go
```diff
@@ -87,12 +87,31 @@ func (p *GenericPlugin) Apply() error {
 			return nil
 		}
 	}
+
+	// Create a map with all the PFs we will need to configure
+	// we need to create it here before we access the host file system using the chroot function
+	// because the SkipConfigVf needs the mstconfig package that exist only inside the sriov-config-daemon file system
+	vfsToSkip := map[string]bool{}
```
e0ne, line 94: nit: it would be good to move this code into the separate function

pkg/plugins/generic/generic_plugin.go
```diff
@@ -87,12 +87,31 @@ func (p *GenericPlugin) Apply() error {
 			return nil
 		}
 	}
+
+	// Create a map with all the PFs we will need to configure
+	// we need to create it here before we access the host file system using the chroot function
```
e0ne, line 92: It's clear to my why do we need this logic but it doesn't look good architecture to have this map and still use SkipConfigVf function in other places.

e0ne (commented):
I still need to test this PR.  With this change we introduce more platform-specific and NIC-specific code which is used across the operator instead of in s specific plugins

pkg/utils/utils_mlx.go
SchSeba: sounds good done

pkg/plugins/generic/generic_plugin.go
SchSeba: done

pkg/plugins/generic/generic_plugin.go
SchSeba: done we use the same function on both places

SchSeba:
> I still need to test this PR. With this change we introduce more platform-specific and NIC-specific code which is used across the operator instead of in s specific plugins

I agree but I was not able to find a different solution, but I am optimistic that the rework I am doing will make stuff simpler

pkg/utils/utils_mlx.go
```diff
@@ -0,0 +1,116 @@
...
+		return bluefield2Dpu, nil
+	} else if strings.Contains(internalCPUPageSupplierstatus, extHostPf) &&
+		strings.Contains(internalCPUEswitchManagerStatus, extHostPf) &&
+		strings.Contains(internalCPUIbVportoStatus, extHostPf) &&
+		strings.Contains(internalCPUOffloadEngineStatus, disabled) &&
+		strings.Contains(internalCPUModelStatus, embeddedCPU) {
+		glog.V(2).Infof("MellanoxBlueFieldMode():card %s in ConnectX mode", PciAddress)
+		return bluefield2Conntexd, nil
+	}
+
```
wizhaoredhat, line 114: Just a suggestion. I believe it would help with debugging if who have some logs here (in the case of failure) what the actual mstconfig is.

pkg/utils/utils_mlx.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/golang/glog"
+)
+
+//BlueField2 mode representation
```
adrianchiris, line 11: the mode and the parameters will be valid for future Bluefield NICs. so maybe drop the "2"

pkg/utils/utils_mlx.go
```diff
@@ -0,0 +1,116 @@
...
+type BlueField2Mode int
+
+const (
+	bluefield2Dpu BlueField2Mode = iota
+	bluefield2Conntexd
```
adrianchiris, line 16: nit "Connexd" -> "ConnectXMode" ?

pkg/utils/switchdev.go
```diff
@@ -29,15 +29,23 @@ func IsSwitchdevModeSpec(spec sriovnetworkv1.SriovNetworkNodeStateSpec) bool {
 }
 
 func WriteSwitchdevConfFile(newState *sriovnetworkv1.SriovNetworkNodeState) (update bool, err error) {
+	// Create a map with all the PFs we will need to SKIP for systemd configuration
+	vfsToSkip, err := GetVfsToSkip(newState)
```
adrianchiris, line 33: PFsToSkip/pfsToSkip ?

pkg/utils/utils.go
adrianchiris: what is the use case you refer ?

pkg/utils/utils.go
```diff
@@ -185,17 +190,48 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
...
+		return true, nil
 	}
-	return false
+
+	return false, nil
+}
+
+func GetVfsToSkip(ns *sriovnetworkv1.SriovNetworkNodeState) (map[string]bool, error) {
```
adrianchiris, line 218: nit: add docstring

adrianchiris (commented):
still need to give this one a deeper look.

one thing in mind is creating an interface to provide NIC information (e,g if its DPU or non DPU)
then keeping the vendor specific logic in plugin.

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -266,25 +258,6 @@ func mlnxNicFromMap(mstData map[string]string) (*mlnxNic, error) {
...
-				regexResult := formatRegex.FindStringSubmatch(line)
-				fwCurrent[attr] = regexResult[3]
-				fwNext[attr] = regexResult[4]
-				break
-			}
-		}
-	}
-	return
-}
-
```
bn222, line 287: This is a refactor, please do that in a separate commit

pkg/utils/utils_mlx.go
```diff
@@ -0,0 +1,116 @@
...
+		strings.Contains(internalCPUOffloadEngineStatus, enabled) &&
+		strings.Contains(internalCPUModelStatus, embeddedCPU) {
+		glog.V(2).Infof("MellanoxBlueFieldMode():card %s in DPU mode", PciAddress)
+		return bluefield2Dpu, nil
+	} else if strings.Contains(internalCPUPageSupplierstatus, extHostPf) &&
+		strings.Contains(internalCPUEswitchManagerStatus, extHostPf) &&
+		strings.Contains(internalCPUIbVportoStatus, extHostPf) &&
+		strings.Contains(internalCPUOffloadEngineStatus, disabled) &&
+		strings.Contains(internalCPUModelStatus, embeddedCPU) {
+		glog.V(2).Infof("MellanoxBlueFieldMode():card %s in ConnectX mode", PciAddress)
```
bn222, line 111: I prefer to have the log happen outside this function

SchSeba:
@adrianchiris about your comment 
> what is the use case you refer ?

what do you mean?

pkg/utils/utils_mlx.go
SchSeba: done

SchSeba:
@bn222 done

pkg/utils/utils_mlx.go
SchSeba: I will not like to add another if outside the function. why do you prefer it outside?

pkg/plugins/mellanox/mellanox_plugin.go
SchSeba: done I separate the commit

pkg/utils/utils.go
```diff
@@ -184,18 +189,51 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
...
+					glog.Errorf("GetPfsToSkip(): fail to check for skip VFs %s: %v.", iface.PciAddress, err)
+					return vfsToSkip, err
+				}
+				vfsToSkip[iface.PciAddress] = skip
+				break
+			}
+		}
+	}
+
+	return vfsToSkip, nil
```
bn222, line 236: Your function is called `GetPfsToSkip` and you're variable is called `vfsToSkip`

SchSeba: done thanks!

SchSeba:
@e0ne @bn222 can you make another round of review so we can merge this PR please :)

pkg/utils/utils_mlx.go
bn222: Just a preference. I expect that a function that checks the mode doesn't print to the log. The function that is asking to check for the mode could of course check. Separation of concerns imo.

e0ne (approved):
Verified with vanilla k8s 1.25

bn222:
lgtm

SchSeba:
Thanks for the review folks I am going to merge this PR

PR #354, opened 2022-08-29
improve drain check

SchSeba:
We need to check in the drain function also if VFs exist
If they exist we need to check if any other configuration is required (link driver and MTU)

If a change to the VF is requested we should drain the node because we can have pods using the VF

### DIFF
--- a/pkg/plugins/generic/generic_plugin.go
+++ b/pkg/plugins/generic/generic_plugin.go
@@ -158,13 +158,8 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
 				// TODO: no need to perform further checks if ifaceStatus.NumVfs equals to 0
 				// once https://github.com/kubernetes/kubernetes/issues/109595 will be fixed
 				configured = true
-				if iface.NumVfs != ifaceStatus.NumVfs {
-					glog.V(2).Infof("generic-plugin needDrainNode(): need drain, expect NumVfs %v, current NumVfs %v", iface.NumVfs, ifaceStatus.NumVfs)
-					needDrain = true
-					return
-				}
-				if iface.Mtu != 0 && iface.Mtu != ifaceStatus.Mtu {
-					glog.V(2).Infof("generic-plugin needDrainNode(): need drain, expect MTU %v, current MTU %v", iface.Mtu, ifaceStatus.Mtu)
+				if utils.NeedUpdate(&iface, &ifaceStatus) {
+					glog.V(2).Infof("generic-plugin needDrainNode(): need drain, PF %s request update", iface.PciAddress)
 					needDrain = true
 					return
 				}
--- a/pkg/plugins/generic/generic_plugin_test.go
+++ b/pkg/plugins/generic/generic_plugin_test.go
@@ -1,0 +1,131 @@
+package generic_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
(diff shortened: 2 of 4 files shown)


zeeke:
I think there is a regression in case the `desired` interface Mtu is 0.

I arranged a unit test with information taken from a test cluster here:
https://github.com/zeeke/sriov-network-operator-1/blob/pr354-test/pkg/plugins/generic/generic_plugin_test.go

It works without changes at [generic_plugin.go:166](https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/354/files#diff-5e18b1bb6723228bb51dcd9afe17e63609c26b293a04e6f0a44075ef0f5cfcc2L166).

Can you please add the test to the PR?

zeeke:
LGTM

pkg/plugins/generic/generic_plugin.go
```diff
@@ -163,8 +163,8 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
 					needDrain = true
 					return
 				}
-				if iface.Mtu != 0 && iface.Mtu != ifaceStatus.Mtu {
-					glog.V(2).Infof("generic-plugin needDrainNode(): need drain, expect MTU %v, current MTU %v", iface.Mtu, ifaceStatus.Mtu)
+				if utils.NeedUpdate(&iface, &ifaceStatus) {
```
adrianchiris, line 166: this method checks also num VFs so you can remove the condition at L161

pkg/plugins/generic/generic_plugin_test.go
```diff
@@ -0,0 +1,94 @@
+package generic
+
+import (
+	"testing"
+
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	"github.com/stretchr/testify/assert"
+)
+
```
adrianchiris, line 9: Any chance to use ginkgo and gomega as we do in other places  ?

also you could define a test package (i.e `generic_test`) and just test the interface calls WDYT ?

pkg/plugins/generic/generic_plugin.go
SchSeba: done

pkg/plugins/generic/generic_plugin_test.go
SchSeba: done

SchSeba:
Hi @adrianchiris please take a look when you have time

e0ne (approved):
/LGTM

pkg/plugins/generic/generic_plugin_test.go
```diff
@@ -0,0 +1,177 @@
...
+			needDrain, needReboot, err := genericPlugin.OnNodeStateChange(networkNodeState)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(needReboot).To(BeFalse())
+			Expect(needDrain).To(BeTrue())
+		})
+	})
+
+})
+
+//func TestNeedDrainNode_InterfaceWithBadMtuShouldDrainTheNode(t *testing.T) {
```
adrianchiris, line 133: nit : delete ?

adrianchiris (approved):
Minor nit. and im LGTM

pkg/plugins/generic/generic_plugin_test.go
SchSeba: miss that :P

Thanks!

adrianchiris (approved):

PR #363, opened 2022-09-21
support graceful node shutdown

pacevedom:
- Remove MachineConfigPool operations.
- Add signal handler for SIGTERM.
- Add an updating flag to signal when the plugins need to drain.
- Add terminationGracePeriodSeconds default value to config daemon.

In order to support [graceful node shutdown](https://kubernetes.io/blog/2021/04/21/graceful-node-shutdown-beta/) we need to be able to endure SIGTERM signals sent to the config daemon.

Whenever the config daemon is applying plugins it required no other operator rebooting/disrupting the node. The way to achieve this is to pause the MCP in order to get exclusive access to node changes, such as reboot commands. When finished, if required, it would turn off kubelet and then reboot the node.

Starting with MCO 4.12 we have native support for graceful node shutdown, which means the shutdown/reboot will be delayed by this feature based on different timeouts and priority class names (done by means of SIGTERM signal handlers). This means there is no need to pause the MCP anymore, as MCO will issue a reboot and wait for SIGTERM/SIGKILL termination if someone else is doing critical work. The way to signal this critical work is to ignore SIGTERM if a non-interruptible operation is taking place.
Kubelet will take care of evicting pods and killing them if needed.

/cc @SchSeba


pkg/daemon/daemon.go
```diff
@@ -924,10 +800,52 @@ func (dn *Daemon) drainNode() error {
...
+	go func() {
+		for sig := range termChan {
+			//nolint:gocritic
+			switch sig {
+			case syscall.SIGTERM:
+				updateActive := dn.getUpdatingFlag()
+				if updateActive {
+					glog.Info("Got SIGTERM, but actively updating")
+				} else {
+					close(signalCh)
```
zeeke, line 841: what about using `stopCh`? is signalCh really needed?

bn222:
Can you add a much more elaborate commit message body to explain why we need this and how this improves stability?

pkg/daemon/daemon.go
pacevedom: `stopCh` is an external source to stop the daemon, and `signalCh` is an internal way of stopping it. In other words, if we receive SIGTERM and we are not doing any drain/apply operation on the plugins, we need to stop the daemon.

pacevedom:
> Can you add a much more elaborate commit message body to explain why we need this and how this improves stability?

Added more info to the commit message.

bn222:
Please split the single commit in this PR into logical commits

pkg/daemon/daemon.go
```diff
@@ -54,6 +53,10 @@ const (
 	maxUpdateBackoff = 60 * time.Second
 )
 
+var (
+	defaultRebootTimeout = 24 * time.Hour
```
bn222, line 57: 24h seems long to me. can we make it shorter? 6h maybe?

bn222:
Also, the commit message still doesn't contain anything new when I checked after pulling your PR onto my machine.

pkg/daemon/daemon.go
pacevedom: Took this from MCO. I don't see a shutdown event being held for that much time, so let's bring it down to 6h.

pkg/daemon/daemon.go
```diff
@@ -924,10 +803,55 @@ func (dn *Daemon) drainNode() error {
...
+	// Catch SIGTERM - if we're actively updating, we should avoid
+	// having the process be killed.
+	go func() {
+		for sig := range termChan {
+			//nolint:gocritic
+			switch sig {
+			case syscall.SIGTERM:
+				updateActive := dn.getUpdatingFlag()
+				if updateActive {
+					glog.Info("Got SIGTERM, but actively updating")
```
bn222, line 844: So what happens with the SIGTERM in this case? We are already updating in the sriov daemon so we just ignore the request to shut down since we will end our updating activity with a reboot? What if we require a drain but not a reboot? If reqDrain, then we call drainNode and we enable the flag to ignore the signal. Is it the case that in such scenario, we were asked to shutdown, but we ignored that completely?

pacevedom: The way this works is:
- Someone issues a reboot/shutdown.
- Kubelet gets the event and creates two lists of pods: critical and non critical, based on priorityClass.
- Starting with non-critical pods, it checks the minimum between the termination period for the pod and the configured timeout in kubelet.
- Send SIGTERM using eviction API to pod. Check for pod termination for the computed value in the previous bullet. If the pod is still alive, send SIGKILL.
- Repeat operation with critical pods.
- Shutdown.

If a drain must happen then SIGTERM is ignored while doing so. The timeout for termination must be properly configured to allow enough time to completion. After that timeout a SIGKILL will end the process.

If a reboot is required then its the same situation but in other pods. The daemon will issue a shutdown that will kickstart this process in the node. In the end everyone will receive a SIGTERM followed by a SIGKILL if the pod is still alive after the termination grace period.

If there is no need to drain/reboot then SIGTERM will end the process.

Once SIGTERM has been received it marks that the pod is bound to terminate some time in the near future.

bn222: Thank you for clarifying. Please add this explanation to the commit message for future reference too.

pacevedom: Added in the first commit.

bn222:
lgtm

pkg/daemon/daemon.go
zeeke: `stopCh` (as well as `exitCh`) is not written nor closed by anyone in production code. For sure we can even clean that in another PR, but at least use a different name for `installSignalHandler(...)` parameter.

zeeke:
any chance to get this change somehow tested?

pkg/daemon/daemon.go
pacevedom: Renamed to `signalStopCh`

pacevedom:
> any chance to get this change somehow tested? I'm afraid it becomes a little bit hard to add further code changes to `daemon.go` without breaking.

Shared details about testing through Slack.

zeeke:
lgtm

bn222:
Please also add a summary of your conversations from slack to this issue as a reply

pacevedom:
> Please also add a summary of your conversations from slack to this issue as a reply

Did manual testing with two scenarios: drain and reboot. For both of them I configured kubelet to enable graceful node shutdown and, using MCO (needs to be 4.12+ version) or a specially crafted pod (able to ignore SIGTERM and issue reboot commands) I checked the changes in SRIOV operator work without any issues.
No automation yet.

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -93,6 +93,7 @@ spec:
           preStop:
             exec:
               command: ["/bindata/scripts/clean-k8s-services.sh"]
+      terminationGracePeriodSeconds: 900
```
e0ne, line 96: Do we want to make it configurable?

e0ne (changes requested):

bindata/manifests/daemon/daemonset.yaml
pacevedom: It needs to be at least as high as the longest expected plugin apply + some buffer for safety. Is there a good value for this?

zeeke:
> /test-all

@e0ne did the tests showed any regression on these changes?

pacevedom:
Any updates @e0ne?

e0ne:
> Any updates @e0ne?

All tests passed

e0ne (approved):

Issue #226, opened 2022-01-14
Parallelize the SRIOV configuration on multiple nodes 


pliurh:
Currently, when we want to configure the SRIOV on multiple nodes, it is executed in sequence, other nodes wait while one node is being configured. And due to the node draining and reboot, it will take quite a long time, if we have a large number of nodes that need to be configured.

To resolve this issue, we may need to figure out how we can parallelize the configuration on multiple nodes. One idea is to introduce a maximum unavailable number of nodes concept. So that we can conduct the SRIOV configuration simultaneously on such an unavailable number of nodes. We may want to learn the implementation from the Machine Config Operator.

SchSeba:
should be fixed via https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/555

SchSeba:
the PR is merged closing this one

Issue #225, opened 2022-01-14
Optimize draining logic


pliurh:
Currently, the SRIOV network config daemon will drain a node, when 
1. a node reboot is required;
2. the `numVF` is changed from non-zero value to another;
3. the 'mtu' is changed;
4. the configuration of the device plugin is changed. 
 
Among all the conditions, draining is only mandatory for the node reboot case. For the other cases, it would be sufficient to just evict the pods with VF attached. With this enhancement, we can shorten the overall reconciliation time when applying new node policies.

SchSeba:
should be fixed via https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/555

PR #365, opened 2022-09-29
Implementation for new systemd configuration method

SchSeba:
Add a new configuration method called systemd mode.

With this new mode all the VFs are created before kubelet even starts this will allow the operator to have a much better recovery time after reboot.

it will also fix issues like https://github.com/kubernetes/kubernetes/issues/109595

The downside of this solution and the reason why for now it will not be the "default" configuration mode is that every configuration change in sriov will require a node reboot.

After this PR is merged we will need to work on 
https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/225
https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/226

to improve the configuration/reconfiguration/update time on large clusters


jjuele:

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -376,6 +414,25 @@ func (p *K8sPlugin) servicesStateUpdate() error {
...
+		}
+	}
+
+	if p.updateTarget.sriovScript {
+		err := ioutil.WriteFile(path.Join(chroot, p.sriovService.Path),
+			[]byte(p.switchdevBeforeNMRunScript.Contents.Inline), 0755)
+		if err != nil {
+			return err
+		}
+	}
```
bn222, line 431: This function is not really clean. Why check twice? The first one still calls systemd directly

controllers/sriovoperatorconfig_controller.go
```diff
@@ -123,7 +123,7 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 	}
```
bn222, line 123: typo in commit name

api/v1/sriovoperatorconfig_types.go
```diff
@@ -39,6 +39,9 @@ type SriovOperatorConfigSpec struct {
 	DisableDrain bool `json:"disableDrain,omitempty"`
 	// Flag to enable OVS hardware offload. Set to 'true' to provision switchdev-configuration.service and enable OpenvSwitch hw-offload on nodes.
 	EnableOvsOffload bool `json:"enableOvsOffload,omitempty"`
+	// Flag to enable the sriov-network-config-daemon to use a systemd mode instead of the regular method
+	// +kubebuilder:validation:Enum=["daemon","systemd"]
```
lmilleri, line 43: Curly brackets?

pkg/host/host.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	return true, nil
+}
+
+func (h *HostManager) isCoreOS() (bool, error) {
+	glog.Infof("isCoreOS(): checking for CoreOS machine")
+	path := redhatReleaseFile
+	if !h.RunOnHost {
+		path = hostPathFromDaemon + path
```
bn222, line 254: https://pkg.go.dev/path ?

pkg/host/host.go
```diff
@@ -0,0 +1,406 @@
...
+	if err != nil && stderr.Len() != 0 {
+		glog.Errorf("installRDMA(): failed to reload mlx4_en kernel module: error %v stderr %s", err, stderr.String())
+		return err
+	}
+
+	_, stderr, err = h.cmd.Run("/bin/sh", "-c", fmt.Sprintf("%s modprobe -r mlx5_core && %s modprobe mlx5_core", chrootDefinition, chrootDefinition))
+	if err != nil && stderr.Len() != 0 {
+		glog.Errorf("installRDMA(): failed to reload mlx5_core kernel module: error %v stderr %s", err, stderr.String())
+		return err
+	}
```
bn222, line 376: Copy paste code, abstract it away

controllers/sriovoperatorconfig_controller.go
```diff
@@ -175,6 +183,12 @@ func (r *SriovOperatorConfigReconciler) syncConfigDaemonSet(dc *sriovnetworkv1.S
 	data.Data["ClusterType"] = utils.ClusterType
 	data.Data["DevMode"] = os.Getenv("DEV_MODE")
 	data.Data["ImagePullSecrets"] = GetImagePullSecrets()
+	if dc.Spec.ConfigurationMode == sriovnetworkv1.SystemdConfigurationMode {
+		data.Data["UsedSystemdMode"] = true
```
bn222, line 187: `data.Data["UsedSystemdMode"] = (dc.Spec.ConfigurationMode == sriovnetworkv1.SystemdConfigurationMode)` ?

cmd/sriov-network-config-daemon/service.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/version"
+)
+
+var (
+	serviceCmd = &cobra.Command{
+		Use:   "service",
+		Short: "Starts SR-IOV service Config",
+		Long:  "",
+		Run:   runServiceCmd,
```
zeeke, line 26: Maybe here it's better to use `cobra.Command.RunE` 

```
$ export DOCKERFILE=Dockerfile.sriov-network-config-daemon                                                                                                                                                                                             
$ export APP_NAME=sriov-network-config-daemon
$ make image
...
$ docker run -it ghcr.io/k8snetworkplumbingwg/sriov-network-config-daemon /usr/bin/sriov-network-config-daemon service 
I1206 11:00:24.618527       1 service.go:49] Starting sriov-config-service
E1206 11:00:24.618655       1 service.go:52] failed to read list of supported nic ids
E1206 11:00:24.618677       1 systemd.go:136] WriteSriovResult(): failed to create sriov result file on path /etc/sriov-operator/sriov-interface-result.yaml: open /etc/sriov-operator/sriov-interface-result.yaml: no such file or directory
E1206 11:00:24.618685       1 service.go:59] sriov-config-service failed to write sriov result file with content {Failed failed to read list of supported nic ids: stat /etc/sriov-operator/sriov-supported-nics-ids.yaml: no such file or directory} error: open /etc/sriov-operator/sriov-interface-result.yaml: no such file or directory
$ echo $?
0
```

Otherwise, systemd unit will be marked as completed with success, even in case of error.

cmd/sriov-network-config-daemon/service.go
```diff
@@ -0,0 +1,166 @@
...
+			return
+		}
+	}
+
+	nodeState := &sriovv1.SriovNetworkNodeState{
+		Spec:   nodeStateSpec.Spec,
+		Status: sriovv1.SriovNetworkNodeStateStatus{Interfaces: ifaceStatuses},
+	}
+
+	_, _, err = configPlugin.OnNodeStateChange(nodeState)
```
zeeke, line 141: This is going to panic in case of `nodeStateSpec.PlatformType == utils.VirtualOpenStack`. 
Is it missing something like
```
configPlugin = virtualplugin.NewVirtualPlugin(true)
```
?

pkg/plugins/intel/intel_plugin.go
```diff
@@ -44,3 +44,10 @@ func (p *IntelPlugin) Apply() error {
 	glog.Info("intel-plugin Apply()")
 	return nil
 }
+
+func (p *IntelPlugin) SetSystemdFlag() {
```
zeeke, line 48: `SetSystemdFlag`/`IsSystemService` pair seems to be unused. Isn't it?

cmd/sriov-network-config-daemon/service.go
SchSeba: right sorry I didn't test this one yet on OpenStack :(

pkg/plugins/intel/intel_plugin.go
SchSeba: done

cmd/sriov-network-config-daemon/service.go
SchSeba: right thanks!

hack/build-go.sh
```diff
@@ -32,7 +32,7 @@ fi
 
 mkdir -p ${BIN_PATH}
 
-CGO_ENABLED=${CGO_ENABLED:-1}
+CGO_ENABLED=${CGO_ENABLED:-0}
```
bn222, line 35: Why do we want static linkage?

controllers/sriovoperatorconfig_controller.go
```diff
@@ -120,6 +122,12 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 		return reconcile.Result{}, err
 	}
 
+	// For Openshift we need to create the systemd files using a machine config
+	if utils.ClusterType == utils.ClusterTypeOpenshift && defaultConfig.Spec.ConfigurationMode == sriovnetworkv1.SystemdConfigurationMode {
```
bn222: Need to handle hypershift too

controllers/sriovoperatorconfig_controller.go
```diff
@@ -309,3 +323,44 @@ func (r *SriovOperatorConfigReconciler) syncK8sResource(cr *sriovnetworkv1.Sriov
...
+			if err != nil {
+				logger.Error(err, "Fail to convert to MachineConfig")
+				return err
+			}
+			mc.Labels = cr.Spec.ConfigDaemonNodeSelector
+			err = scheme.Convert(mc, obj, nil)
+			if err != nil {
+				logger.Error(err, "Fail to convert to Unstructured")
+				return err
+			}
```
bn222, line 378: Seems like this could be a function with a name that tells us what it actually does without having to understand the code. something like `setLabelInsideObject`

hack/build-go.sh
SchSeba: because now the sriov-config-daemon binary runs on the host and I don't want any gcc lib requirements

controllers/sriovoperatorconfig_controller.go
SchSeba: this is how we check hypershift (dn.openshiftContext.IsHypershift()) like here
```
 utils.ClusterType == utils.ClusterTypeOpenshift && !dn.openshiftContext.IsHypershift()
```

bn222: I mean that in the body of the conditional that checks for openshift, we need to handle hypershift differently than normal openshift.

wizhaoredhat:
@SchSeba Do you have a readme planned explaining how to enable systemd?

e0ne (changes requested):
Please, add license headers to new files

hack/build-go.sh
e0ne: do we need it only for config daemon or for operator too?

api/v1/sriovoperatorconfig_types.go
```diff
@@ -39,6 +39,9 @@ type SriovOperatorConfigSpec struct {
 	DisableDrain bool `json:"disableDrain,omitempty"`
 	// Flag to enable OVS hardware offload. Set to 'true' to provision switchdev-configuration.service and enable OpenvSwitch hw-offload on nodes.
 	EnableOvsOffload bool `json:"enableOvsOffload,omitempty"`
+	// Flag to enable the sriov-network-config-daemon to use a systemd mode instead of the regular method
```
e0ne, line 42: Please, add a note which method will be used by default

controllers/sriovoperatorconfig_controller.go
```diff
@@ -309,3 +324,65 @@ func (r *SriovOperatorConfigReconciler) syncK8sResource(ctx context.Context, cr
 	}
 	return nil
 }
+
+// syncSystemdService creates the Machine Config to deploy the systemd service on openshift ONLY
+func (r *SriovOperatorConfigReconciler) syncSystemdService(ctx context.Context, cr *sriovnetworkv1.SriovOperatorConfig) error {
```
e0ne, line 329: IMO, this method name should be OpenShift  specific, e.g.: syncOpenShiftSystemdService

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -37,38 +37,66 @@ spec:
...
+          volumeMounts:
+            - name: cnibin
+              mountPath: /host/opt/cni/bin
+      {{- if .UsedSystemdMode}}
+        - name: sriov-service-copy
+          image: {{.Image}}
+          command:
+            - /bin/bash
+            - -c
+            - mkdir -p /host/var/lib/sriov/ && cp /usr/bin/sriov-network-config-daemon /host/var/lib/sriov/sriov-network-config-daemon && chcon -t bin_t /host/var/lib/sriov/sriov-network-config-daemon | true #Allow systemd to run the file
```
e0ne, line 77: Please, add a comment why `| true` is needed here

pkg/host/host.go
```diff
@@ -0,0 +1,423 @@
+/*
+Copyright 2021.
```
e0ne, line 2: 2023

pkg/host/host.go
```diff
@@ -0,0 +1,423 @@
...
+		cmd:       &utils.Command{},
+	}
+}
+
+func (h *HostManager) LoadKernelModule(name string, args ...string) error {
+	glog.Infof("LoadKernelModule(): try to load kernel module %s with arguments '%s'", name, args)
+	chrootDefinition := getChrootExtention(h.RunOnHost)
+	cmdArgs := strings.Join(args, " ")
+
+	// check if the driver is already loaded in to the system
```
e0ne, line 81: please, move this to a separate function `isKernelModuleLoaded`

pkg/host/host.go
```diff
@@ -0,0 +1,423 @@
...
+}
+
+func NewHostManager(runOnHost bool) HostManagerInterface {
+	return &HostManager{
+		RunOnHost: runOnHost,
+		cmd:       &utils.Command{},
+	}
+}
+
+func (h *HostManager) LoadKernelModule(name string, args ...string) error {
```
e0ne, line 76: maybe it's better to use https://github.com/pmorjan/kmod but I'm not sure on it

pkg/systemd/systemd.go
```diff
@@ -0,0 +1,288 @@
+/*
+Copyright 2021.
```
e0ne, line 2: 2023?

cmd/sriov-network-config-daemon/main.go
```diff
@@ -1,3 +1,18 @@
+/*
+Copyright 2021.
```
e0ne, line 2: 2023?

cmd/sriov-network-config-daemon/service.go
```diff
@@ -0,0 +1,182 @@
+/*
+Copyright 2021.
```
e0ne, line 2: 2023

e0ne (changes requested):

hack/build-go.sh
SchSeba: it's possible to add this only for the config-daemon but what is the cons of having this to all the binaries?

controllers/sriovoperatorconfig_controller.go
SchSeba: done

pkg/host/host.go
SchSeba: that lib doesn't support chroot. but maybe in the future we can switch this implementation with something like that (add a support for that lib to use chroot methods)

pkg/host/host.go
SchSeba: done

pkg/host/host.go
SchSeba: done

pkg/systemd/systemd.go
SchSeba: done

pkg/daemon/plugin.go
```diff
@@ -46,14 +47,14 @@ func enablePlugins(platform utils.PlatformType, ns *sriovnetworkv1.SriovNetworkN
...
-			k8sPlugin, err := K8sPlugin()
+			k8sPlugin, err := K8sPlugin(useSystemdService)
 			if err != nil {
 				glog.Errorf("enableVendorPlugins(): failed to load the k8s plugin error: %v", err)
 				return nil, err
 			}
 			enabledPlugins[k8sPlugin.Name()] = k8sPlugin
 		}
-		genericPlugin, err := GenericPlugin()
+		genericPlugin, err := GenericPlugin(false)
```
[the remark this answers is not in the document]
SchSeba: what you mean here a different way?

pkg/plugins/k8s/k8s_plugin.go
SchSeba: you are right

hack/build-go.sh
bn222: not a fan of static linking even though it's easier.

pkg/host/host.go
SchSeba: done

bindata/manifests/sriov-config-service/openshift/sriov-config-service.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: worker
```
wizhaoredhat, line 5: I am a bit concerned about the impact of a cluster with a mix of workers that have sriov and don't have sriov. This machine config applies to all workers and uses the default worker MCP pool that comes with openshift. This might trigger the cluster to go down for a very long period causing all workers to reboot. I see this as per design with machine config...

wizhaoredhat: Would it be better to use labeling of nodes that want sriov-network-operator in systemd? But then we cannot use the default worker MCP pool which does complicate things.

SchSeba: there is a way to select different pool

https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/365/files#diff-4be06e26c778a4f59b86cf746f89bbe18b78828a97ab43ac0c39c97b1d6532d6R365

wizhaoredhat: Got it thanks!

controllers/sriovoperatorconfig_controller.go
SchSeba: what will be the difference in hypershift?
there is no machine config?

controllers/sriovoperatorconfig_controller.go
SchSeba: I will like to leave this one like it's I think is more readable

cmd/sriov-network-config-daemon/start.go
```diff
@@ -90,7 +90,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	var config *rest.Config
 	var err error
 
-	if os.Getenv("CLUSTER_TYPE") == utils.ClusterTypeOpenshift {
+	if utils.ClusterType == utils.ClusterTypeOpenshift {
```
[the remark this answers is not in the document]
SchSeba: done

pkg/host/host.go
SchSeba: done

main.go
```diff
@@ -82,6 +84,28 @@ func main() {
...
+	if utils.ClusterType == utils.ClusterTypeOpenshift {
+		mcclient := mcclientset.NewForConfigOrDie(restConfig)
+		openshiftFlavor := utils.OpenshiftFlavorDefault
+		infraClient, err := client.New(restConfig, client.Options{
+			Scheme: scheme,
+		})
+		if err != nil {
+			panic(err)
+		}
+		isHypershift, err := utils.IsExternalControlPlaneCluster(infraClient)
```
bn222, line 98: What's wrong with  `r.OpenshiftContext.IsHypershift()`

SchSeba: that one only return the value it doesn't check it :p
we have it also in https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/cmd/sriov-network-config-daemon/start.go#L136

SchSeba: sure I don't follow that one

SchSeba:
waiting for https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/409

api/v1/helper.go
```diff
@@ -57,6 +57,13 @@ const (
 	SupportedNicIDConfigmap = "supported-nic-ids"
 )
 
+type ConfigurationModeType string
+
+const (
+	DaemonConfigurationMode  = "daemon"
```
e0ne, line 63: Maybe it's better to create types consts here:

```
DaemonConfigurationMode ConfigurationModeType  = "daemon"
SystemdConfigurationMode = "systemd" ConfigurationModeType
```

controllers/sriovoperatorconfig_controller.go
```diff
@@ -299,7 +319,7 @@ func (r *SriovOperatorConfigReconciler) deleteK8sResource(ctx context.Context, i
 
 func (r *SriovOperatorConfigReconciler) syncK8sResource(ctx context.Context, cr *sriovnetworkv1.SriovOperatorConfig, in *uns.Unstructured) error {
 	switch in.GetKind() {
-	case "ClusterRole", "ClusterRoleBinding", "MutatingWebhookConfiguration", "ValidatingWebhookConfiguration":
+	case "ClusterRole", "ClusterRoleBinding", "MutatingWebhookConfiguration", "ValidatingWebhookConfiguration", machineConfigCRDName:
```
e0ne, line 322: nit: let's move all CRDs name into constant

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -37,38 +37,66 @@ spec:
...
+          volumeMounts:
+            - name: cnibin
+              mountPath: /host/opt/cni/bin
+      {{- if .UsedSystemdMode}}
+        - name: sriov-service-copy
+          image: {{.Image}}
+          command:
+            - /bin/bash
+            - -c
+            - mkdir -p /host/var/lib/sriov/ && cp /usr/bin/sriov-network-config-daemon /host/var/lib/sriov/sriov-network-config-daemon && chcon -t bin_t /host/var/lib/sriov/sriov-network-config-daemon | true # Allow systemd to run the file, use pipe true to not failed if the system doesn't have selinux enabled
```
e0ne, line 77: selinux => selinux or apparmor

pkg/host/host.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	_, stderr, mlx5Err := h.cmd.Run("/bin/sh", "-c", fmt.Sprintf("grep --quiet 'mlx5_core' <(%s lsmod)", chrootDefinition))
+	if mlx5Err != nil && stderr.Len() != 0 {
+		glog.Errorf("tryEnableRdma(): failed to check for kernel module 'mlx5_core': error: %v stderr %s", mlx5Err, stderr.String())
+		return false, fmt.Errorf(stderr.String())
+	}
+
+	if mlx4Err != nil && mlx5Err != nil {
+		glog.Errorf("tryEnableRdma(): no RDMA capable devices")
```
e0ne, line 155: What about Intel devices? will this check work for them?

pkg/host/host.go
```diff
@@ -0,0 +1,455 @@
...
+	if err != nil {
+		glog.Errorf("tryEnableRdma(): failed to check if the machine is base on Ubuntu: %v", err)
+		return false, err
+	}
+
+	if isUbuntuSystem {
+		return h.enableRDMAOnUbuntuMachine()
+	}
+
+	osName, err := h.getOSPrettyName()
```
e0ne, line 180: line 167 returns result if isRhelSystem
link 176 returns result if isUbuntuSystem

what other OSs are supported to get into this code?

pkg/utils/command.go
```diff
@@ -0,0 +1,41 @@
+/*
+Copyright 2021.
```
e0ne, line 2: Copyright 2023

pkg/service/utils.go
```diff
@@ -5,6 +5,7 @@ import (
 	"strings"
```
e0ne, line 5: nit: it would be good to have more detailed commit message

bindata/manifests/sriov-config-service/kubernetes/sriov-config-service.yaml
```diff
@@ -0,0 +1,15 @@
...
+  
+  [Service]
+  Type=oneshot
+  ExecStart=/var/lib/sriov/sriov-network-config-daemon service -v {{ .LogLevel }}
+  StandardOutput=journal+console
+  
+  [Install]
+  WantedBy=multi-user.target
+enabled: true
+name: sriov-config.service
```
tariq1890, line 15: Just thinking out loud, would we want to add templatisation for dependent services? Is it possible that some users may want to add services in the `After=` field based on their host OS setup?

SchSeba: That is a great Idea the question is how we will expose that template to the user?

main.go
```diff
@@ -69,6 +69,7 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
+
```
e0ne, line 72: nit: this change is not needed

bindata/manifests/sriov-config-service/kubernetes/sriov-config-service.yaml
e0ne: we definately need to have this configurable.  Let's think about thi in a follow-up PR

pkg/host/host.go
e0ne: @SchSeba ^^

e0ne (approved):
Added few comments and nits which could be addressed in a new revision of this PR or in a follow up PR.

I tested these changes on the vanilla k8s only

PR #367, opened 2022-10-03
Update taints for network-resources-injector and operator-webhook

korroot:
Update node selector for sriov-device-plugin and sriov-network-config-daemon

### DIFF
--- a/bindata/manifests/daemon/daemonset.yaml
+++ b/bindata/manifests/daemon/daemonset.yaml
@@ -26,7 +26,7 @@ spec:
       hostPID: true
       nodeSelector:
         beta.kubernetes.io/os: linux
-        node-role.kubernetes.io/worker:
+        node-role.kubernetes.io/worker: "true"
       tolerations:
       - operator: Exists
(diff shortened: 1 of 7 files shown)


jjuele:

SchSeba:
Hi @korroot what k8s version are you using?

korroot:
@SchSeba SRIOV operator was not able to spin up properly on Rancher based deployment running v1.24.4 k8s version. The issue we experience is that Rancher adds additional taints on master nodes:

- node-role.kubernetes.io/control-plane: NoSchedule
- node-role.kubernetes.io/etcd: NoExecute

So that network-resources-injector and operator-webhook are not started at all.

SchSeba:
Hi @korroot maybe a better idea here is to add the toleration to these pods?

SchSeba:
Hi @korroot do you still need this PR or we can close it?

SchSeba:
in general we can add the needed tolerations to the yaml and we can extend the node selector if needed but that is not the case with the current implementation if this PR

SchSeba:
closing

PR #368, opened 2022-10-06
End2End - Pod Security Admission fixes

zeeke:
These changes aim to solve [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) problem when running tests against k8s v1.25+ , as privileged pods are often needed during tests.

### DIFF
--- a/test/util/clean/clean.go
+++ b/test/util/clean/clean.go
@@ -26,9 +26,7 @@ func All() error {
 			return fmt.Errorf("failed to restore node drain state %v", err)
 		}
 	}
-	if !namespaces.Exists(namespaces.Test, clients) {
-		return nil
-	}
+
 	err := namespaces.DeleteAndWait(clients, namespaces.Test, 5*time.Minute)
 	if err != nil {
 		return fmt.Errorf("failed to delete sriov tests namespace %v", err)
--- a/test/util/namespaces/namespaces.go
+++ b/test/util/namespaces/namespaces.go
@@ -20,6 +20,13 @@ import (
 // Test is the namespace to be use for testing
 const Test = "sriov-conformance-testing"
 
+var inhibitSecurityAdmissionLabels = map[string]string{
+	"pod-security.kubernetes.io/audit":               "privileged",
+	"pod-security.kubernetes.io/enforce":             "privileged",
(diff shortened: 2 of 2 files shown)


test/util/clean/clean.go
```diff
@@ -26,9 +26,7 @@ func All() error {
 			return fmt.Errorf("failed to restore node drain state %v", err)
 		}
 	}
-	if !namespaces.Exists(namespaces.Test, clients) {
-		return nil
-	}
+
 	err := namespaces.DeleteAndWait(clients, namespaces.Test, 5*time.Minute)
```
e0ne, line 30: I'm a bit confused by this change and the commit message: will it remove sriov resources from all namespaces or just from the specified one?

SchSeba: this function delete the namespace and all the objects in that namespace. it's for a test namespace we don't run this function on the operator namespace.

There is also a clean sriov function to remove all the sriov related objects like sriovNetwork and sriovNetworkNodePolicy

test/util/namespaces/namespaces.go
```diff
@@ -20,6 +20,13 @@ import (
 // Test is the namespace to be use for testing
 const Test = "sriov-conformance-testing"
 
+var inhibitSecurityAdmissionLabels = map[string]string{
+	"pod-security.kubernetes.io/audit":               "privileged",
```
wizhaoredhat, line 24: We might want to align this with what we have for https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/371

wizhaoredhat: From what I see the labels have a delta.

SchSeba: @wizhaoredhat this is fine for the test-pods we have this in other places where we run tests like https://github.com/openshift-kni/cnf-features-deploy/pull/1238

SchSeba (approved):

test/util/namespaces/namespaces.go
wizhaoredhat: Ah ok, got it. Thanks!

e0ne (approved):

PR #371, opened 2022-10-24
Add support for operator deployment on PSA

SchSeba:
pod-security-admission is a new feature in k8s 1.25

https://kubernetes.io/docs/concepts/security/pod-security-admission/

This change allow to deploy the operator a k8s cluster with PSA enabled. This commit allow add the openshift label to allow the auto labeling

### DIFF
--- a/deploy/namespace.yaml
+++ b/deploy/namespace.yaml
@@ -4,3 +4,6 @@ metadata:
   name: $NAMESPACE
   labels:
     name: $NAMESPACE
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/enforce-version: v1.24
+    security.openshift.io/scc.podSecurityLabelSync: "false"
--- a/deployment/sriov-network-operator/README.md
+++ b/deployment/sriov-network-operator/README.md
@@ -42,6 +42,9 @@ For additional information and methods for installing Helm, refer to the officia
 # Install Operator
 $ helm install -n sriov-network-operator --create-namespace --wait sriov-network-operator ./
 
+# Add PSA label for the sriov network operator namespace
+$ kubectl label ns sriov-network-operator pod-security.kubernetes.io/enforce=privileged
+
 # View deployed resources
 $ kubectl -n sriov-network-operator get pods
 ```


bn222:
/lgtm

only useful when deploying manually with make deploy-setup

deployment/sriov-network-operator/README.md
```diff
@@ -42,6 +42,9 @@ For additional information and methods for installing Helm, refer to the officia
 # Install Operator
 $ helm install -n sriov-network-operator --create-namespace --wait sriov-network-operator ./
 
+# Add PSA label for the sriov network operator namespace
+$ kubectl label ns sriov-network-operator pod-security.kubernetes.io/enforce=privileged
```
wizhaoredhat, line 46: Do we still need to do this step? The deploy/namespace.yaml should take care of it right?

deploy/namespace.yaml
```diff
@@ -4,3 +4,6 @@ metadata:
   name: $NAMESPACE
   labels:
     name: $NAMESPACE
+    pod-security.kubernetes.io/enforce: privileged
```
wizhaoredhat, line 7: Doesn't seem to be working for me in 4.12. I tried yours and this:
diff --git a/deploy/namespace.yaml b/deploy/namespace.yaml
index 37284785..dd8a4ed7 100644
--- a/deploy/namespace.yaml
+++ b/deploy/namespace.yaml
@@ -4,3 +4,8 @@ metadata:
   name: $NAMESPACE
   labels:
     name: $NAMESPACE
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/enforce-version: v1.24
+    security.openshift.io/scc.podSecurityLabelSync: "false"
However the config daemon is not created.
Only when I do:
oc label ns openshift-sriov-network-operator pod-security.kubernetes.io/audit=privileged pod-security.kubernetes.io/enforce=privileged pod-security.kubernetes.io/warn=privileged security.openshift.io/scc.podSecurityLabelSync=false
Then I see the sriov-network-config-daemon being created.

wizhaoredhat: Sorry I forgot to delete the namespace with:
```
oc delete namespace openshift-sriov-network-operator
```
However only this works for me. Your changes don't seem to work. Most notably the podSecurirtyLabelSync has to be false for me:
```
diff --git a/deploy/namespace.yaml b/deploy/namespace.yaml
index 37284785..dd8a4ed7 100644
--- a/deploy/namespace.yaml
+++ b/deploy/namespace.yaml
@@ -4,3 +4,8 @@ metadata:
   name: $NAMESPACE
   labels:
     name: $NAMESPACE
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/enforce-version: v1.24
+    security.openshift.io/scc.podSecurityLabelSync: "false"
```

deployment/sriov-network-operator/README.md
SchSeba: yep because the helm deployment doesn't use the deploy folder

deploy/namespace.yaml
SchSeba: please take a look I change the labelSync to false.

the namespace you use start with `openshift-` :)

wizhaoredhat: Seems to work properly now! Thanks Seb:
```
[root@wsfd-advnetlab28 sriov-network-operator]# git diff
diff --git a/deploy/namespace.yaml b/deploy/namespace.yaml
index 37284785..81303160 100644
--- a/deploy/namespace.yaml
+++ b/deploy/namespace.yaml
@@ -4,3 +4,6 @@ metadata:
   name: $NAMESPACE
   labels:
     name: $NAMESPACE
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/enforce-version: v1.24
+    security.openshift.io/scc.podSecurityLabelSync: "false"
[root@wsfd-advnetlab28 sriov-network-operator]# oc get pods -n openshift-sriov-network-operator
NAME                                      READY   STATUS    RESTARTS   AGE
network-resources-injector-4qxdb          1/1     Running   0          27s
network-resources-injector-jfjs4          1/1     Running   0          27s
network-resources-injector-k4jj8          1/1     Running   0          27s
operator-webhook-nzw8v                    1/1     Running   0          27s
operator-webhook-sm6zd                    1/1     Running   0          27s
operator-webhook-t69w5                    1/1     Running   0          27s
sriov-network-config-daemon-7htv9         3/3     Running   0          27s
sriov-network-config-daemon-8mk6r         3/3     Running   0          27s
sriov-network-config-daemon-ffjgg         3/3     Running   0          27s
sriov-network-config-daemon-gdmrc         3/3     Running   0          27s
sriov-network-config-daemon-lwqqd         3/3     Running   0          27s
sriov-network-operator-7cdf755677-kfscn   1/1     Running   0          36s
[root@wsfd-advnetlab28 sriov-network-operator]# oc describe project openshift-sriov-network-operator
Name:			openshift-sriov-network-operator
Created:		49 seconds ago
Labels:			kubernetes.io/metadata.name=openshift-sriov-network-operator
			name=openshift-sriov-network-operator
			pod-security.kubernetes.io/enforce=privileged
			pod-security.kubernetes.io/enforce-version=v1.24
			security.openshift.io/scc.podSecurityLabelSync=false
Annotations:		kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"v1","kind":"Namespace","metadata":{"annotations":{},"labels":{"name":"openshift-sriov-network-operator","pod-security.kubernetes.io/enforce":"privileged","pod-security.kubernetes.io/enforce-version":"v1.24","security.openshift.io/scc.podSecurityLabelSync":"false"},"name":"openshift-sriov-network-operator"}}
			
			openshift.io/sa.scc.mcs=s0:c26,c10
			openshift.io/sa.scc.supplemental-groups=1000670000/10000
			openshift.io/sa.scc.uid-range=1000670000/10000
Display Name:		<none>
Description:		<none>
Status:			Active
Node Selector:		<none>
Quota:			<none>
Resource limits:	<none>
```

deployment/sriov-network-operator/README.md
wizhaoredhat: Got it thanks!

jjuele:

deployment/sriov-network-operator/README.md
adrianchiris: by default if PSA is enabled AFAIU it will do nothing unless namespace has been labeled.

is my understanding correct ?

if yes, then we should have a section that says, in case PSA is used, privileged policy should be set in enforce label.
an not always add the label.

deploy/namespace.yaml
```diff
@@ -4,3 +4,6 @@ metadata:
...
+    pod-security.kubernetes.io/enforce-version: v1.24
+    security.openshift.io/scc.podSecurityLabelSync: "false"
```
adrianchiris, line 9: this change needed for dev setups ? or there are other consumers of the deploy folder ?
why do we need to add openshift labels here  ?

wizhaoredhat: This is needed for "make deploy-setup". I don't believe there are other consumers. Unfortunately openshift 4.12 with the new pod security feature needs the openshift label.

adrianchiris: it would be great if we have these 3 labels added to namespace only when when `make deploy-setup` (i.e deployed in openshift) and not always since when running `make deploy-setup-k8s` these labels are not needed.

wizhaoredhat:
@SchSeba  Can we get this in?

deploy/namespace.yaml
```diff
@@ -4,3 +4,6 @@ metadata:
   name: $NAMESPACE
   labels:
     name: $NAMESPACE
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/enforce-version: v1.24
```
bn222, line 8: I don't think that "going back to the behavior from 1.24" makes sense for a feature that changed on purpose in k8s. Can we fix it differently? This seems like adding tech debt.

bn222: Read more about it, and it makes sense to use v1.24 here. Ignore my previous message.

bn222: Can we use `latest` instead of v1.24?

bn222: Reading more into this, it seems that `enforce-version` is optional. It's not clear to me why we need it in the first place?

SchSeba: I think we can remove that part not problem

deploy/namespace.yaml
wizhaoredhat: Let's create a new namespace file for Openshift. And only apply this new namespace file in the case of openshift (deploy-setup and openshift namespace) in the Makefile.

bn222: I think the fact that it is required or not depends on how k8s is being configured, not if it is Openshift or not.

wizhaoredhat: Hmm maybe it is better to use kubectl labels for the different scenarios.

bn222: Let's just remove the last label and use something similar to https://github.com/openshift/dpu-network-operator/pull/52/commits/0e8e5511fb0183b1452a8e68af1d2c12e622ea08#diff-ce0d597a9e5f20a8fe61c3ba5e6185a6b90cfef2c00bb5b5ef3e97b15b79f480R6-R8

PR #373, opened 2022-11-02
improve sriov nic selector in tests

SchSeba:
If the sriov is not able in the kernel for intel nic the totalVF will be 0 so we skip the device. That is not the case for Mellanox devices that will report 0 until we configure the sriov interfaces with the mstconfig package

### DIFF
--- a/test/util/cluster/cluster.go
+++ b/test/util/cluster/cluster.go
@@ -104,6 +104,13 @@ func (n *EnabledNodes) FindOneSriovDevice(node string) (*sriovv1.InterfaceExt, e
 				continue
 			}
 
+			// if the sriov is not enable in the kernel for intel nic the totalVF will be 0 so we skip the device
+			// That is not the case for Mellanox devices that will report 0 until we configure the sriov interfaces
+			// with the mstconfig package
+			if itf.Vendor == intelVendorID && itf.TotalVfs == 0 {
+				continue
+			}
+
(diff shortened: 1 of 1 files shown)


test/util/cluster/cluster.go
```diff
@@ -104,6 +104,13 @@ func (n *EnabledNodes) FindOneSriovDevice(node string) (*sriovv1.InterfaceExt, e
 				continue
 			}
 
+			// if the sriov is not able in the kernel for intel nic the totalVF will be 0 so we skip the device
```
adrianchiris, line 107: nit: able -> enable ?

test/util/cluster/cluster.go
```diff
@@ -126,6 +133,13 @@ func (n *EnabledNodes) FindSriovDevices(node string) ([]*sriovv1.InterfaceExt, e
...
```
adrianchiris, line 136: nit: same as above

adrianchiris (approved):

jjuele:

test/util/cluster/cluster.go
SchSeba: done thanks :)

test/util/cluster/cluster.go
SchSeba: done thanks :)

e0ne (approved):
LGTM

PR #374, opened 2022-11-02
Revert "support graceful node shutdown"

SchSeba:
I will like to propose the revert of this PR.

This PR was intended to switch the machine config operator pause mechanism we have in place for openshift.
The problem is the feature is not fully integrated into openshift yet so merging this PR re-introduce the race conditions we had before in openshift between the sriov operator and the machine config operator.

Reverts k8snetworkplumbingwg/sriov-network-operator#363


adrianchiris (approved):

SchSeba:
two maintainers from two different companies approve the change merging

PR #377, opened 2022-11-22
Vdpa introduction

lmilleri:
This PR is the first step on integrating vDPA in kubernetes containers.

API changes: a new attribute has been added to the policy.
The **vdpaType** can assume two values: **virtio** or **vhost**.
This PR implements the virtio mode only, vhost mode will be implemented in the future.

Supported card: NVIDIA Connect6-DX in switchdev mode (OVS HW offloading enabled)

The NIC is configured via the existing systemd services: switchdev-configuration-before-nm.service and switchdev-configuration-after-nm.service.

Webhook validation has been extended to check vDPA specific static configuration.

Note: this PR is depending on https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/450, so the go.mod needs to be amended after this PR is merged to master.
Please also consider that this vDPA solution depends on PR https://github.com/ovn-org/ovn-kubernetes/pull/2664.


wizhaoredhat:
Changes to the machine config, needs my changes here:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/381
Otherwise your changes won't apply to the node upon upgrade.

bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
```diff
@@ -25,11 +43,25 @@ contents:
...
         VfDirs=$(ls /sys/bus/pci/devices/${pci_addr} | grep virtfn)
 
-        # load VF driver
+        # load VF driver and configure vdpa if needed
         for VfDir in $VfDirs
         do
           VfPciAddr=$(basename "$( readlink -f /sys/bus/pci/devices/${pci_addr}/$VfDir )")
           echo $VfPciAddr > /sys/bus/pci/drivers_probe
+
+          # extract VF id from VfDir
```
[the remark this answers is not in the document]
lmilleri: Yes, basically if the vdpaType attribute is not present, the logic for adding a vdpa device is not executed

wizhaoredhat, line 52: Since this service file will run before the operator has a chance to run. Are we sure we don't need to load the vDPA driver here?

lmilleri: I think the vDPA drivers are loaded in the kernel just when the policy is applied, before the node reboots (please check https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/377/files#diff-5e18b1bb6723228bb51dcd9afe17e63609c26b293a04e6f0a44075ef0f5cfcc2 line 87)

api/v1/helper_test.go
```diff
@@ -77,7 +78,8 @@ func newNodePolicy() *v1.SriovNetworkNodePolicy {
 			Name: "p1",
 		},
 		Spec: v1.SriovNetworkNodePolicySpec{
-			DeviceType: "netdevice",
+			DeviceType: consts.DeviceTypeNetDevice,
+			VdpaType:   consts.VdpaTypeVirtio,
```
SchSeba, line 82: can you please have a newPolicy for the vdpa test?
I will like to validate that we continue to support policies without VdpaType configured

pkg/webhook/validate.go
```diff
@@ -150,12 +150,19 @@ func staticValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePol
...
 		return false, fmt.Errorf("'deviceType: vfio-pci' conflicts with 'isRdma: true'; Set 'deviceType' to (string)'netdevice' Or Set 'isRdma' to (bool)'false'")
 	}
 	if strings.EqualFold(cr.Spec.LinkType, constants.LinkTypeIB) && !cr.Spec.IsRdma {
 		return false, fmt.Errorf("'linkType: ib or IB' requires 'isRdma: true'; Set 'isRdma' to (bool)'true'")
 	}
+	// vdpa specific validation
+	if cr.Spec.VdpaType == constants.VdpaTypeVHost {
+		return false, fmt.Errorf("'vdpaType: vhost' is not supported yet")
+	}
+	if cr.Spec.DeviceType == constants.DeviceTypeVfioPci && cr.Spec.VdpaType == constants.VdpaTypeVirtio {
```
SchSeba, line 163: from the PR info I can see that this works only if the device is on switchDev move should we also validate that here and failed if the user didn't create the policy with switch enabled?

SchSeba, line 163: another point today we only support the MLX device should we also check that here?

SchSeba (commented):
overall looks great!

I just add some small comments related to the webhook validation we do

api/v1/helper_test.go
lmilleri: It makes sense, will do.

pkg/webhook/validate.go
lmilleri: Ok, will do

lmilleri: Ok, will do

pkg/webhook/validate.go
```diff
@@ -150,12 +150,29 @@ func staticValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePol
...
-	if cr.Spec.DeviceType == "vfio-pci" && cr.Spec.IsRdma {
+	if cr.Spec.DeviceType == constants.DeviceTypeVfioPci && cr.Spec.IsRdma {
 		return false, fmt.Errorf("'deviceType: vfio-pci' conflicts with 'isRdma: true'; Set 'deviceType' to (string)'netdevice' Or Set 'isRdma' to (bool)'false'")
 	}
 	if strings.EqualFold(cr.Spec.LinkType, constants.LinkTypeIB) && !cr.Spec.IsRdma {
 		return false, fmt.Errorf("'linkType: ib or IB' requires 'isRdma: true'; Set 'isRdma' to (bool)'true'")
 	}
+
+	// vdpa specific validation
+	if cr.Spec.VdpaType == constants.VdpaTypeVHost {
```
adrianchiris, line 161: why not just remove it from enum kubebuilder annotation ? later on it can be added

lmilleri: will do thx

bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
adrianchiris: `utils.LoadKernelModule` essentially calls modprobe [1]. this is not persistent after reboot AFAIK

so after rebooting a node virtio_vdpa driver would not be loaded when this bit runs. do you need it to perform any of the below operations ?

[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/bindata/scripts/load-kmod.sh

bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
```diff
@@ -25,11 +43,25 @@ contents:
...
+          vfid=${VfDir:6}
```
adrianchiris, line 53: why 6 ?

adrianchiris:
@lmilleri would you like adding a `vdpa.md` under doc ? :)
just to provide some general overview, requirements (kernel, hardware)  plus example configuration of SNO

bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
lmilleri: No, I don't need the virtio_vdpa driver for the operations below, the command `vdpa dev add name vdpa:0000:65:00.3 mgmtdev pci/0000:65:00.3` for example, can be executed with no issues. Later on the driver is loaded as expected, I tested that after rebooting the node.

bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
lmilleri: it extracts the VF id from strings like `/sys/bus/pci/devices/0000:65:00.0/virtfn1`. 1 is the 7th character after `virtfn`. Is it enough to add a comment above the line in question?

bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
lmilleri: Only the parent driver is needed at this stage (e.g. mlx5_vdpa)

bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
adrianchiris: a comment would be nice :) thx !

bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
adrianchiris: OK cool thx for clarifying it

lmilleri:
> @lmilleri would you like adding a `vdpa.md` under doc ? :) just to provide some general overview, requirements (kernel, hardware) plus example configuration of SNO

Good idea, will do

lmilleri:
@SchSeba @adrianchiris I have just pushed all the requested changes. Please have a look to it again

adrianchiris (approved):
Thx for addressing my comments @lmilleri LGTM!

lets wait for CI to pass

lmilleri:
@adrianchiris Nvidia e2e test CI is failing. I can't check for details, apparently 13.74.249.42 is not reachable

adrianchiris:
/test-all

there is some issue with CI log server (should be solved during next week only).
also the CI machine had outdated go version, this was fixed. retriggering.

pkg/webhook/validate.go
```diff
@@ -150,12 +150,25 @@ func staticValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePol
...
+	// vdpa: deviceType must be set to 'netdevice'
+	if cr.Spec.DeviceType != constants.DeviceTypeNetDevice && cr.Spec.VdpaType == constants.VdpaTypeVirtio {
+		return false, fmt.Errorf("'deviceType: %s' conflicts with 'vdpaType: virtio'; Set 'deviceType' to (string)'netdevice' Or Remove 'vdpaType'", cr.Spec.DeviceType)
+	}
+	// vdpa: device must be configured in switchdev mode
+	if cr.Spec.VdpaType == constants.VdpaTypeVirtio && cr.Spec.EswitchMode != sriovnetworkv1.ESwithModeSwitchDev {
+		return false, fmt.Errorf("virtio/vdpa requires the device to be configured in switchdev mode")
+	}
+	// vdpa: only mellanox cards are supported
+	if cr.Spec.VdpaType == constants.VdpaTypeVirtio && cr.Spec.NicSelector.Vendor != "" && cr.Spec.NicSelector.Vendor != MellanoxID {
```
SchSeba, line 169: I think is better to add the validation here https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/377/files#diff-8fd5d83d7413d77ee2eab8e5c1a2ea1623d63e2aec69017a95811db9ad9edf52R235
so we can always check it not only if vendorID is configured

SchSeba (commented):
only one small comment

pkg/webhook/validate.go
lmilleri: done, thx

PR #381, opened 2022-11-28
Fix machine config update in sriovnetworkpoolconfig_controller

wizhaoredhat:
These fixes only apply to Openshift clusters with the machine config operator.
    
The controller was not detecting changes to the machine config files. Firstly if the existing MachineConfig's ignition JSON differs from the rendered MachineConfig's ignition JSON then we should update the MachineConfig. Then we should update the MachineConfig with the rendered MachineConfig. This logic had an incorrect implementation priorly.

### DIFF
--- a/controllers/sriovnetworkpoolconfig_controller.go
+++ b/controllers/sriovnetworkpoolconfig_controller.go
@@ -1,9 +1,10 @@
 package controllers
 
 import (
-	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"reflect"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
Investigating that foundMC.Spec.Config.Raw is never equal to mc.Spec.Config.Raw. Even though there are no changes to the machine config files. It is most likely metadata. Posting this PR (in draft state) such that developers with experience with the Machine Config API can take a look.

controllers/sriovnetworkpoolconfig_controller.go
```diff
@@ -175,14 +176,24 @@ func (r *SriovNetworkPoolConfigReconciler) syncOvsHardwareOffloadMachineConfigs(
...
-			if bytes.Equal(foundMC.Spec.Config.Raw, mc.Spec.Config.Raw) {
+			var foundIgn, renderedIgn interface{}
+			// The Raw config JSON string may have the fields reordered.
+			// For example the "path" field may come before the "contents"
+			// field in the rendered ignition JSON; while the found
+			// MachineConfig's ignition JSON would have it the other way around.
+			// Thus we need to unmarshal the JSON for both found and rendered
+			// ignition and compare.
+			json.Unmarshal(foundMC.Spec.Config.Raw, &foundIgn)
+			json.Unmarshal(mc.Spec.Config.Raw, &renderedIgn)
```
bn222, line 187: Can't we deep compare the Spec? iow, something like

```
if !reflect.DeepEqual(foundMC.spec, foundMC.spec,

Comment (wizhaoredhat) [reply controllers/sriovnetworkpoolconfig_controller.go:187]: I believe DeepEqual will always return false if we directly use the spec (I gave this a try). The reason is due to the Raw JSON config having fields reordered. This will cause the comparison to return not equal even though the fields are the same.

File: controllers/sriovnetworkpoolconfig_controller.go [hunk 2]
```diff
@@ -175,9 +175,10 @@ func (r *SriovNetworkPoolConfigReconciler) syncOvsHardwareOffloadMachineConfigs(
 				return fmt.Errorf("couldn't delete MachineConfig: %v", err)
 			}
 		} else {
-			if bytes.Equal(foundMC.Spec.Config.Raw, mc.Spec.Config.Raw) {
+			if !bytes.Equal(foundMC.Spec.Config.Raw, mc.Spec.Config.Raw) {
 				logger.Info("MachineConfig already exists, updating")
-				err = r.Update(context.TODO(), foundMC)
+				mc.SetResourceVersion(foundMC.GetResourceVersion())
+				err = r.Update(context.TODO(), mc)
```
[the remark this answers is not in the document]
wizhaoredhat: Thanks for the hint Zenghui. Adopted the change.

controllers/sriovnetworkpoolconfig_controller.go
bn222: Then how come https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/controllers/sriovibnetwork_controller.go#L155 doesn't have that issue? Did you investigate that?

wizhaoredhat: I tried reflect.DeepEqual(mc.Spec.Config, foundMC.Spec.Config) and reflect.DeepEqual(mc.Spec, foundMC.Spec); the result for both is always false.

wizhaoredhat: I think it is the type "runtime.RawExtension" that is causing this.

SchSeba: I think you are right here @wizhaoredhat that type is just a []byte

PR #384, opened 2022-11-29
Add the OvS max revalidator time to the ovs-vswitchd service

wizhaoredhat:
The OvS default max revalidator time is not enough for OvS Mellanox NICs. This is presumably only on the Mellanox NICs, however Mellanox NICs are the only NICs that currently support OvS HWOL with the SRIOV network operator. With the default OvS 500ms max revalidator time, it was observed during long duration traffic tests that the flows were removed momentarily then added back immediately. The max revalidator time needs to be extended to allow for higher than expected dump durations during the OvS revalidator sweep. This will prevent the flows from being deleted due to these unexpected delays.

### DIFF
--- a/bindata/manifests/switchdev-config/ovs-units/ovs-vswitchd.service.yaml
+++ b/bindata/manifests/switchdev-config/ovs-units/ovs-vswitchd.service.yaml
@@ -3,4 +3,8 @@ dropins:
 - name: 10-hw-offload.conf
   contents: |
     [Service]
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
Currently testing with 1 second and will report findings. However 10 seconds was tested prior and was working both at Red Hat and NVIDIA.

wizhaoredhat:
No longer needed for now. Fix is in the flow stats area in OvS.

PR #388, opened 2022-12-09
Add script to switch BF2 between NIC and DPU mode

bn222:
This patch adds a new script called bf2-switch-mode.sh that allows the user to easily switch the BF2 between Network Interface Controller (NIC) and Data Processing Unit (DPU) modes. This is useful for optimizing the BF2 for different workloads.

Co-authored-by: Costin Gamenț <costin.gament@gmail.com>
Signed-off-by: Balazs Nemeth <bnemeth@redhat.com>

### DIFF
--- a/bindata/scripts/bf2-switch-mode.sh
+++ b/bindata/scripts/bf2-switch-mode.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+set -e
+
+mode=${1:-query}
+device=${2:-"$(mstconfig q | grep "Device:" | awk 'BEGIN {FS="/";} { print $6 }')"}
+
+if [ -z "$device" ]; then
+    echo "Can't find device."
+    exit 120
+else
+    echo Found device: "${device}"
+fi
+
+fwreset () {
+    mstfwreset -y -d "${device}" reset
+    echo "Switched to ${mode} mode."
+    exit 0
+}
+
+query () {
+    current_config=$(mstconfig -d "${device}" q \
+        INTERNAL_CPU_MODEL \
+        INTERNAL_CPU_PAGE_SUPPLIER \
+        INTERNAL_CPU_ESWITCH_MANAGER \
+        INTERNAL_CPU_IB_VPORT0 \
+        INTERNAL_CPU_OFFLOAD_ENGINE | tail -n 5 | awk '{print $2}' | xargs echo)
+    echo "Current DPU configuration: ${current_config}"
+}
+
+query
+case "${mode}" in
+    dpu)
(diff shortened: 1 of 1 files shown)


bindata/scripts/bf2-switch-mode.sh
```diff
@@ -0,0 +1,66 @@
...
+
+if [ -z "$device" ]; then
+    echo "Can't find device."
+    exit 120
+else
+    echo Found device: "${device}"
+fi
+
+fwreset () {
+    mstfwreset -y -d "$(echo "${device}" | awk -F/ '{print $(NF-1)}')" reset
```
wizhaoredhat, line 15: Why do we need the awk here?

bn222:
@adrianchiris

bindata/scripts/bf2-switch-mode.sh
wizhaoredhat: We talked about this offline and decided to improve on the device retrieval.

bindata/scripts/bf2-switch-mode.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
adrianchiris, line 1: lets add a comment explaining where this script is expected to be used and how.

bindata/scripts/bf2-switch-mode.sh
```diff
@@ -0,0 +1,66 @@
...
+set -e
+
+mode=${1:-query}
+device=${2:-"$(mstconfig q | grep "Device:" | awk 'BEGIN {FS="/";} { print $6 }')"}
```
adrianchiris, line 5: should you ensure the selected device is BF2 ?

bn222: This is a good point, I would have added 

```
type=$(mstconfig q $device | grep "Device type:" | awk '{ print $3}')

if [ "${type}" != "BlueField2" ]; then
    echo "Device is not a Bluefield2"
    exit 120
fi
```

bindata/scripts/bf2-switch-mode.sh
bn222: Good point!

bn222:
This will be pushed only downstream for now to https://github.com/openshift/sriov-network-operator

adrianchiris:
> This will be pushed only downstream for now to https://github.com/openshift/sriov-network-operator

Personally i  think its best. for upstream we should potentially have a way to pass down vendor specific configuration to vendor plugins in a way its agnostic to the sriov-network-operator.

when we have more vendors supporting DPU/IPU then we can see how to model it in API .

PR #389, opened 2022-12-12
Remove sh-specific redirections.

iancoolidge:
The "<<<" are not available in dash, used by deb-based OSes. Replaced it with "echo" and "|" which achieves same effect.

Reviewed-by: Chun-Feng (David) Liu <cfliu@google.com>


iancoolidge:
I don't seem to have merge power. It looks like all tests passed

SchSeba (approved):

PR #391, opened 2022-12-14
Make sriov-network-config-daemon crash when configmap cannot be obtained

### LINKED ISSUE #390: sriov-network-config-daemon should crash if configmap with NIC interfaces cannot be obtained
In our K8s cluster occasionally it is observed that sriov-network-config-daemon at its start is not able to obtain the configmap with supported NIC IDs.  The configuration container is unable to communicate with the K8s API (connection is refused), the reason is unknown.  In result, the device (our E810 NIC) keeps being reported as unsupported. Hance, no network resources are configured and enabled in the K8s cluster. Initially, we figured out that deleting the sriov-network-config-daemon pod manually helps as the new pod tries again and is able to obtain the configmap then.  Also, a slight change in the code helps, which makes the pod crash and start again in case the configmap is not obtained successfully.

The suggested change is to add one line 
`panic(err.Error())`

under https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/cmd/sriov-network-config-daemon/start.go#L186

Our clusters are deployed on systems with Ubuntu 20.04 LTS , Kernel 5.4.0, and Kubernetes 1.23

jkossak:
This PR adds a line to crash sriov-network-config-daemon if it can't access the configmap with NIC interfaces. 
Fixes #390

### DIFF
--- a/cmd/sriov-network-config-daemon/start.go
+++ b/cmd/sriov-network-config-daemon/start.go
@@ -184,6 +184,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	var namespace = os.Getenv("NAMESPACE")
 	if err := sriovnetworkv1.InitNicIDMap(kubeclient, namespace); err != nil {
 		glog.Errorf("failed to run init NicIdMap: %v", err)
+		panic(err.Error())
 	}
 
 	// block the deamon process until nodeWriter finish first its run


zeeke:
Makes sense to me. I'm not aware of any scenario in which the config-daemon can work without `InitNicIDs` successfully initialized.

@SchSeba I think this should not clash with #365 , as the service implementation would use another function to obtain the map. Right?

e0ne (approved):

PR #393, opened 2022-12-20
add support for mellanox/rdma-cni

tariq1890:
In addition to `sriov-cni`, we need to deploy `rdma-cni`. It would be great if we can add this as a component to the sriov-config-daemon DS

### DIFF
--- a/bindata/manifests/daemon/daemonset.yaml
+++ b/bindata/manifests/daemon/daemonset.yaml
@@ -38,6 +38,17 @@ spec:
       {{- end }}
       {{- end }}
       containers:
+      - name: rdma-cni
+        image: {{.RDMACNIImage}}
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            cpu: 10m
+            memory: 10Mi
+        volumeMounts:
+          - name: cnibin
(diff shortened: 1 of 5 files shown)


adrianchiris:
to fully support RDMA namespaces in sriov-network-operator it would entail the following:

1. configure rdma subsystem to exclusive mode on node boot[1]
2. deploy rdma-cni[2]
3. SriovNetwork to support chaning rdma-cni (can be done using meta plugins[3] so no change needed here)
4. IbSriovNetwork to support rdmaIsolation ib-sriov-cni attribute [4]

for 2. we should probably look into moving rdma-cni to npwg as now also Intel supports RDMA.

[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/156
[2] https://github.com/Mellanox/rdma-cni
[3] https://github.com/k8snetworkplumbingwg/sriov-network-operator#chaining-cni-metaplugins
[4] https://github.com/k8snetworkplumbingwg/ib-sriov-cni#configuration-reference

adrianchiris:
converting this one  into an issue as it  entails several changes and perhaps a design doc to align on behaviour.

adrianchiris:
#481

PR #397, opened 2023-01-17
Update go to 1.19

bn222:
Signed-off-by: Balazs Nemeth <bnemeth@redhat.com>

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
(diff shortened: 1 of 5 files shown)


adrianchiris:
in project dockerfiles point to base image with go1.18 these should be adjusted as well no ?

bn222:
> in project dockerfiles point to base image with go1.18 these should be adjusted as well no ?

Good catch. Updated it.

adrianchiris (approved):
lgtm

SchSeba (approved):
/lgtm

PR #402, opened 2023-01-31
Enable smfs flow steering by default in switchdev mode.

wizhaoredhat:
It was found that dmfs is slower than smfs. This affects dump duration of flows in OvS. Since smfs should replace dmfs moving forward, we will set smfs to be used by default.

Signed-off-by: William Zhao <wizhao@redhat.com>


bn222:
@adrianchiris

bn222:
@vrindle

adrianchiris (approved):

PR #403, opened 2023-02-06
remove unused volume from sriov-device-plugin DS

tariq1890:
`/sys/class/net` is not mounted, so we can just remove the volume ref


tariq1890:
@e0ne @pliurh @SchSeba  Requesting your reviews

adrianchiris (approved):

e0ne (approved):

SchSeba (approved):

PR #404, opened 2023-02-07
add printercolumn for syncStatus in SriovNetworkNodeState

tariq1890:
It would be useful to see the status of all nodes when querying through `SriovNetworkNodeStates`

### DIFF
--- a/api/v1/sriovnetworknodestate_types.go
+++ b/api/v1/sriovnetworknodestate_types.go
@@ -91,6 +91,7 @@ type SriovNetworkNodeStateStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="Sync Status",type=string,JSONPath=`.status.syncStatus`
+//+kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 
 // SriovNetworkNodeState is the Schema for the sriovnetworknodestates API
 type SriovNetworkNodeState struct {
--- a/api/v1/sriovnetworknodestate_types.go
+++ b/api/v1/sriovnetworknodestate_types.go
@@ -90,6 +90,7 @@ type SriovNetworkNodeStateStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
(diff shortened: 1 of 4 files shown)


SchSeba (approved):
I like the idea :)

adrianchiris (approved):

api/v1/sriovnetworknodestate_types.go
```diff
@@ -90,6 +90,7 @@ type SriovNetworkNodeStateStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Sync Status",type=string,JSONPath=`.status.syncStatus`
```
e0ne, line 93: Please, add something like `// +kubebuilder:printcolumn:name="Age",type=string,JSONPath=`.metadata.creationTimestamp``  (I didn't test this change with sriov operator so this line could be inacurate) to show resource age as we've got it now

e0ne (changes requested):

api/v1/sriovnetworknodestate_types.go
tariq1890: Sure!

tariq1890:
@e0ne Added the printercolumn for Age. Please review

pkg/daemon/daemon.go
```diff
@@ -103,14 +103,15 @@ type Daemon struct {
...
-	syncStatusFailed    = "Failed"
+	rdmaScriptsPath      = "/bindata/scripts/enable-rdma.sh"
+	udevScriptsPath      = "/bindata/scripts/load-udev.sh"
+	annoKey              = "sriovnetwork.openshift.io/state"
+	annoIdle             = "Idle"
+	annoDraining         = "Draining"
+	annoMcpPaused        = "Draining_MCP_Paused"
+	syncStatusSucceeded  = "Succeeded"
+	syncStatusFailed     = "Failed"
+	syncStatusInProgress = "InProgress"
```
e0ne, line 114: This change is not related to the PR but I'm OK with this minor cleanup

tariq1890:
@e0ne Would you be able to review ?

adrianchiris:
@e0ne can we merge this one ?

e0ne (approved):
LGTM. Thanks for addressing my comments

PR #405, opened 2023-02-08
Skip the Redhat virtual nic from udev rule

SchSeba:
we need to remove the Red Hat Virtio network device from the udev rule configuration if we don't remove it when running the config-daemon on a virtual node it will disconnect the node after a reboot even that the operator should not be installed on virtual environments that are not openstack we should not destroy the cluster if the operator is installed there


adrianchiris (approved):
LGTM,

PR #406, opened 2023-02-08
Pretty print CNI config when rendering NetAttachDefs

tariq1890:
When debugging CNI configs, it becomes a lot easier when we can see pretty-printed JSON in NetworkAttachmentDefinitions

### DIFF
--- a/controllers/helper.go
+++ b/controllers/helper.go
@@ -17,6 +17,8 @@ limitations under the License.
 package controllers
 
 import (
+	"bytes"
+	"encoding/json"
 	"os"
 	"strings"
 
@@ -37,4 +39,12 @@ func GetImagePullSecrets() []string {
 	} else {
 		return []string{}
 	}
+}
+
+func formatJSON(str string) (string, error) {
+	var prettyJSON bytes.Buffer
+	if err := json.Indent(&prettyJSON, []byte(str), "", "    "); err != nil {
+		return "", err
+	}
+	return prettyJSON.String(), nil
 }
--- a/controllers/sriovibnetwork_controller.go
+++ b/controllers/sriovibnetwork_controller.go
@@ -117,6 +117,12 @@ func (r *SriovIBNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+	// format CNI config json in CR for easier readability
+	netAttDef.Spec.Config, err = formatJSON(netAttDef.Spec.Config)
+	if err != nil {
+		reqLogger.Error(err, "Couldn't process rendered NetworkAttachmentDefinition config", "Namespace", netAttDef.Namespace, "Name", netAttDef.Name)
+		return reconcile.Result{}, err
+	}
 	if lnns, ok := instance.GetAnnotations()[sriovnetworkv1.LASTNETWORKNAMESPACE]; ok && netAttDef.GetNamespace() != lnns {
 		err = r.Delete(ctx, &netattdefv1.NetworkAttachmentDefinition{
 			ObjectMeta: metav1.ObjectMeta{
--- a/controllers/sriovibnetwork_controller_test.go
+++ b/controllers/sriovibnetwork_controller_test.go
@@ -47,7 +47,7 @@ var _ = Describe("SriovIBNetwork Controller", func() {
 		DescribeTable("should be possible to create/delete net-att-def",
 			func(cr sriovnetworkv1.SriovIBNetwork) {
 				var err error
-				expect := util.GenerateExpectedIBNetConfig(&cr)
+				expect := generateExpectedIBNetConfig(&cr)
 
 				By("Create the SriovIBNetwork Custom Resource")
 				// get global framework variables
@@ -104,7 +104,7 @@ var _ = Describe("SriovIBNetwork Controller", func() {
 					Expect(k8sClient.Delete(goctx.TODO(), &old)).To(Succeed())
 				}()
 				found := &sriovnetworkv1.SriovIBNetwork{}
-				expect := util.GenerateExpectedIBNetConfig(&new)
+				expect := generateExpectedIBNetConfig(&new)
 
 				retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 					// Retrieve the latest version of SriovIBNetwork before attempting update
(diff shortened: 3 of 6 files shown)


tariq1890:
@adrianchiris @SchSeba @e0ne 

Request review on this as well

api/v1/helper.go
```diff
@@ -521,7 +521,7 @@ func (cr *SriovIBNetwork) RenderNetAttDef() (*uns.Unstructured, error) {
 		return nil, err
 	}
 	for _, obj := range objs {
-		raw, _ := json.Marshal(obj)
+		raw, _ := json.MarshalIndent(obj, "", "  ")
```
adrianchiris, line 524: just wondering if this will make reading logs better or worse when they are scraped and sent over to a centralized logging and viewed from a web console e.g graylog / ELK stack or similar.

tariq1890: hmmm, I hadn't thought of that. I know SumoLogic has auto-formatting capabilities for JSON. So the pretty-printing would be idempotent. Not sure of other tools

tariq1890: Just to be clear, this change is so that we can view the plugins config well-formatted in the NetAttachDef CR

tariq1890:
@adrianchiris Any update on this?

tariq1890:
@adrianchiris I have updated the PR to ensure the rendered CNI config in the NetAttachDef is pretty-printed. Requesting your review on this

controllers/sriovibnetwork_controller.go
```diff
@@ -129,6 +129,13 @@ func (r *SriovIBNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			return reconcile.Result{}, err
 		}
 	}
+
+	netAttDef.Spec.Config, err = formatJSON(netAttDef.Spec.Config)
```
adrianchiris, line 133: nit: could you move this to after L119  so its in the same place where netAttDef is created.

controllers/sriovnetwork_controller.go
```diff
@@ -130,6 +130,13 @@ func (r *SriovNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			return reconcile.Result{}, err
 		}
 	}
+
+	netAttDef.Spec.Config, err = formatJSON(netAttDef.Spec.Config)
```
adrianchiris, line 134: same comment as above

tariq1890: done. Thanks!

controllers/sriovibnetwork_controller.go
tariq1890: done. Thanks!

adrianchiris (approved):

test/util/util.go
```diff
@@ -267,3 +268,11 @@ func validateSelector(rc *dptypes.NetDeviceSelectors, ns *sriovnetworkv1.SriovNe
 	}
 	return true
 }
+
+func mustFormatJSON(str string) string {
+	var prettyJSON bytes.Buffer
+	if err := json.Indent(&prettyJSON, []byte(str), "", "    "); err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
tariq1890: I thought panic was justified here, as

1) as the input to this function is deterministic, we are only passing in static json strings that we deem as expected output

2) This function is defined in the test package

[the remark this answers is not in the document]
zeeke: Well, the function is defined in the `test/util` package and it's called by exported functions. Actually, these functions are invoked only by  `controllers/*_test.go` files but `test/util/*` packages are used as a test library from other projects.

If you make the whole chain of function calls package-private, it looks good to me to raise a panic.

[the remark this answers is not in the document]
tariq1890: Hi @zeeke I've just refactored the code to make it all package private. Can you check now?

zeeke (approved):
/lgtm 

@tariq1890 thanks for working on this

waiting for @adrianchiris approval as the code changed a bit since his last review

controllers/helper.go
```diff
@@ -38,3 +41,23 @@ func GetImagePullSecrets() []string {
...
+
+func formatJSON(str string) (string, error) {
+	var prettyJSON bytes.Buffer
+	if err := json.Indent(&prettyJSON, []byte(str), "", "    "); err != nil {
+		return "", err
+	}
+	return prettyJSON.String(), nil
+}
+
+func getLinkState(state string) string {
```
adrianchiris, line 53: why was this method moved ? as its only used in tests it should stay where it was IMO.

controllers/sriovibnetwork_controller_test.go
```diff
@@ -190,3 +190,17 @@ var _ = Describe("SriovIBNetwork Controller", func() {
 		})
 	})
 })
+
+func generateExpectedIBNetConfig(cr *sriovnetworkv1.SriovIBNetwork) string {
```
adrianchiris, line 194: why was it moved from test/utils ?

@fedepaol  @SchSeba do you know if test/utils module being used in other project ?

controllers/sriovnetwork_controller_test.go
```diff
@@ -213,3 +216,34 @@ var _ = Describe("SriovNetwork Controller", func() {
 		})
 	})
 })
+
+func generateExpectedNetConfig(cr *sriovnetworkv1.SriovNetwork) string {
```
adrianchiris, line 220: Moving these methods around is not something that is directly related to this PR so im not a fan of mixing the two.
if they are mixed then at least it should be in separate commits.


also im not too sure if these methods are used externally from /test/util.go
@fedepaol @SchSeba

controllers/helper.go
tariq1890: fair, I will move back this back into the test files

tariq1890:
@adrianchiris I have split the commits into two. Please re-review

adrianchiris:
lets get inputs from @SchSeba @fedepaol regarding the moved methods, i dont remember if these are used in a CI/testing proj in Openshift

zeeke:
I can confirm that methods `GenerateExpectedNetConfig` and `GenerateExpectedIBNetConfig` are not used by Openshift testing project:
- https://github.com/openshift-kni/cnf-features-deploy/tree/master/cnf-tests/testsuites

I asked @tariq1890 to move them because they can panic since this PR, so it's better to break compile compatibility (if any)  than getting a panic inside a library.

tariq1890:
Hello :) how can we move this PR forward? It's been open for a while and I've been through multiple iterations already. Hoping to close this out

adrianchiris:
thx for confirming @zeeke. ill merge this one once CI pass.

PR #408, opened 2023-02-13
Update support for operator deployment on PSA

SalDaniele:
PR #371 provided changes to allow the deployment of the operator when PSA is enabled.

https://kubernetes.io/docs/concepts/security/pod-security-admission/

This commit provides minor updates to address comments on the previous PR.

Signed-off-by: Salvatore Daniele <sdaniele@redhat.com>
Co-authored-by: Sebastian Sch <sebassch@gmail.com>

### DIFF
--- a/deploy/namespace.yaml
+++ b/deploy/namespace.yaml
@@ -4,3 +4,6 @@ metadata:
   name: $NAMESPACE
   labels:
     name: $NAMESPACE
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
--- a/deployment/sriov-network-operator/README.md
+++ b/deployment/sriov-network-operator/README.md
@@ -46,6 +46,11 @@ $ helm install -n sriov-network-operator --create-namespace --wait sriov-network
 $ kubectl -n sriov-network-operator get pods
 ```
 
+In the case that [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) is enabled, the sriov network operator namespace will require a security level of 'privileged'
+```
+$ kubectl label ns sriov-network-operator pod-security.kubernetes.io/enforce=privileged
+```
+
 ## Chart parameters
 
 In order to tailor the deployment of the network operator to your cluster needs


hack/deploy-setup.sh
```diff
@@ -17,7 +17,7 @@ load_manifest() {
 
   pushd ${repo}/deploy
     if ! ${OPERATOR_EXEC} get ns $2 > /dev/null 2>&1 && test -f namespace.yaml ; then
-
```
wizhaoredhat, line 20: Please clean up this file.

wizhaoredhat:
/cc @bn222
/cc @SchSeba
/cc @adrianchiris 

PTAL

hack/deploy-setup.sh
bn222: Let's not touch this file in this commit. Push a separate cleanup PR if you want to clean this up.

deploy/namespace.yaml
```diff
@@ -4,3 +4,6 @@ metadata:
   name: $NAMESPACE
   labels:
     name: $NAMESPACE
+    pod-security.kubernetes.io/enforce: privileged
```
adrianchiris, line 7: why do we need this here ?

if pod-security labels are not added then pod-security mechanism will not kick in for the namespace.
is my understanding wrong ?

bn222: I belive your understanding is wrong. The default behavior depends on how PSA has been configured, hence we can't assume privileged by default.

adrianchiris: OK, i see the built in PSA admission controller [1] has default config, i assume for when these labels are not present on namespace.

[1] https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-admission-controller/

adrianchiris (approved):

deploy/namespace.yaml
bn222: Yes, the default is configurable and that's why we need to be explicit with what permissions we require

SchSeba:
Hi @wizhaoredhat this works for OCP without the other label to disable the OCP sync?

SalDaniele:
Hi @SchSeba when testing on my OCP cluster I did not need to include 'security.openshift.io/scc.podSecurityLabelSync: "false"'

Perhaps this is already set for openshift-sriov-network-operator by default [1].

[1] https://cloud.redhat.com/blog/pod-security-admission-in-openshift-4.11#:~:text=Note%20that%20%E2%80%9Copenshift%2D%E2%80%9D%20prefixed%20namespaces%20do%20not%20get%20synchronized%20by%20default

SchSeba (approved):
/lgtm

PR #409, opened 2023-02-19
use openshift_context instead of internal function in util package

SchSeba:


SchSeba:
as requested by @bn222 I create a new PR to always use the ocp context

https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/365#discussion_r1105800681

pkg/utils/openshift_context.go
```diff
@@ -20,6 +26,36 @@ type OpenshiftContext struct {
...
+		return nil, nil
+	}
+
+	mcclient, err := mcclientset.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	openshiftFlavor := OpenshiftFlavorDefault
+	if ClusterType == ClusterTypeOpenshift {
```
bn222, line 40: Maybe I miss something but if the first conditional in this function checks for `ClusterType != ClusterTypeOpenshift`, how can this ever be false?

controllers/sriovoperatorconfig_controller.go
```diff
@@ -229,11 +230,7 @@ func (r *SriovOperatorConfigReconciler) syncWebhookObjs(ctx context.Context, dc
 		data.Data["CaBundle"] = os.Getenv("WEBHOOK_CA_BUNDLE")
 		data.Data["DevMode"] = os.Getenv("DEV_MODE")
 		data.Data["ImagePullSecrets"] = GetImagePullSecrets()
-		external, err := utils.IsExternalControlPlaneCluster(r.Client)
-		if err != nil {
-			logger.Error(err, "Fail to get control plane topology")
-			return err
-		}
+		external := r.OpenshiftContext.IsHypershift()
```
zeeke, line 233: I suspect this can cause panic:

https://go.dev/play/p/hDtR5cPY6Lj?v=goprev

Maybe it's better to change `IsHypershift()`'s receiver from `OpenshiftContext` to `*OpenshiftContext` and add a nil check inside that function.

bn222: I'm not in favor of passing around pointers just to check for nil. Instead, I think we should pass around the context still, and have it explicitly say it is not Openshift.

pkg/utils/openshift_context.go
SchSeba: you are right I change it

controllers/sriovoperatorconfig_controller.go
SchSeba: good catch!
I update to first check if we are running on ocp

controllers/sriovoperatorconfig_controller.go
```diff
@@ -229,12 +230,13 @@ func (r *SriovOperatorConfigReconciler) syncWebhookObjs(ctx context.Context, dc
 		data.Data["CaBundle"] = os.Getenv("WEBHOOK_CA_BUNDLE")
 		data.Data["DevMode"] = os.Getenv("DEV_MODE")
 		data.Data["ImagePullSecrets"] = GetImagePullSecrets()
-		external, err := utils.IsExternalControlPlaneCluster(r.Client)
-		if err != nil {
-			logger.Error(err, "Fail to get control plane topology")
-			return err
+
+		data.Data["ExternalControlPlane"] = false
+		if utils.ClusterType != utils.ClusterTypeOpenshift {
```
zeeke, line 235: I think this should be `==` ?

@bn222 Were you referring to having this check around the controllers? Or putting it inside `IsHypershift()`?
Something like:

```
func (c *OpenshiftContext) IsHypershift() bool {
        if c == nil {
            return false
        }
	return c.OpenshiftFlavor == OpenshiftFlavorHypershift
}
```

pkg/utils/openshift_context.go
bn222: I don't see the change

pkg/utils/openshift_context.go
```diff
@@ -20,6 +26,36 @@ type OpenshiftContext struct {
 	OpenshiftFlavor OpenshiftFlavor
 }
 
+func NewOpenshiftContext(config *rest.Config, scheme *runtime.Scheme) (*OpenshiftContext, error) {
+	if ClusterType != ClusterTypeOpenshift {
+		return nil, nil
```
bn222, line 31: I really don't like this. Code wise, you will have more checks elsewhere if you return a nil value and that's valid. Just return the context in a way that it conveys the fact that it is not openshift.

controllers/sriovoperatorconfig_controller.go
bn222: Yes, let's not repeat code all over the place. _anything_ related to checking if we are openshift or not, should be in 1 place.

pkg/daemon/daemon.go
```diff
@@ -591,7 +591,7 @@ func (dn *Daemon) completeDrain() error {
 		}
 	}
 
-	if utils.ClusterType == utils.ClusterTypeOpenshift && !dn.openshiftContext.IsHypershift() {
+	if dn.openshiftContext.IsHypershift() && !dn.openshiftContext.IsHypershift() {
```
zeeke, line 594: `true` can't be `false`  :smiley: 
```suggestion
	if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
```

bn222: nice catch :+1:

adrianchiris (approved):

Issue #414, opened 2023-03-07
Make sriov-network-config-daemon crash when nodeInfo cannot be obtained


lizhewei91:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/13405c55b6a77bc2476ad5bf40a3188bb5783244/cmd/sriov-network-config-daemon/start.go#L164

lizhewei91:
If nodeInfo cannot be obtained, Baremetal is always the log platform

--------------------------------------------------------------------------
I0302 11:05:00.113612    7153 start.go:162] starting node writer
W0302 11:05:03.211739    7153 start.go:180] Failed to fetch node state node-17, Get "https://10.222.0.1:443/api/v1/nodes/node-17": dial tcp 10.222.0.1:443: connect: no route to host!
I0302 11:05:03.211791    7153 start.go:182] Running on platform: Baremetal
E0302 11:05:06.283736    7153 start.go:186] failed to run init NicIdMap: Get "https://10.222.0.1:443/api/v1/namespaces/eks-managed/configmaps/supported-nic-ids": dial tcp 10.222.0.1:443: connect: no route to host

PR #415, opened 2023-03-07
Make sriov-network-config-daemon crash when nodeInfo cannot be obtained

lizhewei91:
Change the error level of nodeinfo from warning to fatal. If error is not empty, config-daemon crash. #414


adrianchiris (approved):

adrianchiris:
@zeeke mind giving this one a quick review ?

zeeke (approved):
/lgtm sorry for the late review.

PR #418, opened 2023-03-09
Add the no-wait argument to ovs-vsctl commands for hw-offload

wizhaoredhat:
The "no-wait" argument is needed for ovs-vsctl commands because there is a chance for ovs-vsctl to timeout.

Adding "--no-wait" to the command line would make it such that ovs-vsctl doesn't wait for ovs-vswitchd to update "cur_cfg"

An example of this timeout is shown here:
ovs-vsctl[3265]: ovs|00001|vsctl|INFO|Called as /bin/ovs-vsctl set Open_vSwitch . other_config:hw-offload=true systemd[1]: ovs-vswitchd.service: start-pre operation timed out. Terminating. ovs-vsctl[3265]: 2023-02-22T22:00:20Z|00002|fatal_signal|WARN|terminating with signal 15 (Terminated) ovs-vsctl[3265]: ovs|00002|fatal_signal|WARN|terminating with signal 15 (Terminated) systemd[1]: ovs-vswitchd.service: Failed with result 'timeout'. systemd[1]: Failed to start Open vSwitch Forwarding Unit.

### DIFF
--- a/bindata/manifests/switchdev-config/ovs-units/ovs-vswitchd.service.yaml
+++ b/bindata/manifests/switchdev-config/ovs-units/ovs-vswitchd.service.yaml
@@ -3,4 +3,4 @@ dropins:
 - name: 10-hw-offload.conf
   contents: |
     [Service]
-    ExecStartPre=/bin/ovs-vsctl set Open_vSwitch . other_config:hw-offload=true
+    ExecStartPre=/bin/ovs-vsctl --no-wait set Open_vSwitch . other_config:hw-offload=true
--- a/pkg/render/testdata/machineconfig/units/ovs-vswitchd.service.yaml
+++ b/pkg/render/testdata/machineconfig/units/ovs-vswitchd.service.yaml
@@ -3,4 +3,4 @@ dropins:
 - name: 10-hw-offload.conf
   contents: |
     [Service]
-    ExecStartPre=/bin/ovs-vsctl set Open_vSwitch . other_config:hw-offload=true
+    ExecStartPre=/bin/ovs-vsctl --no-wait set Open_vSwitch . other_config:hw-offload=true


bn222:
This is a great find. Thanks for figuring this one out. I like the idea of adding no-wait we avoid a timeout from systemd. However, I'm wondering what other implications this has. Afaict, not waiting to finish setting hwol to true isn't going to have an impact elsewhere.

/lgtm

bindata/manifests/switchdev-config/ovs-units/ovs-vswitchd.service.yaml
```diff
@@ -3,4 +3,4 @@ dropins:
 - name: 10-hw-offload.conf
   contents: |
     [Service]
-    ExecStartPre=/bin/ovs-vsctl set Open_vSwitch . other_config:hw-offload=true
+    ExecStartPre=/bin/ovs-vsctl --no-wait set Open_vSwitch . other_config:hw-offload=true
```
adrianchiris, line 6: so now OVS might start without this field being set right ?

setting this field after ovs started requires ovs service restart for this config to be applied.

wizhaoredhat: From the docs: http://www.openvswitch.org/support/dist-docs/ovs-vsctl.8.txt

"If you use ovs-vsctl when ovs-vswitchd is not running, use --no-wait." Since we are doing ExecStartPre, this will run before the ovs-vswitchd service starts.

Thus OVS (ovs-vswitchd) will start with this configuration in its database.

It would be a problem if we do ExecStartPost.

wizhaoredhat: E.g.:
Mar 10 06:04:43 localhost systemd[1]: Starting Open vSwitch Forwarding Unit...
Mar 10 06:04:43 localhost ovs-vsctl[2563]: ovs|00001|vsctl|INFO|Called as /bin/ovs-vsctl --no-wait set Open_vSwitch . other_config:hw-offload=true
Mar 10 06:04:43 localhost ovs-ctl[2599]: Inserting openvswitch module.
Mar 10 06:04:43 localhost ovs-ctl[2571]: Starting ovs-vswitchd.

adrianchiris: > i wonder why it doesnt always hang...

that comment was related to before your change.

that is, currently we run without no-wait flag, but we do so before ovs-vswitchd is running so id expect it to always hang.
but apparently it works some times and sometimes it hangs.

wizhaoredhat: I guess the timeout kicks in from systemd which will cause it to exit mid-configuration. ```TimeoutSec=300``` is in our ovs-vswitchd.service. Thus it doesn't hang. It just causes the service to fail, but since the service has ```Restart=on-failure``` the second time, I am guessing the configuration applies very fast (doesn't hit the timeout).

wizhaoredhat: Example:
localhost ovs-vsctl[2663]: ovs|00001|vsctl|INFO|Called as /bin/ovs-vsctl set Open_vSwitch . other_config:hw-offload=true
localhost systemd[1]: ovs-vswitchd.service: start-pre operation timed out. Terminating.
localhost ovs-vsctl[2663]: 2022-12-02T16:51:34Z|00002|fatal_signal|WARN|terminating with signal 15 (Terminated)
localhost ovs-vsctl[2663]: ovs|00002|fatal_signal|WARN|terminating with signal 15 (Terminated)
localhost systemd[1]: ovs-vswitchd.service: Failed with result 'timeout'.
localhost systemd[1]: Failed to start Open vSwitch Forwarding Unit.
localhost systemd[1]: ovs-vswitchd.service: Consumed 23ms CPU time
localhost systemd[1]: ovs-vswitchd.service: Service RestartSec=100ms expired, scheduling restart.
localhost systemd[1]: ovs-vswitchd.service: Scheduled restart job, restart counter is at 1.
localhost systemd[1]: Stopped Open vSwitch Forwarding Unit.
localhost systemd[1]: ovs-vswitchd.service: Consumed 0 CPU time
localhost systemd[1]: Starting Open vSwitch Forwarding Unit...
localhost.localdomain ovs-vsctl[2715]: ovs|00001|vsctl|INFO|Called as /bin/ovs-vsctl set Open_vSwitch . other_config:hw-offload=true
localhost.localdomain ovs-ctl[2760]: Inserting openvswitch module.
localhost.localdomain ovs-ctl[2728]: Starting ovs-vswitchd.
localhost.localdomain ovs-ctl[2728]: Enabling remote OVSDB managers.
localhost.localdomain systemd[1]: Started Open vSwitch Forwarding Unit.

adrianchiris (approved):

bindata/manifests/switchdev-config/ovs-units/ovs-vswitchd.service.yaml
adrianchiris: ack, 10x for clarifying

wizhaoredhat: No problem, Thanks for the ack!

zshi-redhat:
@wizhaoredhat  Is this the fix to the ovs-configuration service failure you found on openshift 4.13?

wizhaoredhat:
> @wizhaoredhat Is this the fix to the ovs-configuration service failure you found on openshift 4.13?

Yes it was in OCP 4.12.

PR #419, opened 2023-03-15
improve the isdraning check for openshift

SchSeba:
there is a case where we annotate the node as draning because we took the leader but on OCP the machine config pool was doing some work so we wait for it. if in the specific stage that we wait for the MCP we reset the daemon the new pod will say that it doesn't need to pause the pool and just start a drain

and we will stuck the cluster if we have some PDB blocking for more then 1 node to drain

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -580,6 +580,14 @@ func (dn *Daemon) nodeHasAnnotation(annoKey string, value string) bool {
 
 func (dn *Daemon) isNodeDraining() bool {
 	if anno, ok := dn.node.Annotations[annoKey]; ok && (anno == annoDraining || anno == annoMcpPaused) {
+		// for openshift cluster draining should be true only if the annotation has MCP paused
+		if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
+			if anno == annoMcpPaused {
+				return true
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -580,6 +580,14 @@ func (dn *Daemon) nodeHasAnnotation(annoKey string, value string) bool {
 
 func (dn *Daemon) isNodeDraining() bool {
 	if anno, ok := dn.node.Annotations[annoKey]; ok && (anno == annoDraining || anno == annoMcpPaused) {
+		// for openshift cluster draining should be true only if the annotation has MCP paused
+		if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
+			if anno == annoMcpPaused {
```
adrianchiris, line 585: I dont know how MCP Pause annot is being used so i cannot comment on the correctess of this and how it would affect all palces which call this method


but some comment about the style :) 

you can check for `annoDraining` and drop the else.

zeeke: Hi @adrianchiris, I addressed your comment about style in:
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/421

I also added some unit tests to be sure we do not forget any use cases.
Can you please refer to that PR?

PR #422, opened 2023-03-20
General cleanup in validate.go

vrindle:
Uses more expressive return types and also reduces cyclomatic complexity.

### DIFF
--- a/pkg/webhook/validate.go
+++ b/pkg/webhook/validate.go
@@ -189,20 +189,9 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
 	for _, node := range nodeList.Items {
 		if cr.Selected(&node) {
 			nodesSelected = true
-			for _, ns := range nsList.Items {
-				if ns.GetName() == node.GetName() {
-					if ok, err := validatePolicyForNodeState(cr, &ns, &node); err != nil || !ok {
-						return false, err
-					}
-				}
-			}
-			// validate current policy against policies in API (may not be converted to SriovNetworkNodeState yet)
-			for _, np := range npList.Items {
-				if np.GetName() != cr.GetName() && np.Selected(&node) {
-					if ok, err := validatePolicyForNodePolicy(cr, &np); err != nil || !ok {
-						return false, err
-					}
-				}
+			err = validatePolicyForNodeStateAndPolicy(nsList, npList, &node, cr)
+			if err != nil {
+				return false, err
 			}
 		}
 	}
@@ -217,44 +206,64 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
 	return true, nil
 }
 
-func validatePolicyForNodeState(policy *sriovnetworkv1.SriovNetworkNodePolicy, state *sriovnetworkv1.SriovNetworkNodeState, node *corev1.Node) (bool, error) {
-	glog.V(2).Infof("validatePolicyForNodeState(): validate policy %s for node %s.", policy.GetName(), state.GetName())
-	for _, iface := range state.Status.Interfaces {
-		if validateNicModel(&policy.Spec.NicSelector, &iface, node) {
-			interfaceSelected = true
-			if policy.GetName() != constants.DefaultPolicyName && policy.Spec.NumVfs == 0 {
-				return false, fmt.Errorf("numVfs(%d) in CR %s is not allowed", policy.Spec.NumVfs, policy.GetName())
-			}
-			if policy.Spec.NumVfs > iface.TotalVfs && iface.Vendor == IntelID {
-				return false, fmt.Errorf("numVfs(%d) in CR %s exceed the maximum allowed value(%d)", policy.Spec.NumVfs, policy.GetName(), iface.TotalVfs)
-			}
-			if policy.Spec.NumVfs > MlxMaxVFs && iface.Vendor == MellanoxID {
-				return false, fmt.Errorf("numVfs(%d) in CR %s exceed the maximum allowed value(%d)", policy.Spec.NumVfs, policy.GetName(), MlxMaxVFs)
-			}
-			// vdpa: only mellanox cards are supported
-			if policy.Spec.VdpaType == constants.VdpaTypeVirtio && iface.Vendor != MellanoxID {
-				return false, fmt.Errorf("vendor(%s) in CR %s not supported for virtio-vdpa", iface.Vendor, policy.GetName())
+func validatePolicyForNodeStateAndPolicy(nsList *sriovnetworkv1.SriovNetworkNodeStateList, npList *sriovnetworkv1.SriovNetworkNodePolicyList, node *corev1.Node, cr *sriovnetworkv1.SriovNetworkNodePolicy) error {
+	for _, ns := range nsList.Items {
+		if ns.GetName() == node.GetName() {
+			if err := validatePolicyForNodeState(cr, &ns, node); err != nil {
+				return err
 			}
 		}
 	}
-	return true, nil
+	// validate current policy against policies in API (may not be converted to SriovNetworkNodeState yet)
+	for _, np := range npList.Items {
+		if np.GetName() != cr.GetName() && np.Selected(node) {
+			if err := validatePolicyForNodePolicy(cr, &np); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func validatePolicyForNodeState(policy *sriovnetworkv1.SriovNetworkNodePolicy, state *sriovnetworkv1.SriovNetworkNodeState, node *corev1.Node) error {
+	glog.V(2).Infof("validatePolicyForNodeState(): validate policy %s for node %s.", policy.GetName(), state.GetName())
+	for _, iface := range state.Status.Interfaces {
+		err := validateNicModel(&policy.Spec.NicSelector, &iface, node)
+		if err == nil {
+			interfaceSelected = true
+			if policy.GetName() != constants.DefaultPolicyName && policy.Spec.NumVfs == 0 {
+				return fmt.Errorf("numVfs(%d) in CR %s is not allowed", policy.Spec.NumVfs, policy.GetName())
+			}
+			if policy.Spec.NumVfs > iface.TotalVfs && iface.Vendor == IntelID {
+				return fmt.Errorf("numVfs(%d) in CR %s exceed the maximum allowed value(%d)", policy.Spec.NumVfs, policy.GetName(), iface.TotalVfs)
+			}
+			if policy.Spec.NumVfs > MlxMaxVFs && iface.Vendor == MellanoxID {
+				return fmt.Errorf("numVfs(%d) in CR %s exceed the maximum allowed value(%d)", policy.Spec.NumVfs, policy.GetName(), MlxMaxVFs)
+			}
+			// vdpa: only mellanox cards are supported
+			if policy.Spec.VdpaType == constants.VdpaTypeVirtio && iface.Vendor != MellanoxID {
+				return fmt.Errorf("vendor(%s) in CR %s not supported for virtio-vdpa", iface.Vendor, policy.GetName())
+			}
+		}
+	}
+	return nil
 }
 
 func validatePolicyForNodePolicy(
 	current *sriovnetworkv1.SriovNetworkNodePolicy,
 	previous *sriovnetworkv1.SriovNetworkNodePolicy,
-) (bool, error) {
+) error {
 	glog.V(2).Infof("validateConflictPolicy(): validate policy %s against policy %s",
 		current.GetName(), previous.GetName())
 
 	if current.GetName() == previous.GetName() {
-		return true, nil
+		return nil
 	}
 
 	for _, curPf := range current.Spec.NicSelector.PfNames {
 		curName, curRngSt, curRngEnd, err := sriovnetworkv1.ParsePFName(curPf)
 		if err != nil {
-			return false, fmt.Errorf("invalid PF name: %s", curPf)
+			return fmt.Errorf("invalid PF name: %s", curPf)
 		}
 		for _, prePf := range previous.Spec.NicSelector.PfNames {
 			// Not validate return err of ParsePFName for previous PF
@@ -262,25 +271,25 @@ func validatePolicyForNodePolicy(
 			preName, preRngSt, preRngEnd, _ := sriovnetworkv1.ParsePFName(prePf)
 			if curName == preName {
 				if curRngEnd < preRngSt || curRngSt > preRngEnd {
-					return true, nil
+					return nil
 				} else {
-					return false, fmt.Errorf("VF index range in %s is overlapped with existing policy %s", curPf, previous.GetName())
+					return fmt.Errorf("VF index range in %s is overlapped with existing policy %s", curPf, previous.GetName())
 				}
 			}
 		}
 	}
-	return true, nil
+	return nil
 }
 
-func validateNicModel(selector *sriovnetworkv1.SriovNetworkNicSelector, iface *sriovnetworkv1.InterfaceExt, node *corev1.Node) bool {
+func validateNicModel(selector *sriovnetworkv1.SriovNetworkNicSelector, iface *sriovnetworkv1.InterfaceExt, node *corev1.Node) error {
 	if selector.Vendor != "" && selector.Vendor != iface.Vendor {
-		return false
+		return fmt.Errorf("selector vendor: %s is not equal to the interface vendor: %s", selector.Vendor, iface.Vendor)
 	}
 	if selector.DeviceID != "" && selector.DeviceID != iface.DeviceID {
-		return false
+		return fmt.Errorf("selector device ID: %s is not equal to the interface device ID: %s", selector.Vendor, iface.Vendor)
 	}
 	if len(selector.RootDevices) > 0 && !sriovnetworkv1.StringInArray(iface.PciAddress, selector.RootDevices) {
-		return false
+		return fmt.Errorf("interface PCI address: %s not found in root devices", iface.PciAddress)
 	}
 	if len(selector.PfNames) > 0 {
 		var pfNames []string
@@ -293,13 +302,13 @@ func validateNicModel(selector *sriovnetworkv1.SriovNetworkNicSelector, iface *s
 			}
 		}
 		if !sriovnetworkv1.StringInArray(iface.Name, pfNames) {
-			return false
+			return fmt.Errorf("interface name: %s not found in physical function names", iface.PciAddress)
 		}
 	}
 
 	// check the vendor/device ID to make sure only devices in supported list are allowed.
 	if sriovnetworkv1.IsSupportedModel(iface.Vendor, iface.DeviceID) {
-		return true
+		return nil
 	}
 
 	// Check the vendor and device ID of the VF only if we are on a virtual environment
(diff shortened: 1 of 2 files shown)


bn222:
As discussed, let's have `validateNicModel` return an error instead. Also, let's get rid of the side-effect way to communicate the result of functions. More precisely, we should at least address `interfaceSelected` and maybe even `nodesSelected`. 

It seems to me that we should also refactor the for loop in the function dynamicValidateSriovNetworkNodePolicy. Some parts of it are nested 5 deep.

pkg/webhook/validate.go
```diff
@@ -293,23 +300,25 @@ func validateNicModel(selector *sriovnetworkv1.SriovNetworkNicSelector, iface *s
...
 	}
 
 	// Check the vendor and device ID of the VF only if we are on a virtual environment
 	for key := range utils.PlatformMap {
 		if strings.Contains(strings.ToLower(node.Spec.ProviderID), strings.ToLower(key)) &&
 			selector.NetFilter != "" && selector.NetFilter == iface.NetFilter &&
 			sriovnetworkv1.IsVfSupportedModel(iface.Vendor, iface.DeviceID) {
-			return true
+			return nil
+		} else {
```
wizhaoredhat, line 318: This is divergent logic. The "for" loop is trying to find something matching. It doesn't mean that if it doesn't match that it is an error.

So the new behaviour of the code you have implemented here is not equivalent to the original.

pkg/webhook/validate.go
```diff
@@ -189,55 +189,60 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
...
+	return nil
+}
+
+func validatePolicyForNodeState(policy *sriovnetworkv1.SriovNetworkNodePolicy, state *sriovnetworkv1.SriovNetworkNodeState, node *corev1.Node) error {
 	glog.V(2).Infof("validatePolicyForNodeState(): validate policy %s for node %s.", policy.GetName(), state.GetName())
 	for _, iface := range state.Status.Interfaces {
-		if validateNicModel(&policy.Spec.NicSelector, &iface, node) {
-			interfaceSelected = true
+		err := validateNicModel(&policy.Spec.NicSelector, &iface, node)
+		if err != nil {
```
wizhaoredhat, line 229: Err contains important information. But it is not passed down or shown to the user.

vrindle: This can be resolved by logging and showing to the user without changing the behavior too much. Latest changes reflect that.

pkg/webhook/validate.go
```diff
@@ -189,55 +189,62 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
...
+			if err != nil {
+				return false, err
 			}
 		}
 	}
 
 	if !nodesSelected {
 		return false, fmt.Errorf("no matched node is selected by the nodeSelector in CR %s", cr.GetName())
 	}
-	if !interfaceSelected {
```
wizhaoredhat, line 213: Is `interfaceSelected` variable used? If not, then remove the variable up top.

pkg/webhook/validate.go
```diff
@@ -189,55 +189,62 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
...
-		return false, fmt.Errorf("no supported NIC is selected by the nicSelector in CR %s", cr.GetName())
-	}
 
 	return true, nil
 }
 
-func validatePolicyForNodeState(policy *sriovnetworkv1.SriovNetworkNodePolicy, state *sriovnetworkv1.SriovNetworkNodeState, node *corev1.Node) (bool, error) {
+func validatePolicyForNodeStateAndPolicy(nsList *sriovnetworkv1.SriovNetworkNodeStateList, npList *sriovnetworkv1.SriovNetworkNodePolicyList, node corev1.Node, cr *sriovnetworkv1.SriovNetworkNodePolicy) error {
```
wizhaoredhat, line 206: Shouldn't be pass node by pointer and not value? E.g. `node *corev1.Node` like in `validatePolicyForNodeState()`

pkg/webhook/validate.go
```diff
@@ -189,55 +189,62 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
...
 			}
+		} else {
+			glog.Errorf("%s", err)
 		}
 	}
-	return true, nil
+	return nil
 }
 
 func validatePolicyForNodePolicy(
```
wizhaoredhat, line 250: Shouldn't we also fix up this function as well? It looks like we always return "true, nil" or "false, error string"

pkg/webhook/validate.go
```diff
@@ -189,55 +187,62 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
...
-				return false, fmt.Errorf("numVfs(%d) in CR %s exceed the maximum allowed value(%d)", policy.Spec.NumVfs, policy.GetName(), MlxMaxVFs)
+				return fmt.Errorf("numVfs(%d) in CR %s exceed the maximum allowed value(%d)", policy.Spec.NumVfs, policy.GetName(), MlxMaxVFs)
 			}
 			// vdpa: only mellanox cards are supported
 			if policy.Spec.VdpaType == constants.VdpaTypeVirtio && iface.Vendor != MellanoxID {
-				return false, fmt.Errorf("vendor(%s) in CR %s not supported for virtio-vdpa", iface.Vendor, policy.GetName())
+				return fmt.Errorf("vendor(%s) in CR %s not supported for virtio-vdpa", iface.Vendor, policy.GetName())
 			}
+		} else {
+			glog.Errorf("%s", err)
```
wizhaoredhat, line 242: So what is the purpose of `validateNicModel`? Is it a function that we call with every interface to see if the policy matches with it? Is it really an error if we don't match?

pkg/webhook/validate.go
```diff
@@ -293,23 +298,23 @@ func validateNicModel(selector *sriovnetworkv1.SriovNetworkNicSelector, iface *s
...
-			return false
+			return fmt.Errorf("interface name: %s not found in physical function names", iface.PciAddress)
 		}
 	}
 
 	// check the vendor/device ID to make sure only devices in supported list are allowed.
 	if sriovnetworkv1.IsSupportedModel(iface.Vendor, iface.DeviceID) {
-		return true
+		return nil
 	}
```
wizhaoredhat, line 308: Isn't it also important to log that the vendor/deviceID is not supported?

pkg/webhook/validate.go
```diff
@@ -272,15 +277,15 @@ func validatePolicyForNodePolicy(
 	return true, nil
 }
 
-func validateNicModel(selector *sriovnetworkv1.SriovNetworkNicSelector, iface *sriovnetworkv1.InterfaceExt, node *corev1.Node) bool {
+func validateNicModel(selector *sriovnetworkv1.SriovNetworkNicSelector, iface *sriovnetworkv1.InterfaceExt, node *corev1.Node) error {
 	if selector.Vendor != "" && selector.Vendor != iface.Vendor {
-		return false
+		return fmt.Errorf("selector vendor: %s is not equal to the interface vendor: %s", selector.Vendor, iface.Vendor)
```
wizhaoredhat, line 282: I don't think we should be considering all these cases as errors. Rather these cases are matching criteria. For example if an Intel NIC being checked does not equal a Nvidia NIC, then it is considered verbose logging to debug an issue. To see which NICs didn't match with the policy.

pkg/webhook/validate.go
vrindle: We do log that. This returns true which means that it is supported. When it reaches it reaches the end of the function we log that vendor/device ID is not supported. https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/422/files#diff-8fd5d83d7413d77ee2eab8e5c1a2ea1623d63e2aec69017a95811db9ad9edf52R319

pkg/webhook/validate.go
```diff
@@ -189,98 +185,105 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
...
-				return false, fmt.Errorf("numVfs(%d) in CR %s exceed the maximum allowed value(%d)", policy.Spec.NumVfs, policy.GetName(), MlxMaxVFs)
+				return fmt.Errorf("numVfs(%d) in CR %s exceed the maximum allowed value(%d)", policy.Spec.NumVfs, policy.GetName(), MlxMaxVFs)
 			}
 			// vdpa: only mellanox cards are supported
 			if policy.Spec.VdpaType == constants.VdpaTypeVirtio && iface.Vendor != MellanoxID {
-				return false, fmt.Errorf("vendor(%s) in CR %s not supported for virtio-vdpa", iface.Vendor, policy.GetName())
+				return fmt.Errorf("vendor(%s) in CR %s not supported for virtio-vdpa", iface.Vendor, policy.GetName())
 			}
+		} else {
+			glog.Infof("%s", err)
```
wizhaoredhat, line 240: So the logic before was if any interface selected was set to true (i.e. validateNicModel returned true) then the  print `fmt.Errorf("no supported NIC is selected by the nicSelector in CR %s", cr.GetName())` won't be activated. However with this new logic, we will be verbosely logging every interface that doesn't match. Do you have a sample log of this that you can share? And it would be apparent.

So we need to somehow store all the debug information going through all the interfaces such that when  `validateNicModel` never returned true then we will give the user all the debug information because nothing matched. This is to improve the verbosity of `no supported NIC is selected by the nicSelector in CR %s` print and not to add verbosity overall.

@bn222  Is this your understanding as well?

pkg/webhook/validate_test.go
```diff
@@ -213,9 +213,8 @@ func TestValidatePolicyForNodeStateWithValidPolicy(t *testing.T) {
 		},
 	}
 	g := NewGomegaWithT(t)
-	ok, err := validatePolicyForNodeState(policy, state, NewNode())
```
wizhaoredhat, line 216: Test coverage is no longer complete because you have new error messages.

SchSeba:
Hi @vrindle can you please fix the commit message?

the tittle is longer then expected for a commit :)

SchSeba:
@vrindle please also add a test to cover the new case where the nic is supported but not initialize and please check the error message

pkg/webhook/validate.go
bn222: This comment should be resolved

pkg/webhook/validate.go
bn222: +1, we need separate commits in this PR. Cleanup PRs that do not change behavior + one commit that actually adds the new error message that we wanted to add in the first place.

pkg/webhook/validate.go
bn222: yes, One approach (and please do give your feedback on this) is to call an additional function that only logs things. It would log which nics it tried and why it didn't work out.

wizhaoredhat: I think I am on the same page as you.

pkg/webhook/validate_test.go
vrindle: Since we only adding errors for debugging/logging information, the test coverage that tests errors should be complete.

pkg/webhook/validate.go
bn222: @vrindle Can you separate out the commits? `git rebase -i` and `git add -p`

pkg/webhook/validate_test.go
wizhaoredhat: We should test that the debugging/logging information is what we expect based on certain parameters.

pkg/webhook/validate.go
wizhaoredhat: @vrindle Removing this print does not solve the issue. There needs to be additional logic to consolidate the information and present it only in the case when no interfaces match.

vrindle: Right I am implementing that in a second commit. This one commit is only to ensure that the general cleanup is done.

bn222: Can we rename the PR so that it's clear that we are doing a cleanup first and do all the refactoring in this first PR. The aim to reduce the size of the PR is that it can then go in. Next, we can have an additional PR after that that only adds the error message we wanted.

vrindle: That sounds good. I renamed the PR. This one only has the cleanup commit. Then there is this PR: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/434

Please ensure this PR gets merged before merging the one above since this is a stacked PR.

pkg/webhook/validate.go
```diff
@@ -189,98 +185,103 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
...
+			if err != nil {
+				return false, err
 			}
 		}
 	}
 
 	if !nodesSelected {
 		return false, fmt.Errorf("no matched node is selected by the nodeSelector in CR %s", cr.GetName())
 	}
-	if !interfaceSelected {
```
wizhaoredhat, line 213: Cleanup should only mean refactoring. You are changing the behaviour of the code here. We want this error to be printed when no interface is found.

vrindle: Okay, so I reverted this to what it was before. You can find the error when no interface is found to be in the other PR.

pkg/webhook/validate.go
```diff
@@ -170,7 +168,6 @@ func staticValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePol
 
 func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePolicy) (bool, error) {
 	nodesSelected = false
-	interfaceSelected = false
```
wizhaoredhat, line 173: This shouldn't be removed. It is still needed.

pkg/webhook/validate.go
```diff
@@ -26,10 +26,8 @@ const (
 	MlxMaxVFs  = 128
 )
 
-var (
-	nodesSelected     bool
-	interfaceSelected bool
-)
+var nodesSelected bool
+var interfaceSelected bool
```
bn222, line 30: Wondering why we still have this here. Isn't the goal of the cleanup to not pass the value of this bool through a global variable? Instead, return it from the function?

vrindle: This is general cleanup. Earlier you requested this to be split into 2 PRs so that is what I did. This one needs to merged before the other PR gets merged.

bn222: ok, sounds good then. Thanks!

pkg/webhook/validate.go
```diff
@@ -189,20 +189,9 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
...
-			// validate current policy against policies in API (may not be converted to SriovNetworkNodeState yet)
-			for _, np := range npList.Items {
-				if np.GetName() != cr.GetName() && np.Selected(&node) {
-					if ok, err := validatePolicyForNodePolicy(cr, &np); err != nil || !ok {
-						return false, err
-					}
-				}
+			err = validatePolicyForNodeStateAndPolicy(nsList, npList, &node, cr)
+			if err != nil {
+				return false, err
```
wizhaoredhat, line 194: We still return bool, err here. Should we clean this up as well?

pkg/webhook/validate_test.go
```diff
@@ -615,14 +602,11 @@ func TestValidatePolicyForNodeStateWithValidNetFilter(t *testing.T) {
...
+	err := validatePolicyForNodeState(policy, state, NewNode())
 	g.Expect(err).NotTo(HaveOccurred())
-	g.Expect(ok).To(Equal(true))
-	g.Expect(interfaceSelected).To(Equal(true))
 }
 
 func TestValidatePolicyForNodeStateWithValidVFAndNetFilter(t *testing.T) {
-	interfaceSelected = false
```
wizhaoredhat, line 625: Don't we still need this?

wizhaoredhat: You removed `interfaceSelected` from the tests when it is still used.

vrindle:
@bn222 
@wizhaoredhat 
@zshi-redhat

pkg/webhook/validate_test.go
```diff
@@ -680,8 +671,7 @@ func TestValidatePolicyForNodeStateWithValidVFAndNetFilter(t *testing.T) {
 		},
 	}
 	g := NewGomegaWithT(t)
-	ok, err := validatePolicyForNodeState(policy, state, NewNode())
+	err := validatePolicyForNodeState(policy, state, NewNode())
 	g.Expect(err).NotTo(HaveOccurred())
-	g.Expect(ok).To(Equal(true))
 	g.Expect(interfaceSelected).To(Equal(true))
 }
```
wizhaoredhat, line 677: Should we add more tests for `validateNicModel`, since now it returns an error instead of a boolean?

vrindle: what sort of tests are you looking to add? Imo, these functions test validateNicModel quite heavily.

wizhaoredhat: `validatePolicyForNodeState` gets tested. But in `err := validateNicModel(&policy.Spec.NicSelector, &iface, node)`, the `err` gets thrown away. I think it would be worth it to test `validateNicModel` indepedently.

vrindle: I agree that we should test that out but maybe not in this PR. In my other PR I  test this when the variable interfaceSelected is removed. Please see this line here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/434/files#diff-ae978883f1787164fb6fa85b31c5fde6421f8a06d9c9f5f013fb72e6f21863a0R560

I think we can leave the tests as is for now, and then we can add this test in the new PR.

wizhaoredhat: Ok

wizhaoredhat:
LGTM

bn222:
@adrianchiris 

This is a pretty small change that doesn't actually change functionality. CI is green and we already have an LGTM from RH. Can you PTAL? Should be pretty quick.

Regards,
Balazs

adrianchiris:
> This is a pretty small change that doesn't actually change functionality. CI is green and we already have an LGTM from RH. 

did not see an LGTM from maintainer, @bn222 @SchSeba you should approve :)
will take a look at this later this week.

bn222:
lgtm

pkg/webhook/validate.go
```diff
@@ -217,70 +206,90 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
...
+		}
+	}
+	return nil
+}
+
+func validatePolicyForNodeState(policy *sriovnetworkv1.SriovNetworkNodePolicy, state *sriovnetworkv1.SriovNetworkNodeState, node *corev1.Node) error {
 	glog.V(2).Infof("validatePolicyForNodeState(): validate policy %s for node %s.", policy.GetName(), state.GetName())
 	for _, iface := range state.Status.Interfaces {
-		if validateNicModel(&policy.Spec.NicSelector, &iface, node) {
+		err := validateNicModel(&policy.Spec.NicSelector, &iface, node)
```
adrianchiris, line 231: nit:

```
if err := validateNicModel; err  == nil {
  ...
}
```

that said, its really a matter of taste here so im fine with keeping as is.

adrianchiris (approved):

adrianchiris:
Thx @vrindle for this improvement !

PR #426, opened 2023-03-28
Put sriov-network-config-daemon contianer on the first place

e0ne:
Make sriov-network-config-daemon container as a default to access logs easier without specifing container name. It's more often use-case rather then checking CNIs logs.

### DIFF
--- a/bindata/manifests/daemon/daemonset.yaml
+++ b/bindata/manifests/daemon/daemonset.yaml
@@ -38,28 +38,6 @@ spec:
       {{- end }}
       {{- end }}
       containers:
-      - name: sriov-cni
-        image: {{.SRIOVCNIImage}}
-        securityContext:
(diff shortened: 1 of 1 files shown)


tariq1890:
Instead of re-ordering the container part in the manifest, you can set the annotation to specify which container to default to when executing `kubectl log/port-forward`.

[kubectl.kubernetes.io/default-container]https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-container

adrianchiris (approved):

adrianchiris:
Thanks @e0ne !

im fine with both approaches (annot as @tariq1890 mentioned or reorder containers).

current approach ultimately ends up in fewer lines of text so thats a plus :) while the other is explicit so its clear what logs we are getting by default.

PR #428, opened 2023-04-10
remove unnecessary method to create SriovOperatorConfig

tariq1890:
This method seems to be unnecessary as the SriovOperatorConfig Reconciler method is already creating the CR when not already specified.

See [here](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/controllers/sriovoperatorconfig_controller.go#L77)

### DIFF
--- a/main.go
+++ b/main.go
@@ -164,17 +164,19 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "SriovNetworkPoolConfig")
 		os.Exit(1)
 	}
+
+	ctx := ctrl.SetupSignalHandler()
 	// +kubebuilder:scaffold:builder
 
 	// Create a default SriovNetworkNodePolicy
-	err = createDefaultPolicy(kubeClient)
+	err = createDefaultPolicy(ctx, kubeClient)
 	if err != nil {
 		setupLog.Error(err, "unable to create default SriovNetworkNodePolicy")
 		os.Exit(1)
 	}
 
 	// Create default SriovOperatorConfig
-	err = createDefaultOperatorConfig(kubeClient)
+	err = createDefaultOperatorConfig(ctx, kubeClient)
 	if err != nil {
 		setupLog.Error(err, "unable to create default SriovOperatorConfig")
 		os.Exit(1)
@@ -189,9 +191,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	stopCh := ctrl.SetupSignalHandler()
 	go func() {
-		if err := mgrGlobal.Start(stopCh); err != nil {
+		if err := mgrGlobal.Start(ctx); err != nil {
 			setupLog.Error(err, "Manager Global exited non-zero")
 			os.Exit(1)
 		}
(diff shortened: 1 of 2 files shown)


zeeke:
Hi @tariq1890 , I'm not sure which piece of code is the one that creates the default `SriovOperatorConfig`. 
I think Reconcile method is not being called if the user doesn't create any resources, right? Did you try these changes in a clean installation?

tariq1890:
@zeeke Great question, and you're right! The Reconciler will only run if the `default` SriovOperatorConfig CR is created. 

What is actually unnecessary is the block of code that Creates the SriovOperatorConfig CR within the `Reconcile` method. If you look at the `Reconcile` method it is only actually designed to response to changes of the `SriovOperatorConfigs` named default and located in the configured Namespace. So it doesn't make sense to have the create logic within the `Reconcile` method

main.go
```diff
@@ -164,17 +164,19 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "SriovNetworkPoolConfig")
 		os.Exit(1)
 	}
+
+	ctx := ctrl.SetupSignalHandler()
 	// +kubebuilder:scaffold:builder
 
 	// Create a default SriovNetworkNodePolicy
-	err = createDefaultPolicy(kubeClient)
+	err = createDefaultPolicy(ctx, kubeClient)
```
tariq1890, line 172: bonus change: pass in `ctx` param :)

rollandf:
> @zeeke Great question, and you're right! The Reconciler will only run if the `default` SriovOperatorConfig CR is created.
> 
> What is actually unnecessary is the block of code that Creates the SriovOperatorConfig CR within the `Reconcile` method. If you look at the `Reconcile` method it is only actually designed to respond to changes of the `SriovOperatorConfigs` named default and located in the configured Namespace. So it doesn't make sense to have the create logic within the `Reconcile` method

Hey, what about a flow that someone delete the `default`  SriovOperatorConfig CR after the pod created it?
I think that the code you removed would have created it again.
WDYT?

SchSeba:
I was going to comment the same thing as @rollandf :)

tariq1890:
Hey all, sorry I didn't reply sooner. From the review comments and code, here's what I have learnt so far

i) The `SriovOperatorConfig` Reconciler is only aimed at reconciling against one specific CR named `default`. If this is the case, why should this even be a CR or have a reconcile loop? Shouldn't this just be a ConfigMap?

ii) Ar argument for not deleting the method, is that you want the reconcile method to re-create the `SriovOperatorConfig` resource on an unintended delete. So a permanence of the Operator config is desired here. Is a reconciler really the best solution for this? Wouldn't it just be better to have a set of environment variables to determine the config of the sriov-network-operator? This way, the config shares the lifetime of the operator. 

iii) When iterating on this PR, I've realised that e2e tests always pass. It does seem like the e2e tests should be catching something if we consider this method necessary

Overall, I do feel it's a bit iffy the way `SriovOperatorConfig` is designed

rollandf:
@tariq1890 
I think that the main thinking is about supporting dynamical changes to the operator configuration.
If this is the case, having a configmap will need also code to watch changes, and environment variables changes requires restart of the Pod.

I assume that e2e tests are not covering 100% of the code flows. Adding more tests is always welcomed.

SchSeba:
Hi @tariq1890 thanks for the comment!

btw another reason to have this as a CRD and not a config map is the validation we can do on it and the documentation we can add in the CRD description instead of just text the is super prune to errors from the user

tariq1890:
Thanks @rollandf @SchSeba !

I will close the PR

PR #432, opened 2023-04-24
Remove imagePullPolicy alltogether and use k8s defaults

Longchuanzheng:
Change sriov-device-plugin DaemonSet ImagePullPolicy to IfNotPresent
  
I find sriov-device-plugin DaemonSet ImagePullPolicy doesn't set, while imagePullPolicy for sriov-network-operator is `IfNotPresent`.  So it makes sense to set sriov-device-plugin DaemonSet ImagePullPolicy as `IfNotPresent`.


rollandf (approved):
LGTM

SchSeba:
+1 @adrianchiris

PR #435, opened 2023-05-08
Allow creating policies that ignore NUMA topology

zeeke:
This PR allows users to create policies that set the sriov-network-device plugin `ExcludeTopology`  field in the advertised resource list.

The feature is helpful in scenarios where the kubelet is configured with `--topology-manager-policy single-numa-node` [2] and some workloads may run with a NIC that is not aligned to the CPU/Memory NUMA node.

[1] https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/master/pkg/types/types.go#L101
[2] https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/#policy-single-numa-node

### DIFF
--- a/api/v1/sriovnetworknodepolicy_types.go
+++ b/api/v1/sriovnetworknodepolicy_types.go
@@ -57,6 +57,8 @@ type SriovNetworkNodePolicySpec struct {
 	// +kubebuilder:validation:Enum=virtio
 	// VDPA device type. Allowed value "virtio"
 	VdpaType string `json:"vdpaType,omitempty"`
+	// Exclude device's NUMA node when advertising this resource by SRIOV network device plugin. Default to false.
+	ExcludeTopology bool `json:"excludeTopology,omitempty"`
 }
 
 type SriovNetworkNicSelector struct {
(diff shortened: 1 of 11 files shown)


api/v1/sriovnetworknodepolicy_types.go
```diff
@@ -57,6 +57,8 @@ type SriovNetworkNodePolicySpec struct {
 	// +kubebuilder:validation:Enum=virtio
 	// VDPA device type. Allowed value "virtio"
 	VdpaType string `json:"vdpaType,omitempty"`
+	// Exclude device's NUMA node when advertising this resource. Default to false.
```
adrianchiris, line 60: nit:  Exclude device's NUMA node when advertising this resource by sriov network device plugin

adrianchiris (commented):
minor nit otherwise lgtm

api/v1/sriovnetworknodepolicy_types.go
zeeke: Looks good, thanks

SchSeba (approved):

adrianchiris (approved):

PR #437, opened 2023-05-09
Add NVIDIA BlueField 3 to supported NICs

e0ne:


SchSeba:
Hi :)
can you please also update the doc with the supported nic and the capability?
from the code looks like it support legacy DPDK and HWoffload right?

e0ne:
> Hi :) can you please also update the doc with the supported nic and the capability? from the code looks like it support legacy DPDK and HWoffload right?

AFAIK, BF3 should contain all features from BF2

adrianchiris (approved):

adrianchiris:
LGTM

PR #440, opened 2023-05-15
Add a costume junit reporter to the conformance suite

liornoy:
Currently, the conformance test can be invoked with ginkgo's --junit-report
flag and produce a default junit report.
    
This commit adds a non-default junit report to the conformance suite
and configures it with specific configurations to produce
clean and easy-to-read report.
    
Added a "junit" flag, that when is not empty, will
produce the junit report in the given path, with the
report name "junit_sriov_conformance.xml"


adrianchiris (approved):
LGTM

zeeke (approved):
LGTM
Thanks @liornoy

PR #441, opened 2023-05-17
Update base images to use CentOS 9 Stream

e0ne:


adrianchiris:
mstflint updated finally in centos 9 :) . LGTM!

adrianchiris (approved):

PR #444, opened 2023-05-23
[Helm] Dont force deployment on control nodes

adrianchiris:
use preferred scheduling instead.

This is desired for a better user experience
on managed k8s clusters where control plane nodes
are not part of the cluster

### DIFF
--- a/deployment/sriov-network-operator/values.yaml
+++ b/deployment/sriov-network-operator/values.yaml
@@ -9,13 +9,16 @@ operator:
   nodeSelector: {}
   affinity:
     nodeAffinity:
-      requiredDuringSchedulingIgnoredDuringExecution:
-        nodeSelectorTerms:
-          - matchExpressions:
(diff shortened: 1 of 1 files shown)


adrianchiris:
@e0ne @Eoghan1232  PTAL, this one is straight forward

ykulazhenkov (approved):

Eoghan1232 (approved):
lgtm.

question: do you want them to have the same weight value assigned? this would mean the scheduler would check both at the same time, and choose whichever one is best to deploy on. 
should one be a higher priority?

adrianchiris:
>question: do you want them to have the same weight value assigned? this would mean the scheduler would check both at the same time, and choose whichever one is best to deploy on.
should one be a higher priority?


in this case its fine to have both with same weight. generally in a cluster you would either have one set for control plane nodes or both.

latest deployments with kubeadm just set the control-plane label. 
i kept the "master" label for backward compat.

PR #446, opened 2023-05-25
small changes in e2e test to run on single node

SchSeba:
on single node the drain is disable so on the last test we must clean the pods.

we don't use AfterEach because it's not possible to run a reporter on that.


liornoy:
are the changes to the timeout (from 1 to 3 and from 3 to 5) required due to re-occurring flakes?

SchSeba:
@liornoy right SNO have high CPU load and stuff may take a bit more time then expected

liornoy:
cool. lgtm

zeeke (approved):
LGTM

PR #448, opened 2023-05-29
Update the k8s reporter

liornoy:
In the validation and conformance test suites,
we update the k8sreporter to the latest version
as an external library. since all of the logic
for the k8s reported is now placed under one repository.

also, we omit the "-dump" flag which is not supported and add the "-reporter" flag, which if not empty
it specifies where to output the reports to.


SchSeba:
do we still need this?

liornoy:
> do we still need this?

yes. soon we'll need the report flag to invoke the k8sreporter and I'll revive this pr

PR #449, opened 2023-05-30
Run mstfwreset when applying config changes

SalDaniele:
As of the most recent update to the Bluefield2 firmware (v24.37.1300)
configuration changes no longer are applied on node reboot without
running mstfwreset().

This can result in a node ending up in a bootloop when applying a new
configuration.

i.e. a sriov-workload-node-policy if applied to updated total vfs. This
change is applied via mstconfig, however the change is not reflected in
NUM_OF_VFS without running mstfwreset(). This resulted in a repeated
call to reboot.

```
    I0523 20:59:44.059046    7496 mellanox_plugin.go:335] Changing TotalVfs 16 to 12, needs reboot
    I0523 20:59:44.059057    7496 mellanox_plugin.go:172] mellanox-plugin needDrain true needReboot true
```
    
Signed-off-by: Salvatore Daniele <sdaniele@redhat.com>

### DIFF
--- a/pkg/plugins/mellanox/mellanox_plugin.go
+++ b/pkg/plugins/mellanox/mellanox_plugin.go
@@ -180,7 +180,13 @@ func (p *MellanoxPlugin) Apply() error {
 		return nil
 	}
 	glog.Info("mellanox-plugin Apply()")
-	return configFW()
+	if err := configFW(); err != nil {
+		return err
+	}
+	if err := resetFW(); err != nil {
+		return err
+	}
+	return nil
 }
 
 func configFW() error {
@@ -209,6 +215,21 @@ func configFW() error {
 		_, err := utils.RunCommand("mstconfig", cmdArgs...)
 		if err != nil {
 			glog.Errorf("mellanox-plugin configFW(): failed : %v", err)
+			return err
+		}
+	}
+	return nil
+}
+
+func resetFW() error {
+	glog.Info("mellanox-plugin resetFW()")
+	for pciAddr := range attributesToChange {
+		cmdArgs := []string{"-d", pciAddr, "-y", "reset"}
+
+		glog.V(2).Infof("mellanox-plugin: resetFW(): %v", cmdArgs)
+		_, err := utils.RunCommand("mstfwreset", cmdArgs...)
+		if err != nil {
+			glog.Errorf("mellanox-plugin resetFW(): failed : %v", err)
 			return err
 		}
 	}
--- a/Dockerfile.sriov-network-config-daemon
+++ b/Dockerfile.sriov-network-config-daemon
@@ -5,7 +5,7 @@ RUN make _build-sriov-network-config-daemon BIN_PATH=build/_output/cmd
 
 FROM quay.io/centos/centos:stream9
 ARG MSTFLINT=mstflint
-RUN ARCH_DEP_PKGS=$(if [ "$(uname -m)" != "s390x" ]; then echo -n ${MSTFLINT} ; fi) && yum -y install hwdata $ARCH_DEP_PKGS && yum clean all
+RUN ARCH_DEP_PKGS=$(if [ "$(uname -m)" != "s390x" ]; then echo -n ${MSTFLINT} ; fi) && yum -y install hwdata pciutils $ARCH_DEP_PKGS && yum clean all
 LABEL io.k8s.display-name="sriov-network-config-daemon" \
       io.k8s.description="This is a daemon that manage and config sriov network devices in Kubernetes cluster"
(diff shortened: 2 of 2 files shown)


wizhaoredhat:
LGTM

wizhaoredhat:
@SalDaniele  Could you also describe the bootloop with the config-daemon we were seeing with the latest BF2 firmware?

adrianchiris:
we are currently not calling mstfwreset in sriov-network-config-daemon, why is this change needed ?

wizhaoredhat:
@adrianchiris Although we aren't using mstfwreset here. The mstflint package as a whole does have pciutils as a dependency. We should add this dependency in.

SalDaniele:
@adrianchiris I just added an additional commit to address the crux of the issue we are encountering. As of the most recent Bluefields2 firmware update (5-18-23) config changes are no longer applied on next boot as they have been previously. This requires us to call mstfwreset manually. Without doing so, we can end up in a boot loop, where a change is applied by the config daemon, the node reboots, the change is not reflected, etc.

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -215,6 +221,20 @@ func configFW() error {
 	return nil
 }
 
+func resetFW() error {
+	glog.Info("mellanox-plugin resetFW()")
+	for pciAddr := range attributesToChange {
+		cmdArgs := []string{"-d", pciAddr, "-y", "reset"}
+
+		_, err := utils.RunCommand("mstfwreset", cmdArgs...)
```
wizhaoredhat, line 229: This will print the stdout of mstfwreset right? If not we may want something like `glog.V(2).Infof("mellanox-plugin: configFW(): %v", cmdArgs)` from above.

SalDaniele: Yes utils.RunCommand will print the stdout of mstfwreset, but I also added this additional log consistent with configFW().

wizhaoredhat:
LGTM
@adrianchiris  PTAL

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -180,7 +180,13 @@ func (p *MellanoxPlugin) Apply() error {
 		return nil
 	}
 	glog.Info("mellanox-plugin Apply()")
-	return configFW()
+	if err := configFW(); err != nil {
+		return err
+	}
+	if err := resetFW(); err != nil {
```
adrianchiris, line 186: im not too keen on always running mstfwreset, perhaps only for DPUs in Embedded mode ?

i need to think about it a bit more

adrianchiris: also for DPU you need to run with --reset-sync=1 flag.

see:
https://docs.nvidia.com/networking/pages/viewpage.action?pageId=52009175

with current implementation, does it work for u ?

SalDaniele: Aahh good to know. I was testing the current implementation on a cluster w/ the BF in NIC mode, which was working. The issue I was seeing was not specific to running in DPU embedded mode, the node would get stuck in a bootloop if I applied a policy to change the number of vfs, since after rebooting the updated configuration would not be applied, triggering another reboot.

adrianchiris: So, you encountered the issue with DPU in NIC mode as well ?

SalDaniele: Yes

SalDaniele: My understanding since submitting this patch is that this is not intended behavior by Nvidia and a firmware patch will come in July to address this issue.

bn222: The issue is that the behavior of NIC mode changed too and is broken in a similar to DPU. With DPU mode --sync 1 causes the DPU to hang. I have code that adds `--sync 1` but I want it to work correctly before pushing the changes.

adrianchiris: > My understanding since submitting this patch is that this is not intended behavior by Nvidia and a firmware patch will come in July to address this issue.

that is correct, its a bug and Nvidia should fix the firmware.

@SalDaniele in that case, i think this PR is no longer required and can be closed ?

SalDaniele: Yes provided this fix is in the next firmware release, we can close this PR

adrianchiris:
PR is not needed, see discussion above

PR #451, opened 2023-06-01
conformance-test: ExcludeTopology field

zeeke:
Add an End2End to validate two different
SriovNetworkNodePolicies can't specify different
`ExcludeTopology` values for the same resource.

Needs:
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/435

### DIFF
--- a/test/conformance/tests/test_sriov_operator.go
+++ b/test/conformance/tests/test_sriov_operator.go
@@ -1368,6 +1368,109 @@ var _ = Describe("[sriov] operator", func() {
 					Expect(stdout).To(ContainSubstring("2 packets transmitted, 2 received, 0% packet loss"))
 				})
 			})
+
+			Context("ExcludeTopology", func() {
+
+				var excludeTopologyTrueResourceXXX, excludeTopologyFalseResourceXXX, excludeTopologyFalseResourceYYY *sriovv1.SriovNetworkNodePolicy
+				var node string
+				var intf *sriovv1.InterfaceExt
+
+				BeforeEach(func() {
+					if discovery.Enabled() {
+						Skip("Test unsuitable to be run in discovery mode")
+					}
+
+					node = sriovInfos.Nodes[0]
+					sriovDeviceList, err := sriovInfos.FindSriovDevices(node)
+					Expect(err).ToNot(HaveOccurred())
+					unusedSriovDevices, err := findUnusedSriovDevices(node, sriovDeviceList)
+					Expect(err).ToNot(HaveOccurred())
+
+					intf = unusedSriovDevices[0]
+
+					excludeTopologyTrueResourceXXX = &sriovv1.SriovNetworkNodePolicy{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      "test-exclude-topology-true-res-xxx",
+							Namespace: operatorNamespace,
+						},
+
+						Spec: sriovv1.SriovNetworkNodePolicySpec{
+							NumVfs:       10,
+							ResourceName: "resourceXXX",
+							NodeSelector: map[string]string{"kubernetes.io/hostname": node},
+							NicSelector: sriovv1.SriovNetworkNicSelector{
+								PfNames: []string{intf.Name + "#0-4"},
+							},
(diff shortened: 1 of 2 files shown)


SchSeba:
This looks great!
can you please add a test to validate that the config-map we create for the device plugin takes this change into account?

zeeke:
> This looks great! can you please add a test to validate that the config-map we create for the device plugin takes this change into account?

I added a test case. I'm not 100% happy with asserting the configuration by matching strings, but the alternative (like deserializing the ResourceConfig) would be more complicated. Do you have opinion on that?

test/conformance/tests/test_sriov_operator.go
```diff
@@ -1368,6 +1368,138 @@ var _ = Describe("[sriov] operator", func() {
...
+				    kind: ConfigMap
+				    metadata:
+				      creationTimestamp: "2023-05-29T09:45:33Z"
+				      name: device-plugin-config
+				      namespace: openshift-sriov-network-operator
+				      resourceVersion: "10437605"
+				      uid: 74dd4f4c-4924-4ea0-bf21-7dd68d7fb8f7
+
+				*/
+				FIt("field is forwarded to the device plugin configuration", func() {
```
SchSeba, line 1392: please remove the `F`

test/conformance/tests/test_sriov_operator.go
```diff
@@ -1368,6 +1368,138 @@ var _ = Describe("[sriov] operator", func() {
...
+							},
+							ExcludeTopology: false,
+						},
+					}
+
+					err = clients.Create(context.Background(), excludeTopologyFalse)
+					Expect(err).ToNot(HaveOccurred())
+
+					assertDevicePluginConfigurationContains(node,
+						`{"resourceName":"resourceXXX","excludeTopology":true,"selectors":{"pfNames":["`+intf.Name+`#0-4"],"IsRdma":false,"NeedVhostNet":false},"SelectorObj":null}`)
```
SchSeba, line 1444: nit: can you use fmt.sprint I think is a bit cleaner to find the variables

test/conformance/tests/test_sriov_operator.go
```diff
@@ -1368,6 +1368,138 @@ var _ = Describe("[sriov] operator", func() {
...
 				})
 			})
+
+			Context("ExcludeTopology", func() {
+				BeforeEach(func() {
+					if discovery.Enabled() {
+						Skip("Test unsuitable to be run in discovery mode")
+					}
+				})
+				/*
```
SchSeba, line 1378: I think you can remove this

SchSeba (commented):
the test looks good.

just small comments

zeeke:
@SchSeba @Eoghan1232 @adrianchiris Can you please take a look ?

test/conformance/tests/test_sriov_operator.go
```diff
@@ -1368,6 +1368,125 @@ var _ = Describe("[sriov] operator", func() {
...
+				It("field is forwarded to the device plugin configuration", func() {
+					node := sriovInfos.Nodes[0]
+					sriovDeviceList, err := sriovInfos.FindSriovDevices(node)
+					Expect(err).ToNot(HaveOccurred())
+					unusedSriovDevices, err := findUnusedSriovDevices(node, sriovDeviceList)
+					Expect(err).ToNot(HaveOccurred())
+
+					intf := unusedSriovDevices[0]
+
+					excludeTopologyTrue := &sriovv1.SriovNetworkNodePolicy{
```
adrianchiris, line 1388: `excludeTopologyTrue`, `excludeTopologyFalse` can be moved to common place under Context ?

i see they both defined the same in both test cases.

adrianchiris:
Overall looks good, one small comment and we can merge IMO

test/conformance/tests/test_sriov_operator.go
zeeke: I moved the declaration under Context. BTW, assignments need the node name and the interface name, so I had to put them under BeforeEach. WDYT?

PR #454, opened 2023-06-06
Add AllMuticast mode support

mlguerrero12:
The sriov cni has a new option to enable allmulticast mode on the VF. These changes allows users to enable this mode via the SriovNetwork object.

### DIFF
--- a/pkg/webhook/validate.go
+++ b/pkg/webhook/validate.go
@@ -48,6 +48,19 @@ func validateSriovOperatorConfig(cr *sriovnetworkv1.SriovOperatorConfig, operati
 	return false, warnings, fmt.Errorf("only default SriovOperatorConfig is used")
 }
 
+func validateSriovNetwork(cr *sriovnetworkv1.SriovNetwork, operation v1.Operation) (bool, []string, error) {
+	glog.V(2).Infof("validateSriovNetwork: %v", cr)
+	var warnings []string
+
+	if operation == v1.Create || operation == v1.Update {
+		if cr.Spec.AllMulticast == sriovnetworkv1.SriovCniStateOn && cr.Spec.Trust != sriovnetworkv1.SriovCniStateOn {
+			return false, warnings, fmt.Errorf("trust must be enabled in order to set allMulticast on")
+		}
+	}
+
+	return true, warnings, nil
+}
+
 func validateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePolicy, operation v1.Operation) (bool, []string, error) {
 	glog.V(2).Infof("validateSriovNetworkNodePolicy: %v", cr)
 	var warnings []string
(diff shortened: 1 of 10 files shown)


zeeke:
LGTM

thanks @mlguerrero12

cgoncalves:
Would it make sense and be possible to add a validation to the webhook that checks that `trust` is enabled as a prerequisite?

SchSeba:
The PR looks good please implement a validation webhook for it

mlguerrero12:
@cgoncalves, @SchSeba, I implemented the validation as requested. Please have a look

cgoncalves (approved):
LGTM

pkg/webhook/validate.go
```diff
@@ -48,6 +48,19 @@ func validateSriovOperatorConfig(cr *sriovnetworkv1.SriovOperatorConfig, operati
 	return false, warnings, fmt.Errorf("only default SriovOperatorConfig is used")
 }
 
+func validateSriovNetwork(cr *sriovnetworkv1.SriovNetwork, operation v1.Operation) (bool, []string, error) {
+	glog.V(2).Infof("validateSriovNetwork: %v", cr)
+	var warnings []string
+
+	if operation == v1.Create || operation == v1.Update {
```
SchSeba, line 55: why we need this? we should only get called by this two operations base on the validationWebhook configuration no?

pkg/webhook/validate_test.go
```diff
@@ -164,6 +164,40 @@ func TestValidateSriovOperatorConfigWithDefaultOperatorConfig(t *testing.T) {
...
+	g := NewGomegaWithT(t)
+	ok, _, err = validateSriovNetwork(network, "CREATE")
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(ok).To(Equal(true))
+
+	ok, _, err = validateSriovNetwork(network, "UPDATE")
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(ok).To(Equal(true))
+
+	network.Spec.Trust = "off"
```
SchSeba, line 190: nit: please have a separate test for the off so it's cleaner thanks :)

SchSeba (commented):
Thanks for implementing the validation.

just two small comments and we should be good to go I think

pkg/webhook/validate.go
mlguerrero12: Correct, this improves readability so that people don't have to check the ValidatingWebhookConfiguration object to understand why only update and create are affected. It also prevents future mistakes when other validations are added.

if this bothers you, I could remove it. But I think it's good to have it.

pkg/webhook/validate_test.go
mlguerrero12: sure, thanks

mlguerrero12: done

PR #421, opened 2023-03-16
improve the isdraning check for openshift

zeeke:
here is a case where we annotate the node as draning because we took the leader but on OCP the machine config pool was doing some work so we wait for it. if in the specific stage that we wait for the MCP we reset the daemon the new pod will say that it doesn't need to pause the pool and just start a drain

and we will stuck the cluster if we have some PDB blocking for more then 1 node to drain

PR #455, opened 2023-06-11
Fix race condition in leader election for OCP

SchSeba:
In OCP we have two steps to prepare a node before we drain it. first we get leader election and annotate the node with Draining and then we pause the MCP and mark the node as MCP paused.

if the config_daemon get a reset between the fist part to the second it will get stuck because one node will take the leader election BUT it will not mark the node as Draining as there is another node already draining. and the node with the draining label will try to get the drain lock again but the first node has it.

with this change if the node as Draning or MCP pause label it will not try to take the lock again and just continue after the reset.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -507,18 +507,22 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 				go dn.getDrainLock(ctx, done)
 				<-done
 			}
+		}
 
-			if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
-				glog.Infof("nodeStateSyncHandler(): pause MCP")
-				if err := dn.pauseMCP(); err != nil {
-					return err
-				}
+		if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
+			glog.Infof("nodeStateSyncHandler(): pause MCP")
+			if err := dn.pauseMCP(); err != nil {
+				return err
 			}
 		}
 
-		glog.Info("nodeStateSyncHandler(): drain node")
-		if err := dn.drainNode(); err != nil {
-			return err
+		if dn.disableDrain {
+			glog.Info("nodeStateSyncHandler(): disable drain is true skipping drain")
+		} else {
+			glog.Info("nodeStateSyncHandler(): drain node")
+			if err := dn.drainNode(); err != nil {
+				return err
+			}
 		}
 	}
 
@@ -578,15 +582,12 @@ func (dn *Daemon) nodeHasAnnotation(annoKey string, value string) bool {
 	return false
 }
 
+// isNodeDraining: check if the node is draining
+// both Draining and MCP paused labels will return true
 func (dn *Daemon) isNodeDraining() bool {
 	anno, ok := dn.node.Annotations[annoKey]
 	if !ok {
 		return false
-	}
-
-	if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
-		// for openshift cluster draining should be true only if the annotation has MCP paused
-		return anno == annoMcpPaused
 	}
 
 	return anno == annoDraining || anno == annoMcpPaused
@@ -895,11 +896,6 @@ func (dn *Daemon) pauseMCP() error {
 }
 
 func (dn *Daemon) drainNode() error {
-	if dn.disableDrain {
-		glog.Info("drainNode(): disable drain is true skipping drain")
-		return nil
-	}
-
 	glog.Info("drainNode(): Update prepared")
 	var err error
 
--- a/pkg/daemon/daemon_test.go
+++ b/pkg/daemon/daemon_test.go
@@ -275,7 +275,7 @@ SUBSYSTEM=="net", ACTION=="add|move", ATTRS{phys_switch_id}!="", ATTR{phys_port_
 			Expect(sut.isNodeDraining()).To(BeFalse())
 
 			sut.node.Annotations["sriovnetwork.openshift.io/state"] = "Draining"
-			Expect(sut.isNodeDraining()).To(BeFalse())
+			Expect(sut.isNodeDraining()).To(BeTrue())
 
 			sut.node.Annotations["sriovnetwork.openshift.io/state"] = "Draining_MCP_Paused"
 			Expect(sut.isNodeDraining()).To(BeTrue())


pkg/daemon/daemon.go
```diff
@@ -507,18 +507,22 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
 				<-done
 			}
+		}
 
-			if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
-				glog.Infof("nodeStateSyncHandler(): pause MCP")
-				if err := dn.pauseMCP(); err != nil {
-					return err
-				}
+		if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() && !dn.isMcpPaused() {
```
adrianchiris, line 512: why do we need this change ? looking at pauseMCP it should handle an already paused MCP object (which is paused by  config daemon based on node annot)


btw, in MCPPause if we manage to pause MCP but fail to update annotation i think we will deadlock.
here: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/109c2c25df7ac335bfd7a76ef3ee43ef69c95c82/pkg/daemon/daemon.go#L843

pkg/daemon/daemon.go
```diff
@@ -507,18 +507,22 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+			glog.Infof("nodeStateSyncHandler(): pause MCP")
+			if err := dn.pauseMCP(); err != nil {
+				return err
 			}
 		}
 
-		glog.Info("nodeStateSyncHandler(): drain node")
-		if err := dn.drainNode(); err != nil {
-			return err
+		if dn.disableDrain {
```
adrianchiris, line 519: disableDrain is already checked in drainNode()

adrianchiris:
>if the config_daemon get a reset between the fist part to the second it will get stuck because one node will take the leader election BUT it will not mark the node as Draining as there is another node already draining. and the node with the draining label will try to get the drain lock again but the first node has it.


marking the node as draining (via node annotation) is done during leader election.
what may happen is that another node got leader elected and has outdated caches then it would also mark its own node as draining.

pkg/daemon/daemon.go
```diff
@@ -599,7 +610,7 @@ func (dn *Daemon) completeDrain() error {
 		}
 	}
 
-	if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
+	if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() && dn.isMcpPaused() {
```
adrianchiris, line 613: i dont think you can reach in openshift a state where you need to complete the drain and mcp is not paused.
because  u will only drain after calling PauseMCP

pkg/daemon/daemon.go
adrianchiris: ah i see that u moved the check here.

pkg/daemon/daemon.go
adrianchiris: regarding PauseMCP doing early return in case node drain annot is mcp_paused: 
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/109c2c25df7ac335bfd7a76ef3ee43ef69c95c82/pkg/daemon/daemon.go#L839

pkg/daemon/daemon.go
```diff
@@ -578,18 +582,25 @@ func (dn *Daemon) nodeHasAnnotation(annoKey string, value string) bool {
...
 
+// isNodeDraining: check if the node is draining
+// both Draining and MCP paused labels will return true
 func (dn *Daemon) isNodeDraining() bool {
 	anno, ok := dn.node.Annotations[annoKey]
 	if !ok {
 		return false
 	}
 
-	if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
```
adrianchiris, line 587: TBH, i think thats the only change needed to solve the race cond issue.

pkg/daemon/daemon.go
SchSeba: I move it because it makes it more clear I always ask my self everytime I look into this code how it works with disable drain and then I find the if inside the function.

I think we should call the function after we check if drain is disabled :P

pkg/daemon/daemon.go
SchSeba: it's just a safe check here :)

because the completeDrain gets called `if dn.isNodeDraining()` that on OCP can be both `Draining` or `MCP_PAUSE` labels so I just wanted to add another safe point

pkg/daemon/daemon.go
SchSeba: if we do that we will return the previews race condition from here [https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/421](https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/421/files#diff-a53b7b593d3d778e62eaeeafa40088656f9212bfa2c2b7991df15fa78e60b0f0R587)

pkg/daemon/daemon.go
adrianchiris: do you think its needed ? is there a scenario in mind where this might happen ?

pkg/daemon/daemon.go
adrianchiris: ack

pkg/daemon/daemon.go
adrianchiris: ack

pkg/daemon/daemon.go
adrianchiris: here is what i think happens

node-1 : 
1. get lock
2. annotate node as draning
3. try to pause mcp
4. reboot (it did not pause mcp), hence it did not set Draining_MCP_Paused

node-2:
1. tries to get lock <- stuck as there is not draning, but it holds the lock

node-1:
1. config daemon comes back up
2. isNodeDraining returns true
3. calls PauseMCP()
4. PauseMCP() now completes successfully
5. config daemon continues with configuration
6. eventually completes drain, unpause MCP and set node draining annot to Idle

node-2:
1. continues its work


what did i get wrong ?


where i think a deadlock may happen:

node-1 : 
1. get lock
2. annotate node as draning
3. try to pause mcp
4. patch mcp object setting pause to true
5. reboot (it did not set Draining_MCP_Paused node annot)

node-2:
1. tries to get lock <- stuck as there is not draning, but it holds the lock

node-1:
1. config daemon comes back up
2. isNodeDraining returns true
3. calls PauseMCP()
4. stuck as we thing someone else paused MCP obj

SchSeba:
Hi @adrianchiris please take another look when you have time :)

PR #456, opened 2023-06-12
Don't start config daemon on unsupported platforms

EmilienM:
The operator is only supported on Baremetal and OpenStack.
We should not start the config daemon elsewhere, as this is not
something we have tested.

On platform-agnostic deployments, there are a few unknowns that could
potentially be problematic especially for when running in virtualized
environments outside of the OpenStack platform type.

### DIFF
--- a/cmd/sriov-network-config-daemon/start.go
+++ b/cmd/sriov-network-config-daemon/start.go
@@ -162,6 +162,10 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	} else {
 		glog.Fatalf("Failed to fetch node state %s, %v!", startOpts.nodeName, err)
 	}
+	// Return an error if the platform is not Baremetal or OpenStack
+	if platformType != utils.Baremetal && platformType != utils.VirtualOpenStack {
+		glog.Fatalf("Unsupported platform type %s!", platformType.String())
+	}
 	glog.V(0).Infof("Running on platform: %s", platformType.String())
 
 	var namespace = os.Getenv("NAMESPACE")


SchSeba:
/hold
please wait with this PR

SchSeba:
This is not always the case for example I deploy a kubernetes cluster with kubeadm on a baremetal node and it doesn't contain the providerID but everything still work as expected.
With this PR the operator will only run on platform that got deployed via supported provider like OpenStack or Metal3.

Instead I think the deamon should not do anything on the system if there are no policies on it. for example not try to configure Udev rules.

e0ne:
> This is not always the case for example I deploy a kubernetes cluster with kubeadm on a baremetal node and it doesn't contain the providerID but everything still work as expected. With this PR the operator will only run on platform that got deployed via supported provider like OpenStack or Metal3.
> 
> Instead I think the deamon should not do anything on the system if there are no policies on it. for example not try to configure Udev rules.

I agree with @SchSeba. This patch breaks current support of vanilla k8s

cmd/sriov-network-config-daemon/start.go
```diff
@@ -162,6 +162,10 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	} else {
 		glog.Fatalf("Failed to fetch node state %s, %v!", startOpts.nodeName, err)
 	}
+	// Return an error if the platform is not Baremetal or OpenStack
+	if platformType != utils.Baremetal && platformType != utils.VirtualOpenStack {
```
e0ne, line 166: This should be re-visited, because Baremetal is our defaults and this statement will be always True

e0ne (changes requested):

PR #457, opened 2023-06-12
Set Open_vSwitch hardware offload in config daemon

SalDaniele:
Previously, we have set ovs hardware offload via the service applied by the Machine Config Operator
(bindata/manifests/switchdev-config/ovs-units.ovs-vswitchd.service).

However, the issue with this approach is that after the user has removed the MachineConfigPool or removed SRIOV all together, ovs hw-offload remains enabled.

This change will instead handle configuring ovs hardware offloading in the config daemon by checking for switchdev devices, and enabling ovs hardware offload only while such devices are detected. We will remove the serviced file responsible for setting ovs-hwol to true on boot. Upon detecting a change in the sriov node policy, we will check for any switchdev devices and set ovs-hwol accordingly. Changes to ovs will result in drain and reboot of the node to ensure changes are applied

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -491,6 +491,7 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 			}
 		}
 	}
+
 	if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
 		if err = dn.getNodeMachinePool(); err != nil {
 			return err
@@ -558,6 +559,11 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 				return err
 			}
 		}
+	}
+	// After reboot, set ovs HWOL dependent on whether there are switchdev mode devices
+	if err := trySetOvsHWOL(utils.SwitchdevDeviceExists(latestState)); err != nil {
+		glog.Errorf("nodeStateSyncHandler(): Failed to set OpenvSwitch offload error: %v", err)
+		return err
 	}
 	glog.Info("nodeStateSyncHandler(): sync succeeded")
 	dn.nodeState = latestState.DeepCopy()
@@ -1077,5 +1083,40 @@ func tryCreateNMUdevRule() error {
 		glog.Errorf("tryCreateNMUdevRule(): fail to write file: %v", err)
 		return err
 	}
+	return nil
+}
+
+func trySetOvsHWOL(switchdevDevExists bool) error {
+	glog.V(2).Infof("trySetOvsHWOL()")
+	desiredHWOLStateStr := strconv.FormatBool(switchdevDevExists)
+
+	exit, err := utils.Chroot("/host")
+	if err != nil {
+		glog.Errorf("trySetOvsHWOL(): failed to chroot to host: %v", err)
+	} else {
+		defer exit()
+	}
+
+	currentState := utils.OvsHWOLisEnabled()
+	// if we are already in desired state do nothing. HWOL should only be set if a switchdev Device exists
+	if currentState == switchdevDevExists {
+		glog.V(2).Infof("trySetOvsHWOL(): ovs hw-offload already in desired state %s, ovs switch not needed", desiredHWOLStateStr)
+		return nil
+	}
+
+	// otherwise switch state and reboot to apply the config changes to ovs
+	glog.V(2).Infof("trySetOvsHWOL(): switching Open_vSwitch hw-offload from %s to %s", currentState, desiredHWOLStateStr)
+	_, err = utils.RunOVSvsctl("set", "Open_vSwitch", ".", fmt.Sprintf("other_config:hw-offload=%s", desiredHWOLStateStr))
+	if err != nil {
+		glog.Errorf("trySetOvsHWOL(): ovs-vsctl set OpenvSwitch offload %s: failed: %v", desiredHWOLStateStr, err)
+		return err
+	}
+
+	err = utils.RestartOVS()
+	if err != nil {
+		glog.Errorf("SD DEBUG failed to restart ovs")
+		return err
+	}
+
 	return nil
 }
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -788,6 +788,22 @@ func RunCommand(command string, args ...string) (string, error) {
 	return stdout.String(), err
 }
 
+func RunOVSvsctl(args ...string) (string, error) {
+	glog.Infof("runOVSvsctl(): ovs-vsctl %v", args)
+	var output bytes.Buffer
+	cmd := exec.Command("ovs-vsctl", args...)
+	cmd.Stdout = &output
+	err := cmd.Run()
+
+	return output.String(), err
+}
+
+func RestartOVS() error {
+	glog.Infof("RestartOVS(): restarting openvswitch.service")
+	cmd := exec.Command("systemctl", "restart", "openvswitch.service")
+	return cmd.Run()
+}
+
 func hasMellanoxInterfacesInSpec(newState *sriovnetworkv1.SriovNetworkNodeState) bool {
 	for _, ifaceStatus := range newState.Status.Interfaces {
 		if ifaceStatus.Vendor == VendorMellanox {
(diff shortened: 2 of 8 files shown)


pkg/daemon/daemon.go
```diff
@@ -491,6 +491,15 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 			}
 		}
 	}
+
+	d, r, err := trySetOvsHWOL(utils.SwitchdevDeviceExists(latestState))
```
bn222, line 495: iiuc, we are adding 1 more reboot when we enable switchdev hw offload since that code path also uses this and in that case, we already have to have an early dev-link command on each boot. Is my understanding correct? In other words, we pay an extra reboot at the time we enable it to gain a way to get rid of it when we disable it. 

Thinking along the same lines, would it even make sense to have ovs set dev-link to switchdev mode? I'm not sure we want that, but then we could go back to one reboot.

pkg/daemon/daemon.go
```diff
@@ -1079,3 +1088,42 @@ func tryCreateNMUdevRule() error {
...
+	// if we are already in desired state, do nothing
+	cmd := exec.Command("ovs-vsctl", "get", "Open_vSwitch", ".", "other_config:hw-offload")
+	cmd.Stdout = &output
+	if err := cmd.Run(); err != nil {
+		// if ovs has not previously been configured this record may not yet exist
+		glog.Errorf("trySetOvsHWOL(): failed to get current Open_vSwitch hw-offload mode : %v", err)
+		currentState = "false"
+	} else {
+		currentState = output.String()
+	}
```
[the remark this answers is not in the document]
bn222: Would also mean we don't need to convert the bool into a string. It makes a bit more sense to compare bool with bool instead of string that should look like a bool with a string that was created from a bool.

pkg/daemon/daemon.go
```diff
@@ -1079,3 +1088,42 @@ func tryCreateNMUdevRule() error {
...
+
+	if strings.Contains(currentState, desiredState) {
+		glog.V(2).Infof("trySetOvsHWOL(): ovs already in desired state %s, ovs switch not needed", desiredState)
+		return false, false, nil
+	}
+
+	// otherwise switch state and reboot to apply the config changes to ovs
+	glog.V(2).Infof("trySetOvsHWOL(): switching Open_vSwitch hw-offload from %s to %s", currentState, desiredState)
+	cmd = exec.Command("ovs-vsctl", "set", "Open_vSwitch", ".", fmt.Sprintf("other_config:hw-offload=%s", desiredState))
```
bn222, line 1123: I'm starting to think we should have a function that wraps `exec.Command("ovs-vsctl", ...)`. I leave it up to you if you think we benefit from not having to copy paste `exec.Command("ovs-vsctl",` every time we want to call ovs-vsctl

pkg/daemon/daemon.go
```diff
@@ -1079,3 +1088,42 @@ func tryCreateNMUdevRule() error {
 	}
 	return nil
 }
+
+func trySetOvsHWOL(newState bool) (bool, bool, error) {
```
bn222, line 1092: newState is too vague. Can we come up with a bbetter name?

pkg/daemon/daemon.go
wizhaoredhat: In that note. If we make the hw-offload config in OVS ephemeral then we can avoid this all together. Having an ephemeral config has it's pros and cons.

SalDaniele: Yes, we are currently paying an extra reboot when the daemon detects a device has been put into switchdev mode.

wizhaoredhat: One potential issue is that pods might become Ready with a VF while HWOL is turned off. I don't think there will be network connectivity for those pods.

pkg/daemon/daemon.go
SalDaniele: Updated to "switchdevDevExists"

pkg/daemon/daemon.go
SalDaniele: I agree, added functions to wrap running ovs-vsctl and restarting ovs.

pkg/daemon/daemon.go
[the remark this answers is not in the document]
SalDaniele: Added "func OvsHWOLisEnabled()"

pkg/daemon/daemon.go
SalDaniele: See the most recent update, we can instead manually restart ovs from the config daemon to avoid the extra restart.

With the new behavior, we will trigger configuring ovs-hwol after a node reboots due to a change in the node state policy. We will detect if the node has booted with or without switchdev devices and set ovs-hwol accordingly

bn222: I'm ok with a restart of ovs. In theory, it should be possible to dynamically take care of this as well to be honest.

pkg/daemon/daemon.go
```diff
@@ -491,6 +491,7 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 			}
 		}
 	}
+
```
e0ne, line 494: this change is not related

pkg/daemon/daemon.go
```diff
@@ -1079,3 +1085,38 @@ func tryCreateNMUdevRule() error {
...
+	desiredHWOLStateStr := strconv.FormatBool(switchdevDevExists)
+
+	exit, err := utils.Chroot("/host")
+	if err != nil {
+		glog.Errorf("trySetOvsHWOL(): failed to chroot to host: %v", err)
+	} else {
+		defer exit()
+	}
+
+	currentState := utils.OvsHWOLisEnabled()
```
e0ne, line 1100: please rename currentState -> hwOffloadEnabled

pkg/utils/utils.go
```diff
@@ -788,6 +788,22 @@ func RunCommand(command string, args ...string) (string, error) {
...
+	cmd := exec.Command("ovs-vsctl", args...)
+	cmd.Stdout = &output
+	err := cmd.Run()
+
+	return output.String(), err
+}
+
+func RestartOVS() error {
+	glog.Infof("RestartOVS(): restarting openvswitch.service")
+	cmd := exec.Command("systemctl", "restart", "openvswitch.service")
```
e0ne, line 803: did you test it ith Ubuntu 22.04? AFAIK, service name in Ubuntu is different

e0ne (changes requested):
That's good improvement but the PR needs to be updated. Please, add unit-tests too

pkg/utils/utils.go
bn222: Actually, this is a good point. Let's also use the systemd package in go to start services

bn222: Actually, we have to cordon berfore and uncordon after like 803.

SchSeba:
Hi @SalDaniele please rebase this PR.

also I have some questions here
1. is the hw configuration on ovs permanent or it needs to be done after every reboot?
2. does it needs to be done BEFORE you have workloads in the system?
3. this will only work if there is the ovs binary commands on the host

@Eoghan1232 be aware this change can impact the `/host` mount reduction you are working on

SalDaniele:
> Hi @SalDaniele please rebase this PR.
> 
> also I have some questions here
> 
> 1. is the hw configuration on ovs permanent or it needs to be done after every reboot?

Whatever we configure ovs-hwol to be will persist after each reboot. IIUC a reboot or restart is needed to apply the configuration once we set it with `ovs-vsctl set Open_vSwitch ... hw-offload=true`

> 2. does it needs to be done BEFORE you have workloads in the system?

This is a good point I will need to look into this. When we set it with the systemd file I believe that would mean ovs would always be configured before the workloads. I was triggering a node reboot after setting ovs-vsctl but this resulted in additional reboots when creating / removing sriov policies, so I opted to instead attempt to restart ovs. I wonder if cordoning before and after the restart as Balazs had mentioned would solve this.

> 3. this will only work if there is the ovs binary commands on the host
> 

Yes, I have some pending changes I will push w/ the rebase to put this behind a check `if utils.ClusterType == utils.ClusterTypeOpenshift`.

> @Eoghan1232 be aware this change can impact the `/host` mount reduction you are working on

bn222:
regarding '2.' we need a cordone/uncordone before/after we restart ovs. This will only happen the first time we enable it. Once we have it enabled, the "hwol == true" will be persistent across reboots. Next time, when ovs starts up, it will already be with hwol enabled and the whole cordone/uncordone dance will be skipped.

The main benefit of this patch is that we get rid of another file we put on the system. This lgtm.

SchSeba:
@SalDaniele can we close this one there is a huge work on enable switchdev in the sriov operator

PR #461, opened 2023-06-20
Add sriovoperatorconfigs resource to webhook

mlguerrero12:
The sriovoperatorconfigs resource was not present in the validating webhook configuration. The logic to validate this object was not triggered.

PR #464, opened 2023-06-28
Validate disable drain

zeeke:
This PR introduces a webhook validation for the `SriovOperatorConfig` to avoid disabling the drain while
the config-daemon is updating a node. Disabling the drain at this stage would prevent the operator to uncordon a node at the end of the update operation, keeping nodes un-schedulable until manual intervention.

Introduce a small refactor of `validateSriovOperatorConfig(...)`

this PR needs
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/461

cc @mlguerrero12


pkg/webhook/validate.go
```diff
@@ -35,17 +37,60 @@ func validateSriovOperatorConfig(cr *sriovnetworkv1.SriovOperatorConfig, operati
...
+		return false, warnings, err
+	}
+
+	return true, warnings, nil
+}
+
+// validateSriovOperatorConfigDisableDrainWhileUpdating checks if the user is setting `.Spec.DisableDrain` from false to true while
+// operator is updating one or more nodes. Disabling the drain at this stage would prevent the operator to uncordon a node at
+// the end of the update operation, keeping nodes un-schedulable until manual intervention.
+func validateSriovOperatorConfigDisableDrainWhileUpdating(cr *sriovnetworkv1.SriovOperatorConfig) error {
```
adrianchiris, line 63: nit: shorten it to: validateSriovOperatorConfigDisableDrain ?

adrianchiris (commented):
This will technically not solve the issue, but narrow down the when it might happen.

minor nit otherwise lgtm.

pkg/webhook/validate.go
zeeke: sure

adrianchiris (approved):

PR #466, opened 2023-07-10
Use `e2e-test` tag for images built in e2e CI

abdallahyas:
Currently for the kind CI, Kind deploys Kubernetes version v1.27.3, in this version the `ImagePullpolicy` of containers follow the procedures in [1], this breaks the CI, since the CI uses the `latest` tag when building the images, and Kubernetes tries to pull the images always instead of using the built image.

This patch changes the tag of the image built by CI to `e2e-test`, so that K8S will set the image policy to be `IfNotPresent`.

[1] https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting

### DIFF
--- a/hack/run-e2e-test-kind.sh
+++ b/hack/run-e2e-test-kind.sh
@@ -2,8 +2,8 @@
 set -eo pipefail
 here="$(dirname "$(readlink --canonicalize "${BASH_SOURCE[0]}")")"
 root="$(readlink --canonicalize "$here/..")"
-export SRIOV_NETWORK_OPERATOR_IMAGE="${SRIOV_NETWORK_OPERATOR_IMAGE:-sriov-network-operator:latest}"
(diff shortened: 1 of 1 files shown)


abdallahyas:
@e0ne @adrianchiris Can you guys PTAL?
Thanks

e0ne (approved):
It seems to be a workaround for a CI. From other side I'm OK to have a specific image tag for e2e tests

adrianchiris:
as the images for e2e tests are built by `run-e2e-test-kind.sh` im fine with having a specific tag for  it.

the alternative is to explicitly add image pull policy when running make deploy.

adrianchiris (approved):
lgtm

rollandf:
LGTM
Seems that this issue was introduced by #432

zeeke (approved):
LGTM

PR #467, opened 2023-07-10
vhost-vdpa support

lmilleri:
This PR implements the changes for vhost-vdpa support.

- SriovNetworkNodePolicy API: VDPA device type can take now the following values: "virtio" and "vhost"
- if vhost is specified, the vhost-vdpa driver is loaded into the kernel and bound to the vdpa device

### DIFF
--- a/bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
+++ b/bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
@@ -58,8 +58,8 @@ contents:
           do
             extract_min_max_ids
             vdpaType=$(jq -c '.vdpaType' -r <<< $group)
-            if [ $vfid -le $maxId ] && [ $vfid -ge $minId ] && [ $vdpaType == "virtio" ]; then
-              vdpa_cmd="vdpa dev add name vdpa:"${VfPciAddr}" mgmtdev pci/"${VfPciAddr}
+            if [ $vfid -le $maxId ] && [ $vfid -ge $minId ] && [ $vdpaType != "" ]; then
+              vdpa_cmd="vdpa dev add name vdpa:"${VfPciAddr}" mgmtdev pci/"${VfPciAddr}" max_vqp 32"
               eval $vdpa_cmd
             fi
           done
--- a/bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
+++ b/bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
@@ -58,9 +58,17 @@ contents:
           do
             extract_min_max_ids
             vdpaType=$(jq -c '.vdpaType' -r <<< $group)
-            if [ $vfid -le $maxId ] && [ $vfid -ge $minId ] && [ $vdpaType != "" ]; then
-              vdpa_cmd="vdpa dev add name vdpa:"${VfPciAddr}" mgmtdev pci/"${VfPciAddr}" max_vqp 32"
+            if [ $vfid -le $maxId ] && [ $vfid -ge $minId ] && ([ $vdpaType == "virtio" ] || [ $vdpaType == "vhost" ]); then
+              vdpa_name=vdpa:${VfPciAddr}
+              vdpa_cmd="vdpa dev add name "${vdpa_name}" mgmtdev pci/"${VfPciAddr}" max_vqp 32"
               eval $vdpa_cmd
+              # set the driver_override. When the specified driver will be loaded in the kernel
+              # it will be automatically bound to the vdpa device
+              driver_override=virtio_vdpa
+              if [ $vdpaType == "vhost" ]; then
+                driver_override=vhost_vdpa
+              fi
+              echo $driver_override > /sys/bus/vdpa/devices/$vdpa_name/driver_override
             fi
           done
         done
--- a/pkg/plugins/generic/generic_plugin.go
+++ b/pkg/plugins/generic/generic_plugin.go
@@ -19,34 +19,67 @@ import (
 
 var PluginName = "generic_plugin"
 
+// driver id
+const (
+	Vfio = iota
+	VirtioVdpa
+)
+
+// driver name
+const (
+	vfioPciDriver    = "vfio_pci"
+	virtioVdpaDriver = "virtio_vdpa"
+)
+
+// function type for determining if a given driver has to be loaded in the kernel
+type needDriver func(state *sriovnetworkv1.SriovNetworkNodeState, driverState *DriverState) bool
+
+type DriverState struct {
+	DriverName     string
+	DeviceType     string
+	VdpaType       string
+	NeedDriverFunc needDriver
+	DriverLoaded   bool
+}
+
+type DriverStateMapType map[uint]*DriverState
+
 type GenericPlugin struct {
-	PluginName           string
-	SpecVersion          string
-	DesireState          *sriovnetworkv1.SriovNetworkNodeState
-	LastState            *sriovnetworkv1.SriovNetworkNodeState
-	LoadVfioDriver       uint
-	LoadVirtioVdpaDriver uint
-	RunningOnHost        bool
-	HostManager          host.HostManagerInterface
+	PluginName     string
+	SpecVersion    string
+	DesireState    *sriovnetworkv1.SriovNetworkNodeState
+	LastState      *sriovnetworkv1.SriovNetworkNodeState
+	DriverStateMap DriverStateMapType
+	RunningOnHost  bool
+	HostManager    host.HostManagerInterface
 }
 
 const scriptsPath = "bindata/scripts/enable-kargs.sh"
 
-const (
-	unloaded = iota
-	loading
-	loaded
-)
-
 // Initialize our plugin and set up initial values
 func NewGenericPlugin(runningOnHost bool) (plugin.VendorPlugin, error) {
+	driverStateMap := make(map[uint]*DriverState)
+	driverStateMap[Vfio] = &DriverState{
+		DriverName:     vfioPciDriver,
+		DeviceType:     constants.DeviceTypeVfioPci,
+		VdpaType:       "",
+		NeedDriverFunc: needDriverCheckDeviceType,
+		DriverLoaded:   false,
+	}
+	driverStateMap[VirtioVdpa] = &DriverState{
+		DriverName:     virtioVdpaDriver,
+		DeviceType:     constants.DeviceTypeNetDevice,
+		VdpaType:       constants.VdpaTypeVirtio,
+		NeedDriverFunc: needDriverCheckVdpaType,
+		DriverLoaded:   false,
+	}
+
 	return &GenericPlugin{
-		PluginName:           PluginName,
-		SpecVersion:          "1.0",
-		LoadVfioDriver:       unloaded,
-		LoadVirtioVdpaDriver: unloaded,
-		RunningOnHost:        runningOnHost,
-		HostManager:          host.NewHostManager(runningOnHost),
+		PluginName:     PluginName,
+		SpecVersion:    "1.0",
+		DriverStateMap: driverStateMap,
+		RunningOnHost:  runningOnHost,
+		HostManager:    host.NewHostManager(runningOnHost),
 	}, nil
 }
 
@@ -60,7 +93,7 @@ func (p *GenericPlugin) Spec() string {
 	return p.SpecVersion
 }
 
-// OnNodeStateChange Invoked when SriovNetworkNodeState CR is created or updated, return if need dain and/or reboot node
+// OnNodeStateChange Invoked when SriovNetworkNodeState CR is created or updated, return if need drain and/or reboot node
 func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
 	glog.Info("generic-plugin OnNodeStateChange()")
 	needDrain = false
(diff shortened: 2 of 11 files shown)


lmilleri:
@bn222 @SchSeba @adrianchiris @zshi-redhat Please review and approve if looks good

pkg/plugins/generic/generic_plugin.go
```diff
@@ -69,7 +71,7 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
 	p.DesireState = new
 
 	needDrain = needDrainNode(new.Spec.Interfaces, new.Status.Interfaces)
-	needReboot = needRebootNode(new, &p.LoadVfioDriver, &p.LoadVirtioVdpaDriver)
+	needReboot = needRebootNode(new, &p.LoadVfioDriver, &p.LoadVirtioVdpaDriver, &p.LoadVhostVdpaDriver)
```
bn222, line 74: This now starts to look cumborsome. Can we start this PR with a refactor of p.LoadVfioDriver and vdpa driver? What's happening here is that the needRebootNode actually returns a bunch of things through it's arguments. That function is now doing much more than "just checking if we need to reboot".

We also have an enum and a small state machine to load drivers. This could be simplified much further.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -69,31 +109,31 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
 	p.DesireState = new
 
 	needDrain = needDrainNode(new.Spec.Interfaces, new.Status.Interfaces)
-	needReboot = needRebootNode(new, &p.LoadVfioDriver, &p.LoadVirtioVdpaDriver)
+	needReboot = needRebootNode(new, p.DriverStateMap)
```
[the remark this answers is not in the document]
bn222: Just want to add that what you did is definitely an improvement so please don't undo that :)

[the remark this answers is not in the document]
bn222: Initially, the state is `unloaded`, then `needRebootNode` get's called which sets the state to `loading` if needed. Next, `Apply()` is called, which results in calling `p.HostManager.LoadKernelModule` and immediately setting to `loaded`. 

Without asking for any change, I'm just wondering if we can rid of this intermediate state? Why do have a tri-state thing. Can't we just have load + loaded without loading?

It seems to me that loading means "not loaded but need to load. Tried it an failed". Maybe the "need to load" part doesn't need to be cached to simplify the code?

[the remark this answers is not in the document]
lmilleri: Maybe I can write a reduced version of needRebootNode that doesn't set the loading state at all, but just calls the tryEnableIommuInKernelArgs function in case vfio needs to be loaded, this is for knowing if the node has to be rebooted or not. This is to maintain compatibility with the existing code in daemon.go (drain and reboot flags are handled at least in two places, daemon.go and generic_plugin.go).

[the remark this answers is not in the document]
bn222: That makes more sense to me. Ideally, we would want two functions, one to figure out if a reboot is needed, and then a separate one that actually sets the kargs. Currently, the "optimization" of setting the kargs and returning if it actually needs a reboot in the process makes the code unnecessary complex.

`nodeRebootNeeded` and `ensureKargsSet` could be the two functions. If we have those two functions, do we still need 3 states?

I realize you just wanted to add vhost-vdpa support and now we are refactoring the code, but if we just keep adding stuff, the code gets worse over time. Let's put in an active effort to improve the code before we make changes.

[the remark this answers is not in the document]
lmilleri: Removed the `loading` state. Hope it is what you meant. Please review again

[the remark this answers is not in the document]
bn222: yes! this looks better to me. I'd like the input of others too of course.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -69,31 +111,41 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
...
 	}
 	return
 }
 
+// ////////////// for testing purposes only ///////////////////////
+func (p *GenericPlugin) GetDriverStateMap() DriverStateMapType {
+	return p.DriverStateMap
+}
+
+func (p *GenericPlugin) TestDriverLoading(state *sriovnetworkv1.SriovNetworkNodeState) {
```
adrianchiris, line 127: Nit: LoadDriverForTests ?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -69,31 +111,41 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
...
-			return err
-		}
-		p.LoadVfioDriver = loaded
-	}
 
-	if p.LoadVirtioVdpaDriver == loading {
-		if err := p.HostManager.LoadKernelModule("virtio_vdpa"); err != nil {
-			glog.Errorf("generic-plugin Apply(): fail to load virtio_vdpa kmod: %v", err)
-			return err
+	for _, driverState := range p.DriverStateMap {
```
adrianchiris, line 141: can you move this to a helper method: `SyncDriverState()` and call it after L157 while at it ?

flow would be clearer IMO

pkg/plugins/generic/generic_plugin.go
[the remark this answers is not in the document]
adrianchiris: IMO removing this tri-state is fine. 

one thing that bothers me, is that a function that just suppose to check for reboot actually does things to the system (modify kargs, writes config files etc...) however thats not related to this PR.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -217,35 +269,35 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
 	return
 }
 
-func needRebootNode(state *sriovnetworkv1.SriovNetworkNodeState, loadVfioDriver *uint, loadVirtioVdpaDriver *uint) (needReboot bool) {
-	needReboot = false
-	if *loadVfioDriver != loaded {
-		if needVfioDriver(state) {
-			*loadVfioDriver = loading
-			update, err := tryEnableIommuInKernelArgs()
+func nodeToRebootIfVfio(state *sriovnetworkv1.SriovNetworkNodeState, driverMap DriverStateMapType) (updateNode bool) {
```
adrianchiris, line 272: nodeToRebootIfVfio ? maybe needRebootIfVfio ?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -217,35 +269,35 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
...
+	updateNode = false
+	for driverID, driverState := range driverMap {
```
adrianchiris, line 274: driverMap is a .... map.

no need to iterate over it. instead just index it by `Vfio`

pkg/plugins/generic/generic_plugin.go
adrianchiris: also lets use "needReboot" as return val as thats what it means.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -217,35 +269,35 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
...
 }
 
-func needRebootNode(state *sriovnetworkv1.SriovNetworkNodeState, loadVfioDriver *uint, loadVirtioVdpaDriver *uint) (needReboot bool) {
-	needReboot = false
-	if *loadVfioDriver != loaded {
-		if needVfioDriver(state) {
-			*loadVfioDriver = loading
-			update, err := tryEnableIommuInKernelArgs()
+func nodeToRebootIfVfio(state *sriovnetworkv1.SriovNetworkNodeState, driverMap DriverStateMapType) (updateNode bool) {
+	updateNode = false
```
adrianchiris, line 273: this is by default false. so can be removed.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -217,35 +269,35 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
...
+	return updateNode
+}
 
-	if *loadVirtioVdpaDriver != loaded {
-		if needVirtioVdpaDriver(state) {
-			*loadVirtioVdpaDriver = loading
-		}
-	}
+func needRebootNode(state *sriovnetworkv1.SriovNetworkNodeState, driverMap DriverStateMapType) (needReboot bool) {
+	needReboot = false
```
adrianchiris, line 290: this is by default false. so can be removed.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -69,31 +111,41 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
...
-	needReboot = needRebootNode(new, &p.LoadVfioDriver, &p.LoadVirtioVdpaDriver)
+	needReboot = needRebootNode(new, p.DriverStateMap)
 
 	if needReboot {
 		needDrain = true
 	}
 	return
 }
 
+// ////////////// for testing purposes only ///////////////////////
```
adrianchiris, line 122: move those to the end of the file ?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -69,32 +111,30 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
...
 }
 
-// Apply config change
-func (p *GenericPlugin) Apply() error {
-	glog.Infof("generic-plugin Apply(): desiredState=%v", p.DesireState.Spec)
-	if p.LoadVfioDriver == loading {
-		if err := p.HostManager.LoadKernelModule("vfio_pci"); err != nil {
-			glog.Errorf("generic-plugin Apply(): fail to load vfio_pci kmod: %v", err)
-			return err
+func syncDriverState(p *GenericPlugin) error {
```
SchSeba, line 122: please have this function as part of the struct no need to pass it

pkg/plugins/generic/generic_plugin.go
```diff
@@ -104,6 +144,11 @@ func (p *GenericPlugin) Apply() error {
 		}
 	}
 
+	err := syncDriverState(p)
```
SchSeba, line 147: please just simply this like

```golang
if err := p.syncDrvierStates(); err != nil {
 return err
}
```

bindata/manifests/switchdev-config/files/switchdev-configuration-after-nm.sh.yaml
```diff
@@ -58,8 +58,8 @@ contents:
           do
             extract_min_max_ids
             vdpaType=$(jq -c '.vdpaType' -r <<< $group)
-            if [ $vfid -le $maxId ] && [ $vfid -ge $minId ] && [ $vdpaType == "virtio" ]; then
-              vdpa_cmd="vdpa dev add name vdpa:"${VfPciAddr}" mgmtdev pci/"${VfPciAddr}
+            if [ $vfid -le $maxId ] && [ $vfid -ge $minId ] && [ $vdpaType != "" ]; then
+              vdpa_cmd="vdpa dev add name vdpa:"${VfPciAddr}" mgmtdev pci/"${VfPciAddr}" max_vqp 32"
```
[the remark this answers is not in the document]
lmilleri: Yes, you're right, this will be configurable in the future by extending the API? For the time being, the idea is to enable the multiqueue support by providing by default 32 queues. We think it is the best value, considering that the virtio-net driver will use the number of cores as a maximum. The reason for enabling multiqueue is to increase the performances of the card and be ready to demonstrate it to Verizon

pkg/plugins/generic/generic_plugin.go
```diff
@@ -69,32 +111,31 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
...
-func (p *GenericPlugin) Apply() error {
-	glog.Infof("generic-plugin Apply(): desiredState=%v", p.DesireState.Spec)
-	if p.LoadVfioDriver == loading {
-		if err := p.HostManager.LoadKernelModule("vfio_pci"); err != nil {
-			glog.Errorf("generic-plugin Apply(): fail to load vfio_pci kmod: %v", err)
-			return err
+func (p *GenericPlugin) syncDriverState() error {
+	for _, driverState := range p.DriverStateMap {
+		if !driverState.DriverLoaded && driverState.NeedDriverFunc(p.DesireState, driverState) {
+			glog.Infof("loading driver %s", driverState.DriverName)
```
adrianchiris, line 125: nit: use glog.V(2).Infof  as @SchSeba suggested below.

adrianchiris (approved):
lgtm from my side.

one small comment RE log verbosity

adrianchiris:
thanks @lmilleri !

adrianchiris:
@SchSeba @zeeke shall we proceeed with merging this one ?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -217,35 +262,46 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
...
+	if updateNode {
 		glog.V(2).Infof("generic-plugin needRebootNode(): need reboot for updating switchdev device configuration")
 	}
-	needReboot = needReboot || update
+	needReboot = needReboot || updateNode
 	return
 }
+
+// ////////////// for testing purposes only ///////////////////////
+func (p *GenericPlugin) GetDriverStateMap() DriverStateMapType {
```
zeeke, line 295: nit: Can you make these two functions package-private?

```
func (p *GenericPlugin) getDriverStateMap() DriverStateMapType {

func (p *GenericPlugin) loadDriverForTests(state *sriovnetworkv1.SriovNetworkNodeState) {
```

zeeke (approved):
Only one nit comment

Overall, LGTM and good testing coverage. Thanks

pkg/plugins/generic/generic_plugin.go
lmilleri: Had to change the test package as well. Thanks!

PR #468, opened 2023-07-11
Reduce config-daemon logging verbosity

zeeke:
This PR reduces the verbosity of the config-daemon loop when the user sets a `SriovOperatorConfig.Spec.LogLevel < 2`.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
@@ -720,10 +721,16 @@ func generateRandomGUID() net.HardwareAddr {
 
 func GetNicSriovMode(pciAddress string) (string, error) {
 	glog.V(2).Infof("GetNicSriovMode(): device %s", pciAddress)
+
 	devLink, err := netlink.DevLinkGetDeviceByName("pci", pciAddress)
 	if err != nil {
+		if errors.Is(err, syscall.ENODEV) {
+			// the device doesn't support devlink
+			return "", nil
+		}
 		return "", err
 	}
+
 	return devLink.Attrs.Eswitch.Mode, nil
 }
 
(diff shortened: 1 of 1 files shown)


liornoy:
lgtm

cgoncalves (commented):
LGTM

pkg/utils/utils.go
```diff
@@ -720,13 +720,42 @@ func generateRandomGUID() net.HardwareAddr {
...
+	supported, err := isDevLinkSupportedBy(pciAddress)
+	if err != nil {
+		return "", err
+	}
+
+	if !supported {
+		return "", nil
+	}
+
 	devLink, err := netlink.DevLinkGetDeviceByName("pci", pciAddress)
```
adrianchiris, line 733: cant we check if device is found or not error ? 

you should get something like:

```
~# devlink dev show pci/0000:04:00.0
kernel answers: No such device
```

zeeke: Considering the actual warning log message:
```
W0703 12:28:47.813124   58755 utils.go:128] DiscoverSriovDevices(): unable to get device mode 0000:3b:00.0 "no such device"
```

If I understood correctly, the check you are suggesting should be:

```
func GetNicSriovMode(pciAddress string) (string, error) {
	glog.V(2).Infof("GetNicSriovMode(): device %s", pciAddress)
	devLink, err := netlink.DevLinkGetDeviceByName("pci", pciAddress)
	if err != nil {
		if err.Error() == "no such device" {
			return "", nil
		}
		return "", err
	}
	return devLink.Attrs.Eswitch.Mode, nil
}
```

(note the N is not capitalized)

I find it a bit weak, but I recognize the value of efficiency. does it looks good to you?

adrianchiris: the error should be a syscall.Errno value
(netlink read/writes from/to unix socket)

you can check it
```
if err != nil {
  if errors.Is(err, syscall.ENODEV) {
    // do something
  }
  // do something else
}
```

SchSeba (approved):
LGTM

thanks for the PR!

Eoghan1232 (approved):

adrianchiris (approved):

PR #469, opened 2023-07-11
Add `NUMANode` to node state

zeeke:
Add and populate the field
`SriovNetworkNodeState.Status.Interfaces[].NUMANode` so that the user knows which NUMA node every device is installed in.

### DIFF
--- a/api/v1/sriovnetworknodestate_types.go
+++ b/api/v1/sriovnetworknodestate_types.go
@@ -65,6 +65,7 @@ type InterfaceExt struct {
 	LinkType    string            `json:"linkType,omitempty"`
 	EswitchMode string            `json:"eSwitchMode,omitempty"`
 	TotalVfs    int               `json:"totalvfs,omitempty"`
+	NUMANode    int               `json:"numaNode,omitempty"`
 	VFs         []VirtualFunction `json:"Vfs,omitempty"`
 }
 type InterfaceExts []InterfaceExt
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -125,6 +125,7 @@ func DiscoverSriovDevices(withUnsupported bool) ([]sriovnetworkv1.InterfaceExt, 
 		if dputils.IsSriovPF(device.Address) {
 			iface.TotalVfs = dputils.GetSriovVFcapacity(device.Address)
 			iface.NumVfs = dputils.GetVFconfigured(device.Address)
+			iface.NUMANode = dputils.GetDevNode(device.Address)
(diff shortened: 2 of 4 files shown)


pkg/utils/utils.go
```diff
@@ -125,6 +125,7 @@ func DiscoverSriovDevices(withUnsupported bool) ([]sriovnetworkv1.InterfaceExt,
 		if dputils.IsSriovPF(device.Address) {
 			iface.TotalVfs = dputils.GetSriovVFcapacity(device.Address)
 			iface.NumVfs = dputils.GetVFconfigured(device.Address)
+			iface.NUMANode = dputils.GetDevNode(device.Address)
```
e0ne, line 128: Do we need to set NUMANode attribute even it's not specified or error occured https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/b3aea760a5e2d92414d73f9195c731a532282dad/pkg/utils/utils.go#L211?

e0ne (changes requested):

pkg/utils/utils.go
bn222: -1 means that it failed to get it, even though it might still be on another numa node then 0. This just exposes whatever the underlying lib reports. I'm not entirely sure why we are exposing this info.

api/v1/sriovnetworknodestate_types.go
```diff
@@ -65,6 +65,7 @@ type InterfaceExt struct {
 	LinkType    string            `json:"linkType,omitempty"`
 	EswitchMode string            `json:"eSwitchMode,omitempty"`
 	TotalVfs    int               `json:"totalvfs,omitempty"`
+	NUMANode    int               `json:"numaNode,omitempty"`
```
adrianchiris, line 68: Why do we need to add this ? what is the use case ?

just for general information for the user, is not good enough IMO. a user may want to know alot of things about the node. perhaps this info should be exposed by node-feature-discovery ?

bn222: if the info is exposed by nfd, the user still needs to map the info exposed by node state to the nfd though. Which might be ok, but it's important to realize this imo.

zeeke: Yes, my idea was to make it available for general information to the user, as the other NIC related properties (PCI, vendor, deviceID). It would also make testing NUMA-related features a bit easier, but I understand it can be too little for an API change.

zeeke:
As discussed during the last community meeting, SriovNetworkNodeState is intended to be used as an internal API, in which information fields are read and written by different operator components. 
Closing

PR #472, opened 2023-07-12
Re-introduce the check if the number of VFs is 0

SchSeba:
now that the issue with kubelet with fixed all the way to k8s 1.25 we can add again the check if the number of existing vfs is 0 there is no need to drain the node because there are no workloads running with vfs from that nic

### DIFF
--- a/pkg/plugins/generic/generic_plugin.go
+++ b/pkg/plugins/generic/generic_plugin.go
@@ -195,15 +195,17 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
 		configured := false
 		for _, iface := range desired {
 			if iface.PciAddress == ifaceStatus.PciAddress {
-				// TODO: no need to perform further checks if ifaceStatus.NumVfs equals to 0
-				// once https://github.com/kubernetes/kubernetes/issues/109595 will be fixed
 				configured = true
+				if ifaceStatus.NumVfs == 0 {
+					glog.V(2).Infof("generic-plugin needDrainNode(): no need drain, for PCI address %s current NumVfs is 0", iface.PciAddress)
+					break
+				}
 				if utils.NeedUpdate(&iface, &ifaceStatus) {
-					glog.V(2).Infof("generic-plugin needDrainNode(): need drain, PF %s request update", iface.PciAddress)
+					glog.V(2).Infof("generic-plugin needDrainNode(): need drain, for PCI address %s request update", iface.PciAddress)
 					needDrain = true
 					return
 				}
-				glog.V(2).Infof("generic-plugin needDrainNode(): no need drain, expect NumVfs %v, current NumVfs %v", iface.NumVfs, ifaceStatus.NumVfs)
+				glog.V(2).Infof("generic-plugin needDrainNode(): no need drain,for PCI address %s expect NumVfs %v, current NumVfs %v", iface.PciAddress, iface.NumVfs, ifaceStatus.NumVfs)
 			}
 		}
 		if !configured && ifaceStatus.NumVfs > 0 {


SchSeba:
cc @adrianchiris @e0ne @zeeke @Eoghan1232 please take a look :)

pkg/plugins/generic/generic_plugin.go
```diff
@@ -194,16 +194,18 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
...
+					if utils.NeedUpdate(&iface, &ifaceStatus) {
+						glog.V(2).Infof("generic-plugin needDrainNode(): need drain, PF %s request update", iface.PciAddress)
+						needDrain = true
+						return
+					}
+					glog.V(2).Infof("generic-plugin needDrainNode(): no need drain, expect NumVfs %v, current NumVfs %v", iface.NumVfs, ifaceStatus.NumVfs)
 				}
-				glog.V(2).Infof("generic-plugin needDrainNode(): no need drain, expect NumVfs %v, current NumVfs %v", iface.NumVfs, ifaceStatus.NumVfs)
+			} else {
+				glog.V(2).Infof("generic-plugin needDrainNode(): no need drain, current NumVfs is 0")
```
zeeke, line 208: can you add context information to this log? like `iface.PciAddress`

SchSeba:
Thanks @zeeke can you give another round?

zeeke (approved):
LGTM

Eoghan1232 (approved):
lgtm!

pkg/plugins/generic/generic_plugin.go
```diff
@@ -195,15 +195,18 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
 		configured := false
 		for _, iface := range desired {
 			if iface.PciAddress == ifaceStatus.PciAddress {
-				// TODO: no need to perform further checks if ifaceStatus.NumVfs equals to 0
-				// once https://github.com/kubernetes/kubernetes/issues/109595 will be fixed
+				if ifaceStatus.NumVfs == 0 {
```
adrianchiris, line 198: nit: can you add the check after `configured` ?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -195,15 +195,18 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
...
+					glog.V(2).Infof("generic-plugin needDrainNode(): no need drain, for PCI address %s current NumVfs is 0", iface.PciAddress)
+					break
```
adrianchiris, line 200: nit: return here like we do below ? WDYT ?

adrianchiris (commented):
small nits, i see it already got 2 LGTMs,

up 2 u if you want to address b4 merging this one :)

pkg/plugins/generic/generic_plugin.go
SchSeba: done

pkg/plugins/generic/generic_plugin.go
SchSeba: no we can't because we need to continue and check all the other status vs desire.

in the NeedUpdate case we return because if we find even one that needs to update we need to drain, but that is not the case if we find one that doesn't need to drain we need to check for all of them

adrianchiris: ah, didnt see that nested loop. (it was not showing in browser) :\

yes you are right !

adrianchiris (approved):

SchSeba:
Thanks for the quick review folks! merging

PR #476, opened 2023-07-13
design doc for the externally-manage-pf support

SchSeba:


SchSeba:
cc @adrianchiris @zeeke

zeeke:
LGTM

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,167 @@
...
+  priority: 99
+  resourceName: sriov_nic_1
+  externallyManaged: true
+```
+
+#### Validation
+The SR-IOV operator will do a validation webhook to check if the requested `numVfs` is equal to what the user allocate
+if not it will reject the policy creation.
+
+The SR-IOV operator will do a validation webhook to check if the requested MTU is equal to what exist on the PF
```
e0ne, line 82: This is valid only if MTU is provided

e0ne (changes requested):

doc/design/externally-manage-pf.md
SchSeba: right that is why I comment that is the requested MTU.
if the user doesn't request any MTU we don't make any verification

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,167 @@
...
+  resourceName: sriov_nic_1
+  externallyManaged: true
+```
+
+#### Validation
+The SR-IOV operator will do a validation webhook to check if the requested `numVfs` is equal to what the user allocate
+if not it will reject the policy creation.
+
+The SR-IOV operator will do a validation webhook to check if the requested MTU is equal to what exist on the PF 
+if not it will reject the policy creation.
```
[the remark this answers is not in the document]
SchSeba: sure same validation will be done in the config-daemon. I added a paragraph for that

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Non-Goals
+
+* Supporting switchdev mode (may change in the future if there is a request)
+
+## Proposal
+
+Create a flow in the SR-IOV operator where the user can request a configuration for a subset of virtual functions.
+
+The operator will first validate and the requested PF contains the requested amount of virtual functions allocated,
```
adrianchiris, line 48: nit: remove "and" ?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+Then the operator will configure the subset of virtual functions with the request driver and will update the device plugin
```
adrianchiris, line 49: what about MTU, GUID, MAC, driver unbind/bind for rdma?

can you add a list of the current operations sriov operator (and config daemon) performs on PF and VF 
and then list the set of operations it will perform on a VF that belongs to externally manager PF ?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+Name              string    `json:"name,omitempty"`
+LinkType          string    `json:"linkType,omitempty"`
+EswitchMode       string    `json:"eSwitchMode,omitempty"`
+VfGroups          []VfGroup `json:"vfGroups,omitempty"`
++ ExternallyManaged bool      `json:"externallyManaged,omitempty"`
+}
+```
+
+### Implementation Details/Notes/Constraints
+
```
adrianchiris, line 164: no need to cache things ? to remember that a PF is externally managed ? i recall there was something like that in the initial PR ?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+Mtu               int       `json:"mtu,omitempty"`
+Name              string    `json:"name,omitempty"`
+LinkType          string    `json:"linkType,omitempty"`
+EswitchMode       string    `json:"eSwitchMode,omitempty"`
+VfGroups          []VfGroup `json:"vfGroups,omitempty"`
++ ExternallyManaged bool      `json:"externallyManaged,omitempty"`
+}
+```
+
+### Implementation Details/Notes/Constraints
```
adrianchiris, line 163: can we have changes that need to be applied to the different components under this section ?

- controller
- webhook
- config daemon

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Goals
+
+* Allow the SR-IOV operator to handle the configure and pod allocation of a subset of virtual functions.
+* Allow the user to Allocate the number of virtual functions he wants for the system and the subset he wants for pods
+* Not resetting the numOfVfs for PFs that the operator didn't configure
+
+### Non-Goals
+
+* Supporting switchdev mode (may change in the future if there is a request)
```
adrianchiris, line 42: how about supporting a case where externally managed means SR-IOV configuration are applied externally as well.
then the role of the operator would be only to manage device plugin configuration  + restart.

do you see a use-case for it ?

if there is agreement this might be a use-case, we should state it as a non-goal (and this proposal should not prevent us from  doing the above IMO if the use-case arrives )

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+```yaml
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetworkNodePolicy
+metadata:
+  name: sriov-nic-1
+  namespace: sriov-network-operator
+spec:
+  deviceType: netdevice
+  nicSelector:
+    pfNames: ["ens3f0#5-9"]
```
adrianchiris, line 69: are we limited to just pfNames ? 
do we need to specify this range of VFs ?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+  nodeSelector:
+    node-role.kubernetes.io/worker: ""
+  numVfs: 10
+  priority: 99
+  resourceName: sriov_nic_1
+  externallyManaged: true
+```
+
+#### Validation
+The SR-IOV operator will do a validation webhook to check if the requested `numVfs` is equal to what the user allocate
```
adrianchiris, line 79: until today, did we prevent the creation of SriovNetworkNodePolicy because an sriovNetworkNodeState did not have status ?

wont this affect UX, say i have a provisioning infra where i deploy sriov operator then create its CRs
i would now need to split between the two steps, add custom wait logic to make sure i can create the CR, this is not automation friendly.

also what happens on scale-up of the cluster ?  (after you already have a bunch of externally managed policies)

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+spec:
+  deviceType: netdevice
+  nicSelector:
+    pfNames: ["ens3f0#5-9"]
+  nodeSelector:
+    node-role.kubernetes.io/worker: ""
+  numVfs: 10
+  priority: 99
+  resourceName: sriov_nic_1
+  externallyManaged: true
```
adrianchiris, line 75: externallyManagedPFs ?

or we fine with keeping it more general ?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Validation
+The SR-IOV operator will do a validation webhook to check if the requested `numVfs` is equal to what the user allocate
+if not it will reject the policy creation.
+
+The SR-IOV operator will do a validation webhook to check if the requested MTU is equal to what exist on the PF 
+if not it will reject the policy creation.
+
+
+*Note:* Same validation will be done in the SR-IOV config-daemon container to cover cases where the user doesn't want to deploy"
```
adrianchiris, line 86: or scale-up use-cases

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+the webhook. If the verification failed in the policy apply stage
+the `sriovNetworkNodeState.status.SyncStatus` field will be report a `Failed` status and the error description will 
+get exposed in `sriovNetworkNodeState.status.LastSyncError`
+
+
+#### Configuration
+
+The SR-IOV operator config daemon will reconcile on the SriovNetworkNodeState update and will follow the regular
+flow of virtual functions *SKIPPING* only the Virtual function allocation.
```
adrianchiris, line 95: is that the only thing ? will it still call vendor plugins ?

i assume it will call `needUpdate` and fail if true is returned.

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,174 @@
...
+
+The SR-IOV operator will update the SR-IOV Network Device Plugin with the pool information
+
+Another change with the operator beavior is when we delete a policy with had `externallyManaged: true` the SR-IOV operator
```
adrianchiris, line 99: what if we then proceed with deleting the sriovnetworknodestate obj ?

this needs to be consistent.

adrianchiris (changes requested):

doc/design/externally-manage-pf.md
SchSeba: done

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,212 @@
...
+creation-date: 12-07-2023
+last-updated: 12-07-2023
+---
+
+# Externally Manage PF
+
+## Summary
+
+Allow the SR-IOV operator to configure and allocate a subset of virtual functions from a physical function not
+configured by the operator
```
wizhaoredhat, line 17: "...not configured by the operator": I think this summary is a bit confusing for a new reader. I think what is meant is by "...from a physical function that is configured externally from SR-IOV network operator" Wdyt?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,212 @@
...
+  - adrianchiris
+creation-date: 12-07-2023
+last-updated: 12-07-2023
+---
+
+# Externally Manage PF
+
+## Summary
+
+Allow the SR-IOV operator to configure and allocate a subset of virtual functions from a physical function not
```
wizhaoredhat, line 16: NIT: SR-IOV network operator.

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,212 @@
...
+configured by the operator
+
+## Motivation
+
+The feature is needed to allow the operator to only configure a subset of virtual functions.
+This allows a third party component like nmstate, kubernetes-nmstate, NetworkManager to handle the creation
+and the usage of the virtual functions on the system.Some of the examples are using the virtual function as the primary
```
wizhaoredhat, line 23: Nit: Missing "space" between "." and "Some".

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,212 @@
...
+
+The changes in the manager for this feature are minimal we only copy the `ExternallyManaged` boolean from the policy
+to the generated `nodeState.Spec`
+
+#### Config Daemon
+
+This is where most of the changes for this feature are implemented.
+
+First step we will do a validation same as on the webhook to check the PF have everything we need to apply the requested
+policy, by checking the `numVfs` and `MTU`.
```
adrianchiris, line 195: linkType as well ?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,212 @@
...
+Next config-daemon will skip all the PF configuration like `numVfs` and `MTU`. he will only preform the virtual function 
+driver allocation, administrative mac allocation and MTU. Last step as always will be to reset the device plugin so
```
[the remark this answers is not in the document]
adrianchiris: also need to update this part with RDMA bits

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,212 @@
...
+kubelet will be able to discover the SR-IOV devices.
+
+The config-daemon will also save on the node a cache of the last applied policy. this is needed to be able and understand
```
adrianchiris, line 200: note: we already write a checkpoint file on daemon start see:

`see: NodeStateStatusWriter.writeCheckpointFile()`

perhaps this can be reused for last applied state

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,212 @@
...
+if we need to reset the PF configuration(`ExternallyManaged` was false) or not when o policy is removed.
+
+### Upgrade & Downgrade considerations
+
+The feature supports both Upgrade and Downgrade as we are introducing a new field in the API
+
+### Test Plan
+
+* Should not allow to create a policy if there are no vfs configured
```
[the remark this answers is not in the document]
SchSeba: done that was the name of the tests I already did on the PR  :)

doc/design/externally-manage-pf.md
SchSeba: that file is only for virtual env and it save something else from what we need

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,246 @@
...
+metadata:
+  name: sriov-nic-1
+  namespace: sriov-network-operator
+spec:
+  deviceType: netdevice
+  nicSelector:
+    pfNames: ["ens3f0#5-9"]
+  nodeSelector:
+    node-role.kubernetes.io/worker: ""
+  numVfs: 10
```
wizhaoredhat, line 96: One of the goals is "Not resetting the numOfVfs for PFs that are externally managed" then should we somehow block out this parameter? We shouldn't need to provide this when we can get the number of VFs created from sysfs sriov_numvfs.

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,246 @@
...
+This allows a third party component like nmstate, kubernetes-nmstate, NetworkManager to handle the creation
+and the usage of the virtual functions on the system. Some of the examples are using the virtual function as the primary
+nic for the k8s SDN network or a storage network.
+
+Before this change the SR-IOV network operator is the only component that should use/configure VFs. not allowing the user
+to use some of the VFs for host networking.
+
+### Use Cases
+
+* As a user I want to use a virtual function for SDN network, for SDN the network need to be configured before
```
wizhaoredhat, line 31: For this particular use case, it is similar to what we have for OvS Hardware Offload. Do you expect for this use case that the use sriov-network-operator in systemd would be needed?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,246 @@
...
+```yaml
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetworkNodePolicy
+metadata:
+  name: sriov-nic-2
+  namespace: sriov-network-operator
+spec:
+  deviceType: netdevice
+  nicSelector:
+    pfNames: ["ens3f0"]
```
wizhaoredhat, line 114: In this case only the PF is externally managed right? Maybe we can add what the objective of this policy example is.

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,246 @@
...
+* in case of RDMA (both for ETH and IB) it will perform an unbind/bind of the VF driver to set RDMA Node/Port GUID.
+
+### Workflow Description
+
+The user will allocate the virtual functions on the system with any third party tool like nmstate, Kubnernetes-nmstate,
+systemd scripts, etc..
+
+Then the user will be able to create a policy telling the operator that the PF is externally managed by the user.
+
+#### Policy Example:
```
wizhaoredhat, line 83: I think we can state the objective of this policy. E.g. The PF and VFs 0-4 are externally managed. For example nmstate will create 10 vfs, but will only consume VF 0 and 4 in its configuration. Nmstate will also manage the MTU and other parameters of the PF.

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,246 @@
...
+  priority: 99
+  resourceName: sriov_nic_1
+  externallyManaged: true
+```
+
+#### Validation
+The SR-IOV network operator will do a validation webhook to check if the requested `numVfs` is equal to what the user allocate
+if not it will reject the policy creation.
+
+The SR-IOV network operator will do a validation webhook to check if the requested MTU is equal to what exist on the PF
```
wizhaoredhat, line 127: I think we should block out any setting of the MTU when externallyManaged = true. In the code we should automatically set the MTU of the VF from the PF.

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,246 @@
...
+
+Another change with the operator beavior is when we delete a policy with had `externallyManaged: true` the SR-IOV network operator
+will *NOT* reset the `numVfs`
+
+### API Extensions
+
+For SriovNetworkNodePolicy
+
+```golang
+// SriovNetworkNodePolicySpec defines the desired state of SriovNetworkNodePolicy
```
wizhaoredhat, line 152: Since the API might change and this copy paste of the API might get out of date, I am thinking maybe just adding "..." for the existing parameters would be enough.

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,246 @@
...
+VfGroups          []VfGroup `json:"vfGroups,omitempty"`
++ ExternallyManaged bool      `json:"externallyManaged,omitempty"`
+}
+```
+
+### Implementation Details/Notes/Constraints
+
+#### Webhook
+For the webhook we add more validations when the policy contains `ExternallyManaged: true`
+* `numVfs` in the policy equal is equal or lower the number of virtual functions on the system
```
wizhaoredhat, line 212: I think it is a better UX if the user no longer needs to provide these values for externally managed. One less configuration that needs to be copied and pasted from another configuration (e.g. nmstate). It would be better to just blindly inherit these values. Wdyt?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,246 @@
...
+The changes in the manager for this feature are minimal we only copy the `ExternallyManaged` boolean from the policy
+to the generated `nodeState.Spec`
+
+#### Config Daemon
+
+This is where most of the changes for this feature are implemented.
+
+First step we will do a validation same as on the webhook to check the PF have everything we need to apply the requested
+policy, by checking the `numVfs`, `MTU` and `LinkType`.
+Next config-daemon will skip all the PF configuration like `numVfs`, `MTU` and `LinkType`. he will only preform the virtual function
```
wizhaoredhat, line 227: SP: "preform" to "perform"

doc/design/externally-manage-pf.md
adrianchiris: even this design doc may get out of date at some point :)

doc/design/externally-manage-pf.md
SchSeba: +1

doc/design/externally-manage-pf.md
SchSeba: no we want to have this so we can still do validation.
with the externally manage the numVfs MUST be lower or equal to the number of vfs configured in the system

doc/design/externally-manage-pf.md
SchSeba: done

doc/design/externally-manage-pf.md
SchSeba: we allow the user to select a different MTU for the VF that must be lower or equal the one from the PF. (because we don't change the MTU on the PF in this mode

doc/design/externally-manage-pf.md
SchSeba: +1

doc/design/externally-manage-pf.md
SchSeba: works for me

doc/design/externally-manage-pf.md
SchSeba: I will like to have this one because we do a lot of validation and stuff on the code with it when we have multiple policies.
we can change this in the future if we see customer complaining

doc/design/externally-manage-pf.md
SchSeba: done

doc/design/externally-manage-pf.md
wizhaoredhat: I thought if the VF MTU is lower than the PF MTU it can cause issues. From the text, highlighted here, it seems that the MTU on the VF must be equal to the MTU on the PF which is also my expectations.

So the user has no choice in MTU values? It is a "must" for the MTU of VF = MTU of PF?

doc/design/externally-manage-pf.md
wizhaoredhat: I think it maybe helpful to put a note in the document saying that systemd mode is recommended for some of the example use cases.

doc/design/externally-manage-pf.md
wizhaoredhat: I thought the numVfs must be equal since the significance of numVfs is configuring the number of VFs via sysfs sriov_numvfs. I can't see a reason which numVfs < sysfs sriov_numvfs.

Wouldn't we always want to have numVfs == sysfs sriov_numvfs?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,214 @@
...
+    node-role.kubernetes.io/worker: ""
+  numVfs: 10
+  priority: 99
+  resourceName: sriov_nic_1
+  externallyManaged: true
+```
+
+The SR-IOV network operator will use all the 10 virtual functions created externally by the user.
+One if the main use cases for this is if the user want to do some custom configuration to the PF and VFs like loading
+out of tree drivers or other stuff that the operator doesn't support, or are specific to a specific vendor nic
```
adrianchiris, line 128: nit: IMO you can remove "or are specific to a specific vendor nic"
the prior sentences already cover it IMO. with the custom configuration bits.

also today we already have vendor specific bits in the operator. this may wrongfully imply that we don't.

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,214 @@
...
+unbind/bind of the VF driver to set RDMA Node/Port GUID.
+Last step as always will be to reset the device plugin so
+kubelet will be able to discover the SR-IOV devices.
+
+The config-daemon will also save on the node a cache of the last applied policy. this is needed to be able and understand
+if we need to reset the PF configuration(`ExternallyManaged` was false) or not when o policy is removed.
+
+### Upgrade & Downgrade considerations
+
+The feature supports both Upgrade and Downgrade as we are introducing a new field in the API
```
adrianchiris, line 207: downgrade will cause the operator to treat an externally managed PF as non externally managed and actually configure PF, this may cause conflicts  in the system.

imo the only concern here is documenting this.

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,214 @@
...
+* `numVfs` in the policy equal is equal or lower the number of virtual functions on the system
+* `MTU` in the policy equals or lower the MTU we discover on the PF
+* `LinkType` in the policy equals the link type we discover on the PF
+
+#### Controller/Manager
+
+The changes in the manager for this feature are minimal we only copy the `ExternallyManaged` boolean from the policy
+to the generated `nodeState.Spec`
+
+#### Config Daemon
```
adrianchiris, line 189: are there any changes planned for vendor plugins ?

SchSeba: for vendor plugins like mlx we now understand the new field and react on it.
meaning if the num of vfs is not eq or lower in the policy the plugin will return error

SchSeba:
Hi @adrianchiris please give it another look

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Upgrade & Downgrade considerations
+
+The feature supports both Upgrade and Downgrade as we are introducing a new field in the API.
+Downgrade will cause the operator to treat an externally managed PF as non externally managed and actually configure PF,
+this may cause conflicts in the system.
+
+### Test Plan
+
+* Should not allow to create a policy with externallyManaged true if there are no vfs configured
```
e0ne, line 216: we need to support a case when VFs will be configured later

e0ne (changes requested):

doc/design/externally-manage-pf.md
SchSeba: it can also be lower then `sysfs sriov_numvfs` to allow increasing the num of vfs

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,219 @@
...
+First step we will do a validation same as on the webhook to check the PF have everything we need to apply the requested
+policy, by checking the `numVfs`, `MTU` and `LinkType`.
+Next config-daemon will skip all the PF configuration like `numVfs`, `MTU` and `LinkType`. he will only perform the virtual function 
+driver binding, administrative mac allocation and MTU. Another step In case of Infiniband link type it will 
+generate random node and port GUID for the interface, same in case of RDMA (both for ETH and IB) it will perform an 
+unbind/bind of the VF driver to set RDMA Node/Port GUID.
+Last step as always will be to reset the device plugin so
+kubelet will be able to discover the SR-IOV devices.
+
+The config-daemon will also save on the node a cache (file) of the last applied policy. this is needed to be able and understand
```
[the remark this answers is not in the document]
SchSeba: if possible, I would like to leave as part of this PR as it's already integrated.

doc/design/externally-manage-pf.md
SchSeba: we can do it. BUT the user will need to disable the webhook because that is not the right way to do it.
the network must be ready before kubelet start of not then just let the operator create the vfs.
or use systemd mode

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,224 @@
...
+metadata:
+  name: sriov-nic-1
+  namespace: sriov-network-operator
+spec:
+  deviceType: netdevice
+  nicSelector:
+    pfNames: ["ens3f0#5-9"]
+  nodeSelector:
+    node-role.kubernetes.io/worker: ""
+  numVfs: 10
```
adrianchiris, line 104: nit: techincally this number can be 5 right ?

doc/design/externally-manage-pf.md
adrianchiris: ack, lets add this in a subsection here.

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,224 @@
...
+#### Controller/Manager
+
+The changes in the manager for this feature are minimal we only copy the `ExternallyManaged` boolean from the policy
+to the generated `nodeState.Spec`
+
+#### Config Daemon
+
+This is where most of the changes for this feature are implemented.
+
+First step we will do a validation same as on the webhook to check the PF have everything we need to apply the requested
```
adrianchiris, line 201: can this be a bullet list ?

doc/design/externally-manage-pf.md
```diff
@@ -0,0 +1,224 @@
...
+The feature supports both Upgrade and Downgrade as we are introducing a new field in the API.
+Downgrade will cause the operator to treat an externally managed PF as non externally managed and actually configure PF,
+this may cause conflicts in the system.
+
+### Test Plan
+
+* Should not allow to create a policy with externallyManaged true if there are no vfs configured
+* Should create a policy if the number of requested vfs is equal
+* Should create a policy if the number of requested vfs is equal and not delete them when the policy is removed
+* should reset the virtual functions if externallyCreated is false
```
adrianchiris, line 224: should we add a test case for the daemon flow to ensure it conforms  to the newly defined flow for externally managed ?

adrianchiris (approved):
overall LGTM, just a few minor comments 

once addressed consider having an LGTM from my side.

e0ne (approved):
Thanks for addressing my comments

doc/design/externally-manage-pf.md
SchSeba: done I added a new case to test the daemon by disabling the webhook

`should to configure a policy with externallyManaged true if there are no vfs configured with disabled webhook`

zeeke (approved):
LGTM

PR #478, opened 2023-07-14
Avoid reconciling `SriovNetworkNodePolicies` multiple times

zeeke:
`SriovNetworkNodePolicy` reconcile function lists all node policies and updates all SriovNetworkNodeState. Hence there is no need to resync all of them periodically.

Avoid triggering multiple reconciliations when creating multiple policies in a row (that is what happens when using yaml files).

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"time"
 
 	errs "github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
@@ -35,11 +36,15 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	kscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	dptypes "github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
 
@@ -48,6 +53,8 @@ import (
 	constants "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/render"
 )
+
+const nodePolicySyncEventName = "node-policy-sync-event"
 
 // SriovNetworkNodePolicyReconciler reconciles a SriovNetworkNodePolicy object
 type SriovNetworkNodePolicyReconciler struct {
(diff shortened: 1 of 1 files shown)


controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -152,8 +163,29 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(ctx context.Context, req ct
...
+
+	squashAndDebounceHandler := handler.Funcs{
+		CreateFunc: func(e event.CreateEvent, q workqueue.RateLimitingInterface) {
+			log.Log.WithName("SriovNetworkNodePolicy").
+				Info("Enqueuing sync for create event", "resource", e.Object.GetName())
+			qHandler(q)
+		},
+		UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+			log.Log.WithName("SriovNetworkNodePolicy").
+				Info("Enqueuing sync for create event", "resource", e.ObjectNew.GetName())
```
SchSeba, line 181: nit: create -> update

SchSeba (commented):
general question is it possible for the user to create an object with that name?

can we use a non existing namespace or something like that to be sure there is not going to be an object like that?

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -152,8 +163,29 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(ctx context.Context, req ct
...
 func (r *SriovNetworkNodePolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	qHandler := func(q workqueue.RateLimitingInterface) {
+		q.AddAfter(reconcile.Request{NamespacedName: types.NamespacedName{
+			Namespace: namespace,
+			Name:      nodePolicySyncEventName,
+		}}, time.Second)
+	}
+
+	squashAndDebounceHandler := handler.Funcs{
+		CreateFunc: func(e event.CreateEvent, q workqueue.RateLimitingInterface) {
```
e0ne, line 174: Why we don't have DeleteFunc here?

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -152,8 +163,29 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(ctx context.Context, req ct
...
 // SetupWithManager sets up the controller with the Manager.
 func (r *SriovNetworkNodePolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	qHandler := func(q workqueue.RateLimitingInterface) {
+		q.AddAfter(reconcile.Request{NamespacedName: types.NamespacedName{
+			Namespace: namespace,
+			Name:      nodePolicySyncEventName,
+		}}, time.Second)
+	}
+
+	squashAndDebounceHandler := handler.Funcs{
```
adrianchiris, line 173: nit: just name it `delayedEventHandler` ?

controllers/sriovnetworknodepolicy_controller.go
adrianchiris: +1

a deleted policy also affects node states spec

adrianchiris:
>general question is it possible for the user to create an object with that name?
>
>can we use a non existing namespace or something like that to be sure there is not going to be an object like that?

why do we care ? its just an "event" to trigger reconcile that is not dependent on the specific obj that triggered the event.

controllers/sriovnetworknodepolicy_controller.go
zeeke: good point! fixing

controllers/sriovnetworknodepolicy_controller.go
zeeke: done

zeeke:
> general question is it possible for the user to create an object with that name?
>
> can we use a non existing namespace or something like that to be sure there is not going to be an object like that?

Good point. If the user creates a node policy called `node-policy-sync-event`, then the Reconcile logic is invoked twice. Not a big deal, but this PR is all about avoiding doing the same things multiple times. Changing.


> why do we care ? its just an "event" to trigger reconcile that is not dependent on the specific obj that triggered the event.

Because the controller (and the Reconcile method) are still triggered by every SriovNetworkNodePolicy, as it's the `.For(...)` argument.

SchSeba (approved):
LGTM

thanks!

PR #479, opened 2023-07-19
Add 'Parallel SR-IOV configuration' design document

e0ne:


README.md
```diff
@@ -10,7 +10,7 @@ SR-IOV network is an optional feature of an Openshift cluster. To make it work,
 
 - Initialize the supported SR-IOV NIC types on selected nodes.
 - Provision/upgrade SR-IOV device plugin executable on selected node.  
-- Provision/upgrade SR-IOV CNI plugin executable on selected nodes.
+- Provision/upgrade SR-IOV CNI plugin executableSriovNetworkPoolConfigon selected nodes.
```
adrianchiris, line 13: unrelated change.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+
+# Parallel SR-IOV configuration
+
+## Summary
+SR-IOV Network Operator configures NICs one by one. That means we’ll need to wait hours or even days to configure all
+NICs on large cluster deployments. With multi-NICs deployments Operator configures NICs one by one on each node which
+leads to a lot of unnecessary node drain calls during SR-IOV configuration.
+
+## Motivation
+Allow SR-IOV Network Operator to configure more than one node at the moment.
```
adrianchiris, line 20: I think thats the summary and what u got in summary is the motivation...

adrianchiris: nit: remove at the moment

adrianchiris: also add that we are moving all draining logic to a centralized place which will reduce chances of race conditions and bugs that were encountered before in sriov-network-config-daemon with draining.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Use Cases
+
+### Goals
```
adrianchiris, line 24: one of the goals IMO should be to move draining logic from config daemon to a centralized place.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+* Number of drainable at the moment nodes should be 1 be default
```
adrianchiris, line 25: i think you can drop "at the moment" here and below.

do you think it adds more context ? if yes we can keep, but i think the meaning of the sentence will remain the same.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+* A node could be included into several pools. In this case pool should be 
+  selected by it’s priority like it’s implemented for SriovNetworkNodePolicy
+
+
+### Non-Goals
+Parallel NICs configuration on the same node is out of scope of this proposal
+
+## Proposal
+
+Introduce nodes pool drain configuration to meet following requirements:
```
adrianchiris, line 37: this paragraph does not add any new information, you can remove it IMO.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+
+```golang
+AnnoIdle            = "Idle"
+AnnoDrainRequired   = "Drain_Required"
+AnnoMcpPaused       = "Draining_MCP_Paused"
+AnnoDraining        = "Draining"
+AnnoDraining        = "Draining_Complete"
+```
+
+Drain controller will watch for node annotation changes, `SriovNetworkPoolConfig` and `SriovNetworkNodeState` changes:
```
adrianchiris, line 57: lets add a sentence before this one on where we are storing the states.

i believe we want it as annotations but not on node, but rather on `sriovNetworkNodeState`.   Thoughts ?

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+```golang
+// +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=SriovNetworkPoolConfig,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=SriovNetworkNodeState,verbs=get;list;watch;update;patch
+```
+
+Since new `SriovNetworkNodeState` will appear with `syncState=InProgess` controller will `Drain_Required` annotation.
```
adrianchiris, line 64: controller will do what ? im missing a verb here.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+nodeSelectorTerms:
+- matchExpressions:
+- key: some-label
+operator: In
+``` 
+
+### Implementation Details/Notes/Constraints
+The implementation of this feature is very complex and requires a lot of changes to different parts of SR-IOV Network
+Operator. To not introduce breaking changes we have to split this effort to several phases:
+* implement new Drain controller without API changes:
```
adrianchiris, line 179: This is inaccurate, given we move to the new drain states.

i think this should say: implement new drain controller without user-facing API changes.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+  it will proceed only one node configuration at the same time and doesn't require API changes
+* introduce new API changes to support pool of nodes to be drained in a parallel:
+  at this phase we introduce new API from one of the proposed options above and modify Drain controller to watch for
+  the specified CRs and proceed drain in a parallel per node pool configuration
+* drop `NodeDrainAnnotation` usage and move this annotation values into the `SriovNetworkNodeState` object
```
adrianchiris, line 184: why do we want to do this last ? and not part of the initial change ? seems like un-needed work to first implement using node annot and then move them to a different k8s obj

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+
+All phases should be implemented one-by-one in a separate PRs in the order above.
+
+### Upgrade & Downgrade considerations
+This feature introduces changes to CRDs which means CRDs should be applied during upgrade or downgrade
```
adrianchiris, line 189: do we care about "version skew"

that is new controller and old config daemon or via versa ?

i think not. its not expected to update sriov-network-operator mid config in cluster and if idle there should be no issues as no drain is requested/required.

we need also to add logic in drain controller to cleanup old drain annotations from nodes given we move them (and after we move them) to sriovnetworknodestate.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+  the specified CRs and proceed drain in a parallel per node pool configuration
+* drop `NodeDrainAnnotation` usage and move this annotation values into the `SriovNetworkNodeState` object
+
+All phases should be implemented one-by-one in a separate PRs in the order above.
+
+### Upgrade & Downgrade considerations
+This feature introduces changes to CRDs which means CRDs should be applied during upgrade or downgrade 
+
+### Test Plan
+* Unit tests should be implemented for new Drain Controller.
```
adrianchiris, line 192: should -> will

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+* Nodes pool should be defined by node selector
+* A node could be included into several pools. In this case pool should be selected by it’s priority like it’s
+  implemented for `SriovNetworkNodePolicy`
+
+
+### Workflow Description
+A new Drain controller will be introduced to manage node drain and cordon procedures. That means we don't need to do
+drain and use `drain lock` in config daemon anymore. The overall drain process will be covered by the following states:
+
+```golang
```
adrianchiris, line 49: lets have a sub section here with the state machine and transitions of the states below.

you can use: https://mermaid.js.org/syntax/stateDiagram.html
or anything else that helps convey the states a configuring node goes through.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+
+Drain controller will watch for node annotation changes, `SriovNetworkPoolConfig` and `SriovNetworkNodeState` changes:
+```golang
+// +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=SriovNetworkPoolConfig,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=SriovNetworkNodeState,verbs=get;list;watch;update;patch
+```
+
+Since new `SriovNetworkNodeState` will appear with `syncState=InProgess` controller will `Drain_Required` annotation.
+Another reconcile loop will check if current `Draining` nodes count is less than `maxUnavailable` and will
```
adrianchiris, line 65: another ? where was the first ?

doc/design/parallel-node-config.md
adrianchiris: also syncstate in-progress does  not nessecarily mean we need to drain.

imo config daemon should be in charge of setting the annotation (state) to drain required.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+Introduce nodes pool drain configuration to meet following requirements:
+* Number of drainable at the moment nodes should be 1 be default
+* Number of drainable at the moment nodes should be configured by pool
+* Nodes pool should be defined by node selector
+* A node could be included into several pools. In this case pool should be selected by it’s priority like it’s
+  implemented for `SriovNetworkNodePolicy`
+
+
+### Workflow Description
+A new Drain controller will be introduced to manage node drain and cordon procedures. That means we don't need to do
```
adrianchiris, line 46: what does this drain controller watch ? what will trigger a reconcile ?

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+AnnoDraining        = "Draining_Complete"
+```
+
+Drain controller will watch for node annotation changes, `SriovNetworkPoolConfig` and `SriovNetworkNodeState` changes:
+```golang
+// +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=SriovNetworkPoolConfig,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=SriovNetworkNodeState,verbs=get;list;watch;update;patch
+```
+
```
adrianchiris, line 63: we should have the flow below as numbered steps for clarity on how reconcile flow will look.

also i believe config daemon also has a part in this flow.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+Since new `SriovNetworkNodeState` will appear with `syncState=InProgess` controller will `Drain_Required` annotation.
+Another reconcile loop will check if current `Draining` nodes count is less than `maxUnavailable` and will
+mark it as `Draining` to start draining procedure. If current Draining nodes count equals to
```
adrianchiris, line 66: nit: equals -> greater equal .

the greater equal is in case maxUnavailable changed for the node and we are now oversubscribed. (i think its ok to be in this state)

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+
+Since new `SriovNetworkNodeState` will appear with `syncState=InProgess` controller will `Drain_Required` annotation.
+Another reconcile loop will check if current `Draining` nodes count is less than `maxUnavailable` and will
+mark it as `Draining` to start draining procedure. If current Draining nodes count equals to 
+`maxUnavailable`, no annotation updates will be applied in the current reconcile loop.
+
+If  `SriovNetworkNodeState` will appear with `syncState=Succeed` and node annotation `Draining_Complete` controller
+should change annotation to `Idle`.
+
+We can add a webhook to block the change of the configuration when the nodes are draining to a lower number. Without
```
adrianchiris, line 72: i think oversubscription in this case is OK, no need to block. we just wont drain anymore nodes.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+validating webhook controller will proceed drain procedure for the current `Draining` nodes until the count of them
+will decrease to specified `maxUnavailable` number.
+
+We need this logic in controller to have it centralized and not to have raise condition issues during annotation update.
+
+Config daemon will not have drain-related logic with will simplify it a lot.
```
adrianchiris, line 78: it will  have minimal drain related logic ( it still needs to read the annot) and possibly set it when requesting drain.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+nodeSelectorTerms:
+- matchExpressions:
+    - key: some-label
+      operator: In
+      values:
+        - val-2
+- matchExpressions:
+    - key: other-label
+      operator: "Exists"
+      DrainConfig:
```
adrianchiris, line 104: wrong indent ?

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+Config daemon will not have drain-related logic with will simplify it a lot.
+
+### API Extensions
+
+#### Option 1: extend existing CR SriovNetworkPoolConfig
+SriovNetworkPoolConfig is used only for OpenShift to provide configuration for
+OVS Hardware Offloading. We can extend it to add configuration for the drain
+pool. E.g.:
+
+```yaml
```
adrianchiris, line 87: please also add a go snippet of the type.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,195 @@
...
+validating webhook controller will proceed drain procedure for the current `Draining` nodes until the count of them
+will decrease to specified `maxUnavailable` number.
+
+We need this logic in controller to have it centralized and not to have raise condition issues during annotation update.
+
+Config daemon will not have drain-related logic with will simplify it a lot.
+
+### API Extensions
+
+#### Option 1: extend existing CR SriovNetworkPoolConfig
```
adrianchiris, line 82: +1 on this option from me.

adrianchiris (changes requested):

doc/design/parallel-node-config.md
e0ne: done

doc/design/parallel-node-config.md
e0ne: fixed

doc/design/parallel-node-config.md
e0ne: thanks for the suggestion!

doc/design/parallel-node-config.md
SchSeba: agree we should move this as part of the first change.

we much check how this will work in case the operator upgrade is done in the middle of the configuration.

doc/design/parallel-node-config.md
SchSeba: I think we should have a support for this as we have users that runs the sriov operator.
we need to be sure a case where the operator gets updated in the middle of a drain is something we support and are able to recover from it

SchSeba (commented):
nice to see this work!
I am really expecting for this feature

SchSeba:
One more comment is we need to see how the machine config will work with parallel node draining for openshift

api/v1/sriovnetworkpoolconfig_types.go
```diff
@@ -6,8 +6,14 @@ import (
 
 // SriovNetworkPoolConfigSpec defines the desired state of SriovNetworkPoolConfig
 type SriovNetworkPoolConfigSpec struct {
```
adrianchiris, line 8: added by mistake ?

doc/design/parallel-node-config.md
adrianchiris: i still see "at the moment" :) will u remove it ?

doc/design/parallel-node-config.md
adrianchiris: @e0ne here you should say that config daemon will have minimal drain related logic ? because it still needs to set / wait on some annotation.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,249 @@
...
+`maxUnavailable`, no `SriovNetworkNodeState` updates will be applied in the current reconcile loop.
+
+Once draining and configuration will be finished config daemon will update `SriovNetworkNodeState.DrainStatus` to
+`Idle`.
+
+We can add a webhook to block the change of the configuration when the nodes are draining to a lower number. Without
+validating webhook controller will proceed drain procedure for the current `Draining` nodes until the count of them
+will decrease to specified `maxUnavailable` number.
+
+We need this logic in controller to have it centralized and not to have raise condition issues during
```
wizhaoredhat, line 79: Typo? "raise" => "race"?

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,249 @@
...
+Drain controller will watch for node annotation, `SriovNetworkNodeState` and `SriovNetworkPoolConfig` changes:
+```golang
+// +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=SriovNetworkPoolConfig,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=SriovNetworkNodeState,verbs=get;list;watch;update;patch
+```
+
+Once `SriovNetworkNodeState.DrainStatus` will be marked as `Drain_Required` by config daemon controller add will start
+draining procedure.  Reconcile loop will check if current `Draining` nodes count is greater or equal than
+`maxUnavailable` and will mark it as `Draining` to start draining procedure. If current Draining nodes count equals to
```
wizhaoredhat, line 69: Where is "maxUnavailable" taken from? Is this from a PodDisruptionBudget definition? If not, it should be made clear that a new parameter is added (since "maxUnavailable" is a common term in kubernates)

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,249 @@
...
+type SriovNetworkPoolConfigSpec struct {
+    // +kubebuilder:validation:Minimum=0
+    // +kubebuilder:validation:Maximum=99
+    // Priority of the SriovNetworkPoolConfig, higher priority policies can override lower ones.
+    Priority int `json:"priority,omitempty"`
+    // OvsHardwareOffloadConfig describes the OVS HWOL configuration for selected Nodes
+    OvsHardwareOffloadConfig OvsHardwareOffloadConfig `json:"ovsHardwareOffloadConfig,omitempty"`
+    // NodeSelectorTerms is a list of node selectors to apply SriovNetworkPoolConfig
+    NodeSelectorTerms *v1.NodeSelector `json:"nodeSelectorTerms,omitempty"`
+    DrainConfit DrainConfigSpec `json:"drainConfig,omitempty"`
```
SchSeba, line 127: DrainConfit -> DrainConfig

doc/design/parallel-node-config.md
e0ne: It's a new field of in `SriovNetworkPoolConfig`

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,249 @@
...
+#### Extend existing CR SriovNetworkNodeState
+SriovNetworkNodeState already contains current state of the configuration progress in a `status` field, so we don't need
+to store drain status in node annotation. `SriovNetworkNodeStateStatus` will be extended to contain drain progress
+instead of node annotation:
+
+```golang
+// SriovNetworkNodeStateStatus defines the observed state of SriovNetworkNodeState
+type SriovNetworkNodeStateStatus struct {
+	Interfaces    InterfaceExts `json:"interfaces,omitempty"`
+	DrainStatus    string        `json:"drainStatus,omitempty"`
```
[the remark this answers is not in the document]
e0ne: done

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,249 @@
...
+
+Default pool should be always created with the lowest priority:
+```yaml
+apiVersion: v1
+kind: SriovNetworkPoolConfig
+metadata:
+name: default
+namespace: network-operator
+spec:
+priority:  100
```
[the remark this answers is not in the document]
e0ne: thanks, fixed priority and correct time

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,249 @@
...
+
+
+### Implementation Details/Notes/Constraints
+The implementation of this feature is very complex and requires a lot of changes to different parts of SR-IOV Network
+Operator. To not introduce breaking changes we have to split this effort to several phases:
+* implement new Drain controller without user-facing API changes:
+  it will proceed only one node configuration at the same time and doesn't require API changes
+* drop `SriovNetworkNodeState` usage and move this annotation values into the `SriovNetworkNodeState` object
+* introduce new API changes to support pool of nodes to be drained in a parallel:
+  at this phase we introduce new API from one of the proposed options above and modify Drain controller to watch for
```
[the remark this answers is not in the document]
e0ne: I updated this paragraph

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,245 @@
...
+After operator upgrade we have to support `sriovnetwork.openshift.io/state` node annotation to
+`SriovNetworkNodeState.DrainStatus` migration. This logic will be implemented in a Drain controller and should be
+supported at least until next major version of the operator.
+
+Node reconcile loop will check for `sriovnetwork.openshift.io/state` annotation and move its value to
+`SriovNetworkNodeState.DrainStatus` field. `sriovnetwork.openshift.io/state` annotation will be deleted by operator once
+`SriovNetworkNodeState` is updated.
+
+### Alternative APIs
+#### Option 1: extend SriovOperatorConfig CRD
```
adrianchiris, line 187: we should add 1-2 sentences on why each option here was not chosen

e0ne: done

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,251 @@
...
+SR-IOV Network Operator configures SR-IOV one node at a time and one nic at a same time. That means we’ll need to wait
+hours or even days to configure all NICs  on large cluster deployments. With multi-NICs deployments Operator configures
+NICs one by one on each node which  leads to a lot of unnecessary node drain calls during SR-IOV configuration. Also
+moving all draining logic to a centralized place which will reduce chances of race conditions and bugs that were
+encountered before in sriov-network-config-daemon with draining.
+
+### Use Cases
+
+### Goals
+* Number of drainable nodes should be 1 bн default
```
adrianchiris, line 27: `bн` -> by ?

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,251 @@
...
+* A node could be included into several pools. In this case pool should be 
+  selected by it’s priority like it’s implemented for SriovNetworkNodePolicy
+* Move all drain-related logic into the centralized place
+
+
+### Non-Goals
+Parallel NICs configuration on the same node is out of scope of this proposal
+
+## Proposal
+
```
adrianchiris, line 39: nit: remove newline

adrianchiris (commented):
@e0ne gave this one another look.

once my remaining comments are addressed, consider it LGTM from my side.

doc/design/parallel-node-config.md
e0ne: fixed

doc/design/parallel-node-config.md
e0ne: done

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,250 @@
...
+maxUnavailable: 1
+```
+
+
+### Implementation Details/Notes/Constraints
+The implementation of this feature is very complex and requires a lot of changes to different parts of SR-IOV Network
+Operator. To not introduce breaking changes we have to split this effort to several phases:
+* implement new Drain controller without user-facing API changes:
+  it will proceed only one node configuration at the same time and doesn't require API changes
+* drop `SriovNetworkNodeState` usage and move this annotation values into the `SriovNetworkNodeState` object
```
SchSeba, line 169: I think here is to drop the annotation from the node object

SchSeba (commented):
great work I am really waiting for this feature to go in!

I left some small comments

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Proposal
+Introduce nodes pool drain configuration and controller to meet Goals targets.
+
+
+### Workflow Description
+A new Drain controller will be introduced to manage node drain and cordon procedures. That means we don't need to do
+drain and use `drain lock` in config daemon anymore. The overall drain process will be covered by the following states:
+
+```golang
```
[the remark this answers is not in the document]
e0ne: done

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,250 @@
...
+```
+
+Drain controller will watch for node annotation, `SriovNetworkNodeState` and `SriovNetworkPoolConfig` changes:
+```golang
+// +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=SriovNetworkPoolConfig,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=SriovNetworkNodeState,verbs=get;list;watch;update;patch
+```
+
+Once `SriovNetworkNodeState.DrainStatus` will be marked as `Drain_Required` by config daemon controller add will start
```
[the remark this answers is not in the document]
e0ne: fixed

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,250 @@
...
+#### Extend existing CR SriovNetworkNodeState
+SriovNetworkNodeState already contains current state of the configuration progress in a `status` field, so we don't need
+to store drain status in node annotation. `SriovNetworkNodeStateStatus` will be extended to contain drain progress
+instead of node annotation:
+
+```golang
+// SriovNetworkNodeStateStatus defines the observed state of SriovNetworkNodeState
+type SriovNetworkNodeStateStatus struct {
+	Interfaces    InterfaceExts `json:"interfaces,omitempty"`
+    // +kubebuilder:validation:Enum=Idle;Draining,Draining_MCP_Paused,Succeeded,Failed,InProgress
```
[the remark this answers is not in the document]
e0ne: fixed

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,250 @@
...
+### Workflow Description
+A new Drain controller will be introduced to manage node drain and cordon procedures. That means we don't need to do
+drain and use `drain lock` in config daemon anymore. The overall drain process will be covered by the following states:
+
+```golang
+DrainIdle        = "Idle"
+DrainRequired     = "Drain_Required"
+DrainMcpPaused   = "Draining_MCP_Paused"
+Draining         = "Draining"
+DrainingComplete = "Draining_Complete"
```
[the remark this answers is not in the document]
e0ne: done

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,250 @@
...
+// SriovNetworkPoolConfigSpec defines the desired state of SriovNetworkPoolConfig
+type SriovNetworkPoolConfigSpec struct {
+    // +kubebuilder:validation:Minimum=0
+    // +kubebuilder:validation:Maximum=99
+    // Priority of the SriovNetworkPoolConfig, higher priority policies can override lower ones.
+    Priority int `json:"priority,omitempty"`
+    // OvsHardwareOffloadConfig describes the OVS HWOL configuration for selected Nodes
+    OvsHardwareOffloadConfig OvsHardwareOffloadConfig `json:"ovsHardwareOffloadConfig,omitempty"`
+    // NodeSelectorTerms is a list of node selectors to apply SriovNetworkPoolConfig
+    NodeSelectorTerms *v1.NodeSelectorTerm `json:"nodeSelectorTerms,omitempty"`
```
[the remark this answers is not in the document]
e0ne: IMO, `NodeSelectorTerm` is more adjustable

doc/design/parallel-node-config.md
e0ne: fixed

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,250 @@
...
+    DrainConfig DrainConfigSpec `json:"drainConfig,omitempty"`
```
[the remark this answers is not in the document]
e0ne: It's introduced in case if we decide add more drain-related config options (e.g. `false=true`) so we won't need to change API in the future

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,254 @@
...
+    - key: some-label
+      operator: In
+      values:
+        - val-2
+  - matchExpressions:
+    - key: other-label
+      operator: "Exists"
+      drainConfig:
+        maxUnavailable: 5
+        ovsHardwareOffloadConfig: {}
```
SchSeba, line 150: the indentation here is not right I think

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,254 @@
...
+
+Default pool should be always created with the lowest priority:
+```yaml
+apiVersion: v1
+kind: SriovNetworkPoolConfig
+metadata:
+name: default
+namespace: network-operator
+spec:
+priority: 0
```
SchSeba, line 161: the indentation here is not right I think

doc/design/parallel-node-config.md
zeeke: Is it the same as `maxParallelNodeConfiguration` ? I'm fine with both names, just trying to get if they refer to the same concept

zeeke:
@e0ne , @SchSeba I was thinking at the following scenario:

```
Given a cluster with 5 worker nodes (A, B, C, D, E)
And two SriovNetworkPoolConfigs:
  pool1 that targets A,B,C, priority 99, maxParallelNodeConfiguration 1
  pool2 that targets C,D,E, priority 1, maxParallelNodeConfiguration 2

When the user creates a policy that applies to C,D,E
Then C, D, E starts configuring immediately, because:
- C belongs to pool1, where there is no other node configuring ATM
- D and E belongs to pool2 and it has maxParalleNodeConfiguration = 2
```

If my considerations are right, please add this scenario to a "Examples" section

adrianchiris:
> @e0ne , @SchSeba, I was thinking of the following scenario:
> 
> ```
> Given a cluster with 5 worker nodes (A, B, C, D, E)
> And two SriovNetworkPoolConfigs:
>   pool1 that targets A,B,C, priority 1, maxParallelNodeConfiguration 1
>   pool2 that targets C,D,E, priority 99, maxParallelNodeConfiguration 2
> 
> When the user creates a policy that applies to C,D,E
> Then C, D, E starts configuring immediately, because:
> - C belongs to pool1 (as it has the highest priority), where there is no other node configuring ATM
> - D and E belongs to pool2 and it has maxParalleNodeConfiguration = 2
> ```
> 
> If my considerations are right, please add this scenario to a "Examples" section.

+1 on emphasizing that a node may belong to at most a single pool (the one with the most important priority)

SchSeba:
yep great example!

SchSeba:
also, another point is if the node selector is good for 2 pools and the priority is equal we will select the first pool in alphabetic order

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,254 @@
...
+
+# Parallel SR-IOV configuration
+
+## Summary
+Allow SR-IOV Network Operator to configure more than one node at the moment.
+
+## Motivation
+SR-IOV Network Operator configures SR-IOV one node at a time and one nic at a same time. That means we’ll need to wait
+hours or even days to configure all NICs  on large cluster deployments. With multi-NICs deployments Operator configures
+NICs one by one on each node which  leads to a lot of unnecessary node drain calls during SR-IOV configuration. Also
```
adrianchiris, line 20: are we really saving up on drain calls ?

even currently sriov-network-config-daemon will configure all relevant nic in the node. it just does it serially.

doc/design/parallel-node-config.md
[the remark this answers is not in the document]
adrianchiris: +1 to sticking with standard k8s NodeSelectorTerms.
these can be translated to machineconfig selector if needed.

WDYT?

doc/design/parallel-node-config.md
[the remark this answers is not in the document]
adrianchiris: +1 on having this a stuct for future extensability.

doc/design/parallel-node-config.md
adrianchiris: yea i believe this needs to be removed.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,254 @@
...
+	DrainStatus    string        `json:"drainStatus,omitempty"`
+    // +kubebuilder:validation:Enum=Idle,Succeeded,Failed,InProgress
+	SyncStatus    string        `json:"syncStatus,omitempty"`
+	LastSyncError string        `json:"lastSyncError,omitempty"`
+}
+```
+
+#### Extend existing CR SriovNetworkPoolConfig
+SriovNetworkPoolConfig is used only for OpenShift to provide configuration for
+OVS Hardware Offloading. We can extend it to add configuration for the drain
```
adrianchiris, line 102: can you add some info that `SriovNetworkPoolConfig` will now be relevant for all cluster types (kubernetes and openshift)

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,250 @@
...
+SriovNetworkNodeState already contains current state of the configuration progress in a `status` field, so we don't need
+to store drain status in node annotation. `SriovNetworkNodeStateStatus` will be extended to contain drain progress
+instead of node annotation:
+
+```golang
+// SriovNetworkNodeStateStatus defines the observed state of SriovNetworkNodeState
+type SriovNetworkNodeStateStatus struct {
+	Interfaces    InterfaceExts `json:"interfaces,omitempty"`
+    // +kubebuilder:validation:Enum=Idle;Draining,Draining_MCP_Paused,Succeeded,Failed,InProgress
+	DrainStatus    string        `json:"drainStatus,omitempty"`
```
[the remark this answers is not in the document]
e0ne: done

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,254 @@
...
+  `SriovNetworkNodeState` object
+* introduce new API changes to support pool of nodes to be drained in a parallel:
+  at this phase we introduce new fields in SriovNetworkPoolConfig and modify Drain controller to watch for
+  the specified CRs and proceed drain in a parallel per node pool configuration
+
+All phases should be implemented one-by-one in a separate PRs in the order above.
+
+### Upgrade & Downgrade considerations
+After operator upgrade we have to support `sriovnetwork.openshift.io/state` node annotation to
+`SriovNetworkNodeState.Status.DrainStatus` migration. This logic will be implemented in a Drain controller and should be
```
adrianchiris, line 182: shouldnt we implement the logic in config daemon as it is the one who triggers drain flow ?

doc/design/parallel-node-config.md
e0ne: fixed

doc/design/parallel-node-config.md
e0ne: Changed to `MaxParallelNodeConfiguration`

doc/design/parallel-node-config.md
e0ne: fixed

doc/design/parallel-node-config.md
e0ne: done

doc/design/parallel-node-config.md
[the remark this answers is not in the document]
SchSeba: can you please change to something like what we have in the PR?

```
// SriovNetworkPoolConfigSpec defines the desired state of SriovNetworkPoolConfig
type SriovNetworkPoolConfigSpec struct {
	// OvsHardwareOffloadConfig describes the OVS HWOL configuration for selected Nodes
	OvsHardwareOffloadConfig OvsHardwareOffloadConfig `json:"ovsHardwareOffloadConfig,omitempty"`

	// nodeSelector specifies a label selector for Nodes
	NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`

	// maxUnavailable defines either an integer number or percentage
	// of nodes in the pool that can go Unavailable during an update.
	//
	// A value larger than 1 will mean multiple nodes going unavailable during
	// the update, which may affect your workload stress on the remaining nodes.
	// Drain will respect Pod Disruption Budgets (PDBs) such as etcd quorum guards,
	// even if maxUnavailable is greater than one.
}
```

PR #484, opened 2023-08-02
Fix Plugin Config Daemon node selector

wizhaoredhat:
When node selectors are added via "configDaemonNodeSelector" via the sriovoperatorconfigs CRD then completely removed, the previous node selectors would remain for the plugin config daemons (sriov-device-plugin). This was not cleaned up, because the function "syncPluginDaemonSet" bails out when "configDaemonNodeSelector" is empty. For example:

```
Step 1) Create 3 labels in node selector:
configDaemonNodeSelector = {"labelA": "", "labelB": "", "labelC": ""}
device-plugin DS NodeSelector = {"labelA": "", "labelB": "", "labelC": ""} 

Step 2) Remove all labels in node selector:
configDaemonNodeSelector = {}
device-plugin DS NodeSelector = {"labelA": "", "labelB": "", "labelC": ""}
```

The expectation is that the device plugin will revert to the default node selectors (e.g. {"node-role.kubernetes.io/worker": "", "kubernetes.io/os": "linux"})

This change will make the node selector field in the sriov-device-plugin to take in a variable. This default node selector value is then shared with the code that updates the node selector.

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -335,6 +335,7 @@ func (r *SriovNetworkNodePolicyReconciler) syncPluginDaemonObjs(ctx context.Cont
 	data.Data["ReleaseVersion"] = os.Getenv("RELEASEVERSION")
 	data.Data["ResourcePrefix"] = os.Getenv("RESOURCE_PREFIX")
 	data.Data["ImagePullSecrets"] = GetImagePullSecrets()
+	data.Data["NodeSelectorField"] = GetDefaultNodeSelector()
 
 	objs, err := renderDsForCR(constants.PluginPath, &data)
 	if err != nil {
--- a/controllers/sriovoperatorconfig_controller.go
+++ b/controllers/sriovoperatorconfig_controller.go
@@ -154,9 +154,6 @@ func (r *SriovOperatorConfigReconciler) syncPluginDaemonSet(ctx context.Context,
 
 	names := []string{"sriov-cni", "sriov-device-plugin"}
 
-	if len(dc.Spec.ConfigDaemonNodeSelector) == 0 {
-		return nil
-	}
 	for _, name := range names {
 		err := r.Client.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, ds)
 		if err != nil {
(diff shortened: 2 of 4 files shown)


wizhaoredhat:
@bn222

controllers/sriovoperatorconfig_controller.go
```diff
@@ -147,16 +147,48 @@ func (r *SriovOperatorConfigReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Complete(r)
 }
 
+// This function will attempt to render the plugin daemon sets to retrieve the default node selector
+// when successful it will set the input node selector.
+func (r *SriovOperatorConfigReconciler) getDefaultPluginDsNodeSelector(name string, m *map[string]string) error {
```
SchSeba, line 152: I think a better solution is just to have the default nodeSelector for the device plugin as a variable in the operator and just render it

bn222: Do you mean that in the template, we should hardcode the default nodeSelector? If yes, I agree. That would be concise.

SchSeba: I mean to have the nodeSelector in the API helper as a map
```
nodeSelector:
        kubernetes.io/os: linux
        node-role.kubernetes.io/worker:
```
and in the yaml we have 

```
nodeSelector:
  {nodeSelector}
```

so we don't need to parse the yaml just for that

wizhaoredhat: @SchSeba Thanks for the suggestion. I did think about that but in Openshift, we have "beta.kubernetes.io/os=linux", but in upstream we have "kubernetes.io/os=linux". Thus it doesn't match correctly in OCP.

Not sure at which point OCP coverts kubernetes.io/os=linux => beta.kubernetes.io/os=linux

wizhaoredhat: ^ Sorry scratch this comment. Trying a different method now.

wizhaoredhat: @SchSeba Please see the new solution now. I think it is a lot cleaner.

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -335,6 +335,7 @@ func (r *SriovNetworkNodePolicyReconciler) syncPluginDaemonObjs(ctx context.Cont
 	data.Data["ReleaseVersion"] = os.Getenv("RELEASEVERSION")
 	data.Data["ResourcePrefix"] = os.Getenv("RESOURCE_PREFIX")
 	data.Data["ImagePullSecrets"] = GetImagePullSecrets()
+	data.Data["NodeSelectorField"] = GetDefaultNodeSelector()
```
zeeke, line 338: I think we are missing a scenario here. Consider the following steps:
a. The user installs the operator
b. the user configures SriovOperatorConfig.Spec.ConfigDaemonNodeSelector = map("x": "y")
c. the config reconciler sets the plugin DaemonSet NodeSelector to "x: y"
d. the user configures a policy
c. the sriovnetworknodepolicy reconciler sets the DaemonSet node selector back to  `GetDefaultNodeSelector()`

Am I missing something here?

zeeke: I believe you are right. sriovoperatorconfig_controller.go reacts on DeamonSet objects.

zeeke (approved):
LGTM

wizhaoredhat:
@adrianchiris @e0ne  Please take a look when you have a chance. Thanks!

SchSeba:
Hi @adrianchiris,
by default the config-daemon have node selector on `workers` you can check in the bin/data folder

e0ne (approved):

PR #486, opened 2023-08-03
Kernel params set by the config daemon should be verified

wizhaoredhat:
In the case of vfio-pci, the config daemon might not notice it has to reboot to add kernel parameters. This could happen if the vfio driver state was updated but the config daemon was interrupted (e.g policy was removed and readded quickly).

This could lead to a pending change to the kernel parameters via grubby or os-tree but without a reboot, these changes wouldn't be applied.

This change introduces a map to hold the desired kernel parameters that the config daemon wishes to apply. This map is then validated against "/proc/cmdline" to see if the desired kernel parameters are actually applied. If not then there will be a subsequent attempt to apply these parameters.

We need to make sure that the kernel parameters are set, thus we should pass the errors up to the config daemon instead of silently consuming the errors when setting kernel parameters.

### DIFF
--- a/pkg/plugins/generic/generic_plugin.go
+++ b/pkg/plugins/generic/generic_plugin.go
@@ -2,6 +2,7 @@ package generic
 
 import (
 	"bytes"
+	"errors"
 	"os/exec"
 	"reflect"
 	"strconv"
@@ -116,7 +117,7 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
 	needDrain = p.needDrainNode(new.Spec.Interfaces, new.Status.Interfaces)
 	needReboot, err = p.needRebootNode(new)
 	if err != nil {
-		return false, false, err
+		return needDrain, needReboot, err
 	}
 
 	if needReboot {
@@ -173,6 +174,10 @@ func (p *GenericPlugin) Apply() error {
 	}
 
 	if err := utils.SyncNodeState(p.DesireState, pfsToSkip); err != nil {
+		// Catch the "cannot allocate memory" error and try to use PCI realloc
+		if errors.Is(err, syscall.ENOMEM) {
+			p.addToDesiredKernelArgs(utils.KernelArgPciRealloc)
+		}
 		return err
 	}
 	p.LastState = &sriovnetworkv1.SriovNetworkNodeState{}
--- a/pkg/plugins/generic/generic_plugin.go
+++ b/pkg/plugins/generic/generic_plugin.go
@@ -47,14 +47,15 @@ type DriverState struct {
 type DriverStateMapType map[uint]*DriverState
 
 type GenericPlugin struct {
-	PluginName     string
-	SpecVersion    string
-	DesireState    *sriovnetworkv1.SriovNetworkNodeState
-	LastState      *sriovnetworkv1.SriovNetworkNodeState
-	DriverStateMap DriverStateMapType
-	RunningOnHost  bool
-	HostManager    host.HostManagerInterface
-	StoreManager   utils.StoreManagerInterface
+	PluginName        string
+	SpecVersion       string
+	DesireState       *sriovnetworkv1.SriovNetworkNodeState
+	LastState         *sriovnetworkv1.SriovNetworkNodeState
+	DriverStateMap    DriverStateMapType
+	DesiredKernelArgs map[string]bool
+	RunningOnHost     bool
+	HostManager       host.HostManagerInterface
+	StoreManager      utils.StoreManagerInterface
 }
 
 const scriptsPath = "bindata/scripts/enable-kargs.sh"
@@ -84,12 +85,13 @@ func NewGenericPlugin(runningOnHost bool, hostManager host.HostManagerInterface,
 		DriverLoaded:   false,
 	}
 	return &GenericPlugin{
-		PluginName:     PluginName,
-		SpecVersion:    "1.0",
-		DriverStateMap: driverStateMap,
-		RunningOnHost:  runningOnHost,
-		HostManager:    hostManager,
-		StoreManager:   storeManager,
+		PluginName:        PluginName,
+		SpecVersion:       "1.0",
+		DriverStateMap:    driverStateMap,
+		DesiredKernelArgs: make(map[string]bool),
+		RunningOnHost:     runningOnHost,
+		HostManager:       hostManager,
+		StoreManager:      storeManager,
 	}, nil
 }
 
@@ -112,7 +114,10 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
 	p.DesireState = new
 
 	needDrain = p.needDrainNode(new.Spec.Interfaces, new.Status.Interfaces)
-	needReboot = needRebootNode(new, p.DriverStateMap)
+	needReboot, err = p.needRebootNode(new)
+	if err != nil {
+		return false, false, err
+	}
 
 	if needReboot {
 		needDrain = true
@@ -197,28 +202,28 @@ func needDriverCheckVdpaType(state *sriovnetworkv1.SriovNetworkNodeState, driver
 	return false
 }
 
-func tryEnableIommuInKernelArgs() (bool, error) {
-	glog.Info("generic-plugin tryEnableIommuInKernelArgs()")
-	args := [2]string{"intel_iommu=on", "iommu=pt"}
+// setKernelArg Tries to add the kernel args via ostree or grubby.
+func setKernelArg(karg string) (bool, error) {
+	glog.Info("generic-plugin setKernelArg()")
 	var stdout, stderr bytes.Buffer
-	cmd := exec.Command("/bin/sh", scriptsPath, args[0], args[1])
+	cmd := exec.Command("/bin/sh", scriptsPath, karg)
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 
 	if err := cmd.Run(); err != nil {
 		// if grubby is not there log and assume kernel args are set correctly.
 		if isCommandNotFound(err) {
-			glog.Error("generic-plugin tryEnableIommuInKernelArgs(): grubby command not found. Please ensure that kernel args intel_iommu=on iommu=pt are set")
+			glog.Errorf("generic-plugin setKernelArg(): grubby or ostree command not found. Please ensure that kernel arg %s are set", karg)
 			return false, nil
 		}
-		glog.Errorf("generic-plugin tryEnableIommuInKernelArgs(): fail to enable iommu %s: %v", args, err)
+		glog.Errorf("generic-plugin setKernelArg(): fail to enable kernel arg %s: %v", karg, err)
 		return false, err
 	}
 
 	i, err := strconv.Atoi(strings.TrimSpace(stdout.String()))
 	if err == nil {
 		if i > 0 {
-			glog.Infof("generic-plugin tryEnableIommuInKernelArgs(): need to reboot node")
+			glog.Infof("generic-plugin setKernelArg(): need to reboot node for kernel arg %s", karg)
 			return true, nil
 		}
 	}
@@ -232,6 +237,48 @@ func isCommandNotFound(err error) bool {
 		}
 	}
 	return false
+}
+
+// addToDesiredKernelArgs Should be called to queue a kernel arg to be added to the node.
+func (p *GenericPlugin) addToDesiredKernelArgs(karg string) {
+	if _, ok := p.DesiredKernelArgs[karg]; !ok {
+		glog.Infof("generic-plugin addToDesiredKernelArgs(): Adding %s to desired kernel arg", karg)
+		p.DesiredKernelArgs[karg] = false
+	}
+}
+
+// syncDesiredKernelArgs Should be called to set all the kernel arguments. Returns bool if node update is needed.
+func (p *GenericPlugin) syncDesiredKernelArgs() (bool, error) {
+	needReboot := false
+	if len(p.DesiredKernelArgs) == 0 {
+		return false, nil
+	}
+	kargs, err := utils.GetCurrentKernelArgs(false)
+	if err != nil {
+		return false, err
+	}
+	for desiredKarg, attempted := range p.DesiredKernelArgs {
+		set := utils.IsKernelArgsSet(kargs, desiredKarg)
+		if !set {
+			if attempted {
+				glog.V(2).Infof("generic-plugin syncDesiredKernelArgs(): previously attempted to set kernel arg %s", desiredKarg)
+			}
+			// There is a case when we try to set the kernel argument here, the daemon could decide to not reboot because
+			// the daemon encountered a potentially one-time error. However we always want to make sure that the kernel
+			// argument is set once the daemon goes through node state sync again.
+			update, err := setKernelArg(desiredKarg)
+			if err != nil {
+				glog.Errorf("generic-plugin syncDesiredKernelArgs(): fail to set kernel arg %s: %v", desiredKarg, err)
+				return false, err
+			}
+			if update {
+				needReboot = true
+				glog.V(2).Infof("generic-plugin syncDesiredKernelArgs(): need reboot for setting kernel arg %s", desiredKarg)
+			}
+			p.DesiredKernelArgs[desiredKarg] = true
+		}
+	}
+	return needReboot, nil
 }
 
 func (p *GenericPlugin) needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.InterfaceExts) (needDrain bool) {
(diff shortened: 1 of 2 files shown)


wizhaoredhat:
@bn222

pkg/plugins/generic/generic_plugin.go
```diff
@@ -26,6 +26,7 @@ type GenericPlugin struct {
 	LastState            *sriovnetworkv1.SriovNetworkNodeState
 	LoadVfioDriver       uint
 	LoadVirtioVdpaDriver uint
+	DesiredKernelParams  map[string]uint
```
SchSeba, line 29: in general I don't think we need this if we want to be sure it's loaded lets just check the proc cmdline file.

bn222:
I don't think this is fixing the issue. We aren't **reconciling** the state of the system to a desired state **whenever** the current state is different. What we want is to trigger setting the desired kargs at any point when they are set to something else then we expect.

Reference for this: OCPBUGS-16909

It's a step in the good direction though, by performing the necessary check when we are making a change. We just need to have this check be executed more frequently.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -124,6 +124,9 @@ func (p *GenericPlugin) Apply() error {
 	}
 
 	if err := utils.SyncNodeState(p.DesireState, pfsToSkip); err != nil {
+		if strings.Contains(err.Error(), "cannot allocate memory") {
+			p.AddToDesiredKernelParams(utils.KernelParamPciRealloc)
+		}
```
bn222, line 129: iiuc, we might reboot twice?

bn222:
We have one remaining situation to fix due to how enable-kargs.sh works. Let's say we've run enable-kargs.sh, and we returned non-zero, i.e. we made a change and we need to reboot. Now, before we actually get to the point of the reboot, the daemon is restarted. Now, assume again execution gets to the point of running enable-kargs.sh but this time, since rpm-ostree kargs contains the _pending_ kargs, it will return 0 since from the perspective of rpm-ostree we are good and we won't reboot. We need to fix this on line 10 in enable-kargs.sh

pkg/utils/utils.go
```diff
@@ -32,6 +32,7 @@ const (
 	sysBusPciDrivers      = "/sys/bus/pci/drivers"
 	sysBusPciDriversProbe = "/sys/bus/pci/drivers_probe"
 	sysClassNet           = "/sys/class/net"
+	sysKernelCmdLine      = "/proc/cmdline"
```
bn222, line 35: I don't like the sys in the name of the variable. It's not `/sys`

pkg/plugins/generic/generic_plugin.go
wizhaoredhat: We need to record what kernel parameters we want to load somewhere. It would be nice for this to happen anywhere in the generic plugin. This is why I think it is still needed.

pkg/plugins/generic/generic_plugin.go
wizhaoredhat: Yes it is possible to reboot multiple times depending on when AddToDesiredKernelParams gets a unique kernel param.

pkg/utils/utils.go
```diff
@@ -51,6 +55,27 @@ func init() {
...
+	if !chroot {
+		path = "/host" + path
+	}
+	cmdLine, err := os.ReadFile(path)
+	if err != nil {
+		glog.Warningf("IsKernelCmdLineParamSet(): Error reading %s, Please check the Kernel Params manually!", sysKernelCmdLine)
+		// Although intuitively we should report false here. However if we cannot reliably read the Kernel's cmd line params
+		// then it is unknown whether it was successfully set, and to avoid unnecessary reboots we should just post a warning
+		// and return true.
+		return true
```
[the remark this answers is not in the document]
wizhaoredhat: I see what you mean. I wanted to avoid exiting the daemon with an error, but you make a good point. Let's error out.

pkg/utils/utils.go
wizhaoredhat: My mistake, copy paste issue =) .
I will rename it to procKernelCmdLine

wizhaoredhat:
> We have one remaining situation to fix due to how enable-kargs.sh works. Let's say we've run enable-kargs.sh, and we returned non-zero, i.e. we made a change and we need to reboot. Now, before we actually get to the point of the reboot, the daemon is restarted. Now, assume again execution gets to the point of running enable-kargs.sh but this time, since rpm-ostree kargs contains the _pending_ kargs, it will return 0 since from the perspective of rpm-ostree we are good and we won't reboot. We need to fix this on line 10 in enable-kargs.sh

@bn222 
I thought the logic works in your case. But maybe I misunderstood.
```
        if [[ $args != *${t}* ]];then                                                                             // This would be true in your case because cmdLine does not have the kernel parameter (we made a change and we didn't reboot). This will always be true because we didn't reboot.
            ...
            let ret++                                                                                                        // We increment this value and return it.
        fi
```
If ret > 0 then we will be marked for reboot.

wizhaoredhat:
> I don't think this is fixing the issue. We aren't **reconciling** the state of the system to a desired state **whenever** the current state is different. What we want is to trigger setting the desired kargs at any point when they are set to something else then we expect.
> 
> Reference for this: OCPBUGS-16909
> 
> It's a step in the good direction though, by performing the necessary check when we are making a change. We just need to have this check be executed more frequently.

Yes, I agree this is an issue. I was thinking of calling it earlier in `nodeStateSyncHandler`.

bn222:
> > We have one remaining situation to fix due to how enable-kargs.sh works. Let's say we've run enable-kargs.sh, and we returned non-zero, i.e. we made a change and we need to reboot. Now, before we actually get to the point of the reboot, the daemon is restarted. Now, assume again execution gets to the point of running enable-kargs.sh but this time, since rpm-ostree kargs contains the _pending_ kargs, it will return 0 since from the perspective of rpm-ostree we are good and we won't reboot. We need to fix this on line 10 in enable-kargs.sh
> 
> @bn222 I thought the logic works in your case. But maybe I misunderstood.
> 
> ```
>         if [[ $args != *${t}* ]];then                                                                             // This would be true in your case because cmdLine does not have the kernel parameter (we made a change and we didn't reboot). This will always be true because we didn't reboot.
>             ...
>             let ret++                                                                                                        // We increment this value and return it.
>         fi
> ```
> 
> If ret > 0 then we will be marked for reboot.

Yes, that happens in normal execution. What happens when we kill the pod immediately after we are marked for reboot? I think we will run through the whole logic from start again, but this time we will not mark for reboot, and we will just wait. iow, I believe there is an edge case due to non-idempotency.

SchSeba:
Hi folks,

let me add some ideas here.

we can improve this one with a simple fix I think.
instead of checking the rpm-ostree command return to see if we change it or not we just need to look on what is really in the system by doing `cat /proc/cmdline`

and I don't think we need to revalidate this every few seconds it's a waster of resource as this variable is loaded or unloaded after reboot you can't do it when the system is already running.

meaning if the user disable it but didn't reboot all good we still have it.
and after reboot the operator will see that the variable is not there and will apply it and reboot.

so again I think the only change needed is to check if we need to reboot by looking on the source of true that is /proc/cmdline and not the output of the rpm-ostree or grubby

bn222:
that is correct and also what I'm proposing. No need to recheck that every time. A reboot implies a start of execution from the beginning, but we do need to check /proc/cmdline instead.

While at it, please move the .sh to .go (maybe as a separate patch). I don't see any need to use .sh here.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -246,6 +277,6 @@ func needRebootNode(state *sriovnetworkv1.SriovNetworkNodeState, loadVfioDriver
 	if update {
 		glog.V(2).Infof("generic-plugin needRebootNode(): need reboot for updating switchdev device configuration")
 	}
-	needReboot = needReboot || update
+	needReboot = needReboot || update || p.SetAllDesiredKernelParams()
```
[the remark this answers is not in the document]
wizhaoredhat: Good point, I will pass by reference `needReboot` into `SetAllDesiredKernelParams` to make it clearer.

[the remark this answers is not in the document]
bn222: no, that is still weird. Can we have two functions? One to check if we need to do something, and the other do actually do something?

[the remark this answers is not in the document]
bn222: So can you have each function call a third function that extracts the same logic part? Genuine question, not a request.

bn222:
LGTM from me on this. We would want to have a separate go function that does the configuring and another that checks if the config needs to be done), but we will address that in a PR that cleans up the .sh file (by moving that functionality into .go) and separates out the two tasks (check & set).

wizhaoredhat:
@adrianchiris @e0ne Could you PTAL, Thanks!

pkg/plugins/generic/generic_plugin.go
zeeke: > @zeeke Could you elaborate more on why adding a state to GeneralPlugin will make debugging sessions harder?

When looking at logs, we will not be able to fully understand the state of the operator, as it will have internal variables.

> Could you explain by adding the map and functions in its own private package, what exactly that helps to do?

It improves the design, as part of logic encapsulation. Making that public means they are meant to be called by other packages, while I suppose it's a bad idea to call `SetAllDesiredKernelParams()` and `AddToDesiredKernelParams()` from other components, isn't it?

wizhaoredhat: @zeeke I see what you mean by adding state variables is not very good. I would like to compare `rpm-ostree kargs` vs. `/proc/cmdline`. BUT even when rebooting there is no grantee that `rpm-ostree kargs = /proc/cmdline`. I believe this is what @SchSeba was also thinking. But in some environments:
`--import-proc-cmdline             Instead of modifying old kernel arguments, we modify args from current /proc/cmdline (the booted deployment)`
is not used with `rpm-ostree kargs` (e.g. OCP).

I made the functions private to the the GenericPlugin such that they can't be called externally. Thanks for catching this!

wizhaoredhat:
@adrianchiris @e0ne  Please take a look when you have a chance. Thanks!

pkg/utils/utils.go
```diff
@@ -51,6 +55,23 @@ func init() {
 	ClusterType = os.Getenv("CLUSTER_TYPE")
 }
 
+// IsKernelCmdLineParamSet This checks if the kernel cmd line is set properly.
+func IsKernelCmdLineParamSet(param string, chroot bool) (bool, error) {
```
bn222, line 60: is `chroot` ever not false?

wizhaoredhat: I am just following utils.go : IsKernelLockdownMode() styling.
If the code chooses to call utils.go "Chroot(path string)" then we need it to be set true.

bn222:
LGTM

pkg/plugins/generic/generic_plugin.go
```diff
@@ -233,6 +238,46 @@ func isCommandNotFound(err error) bool {
 	return false
 }
 
+// addToDesiredKernelParams Should be called to queue a kernel param to be added to the node.
+func (p *GenericPlugin) addToDesiredKernelParams(kparam string) {
+	if _, ok := p.DesiredKernelParams[kparam]; !ok {
+		glog.Infof("generic-plugin addToDesiredKernelParams(): Adding %s to desired kernel params", kparam)
+		// element "uint" is a counter of number of attempts to set the kernel param
+		p.DesiredKernelParams[kparam] = 0
```
SchSeba, line 246: why do we need a counter?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -196,28 +201,28 @@ func needDriverCheckVdpaType(state *sriovnetworkv1.SriovNetworkNodeState, driver
 	return false
 }
 
-func tryEnableIommuInKernelArgs() (bool, error) {
-	glog.Info("generic-plugin tryEnableIommuInKernelArgs()")
-	args := [2]string{"intel_iommu=on", "iommu=pt"}
+// trySetKernelParam Tries to add the kernel param via ostree or grubby.
+func trySetKernelParam(kparam string) (bool, error) {
```
adrianchiris, line 205: shall this be renamed to setKernelParam ?
we should never call this one for an already set parameter now.

also now, i dont think we need to return the number of parameters configured in enable-kargs.sh.
so this bit can be removed from script IMO.

pkg/utils/utils.go
```diff
@@ -51,6 +56,23 @@ func init() {
...
```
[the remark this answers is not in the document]
bn222: +1 great idea. imo better name is then: `getCurrentKernelArgs`. I prefer "args" instead of "params" for sure (the difference is quite formal).  In addition, since we are reading /proc/cmdline we are getting the currently running cmdline args so I prefer to add `Current` in the name of the function. That would allow us to also create a similar function that gets the kernel args for the next boot.

Btw, also rename the PR to "Kernel args set by.."

pkg/plugins/generic/generic_plugin.go
bn222: +1 and same comment wrt param vs args (please use latter terminology).

pkg/plugins/generic/generic_plugin.go
wizhaoredhat: It is an inexpensive way to debug how many times we have tried to set the kernel arguments without success. I could change it to a bool, but we would lose that debug element.

pkg/plugins/generic/generic_plugin.go
wizhaoredhat: I agree, renamed to setKernelArg

pkg/plugins/generic/generic_plugin.go
```diff
@@ -167,6 +169,9 @@ func (p *GenericPlugin) Apply() error {
 	}
 
 	if err := utils.SyncNodeState(p.DesireState, pfsToSkip); err != nil {
+		if strings.Contains(err.Error(), "cannot allocate memory") {
```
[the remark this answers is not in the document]
wizhaoredhat: Thanks for catching this. Made the change.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -111,7 +113,7 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
 	p.DesireState = new
 
 	needDrain = needDrainNode(new.Spec.Interfaces, new.Status.Interfaces)
-	needReboot = needRebootNode(new, p.DriverStateMap)
+	needReboot, err = p.needRebootNode(new)
```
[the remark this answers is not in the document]
wizhaoredhat: I agree. Now doing early return.

pkg/utils/utils.go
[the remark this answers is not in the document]
wizhaoredhat: Yes I made this change to GetCurrentKernelArgs(chroot bool) (string, error) . I am keeping it as a string because the "cmdLine" could have the same argument repeated.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -233,6 +238,49 @@ func isCommandNotFound(err error) bool {
...
+		glog.Infof("generic-plugin addToDesiredKernelArgs(): Adding %s to desired kernel arg", karg)
+		// element "uint" is a counter of number of attempts to set the kernel argument
+		p.DesiredKernelArgs[karg] = 0
+	}
+}
+
+// syncDesiredKernelArgs Should be called to set all the kernel arguments. Returns bool if node update is needed.
+func (p *GenericPlugin) syncDesiredKernelArgs() (bool, error) {
+	needReboot := false
+	if len(p.DesiredKernelArgs) != 0 {
```
bn222, line 258: This looks like an optimization. All the following lines increase nesting. I prefer to have if len() == 0 { return false, nil} at the top since that's more in line with the rest of the repo.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -233,6 +238,49 @@ func isCommandNotFound(err error) bool {
...
+				if err != nil {
+					glog.Errorf("generic-plugin syncDesiredKernelArgs(): fail to set kernel arg %s: %v", desiredKarg, err)
+					return false, err
+				}
+				if update {
+					needReboot = true
+					glog.V(2).Infof("generic-plugin syncDesiredKernelArgs(): need reboot for setting kernel arg %s", desiredKarg)
+				}
+				// Update the number of attempts we tried to set the kernel argument.
+				p.DesiredKernelArgs[desiredKarg]++
```
bn222, line 282: I'm somewhat confused again by tracking attempts. Can you provide an example of when it will be > 1?

bn222: Correct me if I'm wrong, but I don't think it will be more than 1 at any point. It seems to me that a bool is a better choice here.

pkg/plugins/generic/generic_plugin.go
wizhaoredhat: Good point, thanks! I made the changes.

pkg/plugins/generic/generic_plugin.go
wizhaoredhat: Reading through the code, I believe that is correct. I just wanted a way to track it in case it ever happens. But bool should suffice here. made the changes.

pkg/utils/utils.go
```diff
@@ -51,6 +55,25 @@ func init() {
...
+	if err != nil {
+		return "", fmt.Errorf("GetCurrentKernelArgs(): Error reading %s: %v", procKernelCmdLine, err)
+	}
+	return string(cmdLine), nil
+}
+
+// IsKernelArgsSet This checks if the kernel cmd line is set properly. Please note that the same key could be repeated
+// several times in the kernel cmd line. We can only ensure that the kernel cmd line has the key/val kernel arg that we set.
+func IsKernelArgsSet(cmdLine string, karg string) bool {
+	return strings.Contains(cmdLine, karg)
```
bn222, line 75: I'm wondering, let's say we want to check if "a=b" is set, and we have another already set "ba=a", then is it the case that this will return true? I think we might want to split the string on spaces and than compare in each string.

wizhaoredhat: Good catch! I split it in an array + compare.

bn222:
LGTM

wizhaoredhat:
@adrianchiris Please take a look, much appreciated!

Eoghan1232:
@wizhaoredhat I see the CI failed, could you take a look into that?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -285,33 +337,39 @@ func (p *GenericPlugin) needDrainNode(desired sriovnetworkv1.Interfaces, current
 	return
 }
 
-func needRebootIfVfio(state *sriovnetworkv1.SriovNetworkNodeState, driverMap DriverStateMapType) (needReboot bool) {
-	driverState := driverMap[Vfio]
+func (p *GenericPlugin) addVfioDesiredKernelArg(state *sriovnetworkv1.SriovNetworkNodeState) {
```
Eoghan1232, line 340: nit: addVfioDesiredKernelArg -> addVfioDesiredKernelArgs to keep it the same as other methods ? WDYT.
Although, we are only adding a single arg here....

Eoghan1232 (approved):
overall, PR lgtm. 
Can you confirm CI failure is due to CI itself and not this PR?

pkg/plugins/generic/generic_plugin.go
bn222: I prefer singular. It's only adding 1 arg.

zeeke:
> overall, PR lgtm. Can you confirm CI failure is due to CI itself and not this PR?

`k8s` and `ocp` jobs also fail in #518. Not related to this job

pkg/plugins/generic/generic_plugin.go
```diff
@@ -234,6 +244,48 @@ func isCommandNotFound(err error) bool {
...
+	}
+	kargs, err := utils.GetCurrentKernelArgs(false)
+	if err != nil {
+		return false, err
+	}
+	for desiredKarg, attempted := range p.DesiredKernelArgs {
+		set := utils.IsKernelArgsSet(kargs, desiredKarg)
+		if !set {
+			if attempted {
+				glog.V(2).Infof("generic-plugin syncDesiredKernelArgs(): previously attempted to set kernel arg %s", desiredKarg)
```
adrianchiris, line 269: @wizhaoredhat if you previously attempted to set the arg, it means setKernlelArg succeeded.
then why do we need to set it again, shouldnt we "continue" here ?

pkg/plugins/generic/generic_plugin.go
adrianchiris: its adding two, what am i missing ?

L343, L344

pkg/plugins/generic/generic_plugin.go
wizhaoredhat: This captures the case where we hit this case, but either the kernel arg wasn't set properly or if the reboot did not occur for some reason. This results in `IsKernelArgsSet` returning false even though it was set before. So we should try again.

pkg/plugins/generic/generic_plugin.go
wizhaoredhat: Sorry I grouped IOMMU as singular. There is infact Intel IOMMU and the general IOMMU. However this is a small NIT on naming.

PR #487, opened 2023-08-07
Render SriovNetworkNodeState before Device Plugin ConfigMap

e0ne:
Device Plugin ConfigMap relies on SriovNetworkNodeState object, so we need to render it earlier.

The issue is occured only if node stops to correspond to configDaemonNodeSelector during SR-IOV configuration, so controller will delete SriovNetworkNodeState and won't re-create it again once node starts correspond to configDaemonNodeSelector again.

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -132,16 +132,16 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(ctx context.Context, req ct
 
 	// Sort the policies with priority, higher priority ones is applied later
 	sort.Sort(sriovnetworkv1.ByPriority(policyList.Items))
+	// Sync SriovNetworkNodeState objects
+	if err = r.syncAllSriovNetworkNodeStates(ctx, defaultPolicy, policyList, nodeList); err != nil {
+		return reconcile.Result{}, err
+	}
 	// Sync Sriov device plugin ConfigMap object
 	if err = r.syncDevicePluginConfigMap(ctx, policyList, nodeList); err != nil {
 		return reconcile.Result{}, err
 	}
 	// Render and sync Daemon objects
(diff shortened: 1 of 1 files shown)


adrianchiris (approved):
device plugin config  map depends on SriovNetworkNodeState of that node, so it makes sense to sync it first.

zeeke (approved):
LGTM

Though the device plugin config map rendering logic depends on `SriovNetworkNodeState.Status`, that is written by the config-daemon. BTW, this PR avoid raising most common errors IMO

adrianchiris:
> this PR avoid raising most common errors IMO

can you elaborate ?

zeeke:
> Thanks for the review ! can you elaborate ?

Sure:
`renderDevicePluginConfigData(...)` is invoked on every reconcile loop and it tries to retrieve the SriovNetworkNodeState. ([sriovnetworknodepolicy_controller.go#L226 ](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/controllers/sriovnetworknodepolicy_controller.go#L226) ).

That value is used only if the policy has `Spec.NicSelector.NetFilter != ""` ([#L783](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/controllers/sriovnetworknodepolicy_controller.go#L783)), which reads the Status field. If it's just rendered by `syncAllSriovNetworkNodeStates(...)`, the Status is not yet populated, as that function sets the Spec fields.

But in the end, it probably works even in the above scenario: the policies are reconciled every 5 minutes, and at the second loop every resource should be in place and correctly populated. If that's the case, sorry for the noise.

adrianchiris:
> But in the end, it probably works even in the above scenario: the policies are reconciled every 5 minutes, and at the second loop every resource should be in place and correctly populated. If that's the case, sorry for the noise.

That is the  case to my understanding.

however i think we should fail (and retry reconcile) if nodestate has no status field if its a pre-req for rendering device plugin config.

PR #490, opened 2023-08-14
Return reconcile error if node state is not updated

e0ne:
NetFilter selector depends on PCI address of NICs from the node state. After PR #487 is merged we need to check if node state is updated or return an reconcile error to render device plugin config faster.


adrianchiris (approved):
LGTM

adrianchiris:
@zeeke PTAL

PR #491, opened 2023-08-16
Remove use of ioutil deprecated functions

adrianchiris:
use equivalent functions from os and io package

no more strikethrough lines in your IDE for deprecated functions ! 💪


Eoghan1232 (approved):
The changes LGTM!

The error from CodeQL doesn't have anything to do with PR IMO, so I am okay to merge this, and a follow up PR can be done to fix that issue.
Unless you want to do a quick fix for it.

PR #488, opened 2023-08-09
Drain controller

e0ne:

PR #494, opened 2023-08-23
config-daemon: Improve logging of draining state

zeeke:
Executions of `nodeUpdateHandler` are not tracked by logs,
that make debugging problems harder.

Also, adding a log line to write the node name the config-daemon is working on.

**Note**: part of this work will be superseded by
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/488

but it's still useful to deal with draining issue while waiting for that PR.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -400,16 +400,31 @@ func (dn *Daemon) nodeUpdateHandler(old, new interface{}) {
 		return
 	}
 	dn.node = node.DeepCopy()
+
 	nodes, err := dn.nodeLister.List(labels.Everything())
 	if err != nil {
+		glog.Errorf("nodeUpdateHandler(): failed to list nodes: %v", err)
 		return
 	}
-	for _, node := range nodes {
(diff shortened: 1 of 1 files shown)


SchSeba (approved):
nice work!

pkg/daemon/daemon.go
```diff
@@ -400,16 +402,31 @@ func (dn *Daemon) nodeUpdateHandler(old, new interface{}) {
...
+		drainingAnnotationValue := otherNode.Annotations[annoKey]
+		if drainingAnnotationValue == annoDraining || drainingAnnotationValue == annoMcpPaused {
+			glog.V(2).Infof("nodeUpdateHandler(): node is not drainable as [%s] has [%s == %s] ", otherNode.Name, annoKey, drainingAnnotationValue)
 			dn.drainable = false
 			return
 		}
 	}
+
+	if !dn.drainable {
+		glog.V(2).Infof("nodeUpdateHandler(): node is drainable")
```
adrianchiris, line 427: nit: node is now drainable ? 

as you only log the transition from false to true here

adrianchiris (approved):
overall LGTM, small nit

pkg/daemon/daemon.go
zeeke: agree

PR #496, opened 2023-09-04
set mellanox ib vf state to Follow

wangyd1988:
According to the mellanox 'https://enterprise-support.nvidia.com/s/article/howto-configure-sr-iov-for-connect-ib-connectx-4-with-kvm--infiniband-x' documentation, we must set the mellanox ib vf state to Follow when using the mellanox ib network card

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"math/rand"
 	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


SchSeba:
Hi @wangyd1988 

should this change be part of the https://github.com/k8snetworkplumbingwg/ib-sriov-cni instead of the operator?

adrianchiris:
@wangyd1988 are you running VM with an IB Sriov VF ?

wangyd1988:
> Hi @wangyd1988
> 
> should this change be part of the https://github.com/k8snetworkplumbingwg/ib-sriov-cni instead of the operator?

I think it's part of the operator, because pf also needs to configure state, and it's part of the vm configuration

wangyd1988:
> > Hi @wangyd1988
> > should this change be part of the https://github.com/k8snetworkplumbingwg/ib-sriov-cni instead of the operator?
> 
> I think it's part of the operator, because pf also needs to configure state, and it's part of the vm configuration

yes,

> @wangyd1988 are you running VM with an IB Sriov VF ?
> 
> this probably needs to be part of ib-sriov-cni as @SchSeba pointed out.

Yes, I ran it in a vm

adrianchiris:
@wangyd1988 can you explain your use-case and what exactly is not working ?

if you use `ib-sriov-cni` you can set the `link_state` configuration if you set it to `auto` it will have the same effect. (it also relies on upstream APIs and not driver specific)

adrianchiris:
@wangyd1988 please see the alternatives above, the recommended way is to use ib-sriov-cni to set the link state.
which utilizes non driver-specific APIs.

Issue #480, opened 2023-07-20
inconsistent logging


bn222:
We should use the same logging lib throughout the operator.

zeeke:
As discussed during last community meeting, here is a summary of logs around the operator.

### Current state:

- a. `config-daemon` uses [github.com/golang/glog](github.com/golang/glog) with verbosisty levels 0 (info), 1 (debug) and 2 (trace)
- b. SriovOperatorConfig.Spec.LogLevel 
    - b1. The field has the following documentation:
> Flag to control the log verbose level of the operator. Set to '0' to show only the basic logs. And set to '2' to show all the available logs.

- c. The field only affects the config-daemon, which changes the logging level at runtime in [pkg/daemon/daemon.go#L437](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/14c6765359f208cea9604c694e70942ce56e9a88/pkg/daemon/daemon.go#L437)


- d. The operator uses `controller-runtime/pkg/log`:
    - d1. "sigs.k8s.io/controller-runtime/pkg/log" is a proxy for "github.com/go-logr/logr", which in turn is an abstract logger
    - d2. Logger implementation is set to "go.uber.org/zap/zapcore" in [main.go#L85](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/4dd75bc7ae339e8397266580ff1b97a6b2125bdb/main.go#L85)
        - d2a. We set `zap.Options.Development = true`, that forces `encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn` [zap.go#L255](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/14c6765359f208cea9604c694e70942ce56e9a88/vendor/sigs.k8s.io/controller-runtime/pkg/log/zap/zap.go#L255)
   - d3. Every log line is logged with V(0), verbosity info
- e. The operator has a few `log.Printf` calls: (e.g. [pkg/apply/apply.go#L27](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/8074d2846191223821747c4fcd9b372c7feda577/pkg/apply/apply.go#L27))


### My proposal here includes:

- p1. Reconfigure the `zap` logging level **at runtime** of the operator following `SriovOperatorConfig.Spec.LogLevel` value.
- p2. Remove `zap.Options.Development = true` as it inhibits any loglevel change (tested manually)
[...]

PR #497, opened 2023-09-05
Parallel NICs configuration

e0ne:


e0ne:
UPDATE: environment variables exist in pod after node reboot

api/v1/helper.go
```diff
@@ -121,7 +121,7 @@ func IsSupportedModel(vendorID, deviceID string) bool {
 			return true
 		}
 	}
-	log.Info("IsSupportedModel():", "Unsupported model:", "vendorId:", vendorID, "deviceId:", deviceID)
+	glog.Infof("IsSupportedModel(): Unsupported model: vendorId=%s, deviceId=%s", vendorID, deviceID)
```
adrianchiris, line 124: we really need everything to use container runtime logger, as this  helper is used by the controller as well, we dont want to use two different loggers.

maybe we need to tackle  switching config daemon to controller runtime logger as a separate PR ?

bn222: and in that separate PR, finally close https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/480 ;)

pkg/utils/utils.go
```diff
@@ -197,8 +198,71 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
 }
 
 // SyncNodeState Attempt to update the node state to match the desired state
-func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState, pfsToConfig map[string]bool) error {
-	return ConfigSriovInterfaces(newState.Spec.Interfaces, newState.Status.Interfaces, pfsToConfig)
+func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState, pfsToConfig map[string]bool, parallelNicConfig bool) error {
+	if !parallelNicConfig {
+		return ConfigSriovInterfaces(newState.Spec.Interfaces, newState.Status.Interfaces, pfsToConfig)
+	}
+	return ConfigSriovInterfacesInParallel(newState.Spec.Interfaces, newState.Status.Interfaces, pfsToConfig)
```
zeeke, line 205: nit: use positve condition for readability
```suggestion
	if parallelNicConfig {
		return ConfigSriovInterfacesInParallel(newState.Spec.Interfaces, newState.Status.Interfaces, pfsToConfig)		
	}

	return ConfigSriovInterfaces(newState.Spec.Interfaces, newState.Status.Interfaces, pfsToConfig)
```

pkg/utils/utils.go
```diff
@@ -197,8 +198,71 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
+				go func(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt) {
+					if err := configSriovDevice(iface, ifaceStatus); err != nil {
+						glog.Errorf("ConfigSriovInterfacesInParallel(): fail to configure sriov interface %s: %v. resetting interface.", iface.PciAddress, err)
+						result = err
+						if resetErr := resetSriovDevice(*ifaceStatus); resetErr != nil {
+							glog.Errorf("SyncNodeState(): fail to reset on error SR-IOV interface: %s", resetErr)
+							result = resetErr
+						}
+					}
+					wg.Done()
```
zeeke, line 242: Please, put it in a `defer` statement at the beginning of the function

pkg/utils/utils.go
```diff
@@ -197,8 +198,71 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
+				if !NeedUpdate(&iface, &ifaceStatus) {
+					glog.V(2).Infof("syncNodeState(): no need update interface %s", iface.PciAddress)
+					break
+				}
+
+				wg.Add(1)
+				go func(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt) {
+					if err := configSriovDevice(iface, ifaceStatus); err != nil {
+						glog.Errorf("ConfigSriovInterfacesInParallel(): fail to configure sriov interface %s: %v. resetting interface.", iface.PciAddress, err)
+						result = err
```
zeeke, line 236: What about using [`errgroup.Group`](https://cs.opensource.google/go/x/sync/+/v0.4.0:errgroup/errgroup.go;l=22) to handle this synchronization? I think it would be useful, as we are dealing with errors here.

pkg/utils/utils.go
```diff
@@ -197,8 +198,71 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
+						if resetErr := resetSriovDevice(*ifaceStatus); resetErr != nil {
+							glog.Errorf("SyncNodeState(): fail to reset on error SR-IOV interface: %s", resetErr)
+							result = resetErr
+						}
+					}
+					wg.Done()
+				}(&iface, &ifaceStatus)
```
zeeke, line 243: It's not a good idea to use pointers here, as the variable address is always the same. Please, take a look at this modified example of WaitGroup: https://go.dev/play/p/8CuE98kz-Eg. If you try running it, you'll get the same print every time:
```
http://www.example.com/
http://www.example.com/
http://www.example.com/

Program exited.
```

zeeke (commented):
Left a few comments

api/v1/helper.go
adrianchiris: #527 addresses most issues. #480 was updated to reflect remaining work.

pkg/utils/utils.go
```diff
@@ -200,8 +201,71 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
-func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState, pfsToConfig map[string]bool) error {
-	return ConfigSriovInterfaces(newState.Spec.Interfaces, newState.Status.Interfaces, pfsToConfig)
+func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState, pfsToConfig map[string]bool, parallelNicConfig bool) error {
+	if !parallelNicConfig {
+		return ConfigSriovInterfaces(newState.Spec.Interfaces, newState.Status.Interfaces, pfsToConfig)
+	}
+	return ConfigSriovInterfacesInParallel(newState.Spec.Interfaces, newState.Status.Interfaces, pfsToConfig)
+}
+
+func ConfigSriovInterfacesInParallel(interfaces []sriovnetworkv1.Interface, ifaceStatuses []sriovnetworkv1.InterfaceExt, pfsToConfig map[string]bool) error {
```
adrianchiris, line 211: generally much of the logic is the same as ConfigSriovInterfaces

i would try to put the logic which configures a single interface which is common between the two in its own method and use it in both.

adrianchiris, line 211: also need to take into account externally managed and the use of storeManager.

need to make storeManager  concurrency safe (mutex for  prevent issues when loading/storing data)

pkg/utils/utils.go
```diff
@@ -200,8 +201,71 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
+
+				break
+			}
+		}
+		if !configured && ifaceStatus.NumVfs > 0 {
+			if skip := pfsToConfig[ifaceStatus.PciAddress]; skip {
+				continue
+			}
+
+			if err := resetSriovDevice(ifaceStatus); err != nil {
```
adrianchiris, line 256: this can also happen in parallel


i would do an initial calculation of which interfaces needs configuring and which need reset then run both parallel
wait for them to finish and report errors if any.

pkg/utils/utils.go
```diff
@@ -200,8 +201,71 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
+func ConfigSriovInterfacesInParallel(interfaces []sriovnetworkv1.Interface, ifaceStatuses []sriovnetworkv1.InterfaceExt, pfsToConfig map[string]bool) error {
+	log.Log.V(2).Info("ConfigSriovInterfacesInParallel(): start sriov configuration")
+	if IsKernelLockdownMode(true) && hasMellanoxInterfacesInSpec(ifaceStatuses, interfaces) {
+		log.Log.Error(nil, "cannot use mellanox devices when in kernel lockdown mode")
+		return fmt.Errorf("cannot use mellanox devices when in kernel lockdown mode")
+	}
+	// TODO(e0ne): store all errors in SriovNetworkNodeState
+	var result error
+	wg := sync.WaitGroup{}
+	for _, ifaceStatus := range ifaceStatuses {
```
mlguerrero12, line 220: I believe the concurrency should start here, for each one of the "ifaceStatuses". This way, you'll also cover the resetting of the sriov as @adrianchiris mentioned. 

There is also a well-known pattern in concurrency Go for this type of operations. Basically, the routines send the result through a channel and there is a for loop waiting and aggregating the results. I'll paste a working example here where I also address the defer and not reference for the loop var comments of @zeeke. 

func ConfigSriovInterfaces(interfaces []sriovnetworkv1.Interface, ifaceStatuses []sriovnetworkv1.InterfaceExt, pfsToConfig map[string]bool) error {
	if IsKernelLockdownMode(true) && hasMellanoxInterfacesInSpec(ifaceStatuses, interfaces) {
		log.Log.Error(nil, "cannot use mellanox devices when in kernel lockdown mode")
		return fmt.Errorf("cannot use mellanox devices when in kernel lockdown mode")
	}

	// we are already inside chroot, so we initialize the store as running on host
	storeManager, err := NewStoreManager(true)
	if err != nil {
		return fmt.Errorf("SyncNodeState(): error initializing storeManager: %v", err)
	}

	var wg sync.WaitGroup
	resChannel := make(chan error)
	var result []error

	for _, ifaceStatus := range ifaceStatuses {
		wg.Add(1)
		go func(ifaceStatus sriovnetworkv1.InterfaceExt) {
			var err error
			defer func() {
				resChannel <- err
				wg.Done()
			}()

			configured := false
			for _, iface := range interfaces {
					configured = true

					if skip := pfsToConfig[iface.PciAddress]; skip {
						break
					}

					if !NeedUpdate(&iface, &ifaceStatus) {
						log.Log.V(2).Info("syncNodeState(): no need update interface", "address", iface.PciAddress)

						// Save the PF status to the host
						err = storeManager.SaveLastPfAppliedStatus(&iface)
						if err != nil {
							log.Log.Error(err, "SyncNodeState(): failed to save PF applied config to host")
							return
						}

						break
					}
					if err = configSriovDevice(&iface, &ifaceStatus); err != nil {
						log.Log.Error(err, "SyncNodeState(): fail to configure sriov interface. resetting interface.", "address", iface.PciAddress)
						if iface.ExternallyManaged {
							log.Log.Info("SyncNodeState(): skipping device reset as the nic is marked as externally created")
						} else {
							if resetErr := resetSriovDevice(ifaceStatus); resetErr != nil {
								log.Log.Error(resetErr, "SyncNodeState(): failed to reset on error SR-IOV interface")
							}
						}
						return
					}

					// Save the PF status to the host
					err = storeManager.SaveLastPfAppliedStatus(&iface)
					if err != nil {
						log.Log.Error(err, "SyncNodeState(): failed to save PF applied config to host")
						return
					}
					break
				}
			}
			if !configured && ifaceStatus.NumVfs > 0 {
				if skip := pfsToConfig[ifaceStatus.PciAddress]; skip {
					return
				}

				// load the PF info
				var pfStatus *sriovnetworkv1.Interface
				var exist bool
				pfStatus, exist, err = storeManager.LoadPfsStatus(ifaceStatus.PciAddress)
				if err != nil {
					log.Log.Error(err, "SyncNodeState(): failed to load info about PF status for device",
						"address", ifaceStatus.PciAddress)
					return
				}

				if !exist {
					log.Log.Info("SyncNodeState(): PF name with pci address has VFs configured but they weren't created by the sriov operator. Skipping the device reset",
						"pf-name", ifaceStatus.Name,
						"address", ifaceStatus.PciAddress)
					return
				}

				if pfStatus.ExternallyManaged {
					log.Log.Info("SyncNodeState(): PF name with pci address was externally created skipping the device reset",
						"pf-name", ifaceStatus.Name,
						"address", ifaceStatus.PciAddress)
					return
				} else {
					err = RemoveUdevRule(ifaceStatus.PciAddress)
					if err != nil {
						return
					}
				}

				if err = resetSriovDevice(ifaceStatus); err != nil {
					return
				}
			}
		}(ifaceStatus)
	}

	// Wait for routines to finish and close resChannel to exit below for loop.
	// Routines only decrement counter when error is read in the for loop.
	go func() {
		wg.Wait()
		close(resChannel)
	}()

	// Fetch error returned by the routines. This will exit after channel is closed. No more pending routines.
	for routineErr := range resChannel {
		if routineErr != nil {
			result = append(result, routineErr)
		}
	}

	// Only first arriving error is returned.
	if len(result) != 0 {
		log.Log.Error(result[0], "ConfigSriovInterfacesInParallel(): fail to configure sriov interfaces. Only first arriving error is returned. There might be more. Please check the logs.")
		return result[0]
	}

	return nil
}

Note that I'm returning the first error of the result list of errors. For other projects, I've printed and returned this whole list. But printing the errors is already done separately and for returning a list, we will need other changes.

For the storage manager, I used a mutex and lock unlock in save and load.

type StoreManager struct {
	RunOnHost bool
	sync.Mutex
}

func (s *StoreManager) SaveLastPfAppliedStatus(PfInfo *sriovnetworkv1.Interface) error {
	s.Lock()
	defer s.Unlock()
}

Also note that everything after "configured := false" can be encapsulated in a function. So we could avoid duplication.

I've tested this but not extensively. Just wanted to show you the pattern I've used in other projects.

pkg/utils/utils.go
e0ne: done

pkg/utils/utils.go
e0ne: done

pkg/utils/utils.go
e0ne: we need to re-visit our API to include muliple results in SriovNetworkNodeState

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -94,9 +94,12 @@ spec:
           privileged: true
         args:
           - "start"
-        {{- if .UsedSystemdMode}}
+        {{- if .UsedSystemdMode }}
           - --use-systemd-service
         {{- end }}
+        {{- if .ParallelNicConfig }}
```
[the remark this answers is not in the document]
e0ne: I don't think we need it since the source of truth for  these options is CRD

pkg/utils/utils.go
e0ne: Is it OK to implement it in a follow up PR? if not, I'll push it in an additional commit

pkg/utils/utils.go
e0ne: we don't use storeManager in goroutine now

e0ne: done

pkg/utils/utils.go
adrianchiris: i feel like reset is also part of NIC configuration. do you see issue to handle it in the same PR ?

pkg/utils/utils.go
```diff
@@ -206,105 +207,187 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
-
 	// we are already inside chroot, so we initialize the store as running on host
 	storeManager, err := NewStoreManager(true)
 	if err != nil {
 		return fmt.Errorf("SyncNodeState(): error initializing storeManager: %v", err)
 	}
 
-	for _, ifaceStatus := range ifaceStatuses {
+	// TODO(e0ne): store all errors in SriovNetworkNodeState
+	var result error
```
adrianchiris, line 230: this is not being set anywhere

pkg/utils/utils.go
```diff
@@ -206,105 +207,187 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
+					defer wg.Done()
+					if err := configSriovDevice(iface, ifaceStatus); err != nil {
+						log.Log.Error(err, "ConfigSriovInterfacesInParallel(): fail to configure sriov interface. resetting interface.", "address", iface.PciAddress)
 					}
+				}(&interfaces[ifaceIndex], &ifaceStatuses[statusIndex])
+				// Save the PF status to the host
+				err = storeManager.SaveLastPfAppliedStatus(&iface)
+				if err != nil {
+					log.Log.Error(err, "ConfigSriovInterfacesInParallel(): failed to save PF applied config to host")
+					return err
```
adrianchiris, line 262: what if we have goroutines already running ?

pkg/utils/utils.go
```diff
@@ -206,105 +207,187 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
 
-					// Save the PF status to the host
-					err = storeManager.SaveLastPfAppliedStatus(&iface)
-					if err != nil {
-						log.Log.Error(err, "SyncNodeState(): failed to save PF applied config to host")
-						return err
+				wg.Add(1)
+				go func(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt) {
+					defer wg.Done()
+					if err := configSriovDevice(iface, ifaceStatus); err != nil {
```
adrianchiris, line 254: there is an error log msg emitted in configSriovDevice  so no need to do so here.

we do need to have a way to handle errors so we are able to fail after waiting for routines to complete

pkg/utils/utils.go
```diff
@@ -206,105 +207,187 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
+				if err != nil {
+					log.Log.Error(err, "ConfigSriovInterfacesInParallel(): failed to save PF applied config to host")
+					return err
+				}
+				break
+			}
+		}
+		if err := checkForConfigAndReset(configured, ifaceStatus, storeManager, pfsToConfig); err != nil {
+			log.Log.Error(err, "ConfigSriovInterfacesInParallel(): failed to save PF applied config to host")
+			return err
```
adrianchiris, line 269: what if we have goroutines already running ?

pkg/utils/utils.go
```diff
@@ -206,105 +207,187 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
+	wg.Wait()
+	if result != nil {
+		log.Log.Error(result, "ConfigSriovInterfacesInParallel(): fail to configure sriov interface")
+		return result
+	}
+	log.Log.V(2).Info("ConfigSriovInterfacesInParallel(): sriov configuration finished")
+	return nil
+}
+
+func ConfigSriovInterfaces(interfaces []sriovnetworkv1.Interface, ifaceStatuses []sriovnetworkv1.InterfaceExt, pfsToConfig map[string]bool) error {
```
adrianchiris, line 281: nit: add docstring

pkg/utils/utils.go
```diff
@@ -206,105 +207,187 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
-func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState, pfsToConfig map[string]bool) error {
+func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState, pfsToConfig map[string]bool, parallelNicConfig bool) error {
+	if parallelNicConfig {
+		return ConfigSriovInterfacesInParallel(newState.Spec.Interfaces, newState.Status.Interfaces, pfsToConfig)
+	}
 	return ConfigSriovInterfaces(newState.Spec.Interfaces, newState.Status.Interfaces, pfsToConfig)
 }
 
-func ConfigSriovInterfaces(interfaces []sriovnetworkv1.Interface, ifaceStatuses []sriovnetworkv1.InterfaceExt, pfsToConfig map[string]bool) error {
+func ConfigSriovInterfacesInParallel(interfaces []sriovnetworkv1.Interface, ifaceStatuses []sriovnetworkv1.InterfaceExt, pfsToConfig map[string]bool) error {
```
adrianchiris, line 217: nit: add docstring

controllers/sriovoperatorconfig_controller.go
```diff
@@ -194,10 +194,13 @@ func (r *SriovOperatorConfigReconciler) syncConfigDaemonSet(ctx context.Context,
 	data.Data["ClusterType"] = utils.ClusterType
 	data.Data["DevMode"] = os.Getenv("DEV_MODE")
 	data.Data["ImagePullSecrets"] = GetImagePullSecrets()
+	data.Data["UsedSystemdMode"] = false
```
adrianchiris, line 197: ah typo (used instead of use) :( . anyhow, not related to this PR

pkg/utils/utils.go
```diff
@@ -206,105 +207,187 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
-						return err
+				wg.Add(1)
+				go func(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt) {
+					defer wg.Done()
+					if err := configSriovDevice(iface, ifaceStatus); err != nil {
+						log.Log.Error(err, "ConfigSriovInterfacesInParallel(): fail to configure sriov interface. resetting interface.", "address", iface.PciAddress)
 					}
+				}(&interfaces[ifaceIndex], &ifaceStatuses[statusIndex])
+				// Save the PF status to the host
+				err = storeManager.SaveLastPfAppliedStatus(&iface)
```
adrianchiris, line 259: originally this was saved only if pf configuration suceeded. is that OK to save regardless ?

pkg/utils/utils.go
```diff
@@ -206,105 +207,187 @@ func DiscoverSriovDevices(withUnsupported bool, storeManager StoreManagerInterfa
...
+
+		if err = resetSriovDevice(ifaceStatus); err != nil {
+			return err
+		}
 	}
+
 	return nil
 }
 
+func checkNeedUpdate(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt, storeManager StoreManagerInterface) (bool, error) {
```
adrianchiris, line 375: checkNeedUpdate returning true when **no** update is needed is confusing

pkg/utils/utils.go
```diff
@@ -409,6 +492,24 @@ func NeedUpdate(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.Int
...
+				log.Log.Error(resetErr, "ConfigSriovInterfacesInParallel(): failed to reset on error SR-IOV interface")
+				result = resetErr
+			}
+		}
+	}
+
+	return result
+}
+
+func configPf(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt) error {
```
adrianchiris, line 512: all the logs have `configSriovDevice()` that OK ?

pkg/utils/utils.go
```diff
@@ -558,6 +659,7 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 			return err
 		}
 	}
+	log.Log.V(2).Info("configSriovDevice(): config interface completed", "address", ifaceStatus.PciAddress)
```
SchSeba, line 662: nit: address -> pciAddress

pkg/utils/utils.go
SchSeba, line 512: nit: configPF

SchSeba: +1 please change the log function name

pkg/utils/utils.go
```diff
@@ -409,6 +492,24 @@ func NeedUpdate(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.Int
 }
 
 func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt) error {
+	var result error
+	if err := configPf(iface, ifaceStatus); err != nil {
+		log.Log.Error(err, "ConfigSriovInterfacesInParallel(): fail to configure sriov interface. resetting interface.", "address", iface.PciAddress)
```
SchSeba, line 497: nit: PciAddress

SchSeba, line 497: please update the function name in the logs

pkg/utils/utils.go
SchSeba, line 254: I think you want to save the failed error from here so we can report back to the daemon that we failed

pkg/utils/utils.go
SchSeba, line 259: we should not save the device info if we failed

pkg/utils/utils.go
SchSeba, line 217: general idea for this function:
create a channel with the return information + error and then have a for with select using the number of go routines we start
this way we can have a logic after the routine finish

e0ne: done

adrianchiris:
@e0ne can you rebase this one ? then we can re-review

pkg/host/internal/sriov/sriov.go
```diff
@@ -437,92 +437,191 @@ func (s *sriov) ConfigSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *
 	return nil
 }
 
-func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
-	interfaces []sriovnetworkv1.Interface, ifaceStatuses []sriovnetworkv1.InterfaceExt, pfsToConfig map[string]bool) error {
+func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface, interfaces []sriovnetworkv1.Interface, ifaceStatuses []sriovnetworkv1.InterfaceExt, pfsToConfig map[string]bool) error {
```
adrianchiris, line 440: discussed with @e0ne offline.

He will check if we can simplify the logic:

1. move check for lockdown here
2. compute interfaces that need configuration and reset
3. configure/rest interfaces either serially or in parallel

that way we will avoid duplicate logic as much as possible.

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+}
+
+func (s *sriov) resetSriovInterfaces(storeManager store.ManagerInterface, interfaces []sriovnetworkv1.InterfaceExt) error {
+	for _, iface := range interfaces {
+		if err := s.checkForConfigAndReset(iface, storeManager); err != nil {
+			log.Log.Error(err, "resetSriovInterfaces(): fail to configure sriov interface. resetting interface.", "address", iface.PciAddress)
+			return err
+		}
+	}
+	log.Log.V(2).Info("resetSriovInterfaces(): sriov reset finished")
```
adrianchiris, line 631: can you add additional logs when starting and finishing configuring / resetting an interface both in parallel and serially ?

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+	return nil
+}
+
+func checkNeedUpdate(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt, storeManager store.ManagerInterface) (bool, error) {
```
adrianchiris, line 635: can you add docstring for added methods ?

also from the name "checkNeedUpdate" i expected that if true is returned then we need to update, but its actually the opposite... but i see also `sriovnetworkv1.NeedToUpdateSriov` has same meaning...

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+	if !sriovnetworkv1.NeedToUpdateSriov(iface, ifaceStatus) {
+		log.Log.V(2).Info("ConfigSriovInterfaces(): no need update interface", "address", iface.PciAddress)
+
+		// Save the PF status to the host
+		err := storeManager.SaveLastPfAppliedStatus(iface)
+		if err != nil {
+			log.Log.Error(err, "ConfigSriovInterfaces(): failed to save PF applied config to host")
```
adrianchiris, line 642: nit: update log msg

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+			"address", ifaceStatus.PciAddress)
+		return nil
+	}
+
+	if pfStatus.ExternallyManaged {
+		log.Log.Info("checkForConfigAndReset(): PF name with pci address was externally created skipping the device reset",
+			"pf-name", ifaceStatus.Name,
+			"address", ifaceStatus.PciAddress)
+		return nil
+	} else {
```
adrianchiris, line 672: true branch returns, you can simplify by removing else block and moving the removal of udev rule to the containing block.

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+	}
+	log.Log.V(2).Info("resetSriovInterfacesInParallel(): sriov reset finished")
+
+	return nil
+}
+
+func (s *sriov) configSriovInterfaces(storeManager store.ManagerInterface, interfaces []interfaceToConfigure) error {
+	for _, iface := range interfaces {
+		if err := s.ConfigSriovDevice(&iface.iface, &iface.ifaceStatus); err != nil {
+			log.Log.Error(err, "ConfigSriovInterfaces(): fail to configure sriov interface. resetting interface.", "address", iface.iface.PciAddress)
```
adrianchiris, line 603: update log msg here and below (method name)

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+}
 
-			// load the PF info
-			pfStatus, exist, err := storeManager.LoadPfsStatus(ifaceStatus.PciAddress)
-			if err != nil {
-				log.Log.Error(err, "SyncNodeState(): failed to load info about PF status for device",
-					"address", ifaceStatus.PciAddress)
-				return err
+func (s *sriov) configSriovInterfacesInParallel(storeManager store.ManagerInterface, interfaces []interfaceToConfigure) error {
+	log.Log.V(2).Info("ConfigSriovInterfacesInParallel(): start sriov configuration")
```
adrianchiris, line 527: nit: update log msgs in this method. (method name)

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+	var result error
+	errChannel := make(chan error)
+	interfacesToReset := 0
+	for ifaceIndex := range interfaces {
+		interfacesToReset += 1
+		go func(iface *sriovnetworkv1.InterfaceExt) {
+			if err := s.checkForConfigAndReset(*iface, storeManager); err != nil {
+				log.Log.Error(err, "resetSriovInterfacesInParallel(): fail to configure sriov interface. resetting interface.", "address", iface.PciAddress)
+				errChannel <- err
+			} else {
```
adrianchiris, line 581: you can remove the else block and just always `errChannel <- err`
if no error then err will be nil.

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
-				log.Log.Info("SyncNodeState(): PF name with pci address was externally created skipping the device reset",
-					"pf-name", ifaceStatus.Name,
-					"address", ifaceStatus.PciAddress)
-				continue
+	for i := 0; i < interfacesToReset; i++ {
+		errMsg := <-errChannel
+		result = errors.Join(result, errMsg)
+	}
+	if result != nil {
+		log.Log.Error(result, "resetSriovInterfacesInParallel(): fail to reset sriov interface")
```
adrianchiris, line 592: nit: interface -> interfaces

there may be more than one.

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+		log.Log.Error(result, "ConfigSriovInterfacesInParallel(): fail to configure sriov interface")
+		return result
+	}
+	log.Log.V(2).Info("ConfigSriovInterfacesInParallel(): sriov configuration finished")
+	return nil
+}
+
+func (s *sriov) resetSriovInterfacesInParallel(storeManager store.ManagerInterface, interfaces []sriovnetworkv1.InterfaceExt) error {
+	var result error
+	errChannel := make(chan error)
```
adrianchiris, line 573: you could initialize this with capacity of `len(interfaces)` so that go routines don block on errChannel.
not a must, up to u.

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+	interfacesToReset := 0
+	for ifaceIndex := range interfaces {
+		interfacesToReset += 1
+		go func(iface *sriovnetworkv1.InterfaceExt) {
+			if err := s.checkForConfigAndReset(*iface, storeManager); err != nil {
+				log.Log.Error(err, "resetSriovInterfacesInParallel(): fail to configure sriov interface. resetting interface.", "address", iface.PciAddress)
```
adrianchiris, line 579: nit: failed to reset interface ?

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
-			// load the PF info
-			pfStatus, exist, err := storeManager.LoadPfsStatus(ifaceStatus.PciAddress)
-			if err != nil {
-				log.Log.Error(err, "SyncNodeState(): failed to load info about PF status for device",
-					"address", ifaceStatus.PciAddress)
-				return err
+func (s *sriov) configSriovInterfacesInParallel(storeManager store.ManagerInterface, interfaces []interfaceToConfigure) error {
+	log.Log.V(2).Info("ConfigSriovInterfacesInParallel(): start sriov configuration")
+
+	// TODO(e0ne): store all errors in SriovNetworkNodeState
```
adrianchiris, line 529: we do that now right ?

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
-				log.Log.Info("SyncNodeState(): PF name with pci address has VFs configured but they weren't created by the sriov operator. Skipping the device reset",
-					"pf-name", ifaceStatus.Name,
-					"address", ifaceStatus.PciAddress)
-				continue
+	for i := 0; i < interfacesToConfigure; i++ {
+		errMsg := <-errChannel
+		result = errors.Join(result, errMsg)
+	}
+	if result != nil {
+		log.Log.Error(result, "ConfigSriovInterfacesInParallel(): fail to configure sriov interface")
```
adrianchiris, line 564: nit: interface -> interfaces

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+		return result
+	}
+	log.Log.V(2).Info("ConfigSriovInterfacesInParallel(): sriov configuration finished")
```
adrianchiris, line 567: nit: fix method name in logs of this method.

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+			if err := s.ConfigSriovDevice(&iface.iface, &iface.ifaceStatus); err != nil {
+				log.Log.Error(err, "ConfigSriovInterfacesInParallel(): fail to configure sriov interface. resetting interface.", "address", iface.iface.PciAddress)
+				if iface.iface.ExternallyManaged {
+					log.Log.Info("ConfigSriovInterfacesInParallel(): skipping device reset as the nic is marked as externally created")
+				} else {
+					if resetErr := s.ResetSriovDevice(iface.ifaceStatus); resetErr != nil {
+						log.Log.Error(resetErr, "ConfigSriovInterfacesInParallel(): failed to reset on error SR-IOV interface")
+					}
+				}
+				errChannel <- err
```
adrianchiris, line 545: nit you could send to channel once in code in outer block

`errChannel <- err`

if no error, err will be nil

adrianchiris (commented):
overall looks good @e0ne ! 
added some comments, nothing major.

controllers/sriovoperatorconfig_controller.go
```diff
@@ -187,6 +187,10 @@ func (r *SriovOperatorConfigReconciler) syncConfigDaemonSet(ctx context.Context,
 	} else {
 		data.Data["UsedSystemdMode"] = false
 	}
+	data.Data["ParallelNicConfig"] = false
+	if parallelConfig, ok := dc.Spec.FeatureGates["parallelNicConfig"]; ok {
```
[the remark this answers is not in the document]
e0ne: done

pkg/host/internal/sriov/sriov.go
e0ne: done

pkg/host/internal/sriov/sriov.go
e0ne: done

pkg/host/internal/sriov/sriov.go
e0ne: done

pkg/host/internal/sriov/sriov.go
e0ne: in this case we need to devicne `err` variable before `if` statement. Do we want it?

pkg/host/internal/sriov/sriov.go
e0ne: done

pkg/host/internal/sriov/sriov.go
adrianchiris: IMO yes, if it makes sense to u.

pkg/host/internal/sriov/sriov.go
e0ne: done

pkg/host/internal/sriov/sriov.go
e0ne: I agree, it will make code less complex

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,184 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
-					log.Log.Error(err, "SyncNodeState(): failed to save PF applied config to host")
-					return err
+					log.Log.Error(err, "getConfigureAndReset(): failed to check interface")
+					return nil, nil, err
 				}
-				break
+				if skip {
+					break
+				}
+				toBeConfigured = append(toBeConfigured, interfaceToConfigure{iface: iface, ifaceStatus: ifaceStatus})
```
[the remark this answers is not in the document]
e0ne: done

pkg/host/internal/sriov/sriov.go
```diff
@@ -450,13 +455,43 @@ func (s *sriov) ConfigSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *
...
+		return fmt.Errorf("cannot configure sriov interfaces")
+	}
+
+	if vars.ParallelNicConfig {
+		err = s.resetSriovInterfacesInParallel(storeManager, toBeResetted)
+	} else {
+		err = s.resetSriovInterfaces(storeManager, toBeResetted)
+	}
+	if err != nil {
+		log.Log.Error(err, "cannot reset reset interfaces")
```
adrianchiris, line 486: can you fix the typo here : "reset reset" ?

pkg/host/internal/sriov/sriov.go
e0ne: done

pkg/host/internal/sriov/sriov.go
e0ne: done

e0ne: done

adrianchiris:
@zeeke could take a look ?

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,185 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+				log.Log.Error(err, "configSriovInterfacesInParallel(): fail to configure sriov interface. resetting interface.", "address", iface.iface.PciAddress)
+				if iface.iface.ExternallyManaged {
+					log.Log.V(2).Info("configSriovInterfacesInParallel(): skipping device reset as the nic is marked as externally created")
+				} else {
+					if resetErr := s.ResetSriovDevice(iface.ifaceStatus); resetErr != nil {
+						log.Log.Error(resetErr, "configSriovInterfacesInParallel(): failed to reset on error SR-IOV interface")
+					}
+				}
 			}
+			errChannel <- nil
```
zeeke, line 548: `errChannel` writes are always `nil` . `configSriovInterfacesInParallel` will never return an error.

pkg/host/internal/sriov/sriov.go
```diff
@@ -467,76 +502,185 @@ func (s *sriov) ConfigSriovInterfaces(storeManager store.ManagerInterface,
...
+		}(&interfaces[ifaceIndex])
+		// Save the PF status to the host
+		err := storeManager.SaveLastPfAppliedStatus(&iface.iface)
+		if err != nil {
+			log.Log.Error(err, "configSriovInterfacesInParallel(): failed to save PF applied config to host")
+			return err
+		}
+		break
```
zeeke, line 556: this `break` can be removed.

zeeke (commented):
Left a few minor comments

pkg/host/internal/sriov/sriov.go
e0ne: fixed

pkg/host/internal/sriov/sriov.go
e0ne: done

zeeke (approved):
LGTM. merging

PR #500, opened 2023-09-07
Update device plugins volumes mounts

e0ne:
There is no need to mount the whole '/var/lib/kubelet/' directory into a pod. It's enough to mount only '/var/lib/kubelet/device-plugins' and '/var/lib/kubelet/plugins_registry' directories.


bn222:
LGTM

PR #502, opened 2023-09-07
Remove bundle since we use helm chart instead in upstream

bn222:


zeeke (approved):
LGTM

These files relates to openshift installation only.

bn222:
@e0ne can I get an ack on this?

e0ne (approved):

PR #504, opened 2023-09-18
Add context to `validatePolicyForNodeState` errors

zeeke:
`validatePolicyForNodeState(...)` errors should contain the node they were validated against, making the user understand where the configuration problem is located.

Examples of how error messages changed:

```
- numVfs(65) in CR p1 exceed the maximum allowed value(64)           
+ numVfs(65) in CR p1 exceed the maximum allowed value(64) interface(ens803f0) node(worker-1)
                        
- vendor(8086) in CR p1 not supported for vdpa
+ vendor(8086) in CR p1 not supported for vdpa interface(ens803f0) node(worker-1)
```

--- 
_(Note: I put literals instead of `fmt.Sprintf(...)` in assertion calls to simplify test code, [as suggested here](https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html))_


SchSeba (approved):
LGTM
thanks!

rollandf:
LGTM

PR #509, opened 2023-09-21
Support loglevel and logfile for SR-IOV plugin

andreaskaris:
The SR-IOV device plugin now supports the configuration of loglevel and logfile. Expose these settings through the API.

Depends on  https://github.com/k8snetworkplumbingwg/sriov-cni/pull/276


api/v1/sriovnetwork_types.go
```diff
@@ -60,6 +60,13 @@ type SriovNetworkSpec struct {
 	// MetaPluginsConfig configuration to be used in order to chain metaplugins to the sriov interface returned
 	// by the operator.
 	MetaPluginsConfig string `json:"metaPlugins,omitempty"`
+	// LogLevel sets the loglevel of the SRIOV device plugin - either of panic, error, warning, info, debug. Defaults to
+	// error if left blank.
+	// +kubebuilder:validation:Enum={"panic", "error","warning","info","debug",""}
+	LogLevel string `json:"loglevel,omitempty"`
+	// LogFile sets the log file of the SRIOV device plugin logs. If unset (default), this will log to stderr and thus
+	// to crio logs.
```
SchSeba, line 68: crio/multus pod

SchSeba (commented):
nice work!

just a small comment and lets wait for the merge in the sriov-cni

api/v1/sriovnetwork_types.go
e0ne: crio -> container runtime

api/v1/helper.go
```diff
@@ -528,13 +528,16 @@ func (cr *SriovIBNetwork) RenderNetAttDef() (*uns.Unstructured, error) {
 		data.Data["MetaPlugins"] = cr.Spec.MetaPluginsConfig
 	}
```
adrianchiris, line 530: please add comment that currently its not supported by ib-sriov-cni

api/v1/sriovnetwork_types.go
```diff
@@ -60,6 +60,13 @@ type SriovNetworkSpec struct {
 	// MetaPluginsConfig configuration to be used in order to chain metaplugins to the sriov interface returned
 	// by the operator.
 	MetaPluginsConfig string `json:"metaPlugins,omitempty"`
+	// LogLevel sets the log level of the SRIOV device plugin - either of panic, error, warning, info, debug. Defaults
```
mlguerrero12, line 63: What do you think about setting the default explicitly to info using +kubebuilder:default:= "info".

This way, the parameter will always be populated and users will clearly see what log level they are in and where to change it. (I know they can read the docs to find it out, but this improves user experience in my opinion).

api/v1/helper.go
```diff
@@ -659,6 +663,17 @@ func (cr *SriovNetwork) RenderNetAttDef() (*uns.Unstructured, error) {
...
 
+	data.Data["LogLevelConfigured"] = false
+	if cr.Spec.LogLevel != "" {
+		data.Data["LogLevelConfigured"] = true
+		data.Data["LogLevel"] = cr.Spec.LogLevel
+	}
```
bn222, line 670: Can't we just write this as as 2 lines?

```
data.Data["LogLevelConfigured"] = (cr.Spec.LogLevel != "")
data.Data["LogLevel"] = cr.Spec.LogLevel
```

andreaskaris: yep - done

zeeke (approved):
LGTM

PR #514, opened 2023-09-29
conformance-test: Filter NIC by environment variable

zeeke:
In order to verify a specific NIC works well with
the operator, the user who runs the test suite should be able to select involved devices.

This commit introduces the environment variable
`SRIOV_DEVICE_NAME_FILTER` to implement the selection behavior.

examples:
`SRIOV_DEVICE_NAME_FILTER=ens1f1` tells the test suite to use a specific device selecting it by its name

`SRIOV_DEVICE_NAME_FILTER="ens1*"` tells the test suite that every device whose name starts with "ens1" can be used by the tests.

### DIFF
--- a/test/util/cluster/cluster.go
+++ b/test/util/cluster/cluster.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


test/util/cluster/cluster.go
```diff
@@ -93,33 +96,47 @@ func DiscoverSriov(clients *testclient.ClientSet, operatorNamespace string) (*En
...
+		if match {
+			filteredDevices = append(filteredDevices, device)
 		}
 	}
-	return nil, fmt.Errorf("unable to find sriov devices in node %s", node)
+
+	return filteredDevices, nil
 }
 
 // FindSriovDevices retrieves all valid sriov devices for the given node.
```
SchSeba, line 138: nit please update the function comment

SchSeba (commented):
I really like it!

I have one idea let me know what you think about.
we can have a device with the same name but on two different nodes for example:

node0 -> eno1 -> mlx
node1 -> eno1 -> intel

to over come this I will like to propose to extend the env variable to something like

`<node>:<device>`
both regex support

let me know what do you think

zeeke:
> I have one idea let me know what you think about. we can have a device with the same name but on two different nodes for example:
> 
> node0 -> eno1 -> mlx 
> node1 -> eno1 -> intel
> 
> to over come this I will like to propose to extend the env variable to something like
> 
> `<node>:<device>` both regex support
> 
> let me know what do you think

It sounds good to me

test/util/cluster/cluster.go
```diff
@@ -35,6 +36,24 @@ var (
 	intelVendorID      = "8086"
 )
 
+// Name of environment variable to filter which decives can be discovered by FindSriovDevices and FindOneSriovDevice.
+// The filter is a regexp matched against node names and device name in the form <node_name>:<device_name>
+//
+// For example, givend the following devices in the cluster:
```
bn222, line 42: givend

bn222:
LGTM

SchSeba (approved):
/lgtm

e0ne (approved):

PR #515, opened 2023-10-02
improve github actions

SchSeba:
add cancel on progress and also run the virtual cluster tests only if all the light lanes like build and unit tests pass to save resources


bn222:
Can you change the commit message to:

```
Speed up GitHub action runs and save resources

- Require short running actions (golangci, test, build) before starting slower virtual-k8s-cluster and only run virtual-ocp when virtual-k8s-cluster succeed
- Allow cancellation

These two changes together reduce the amount of resources used

```

bn222:
LGTM

SchSeba:
/hold 
this doesn't run the virtual cluster now ;(

SchSeba:
I moved the virtual cluster to the same file as the build and unit test so we can have a dependency.
I also changed the name from Go to Test SR-IOV Operator to better represent the flow

zeeke (approved):
LGTM

hack/run-e2e-conformance-virtual-cluster.sh
```diff
@@ -130,18 +130,67 @@ kcli ssh $cluster_name-ctlplane-0 << EOF
 sudo su
 echo '$insecure_registry' > /etc/containers/registries.conf.d/003-internal.conf
 systemctl restart crio
+
+echo '[connection]
```
adrianchiris, line 134: nit: define function and call it for each node ?

e0ne (approved):

hack/run-e2e-conformance-virtual-cluster.sh
SchSeba: done

PR #516, opened 2023-10-02
virtual: get real PCI address for each device found

EmilienM:
We can't rely on the PCI address from the metadata so we will lookup the real PCI address
for the NIC that matches the MAC address.

Libvirt/QEMU cannot guarantee that the address specified in the XML will match the address seen by the guest.
This is a well known limitation: https://libvirt.org/pci-addresses.html
When using the q35 machine type, it highlights this issue due to the change from using PCI to PCI-E bus for virtual devices.

With that said, the PCI value in Nova Metadata is a best effort hint due to the limitations mentioned above. Therefore
we will lookup the real PCI address for the NIC that matches the MAC address.

This PR also adds  Virtio 1.0 network device as supported

https://devicehunt.com/view/type/pci/vendor/1AF4/device/1041
https://qemu.readthedocs.io/en/master/specs/pci-ids.html

Modern QEMU brings up virtio 1.0 network device, we need to support that
for DPDK.

### DIFF
--- a/pkg/utils/utils_virtual.go
+++ b/pkg/utils/utils_virtual.go
@@ -8,10 +8,12 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/hashicorp/go-retryablehttp"
 	"github.com/jaypipes/ghw"
+	"github.com/jaypipes/ghw/pkg/net"
 
 	dputils "github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/utils"
 
@@ -53,10 +55,13 @@ const (
 	ospMetaDataBaseURL     = "http://169.254.169.254/openstack/2018-08-27"
 	ospHostNetworkDataFile = ospHostMetaDataDir + "/network_data.json"
 	ospHostMetaDataFile    = ospHostMetaDataDir + "/meta_data.json"
-	ospNetworkDataFile     = ospMetaDataDir + "/network_data.json"
-	ospMetaDataFile        = ospMetaDataDir + "/meta_data.json"
 	ospNetworkDataURL      = ospMetaDataBaseURL + "/network_data.json"
 	ospMetaDataURL         = ospMetaDataBaseURL + "/meta_data.json"
+)
+
+var (
+	ospNetworkDataFile = ospMetaDataDir + "/network_data.json"
+	ospMetaDataFile    = ospMetaDataDir + "/meta_data.json"
 )
 
 // OSPMetaDataDevice -- Device structure within meta_data.json
@@ -117,7 +122,39 @@ func GetOpenstackData(useHostPath bool) (metaData *OSPMetaData, networkData *OSP
 	metaData, networkData, err = getOpenstackDataFromConfigDrive(useHostPath)
 	if err != nil {
 		metaData, networkData, err = getOpenstackDataFromMetadataService()
+		if err != nil {
+			return metaData, networkData, fmt.Errorf("GetOpenStackData(): error getting OpenStack data: %w", err)
+		}
 	}
+
+	// We can't rely on the PCI address from the metadata so we will lookup the real PCI address
+	// for the NIC that matches the MAC address.
+	//
+	// Libvirt/QEMU cannot guarantee that the address specified in the XML will match the address seen by the guest.
+	// This is a well known limitation: https://libvirt.org/pci-addresses.html
+	// When using the q35 machine type, it highlights this issue due to the change from using PCI to PCI-E bus for virtual devices.
+	//
+	// With that said, the PCI value in Nova Metadata is a best effort hint due to the limitations mentioned above. Therefore
+	// we will lookup the real PCI address for the NIC that matches the MAC address.
+	netInfo, err := ghw.Network()
+	if err != nil {
+		return metaData, networkData, fmt.Errorf("GetOpenStackData(): error getting network info: %w", err)
+	}
+	for i, device := range metaData.Devices {
+		realPCIAddr, err := getPCIAddressFromMACAddress(device.Mac, netInfo.NICs)
+		if err != nil {
+			// If we can't find the PCI address, we will just print a warning, return the data as is with no error.
+			// In the future, we'll want to drain the node if sno-initial-node-state.json doesn't exist when daemon is restarted and when we have SR-IOV
+			// allocated devices already.
+			glog.Warningf("GetOpenstackData(): error getting PCI address for device %s: %v", device.Mac, err)
+			return metaData, networkData, nil
+		}
+		if realPCIAddr != device.Address {
+			glog.V(2).Infof("GetOpenstackData(): PCI address for device %s does not match Nova metadata value %s, it'll be overwritten with %s", device.Mac, device.Address, realPCIAddr)
+			metaData.Devices[i].Address = realPCIAddr
+		}
+	}
+
 	return metaData, networkData, err
 }
 
@@ -203,6 +240,26 @@ func getOpenstackDataFromMetadataService() (metaData *OSPMetaData, networkData *
 		return metaData, networkData, fmt.Errorf("error unmarshalling raw bytes %v from %s", err, ospNetworkDataURL)
 	}
 	return metaData, networkData, nil
+}
+
+// getPCIAddressFromMACAddress returns the PCI address of a device given its MAC address
+func getPCIAddressFromMACAddress(macAddress string, nics []*net.NIC) (string, error) {
+	var pciAddress string
+	for _, nic := range nics {
+		if strings.EqualFold(nic.MacAddress, macAddress) {
+			if pciAddress == "" {
+				pciAddress = *nic.PCIAddress
+			} else {
+				return "", fmt.Errorf("more than one device found with MAC address %s is unsupported", macAddress)
+			}
+		}
+	}
+
+	if pciAddress != "" {
+		return pciAddress, nil
+	}
+
+	return "", fmt.Errorf("no device found with MAC address %s", macAddress)
 }
 
 // CreateOpenstackDevicesInfo create the openstack device info map
--- a/pkg/utils/utils_virtual_test.go
+++ b/pkg/utils/utils_virtual_test.go
@@ -1,0 +1,58 @@
+package utils
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/jaypipes/ghw"
+	"github.com/jaypipes/ghw/pkg/net"
+	"github.com/jaypipes/ghw/pkg/option"
+	"k8s.io/utils/pointer"
+)
+
+func TestUtils(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Utils")
+}
+
+var _ = Describe("Virtual", func() {
+
+	Context("GetOpenstackData", func() {
+		It("PCI address replacement based on MAC address", func() {
+			ospNetworkDataFile = "./testdata/network_data.json"
+			ospMetaDataFile = "./testdata/meta_data.json"
+			DeferCleanup(func() {
+				ospNetworkDataFile = ospMetaDataDir + "/network_data.json"
+				ospMetaDataFile = ospMetaDataDir + "/meta_data.json"
+			})
+
+			ghw.Network = func(opts ...*option.Option) (*net.Info, error) {
+				return &net.Info{
+					NICs: []*net.NIC{{
+						MacAddress: "fa:16:3e:00:00:00",
+						PCIAddress: pointer.String("0000:04:00.0"),
+					}, {
+						MacAddress: "fa:16:3e:11:11:11",
+						PCIAddress: pointer.String("0000:99:99.9"),
+					}},
+				}, nil
+			}
+
+			DeferCleanup(func() {
+				ghw.Network = net.New
+			})
+
+			metaData, _, err := GetOpenstackData(false)
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(metaData.Devices).To(HaveLen(2))
+			Expect(metaData.Devices[0].Mac).To(Equal("fa:16:3e:00:00:00"))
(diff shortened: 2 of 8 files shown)


pkg/utils/utils_virtual.go
```diff
@@ -118,6 +118,24 @@ func GetOpenstackData(useHostPath bool) (metaData *OSPMetaData, networkData *OSP
...
+	// When using the q35 machine type, it highlights this issue due to the change from using PCI to PCI-E bus for virtual devices.
+	//
+	// With that said, the PCI value in Nova Metadata is a best effort hint due to the limitations mentioned above. Therefore
+	// we will lookup the real PCI address for the NIC that matches the MAC address.
+	for _, device := range metaData.Devices {
+		realPCIAddr, err := getPCIAddressFromMACAddress(device.Mac)
+		if err != nil {
+			return metaData, networkData, fmt.Errorf("GetOpenstackData(): error getting PCI address for device %s: %v", device.Mac, err)
+		}
+		device.Address = realPCIAddr
```
stephenfin, line 136: Rather than blindly overwriting it, could you add a debug-level warning if there's a mismatch? Eventually this will get fixed in Nova (since the fix will likely be there in both cases) and until then, any little encouragement to resolve this is a good thing.

pkg/utils/utils_virtual.go
```diff
@@ -205,6 +223,20 @@ func getOpenstackDataFromMetadataService() (metaData *OSPMetaData, networkData *
...
 
+// getPCIAddressFromMACAddress returns the PCI address of a device given its MAC address
+func getPCIAddressFromMACAddress(macAddress string) (string, error) {
+	nics, err := ghw.Network()
+	if err != nil {
+		return "", fmt.Errorf("error getting network info: %v", err)
+	}
+	for _, nic := range nics.NICs {
+		if nic.MacAddress == macAddress {
+			return *nic.PCIAddress, nil
```
stephenfin, line 234: This will return on the first match. What do we do if there are two devices with the same MAC address? Apparently this is a thing (per @SeanMooney)

stephenfin (changes requested):
Aye, this was the approach I'd be advocating for. Looks pretty reasonable to me, though I do have two comments inline.

pkg/utils/utils_virtual.go
EmilienM: Good point. I'll look at that.

EmilienM:
/hold
under testing

pkg/utils/utils_virtual.go
```diff
@@ -205,6 +226,33 @@ func getOpenstackDataFromMetadataService() (metaData *OSPMetaData, networkData *
...
+// getPCIAddressFromMACAddress returns the PCI address of a device given its MAC address
+func getPCIAddressFromMACAddress(macAddress string) (string, error) {
+	nics, err := ghw.Network()
+	if err != nil {
+		return "", fmt.Errorf("error getting network info: %v", err)
+	}
+	// Check if there are more than one device with the same MAC address
+	// and return an error if that's the case, we don't support that scenario
+	// for now.
+	numberOfDevices := 0
```
SeanMooney, line 238: as noted on slack instead of looping over nics.NICs twice
you can instead declare a var here to hold the PCI address
then 

do
```
for _, nic := range nics.NICs {
		if nic.MacAddress == macAddress {
			if pciAddress == nil {
               pciAddress = nic.PCIAddress
            } else {
             return "", fmt.Errorf("more than one device found with MAC address %s", macAddress)
            }
		}
	}
if pciAdress != nil {
    return *nic.PCIAddress, nil
}
return "", fmt.Errorf("no device found with MAC address %s", macAddress)
```

EmilienM: nice catch, thanks

EmilienM:
Downstream's CI has seen no regression: https://github.com/openshift/sriov-network-operator/pull/837
The OpenStack jobs have passed.

However, I'll confirm with the initial bug reporter that this works in newer version of qemu.

pkg/utils/utils_virtual.go
```diff
@@ -118,6 +118,27 @@ func GetOpenstackData(useHostPath bool) (metaData *OSPMetaData, networkData *OSP
...
+	// Libvirt/QEMU cannot guarantee that the address specified in the XML will match the address seen by the guest.
+	// This is a well known limitation: https://libvirt.org/pci-addresses.html
+	// When using the q35 machine type, it highlights this issue due to the change from using PCI to PCI-E bus for virtual devices.
+	//
+	// With that said, the PCI value in Nova Metadata is a best effort hint due to the limitations mentioned above. Therefore
+	// we will lookup the real PCI address for the NIC that matches the MAC address.
+	for _, device := range metaData.Devices {
+		realPCIAddr, err := getPCIAddressFromMACAddress(device.Mac)
+		if err != nil {
+			return metaData, networkData, fmt.Errorf("GetOpenstackData(): error getting PCI address for device %s: %v", device.Mac, err)
```
mandre, line 134: I believe it's more idiomatic to use the `%w` verb for errors.

pkg/utils/utils_virtual.go
```diff
@@ -118,6 +118,27 @@ func GetOpenstackData(useHostPath bool) (metaData *OSPMetaData, networkData *OSP
...
+		}
+		if realPCIAddr != device.Address {
+			glog.Warningf("GetOpenstackData(): PCI address for device %s does not match Nova metadata value %s, it'll be overwritten with %s", device.Mac, device.Address, realPCIAddr)
```
mandre, line 137: `Warning` level seems too much, as we can't rely on the nova metadata for the PCI address and this would then be an expected scenario to have the PCI address differ. What do you think of lowering the log level?

pkg/utils/utils_virtual.go
```diff
@@ -205,6 +226,33 @@ func getOpenstackDataFromMetadataService() (metaData *OSPMetaData, networkData *
...
+// getPCIAddressFromMACAddress returns the PCI address of a device given its MAC address
+func getPCIAddressFromMACAddress(macAddress string) (string, error) {
+	nics, err := ghw.Network()
+	if err != nil {
+		return "", fmt.Errorf("error getting network info: %v", err)
+	}
+
+	var pciAddress string
+	for _, nic := range nics.NICs {
+		if nic.MacAddress == macAddress {
```
mandre, line 238: I haven't checked how these are returned, but wouldn't it be safer to lowercase the MAC addresses (both `nic.MacAddress` and `macAddress`) before comparing?

EmilienM:
Weird, the webhook is failing: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_sriov-network-operator/837/pull-ci-openshift-sriov-network-operator-master-e2e-openstack-nfv/1709659020530290688/artifacts/e2e-openstack-nfv/gather-extra/artifacts/pods/openshift-sriov-network-operator_operator-webhook-7fvmg_webhook-server.log

pkg/utils/utils_virtual.go
```diff
@@ -205,6 +227,33 @@ func getOpenstackDataFromMetadataService() (metaData *OSPMetaData, networkData *
 	return metaData, networkData, nil
 }
 
+// getPCIAddressFromMACAddress returns the PCI address of a device given its MAC address
+func getPCIAddressFromMACAddress(macAddress string) (string, error) {
+	nics, err := ghw.Network()
+	if err != nil {
+		return "", fmt.Errorf("error getting network info: %v", err)
```
mandre, line 234: There too, we should replace by `%w` if possible.

SchSeba:
Hi @EmilienM I don't think this work as expected in all cases.

can you try this
1. create policy
2. start a pod that use the device
3. restart the config-daemon

I think it this case it will not work as expected because we will not be able to find the device on the host as it's inside the container. or even if we switch the nic driver to be vfio for example no?

EmilienM:
> Hi @EmilienM I don't think this work as expected in all cases.
> 
> can you try this
> 
>     1. create policy
> 
>     2. start a pod that use the device
> 
>     3. restart the config-daemon
> 
> 
> I think it this case it will not work as expected because we will not be able to find the device on the host as it's inside the container. or even if we switch the nic driver to be vfio for example no?

we tried together today and it worked fine.

EmilienM:
This was well tested on OpenStack: https://github.com/openshift/sriov-network-operator/pull/837

zeeke (approved):
LGTM

@adrianchiris @Eoghan1232 please take a look at this when you have time

pkg/utils/utils_virtual.go
```diff
@@ -205,6 +227,33 @@ func getOpenstackDataFromMetadataService() (metaData *OSPMetaData, networkData *
 	return metaData, networkData, nil
 }
 
+// getPCIAddressFromMACAddress returns the PCI address of a device given its MAC address
+func getPCIAddressFromMACAddress(macAddress string) (string, error) {
+	nics, err := ghw.Network()
```
adrianchiris, line 232: this will keep reading all NICs information from the filesystem/ethtool for every MAC address

pkg/utils/utils_virtual.go
```diff
@@ -205,6 +227,33 @@ func getOpenstackDataFromMetadataService() (metaData *OSPMetaData, networkData *
...
+	}
+
+	var pciAddress string
+	for _, nic := range nics.NICs {
+		if strings.EqualFold(nic.MacAddress, macAddress) {
+			if pciAddress == "" {
+				pciAddress = *nic.PCIAddress
+			} else {
+				// Check if there are more than one device with the same MAC address
+				// and return an error if that's the case, we don't support that scenario for now.
```
adrianchiris, line 244: comment is not needed IMO. i think the error below is clear. feel free to throw in  "unsupported" somewhere below if you  you think its worth it.

pkg/utils/utils_virtual_test.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"testing"
+
+	"github.com/jaypipes/ghw"
+	"github.com/jaypipes/ghw/pkg/net"
+	"github.com/jaypipes/ghw/pkg/option"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/utils/pointer"
+)
+
```
adrianchiris, line 12: can you make this one a ginkgo test ?

like we have for daemon package (daemon_test.go)

adrianchiris (commented):
i believe current approach is also problematic in some cases. (see my comments)

the way i see it there are two options:
1.  some different solution/approach
2. make this a "best-effort" change so that we are no worse than where we are today

pkg/utils/utils_virtual.go
```diff
@@ -118,6 +119,27 @@ func GetOpenstackData(useHostPath bool) (metaData *OSPMetaData, networkData *OSP
...
+	// Libvirt/QEMU cannot guarantee that the address specified in the XML will match the address seen by the guest.
+	// This is a well known limitation: https://libvirt.org/pci-addresses.html
+	// When using the q35 machine type, it highlights this issue due to the change from using PCI to PCI-E bus for virtual devices.
+	//
+	// With that said, the PCI value in Nova Metadata is a best effort hint due to the limitations mentioned above. Therefore
+	// we will lookup the real PCI address for the NIC that matches the MAC address.
+	for i, device := range metaData.Devices {
+		realPCIAddr, err := getPCIAddressFromMACAddress(device.Mac)
+		if err != nil {
+			return metaData, networkData, fmt.Errorf("GetOpenstackData(): error getting PCI address for device %s: %w", device.Mac, err)
```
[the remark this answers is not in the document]
SchSeba: agree with @adrianchiris in case the SNO file is deleted and we restart the daemon AND there is a workload running using the device we will not find it so we should not return and error in case it was not found (let's put a warning)

A better solution we can implement in case:
1. SNO doesn't exist
2. daemon is restarted
3. we have sriov allocated devices

is to drain the node so will remove the devices but it's a pretty big change in the code today so it can be implemented in a separate PR because it's not related to this fix, it's a bug that exist even before this.

here we just need to add a warning if the function `getPCIAddressFromMACAddress` returns "device not found base on mac" and not return error that will kill the pod

e0ne (approved):
LGTM

pkg/utils/utils_virtual.go
adrianchiris: i dont follow, this is plain inefficient from software perspective.
while  query from the  system the same information over and over again while expecting to get the same result.

bn222: 1. How do you know upfront it's a performance issue in the system as a whole. I understand that it feels inefficient, and it probably is. However, does it mean we need to fix it now before it has been identified as a real issue?
2. A solution is a caching layer that memoizes calls to this functions (i.e. store a map of macAddress to PCI address in memory). While this is not complex in and of itself, it does mean more code that we need to maintain.

Maybe @adrianchiris has a different idea than 2 in mind? 2 will only help if we actually call this function multiple times frequently enough for it to matter.

adrianchiris: GetOpenstackData is called once via writer when daemon starts. so im not aiming at caching it.

what id like to see is essentially moving this line to the caller so that listing the nic happens only once, then the list is passed as parameter to this function.

when there  are multiple interfaces passed  to the VM (multiple NICs) this becomes quatratic
(instead of listing e.g 16 nics  once you would list 16 nics 16  times)

each time triggering OS operations (ethtool / sysfs).

so my idea is essentially a 2 line change for which we spent multiple lines discussing it :)

pkg/utils/utils_virtual.go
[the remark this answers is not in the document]
EmilienM: Good for me Seb, I'm adding a comment as well about this plan.

adrianchiris (approved):
Thx for addressing my comments ! LGTM

some minor nit but would not block on it.

EmilienM:
the k8s and ocp failures can be ignore I believe, none of them test SNO on OpenStack. They were passing a few days ago, I wonder if something broke?

SchSeba:
HI @EmilienM please add the ID here https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L1141

if not we will have again the issue after reboot on non shiftstack deployments

SchSeba (approved):

PR #521, opened 2023-10-18
Avoid recording multiple events

adrianchiris:
In case update operation fail in node status writer we should not record an event.

move event sending logic to its own method
and call it after status has been successfully updated.


adrianchiris:
@rollandf PTAL.

still need to test this one on my setup.

rollandf (approved):
LGTM

zeeke (approved):
LGTM

Good catch

adrianchiris:
This is working as expected.

There is another issue with roles in the operator following events introduction, ill send fix in a separarte PR.

PR #522, opened 2023-10-18
webhook: Disable HTTP2 by default

cgoncalves:
To address CVE-2023-44487.

### DIFF
--- a/cmd/webhook/start.go
+++ b/cmd/webhook/start.go
@@ -17,9 +17,10 @@ import (
 )
 
 var (
-	certFile string
-	keyFile  string
-	port     int
+	certFile    string
+	keyFile     string
+	port        int
+	enableHTTP2 bool
 )
 
 var (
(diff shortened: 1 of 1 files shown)


bn222:
please put the right OCPBUGS in front of the tiel, and re-assign it to yourself.

adrianchiris:
seems some apis from controller-runtime have changed as build is failling

cgoncalves:
controller-runtime v0.15.3 will include the x/net bump to 0.17.0 but it hasn't been released yet.
controller-runtime v0.16.3 breaks APIs (around metrics and namespaces, easily fixable) but may not be desired at this point in time.
Andrea suggested disabling HTTP/2 on the webhook server. Other projects are doing so as well for the time being.

SchSeba (commented):
LGTM

the only comment I have is we don't have a way to change this variable so it will always by disable.

maybe we can just disable it for now until the controller-runtime gets updated?

zeeke (approved):
> the only comment I have is we don't have a way to change this variable so it will always by disable.
> 
> maybe we can just disable it for now until the controller-runtime gets updated?

u/s user can enable http2 back by editing the deployment files.

LGTM

SchSeba (approved):
LGTM

zeeke:
@e0ne , @adrianchiris , @Eoghan1232 Please, take a look. we need this to fix a CVE

Eoghan1232 (approved):
lgtm

PR #528, opened 2023-10-22
Support number of workers for the virtual cluster

SchSeba:
also add a make target to redeploy the containers on the cluster for an easy development cycle

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,6 +197,9 @@ test-e2e-conformance-virtual-ocp-cluster-ci:
 test-e2e-conformance-virtual-ocp-cluster:
 	SKIP_DELETE=TRUE ./hack/run-e2e-conformance-virtual-ocp.sh
 
+redeploy-operator-virtual-cluster:
+	./hack/virtual-cluster-redeploy.sh
+
 test-e2e-validation-only:
 	SUITE=./test/validation ./hack/run-e2e-conformance.sh	
 
--- a/hack/run-e2e-conformance-virtual-cluster.sh
+++ b/hack/run-e2e-conformance-virtual-cluster.sh
@@ -11,6 +11,14 @@ HOME="/root"
 here="$(dirname "$(readlink --canonicalize "${BASH_SOURCE[0]}")")"
 root="$(readlink --canonicalize "$here/..")"
 
+NUM_OF_WORKERS=${NUM_OF_WORKERS:-2}
+total_number_of_nodes=$((1 + NUM_OF_WORKERS))
+
+if [ "$NUM_OF_WORKERS" -lt 2 ]; then
+    echo "Min number of workers is 2"
+    exit 1
+fi
+
 check_requirements() {
   for cmd in kcli virsh virt-edit podman make go; do
     if ! command -v "$cmd" &> /dev/null; then
@@ -49,7 +57,7 @@ api_ip: $api_ip
 virtual_router_id: $virtual_router_id
 domain: $domain_name
 ctlplanes: 1
-workers: 2
+workers: $NUM_OF_WORKERS
 ingress: false
 machine: q35
 engine: crio
@@ -92,7 +100,7 @@ sleep_time=10
 until $ready || [ $ATTEMPTS -eq $MAX_ATTEMPTS ]
 do
     echo "waiting for cluster to be ready"
-    if [ `kubectl get node | grep Ready | wc -l` == 3 ]; then
+    if [ `kubectl get node | grep Ready | wc -l` == $total_number_of_nodes ]; then
         echo "cluster is ready"
         ready=true
     else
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -8,6 +8,7 @@ TARGET_DIR=$(CURPATH)/build/_output
 BIN_DIR=$(CURPATH)/bin
 KUBECONFIG?=$(HOME)/.kube/config
 export OPERATOR_EXEC?=oc
+export OCP_VERSION=4.14.0-rc.6
```
bn222, line 11: This should not be part of upstream

hack/run-e2e-conformance-virtual-cluster.sh
```diff
@@ -49,7 +57,7 @@ api_ip: $api_ip
 virtual_router_id: $virtual_router_id
 domain: $domain_name
 ctlplanes: 1
-workers: 2
+workers: $NUM_OF_WORKERS
```
bn222, line 60: To me, this file should be downstream only since it's using kcli to deploy an OCP cluster. That goes for the Makefile targets too. No-one from upstream can use this file without OCP. The only reason to have it here is to reduce the effort of deviating upstream from downstream. However, I think if this file is downstream only, there is nothing to maintain upstream for that.

zeeke: Here my thoughts on this.:

[kcli](https://github.com/karmab/kcli) can be seen as an upstream component, though it is primarily used in some openshift deployment. 

File [`hack/run-e2e-conformance-virtual-cluster.sh`](https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/528/files#diff-7e6d81500bdbfcb0341bce1452f665c6c78edc5ab3125d0ac217bde9e7cbe02d) deploys vanilla kubernetes clusters, so it's fine for it to stay u/s.

You're probably referring to [`hack/run-e2e-conformance-virtual-ocp.sh`](https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/528/files#diff-90e627cba9db282e04456e7e550bd478d2c12fbbbce4104f9421d640eb3a6447), which deploys and tests PR against an openshift cluster. In this case, having it u/s helps to intercept breaking changes early in the pipeline, as the code has plenty of `IsOpenshiftCluster()`.

bn222: I understand that we have plenty of `IsOpenshiftCluster()` in upstream, but the question is more fundamental: should we have such things in upstream? It's not black and white either in my opinion. The less of that we have upstream, the better. In this particular case, it seems that we could move that complete file to downstream only.

SchSeba: Hi @bn222 take a look we have a CI system running here
https://github.com/k8snetworkplumbingwg/sriov-network-operator/actions/runs/6606102537/job/17942015613?pr=528

https://github.com/k8snetworkplumbingwg/sriov-network-operator/actions/runs/6606102537/job/17942015613?pr=528

and we want to test OCP on u/s as the operator should support both vanilla k8s and OCP and we want to catch bugs as close to the source as possible

bn222: I know we want to catch bugs early as possible, however the question I'm posing is: should we really test downstream stuff in the upstream repo? Can't we have something that is doing that separately (or as separate as possible) from upstream?

bn222: In summary, my view on this is: either we are missing some tests in upstream that don't tie into OCP, or we need to add those OCP-specific tests to downstream if the things it tests for are downstream specific.

SchSeba: I am sorry I don't think I am following here.

This is like Openstack that it can run on multiple platforms and every vendor that runs it wants to be sure a PR doesn't break is OpenStack deployment.

so here is the same case we treat OCP as a vendor CI meaning any pull request with bug fixes or new features should work on both vanilla k8s and openshift. there is no point in waiting for the changes to be synced into the openshift repos to find bugs.

that will be the same expectation if now k3s when to be sure the operator works on is platform and continue to work.

adrianchiris:
lets discuss this one in a meeting.

my  thoughts:
tests should be agnostic as possible to the k8s distro (upstream, ocp)
any ocp specific prep scripts or ocp specific tests should be out of tree in a different repo.
GH Actions can easily clone more than one repo.

SchSeba:
@adrianchiris I agree that is why we have a separate repo for OCP specific tests using the sriov operator under the cnf-feature-deploy repo

https://github.com/openshift-kni/cnf-features-deploy/blob/master/cnf-tests/testsuites/e2esuite/dpdk/dpdk.go
https://github.com/openshift-kni/cnf-features-deploy/blob/master/cnf-tests/testsuites/e2esuite/sctp/sctp_sriov.go
https://github.com/openshift-kni/cnf-features-deploy/blob/master/cnf-tests/testsuites/e2esuite/vrf/vrf_sriov.go

and I may in the future clone that repo as part of CI system here, but the script to deploy the cluster (can be vanilla ocp k3s or others) with the PR build operator should be in the operator repo if you ask me.

bn222:
To me, it makes more sense to have as few as possible references to non-vanille k8s in this repo. Again, it's not black and white. I understand the need to trigger a job that is testing a vendor-specific release. That's ok imo as long as there is a single entry point to kick off all kinds of vendor-specific testing. In short, I'm in line with @adrianchiris 

Regarding the actual code, I never liked the fact that we had `if openshift` in upstream. Better to abstract those things away too and call a generic thing instead that can be dynamically replaced by an instance of something specific downstream.

bn222:
To add some detail to my previous comment, it's important to realize that there are two parts: "vendor specific testing" and "code related stuff". For this PR, let's focus on the first part only.

adrianchiris (approved):

zeeke (approved):
LGTM

PR #529, opened 2023-10-25
Increase end-to-end test coverage

zeeke:
Add a basic assertion in the end-to-end suite to
verify that the config-daemon can publish
k8s Events.
cc @rollandf 

Add an end-to-end test to verify that configuring the LogLevel
field of a SriovNetwork makes sriov cni Add command logging
to multus pods
cc @andreaskaris

### DIFF
--- a/test/conformance/tests/test_sriov_operator.go
+++ b/test/conformance/tests/test_sriov_operator.go
@@ -930,6 +930,36 @@ var _ = Describe("[sriov] operator", func() {
 				}, 3*time.Minute, time.Second).Should(Equal(corev1.PodRunning))
 			})
 		})
+
+		Context("CNI Logging level", func() {
+			It("Debug logging should be visible in multus pod", func() {
+				sriovNetworkName := "test-log-level-debug-no-file"
+				err := network.CreateSriovNetwork(clients, sriovDevice, sriovNetworkName,
+					namespaces.Test, operatorNamespace, resourceName, ipamIpv4,
+					func(sn *sriovv1.SriovNetwork) {
+						sn.Spec.LogLevel = "debug"
+					})
+				Expect(err).ToNot(HaveOccurred())
+
+				podDeployTime := time.Now()
+
+				testPod := createTestPod(node, []string{sriovNetworkName})
+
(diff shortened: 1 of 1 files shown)


andreaskaris (changes requested):
Looking good for the LogLevel stuff minus the "FIt" / "It" in test_sriov_operator.go on line 111

test/conformance/tests/test_sriov_operator.go
```diff
@@ -108,7 +108,7 @@ var _ = Describe("[sriov] operator", func() {
 	Describe("No SriovNetworkNodePolicy", func() {
 		Context("SR-IOV network config daemon can be set by nodeselector", func() {
 			// 26186
-			FIt("Should schedule the config daemon on selected nodes", func() {
+			It("Should schedule the config daemon on selected nodes", func() {
```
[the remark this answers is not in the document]
zeeke: sure, it's better to make it clean

andreaskaris:
From my side (sriov logging) lgtm

test/conformance/tests/test_sriov_operator.go
```diff
@@ -913,6 +930,36 @@ var _ = Describe("[sriov] operator", func() {
...
+
+				testPod := createTestPod(node, []string{sriovNetworkName})
+
+				recentMultusLogs := getMultusPodLogs(testPod.Spec.NodeName, podDeployTime)
+
+				Expect(recentMultusLogs).To(
+					ContainElement(
+						// Assert against multiple ContainSubstring condition because we can't make assumption on the order of the chunks
+						And(
+							ContainSubstring(`level="debug"`),
```
mlguerrero12, line 954: the problem with this is that a simple change of the log message in the sriov-cni will break ci here. Perhaps we could use something that is less likely to change. The first message of cmdAdd for example. What do you think?

	logging.Debug("function called",
		"func", "cmdAdd",
		"args.Path", args.Path, "args.StdinData", string(args.StdinData), "args.Args", args.Args)

zeeke: sounds good

adrianchiris (approved):

mlguerrero12:
LGTM

SchSeba (approved):

PR #530, opened 2023-10-26
Verify status changes on managed interfaces

mlguerrero12:
Users could modify the settings of VFs which have been configured by the sriov operator. This PR starts the reconciliation loop when these changes are detected in the generic plugin.


SchSeba:
Hi @mlguerrero12 thanks for the PR!
question this PR is to cover a bug? asking because now everytime we will go over a much heavier function to understand that we don't need todo anything then just check the the last apply configuration

dmoessne:
> Hi @mlguerrero12 thanks for the PR! question this PR is to cover a bug? asking because now everytime we will go over a much heavier function to understand that we don't need todo anything then just check the the last apply configuration

Hello @SchSeba indeed this is supposed to mitigate https://issues.redhat.com/browse/OCPBUGS-22337

adrianchiris:
@dmoessne can you share the details publicly ?

currently the assumption is that config daemon owns sriov configuration (unless otherwise instructed e.g via externally managed feature) so no one is expected to change the configuration.

based on that assumption comparing the states is expected to be sufficient. that said, im not against checking the system for the actual state however we need to understand IF we actually need this as it brings with it overhead in the daemon reconcile loop.

mlguerrero12:
@SchSeba, @adrianchiris, we have a customer that manually changed the mtu of some VFs that were previously configured with a sriov policy. The sriov config daemon detects these changes (via pollNicStatus which runs every 30s) and updates the status of the sriov network node state object. However, the reconciliation loop doesn't trigger because it only does when the generation field changes (this is when the spec of the sriov-network-node-state changes). We already informed the customer that they shouldn't modify the mtu without using sriov policies. I'm currently working on a PR to also make the operator react when the status doesn't match what was set in the spec.

This customer was relying on this bug. The problem was that due to a recent issue with etcd encryption, the kube-api server pods were restarted (weekly). The resource version of objects were incremented due to this, including the one for the "device-plugin-config" config map. The resource version of this config map is part of the sriov-network-node-state spec. Due to this, the sriov config daemon started its reconciliation loop. The function OnNodeStateChange of the generic plugin detected changes in the MTU and returned "needDrain = true". The node was drained and this is the reason why the customer complained. They didn't understand why the node was repeatedly draining. The Apply function of the generic plugin never reconciled the status after the draining because it only does when the spec changes. This is what I'm fixing with this PR. Regarding the cost, yes, it has an extra cost but it's just some for loops to compare some parameters of the interfaces. Nothing to be concerned about I would think but I could speed this up with some routines.

To sum up, customer shouldn't be changing the mtu without using sriov policies of VFs that are managed by the sriov operator. We identified some issues in the logic of the sriov daemon. One of them is solved with this PR.

mlguerrero12:
I added a commit to start the reconciliation loop of the sriov config daemon when status of interfaces are externally modified (see my previous comment). Updated title and description of PR.

mlguerrero12:
Emulated PF doesn't seem to react to changes of mtu with echo 2000 > /sys/bus/pci/devices/%s/net/%s/mtu

pkg/plugins/generic/generic_plugin.go
```diff
@@ -124,6 +124,31 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
 	return
 }
 
+// OnNodeStatusChange verify whether SriovNetworkNodeState CR status present changes on configured VFs.
+func (p *GenericPlugin) OnNodeStatusChange(new *sriovnetworkv1.SriovNetworkNodeState) bool {
```
zeeke, line 128: I see no need to create this new interface method on the plugin interface: there is only one implementation, explicitly invoked on the generic instance. What about moving this method next to `utils.NeedUpdate(...)`?

test/conformance/tests/test_sriov_operator.go
```diff
@@ -200,10 +200,13 @@ var _ = Describe("[sriov] operator", func() {
 				discoveryFailed = node == "" || resourceName == "" || numVfs < 5
 			} else {
 				node = sriovInfos.Nodes[0]
-				createVanillaNetworkPolicy(node, sriovInfos, numVfs, resourceName)
+				sriovDevice = createVanillaNetworkPolicy(node, sriovInfos, numVfs, resourceName)
```
zeeke, line 203: Value of `sriovDevice`  is never read. What about reverting the BeforeEach part?

test/conformance/tests/test_sriov_operator.go
```diff
@@ -912,6 +915,53 @@ var _ = Describe("[sriov] operator", func() {
 					return runningPodB.Status.Phase
 				}, 3*time.Minute, time.Second).Should(Equal(corev1.PodRunning))
 			})
+
+			It("should reconcile managed VF if status changes", func() {
+				vf := sriovDevice.VFs[0]
+				mtu := vf.Mtu
```
zeeke, line 921: nit: 
```suggestion
				originalMtu := vf.Mtu
```

zeeke (commented):
Left a few comments

pkg/plugins/generic/generic_plugin.go
mlguerrero12: That's what I was planning to do initially, but like you said, this is something specific to the generic plugin. It is wrong to expose it generally. For instance, when the platform type is VirtualOpenstack, we don't load the generic plugin at all. This also provides a framework for other plugins to perform their own implementation when status changes. I see that Mellanox also inspects the interfaces or perhaps something in the virtual plugin. That can be added in the future.

test/conformance/tests/test_sriov_operator.go
mlguerrero12: correct, this is a leftover of a previous implementation. Don't understand what part you want me to revert. What I'm adding is something missing. After creating the general sriov policy, the info about interfaces is not updated.

test/conformance/tests/test_sriov_operator.go
mlguerrero12: ack, thanks

pkg/plugins/generic/generic_plugin.go
bn222: I haven't understood the full PR yet, but one comment is that **IF** we decide to add `OnNodeStatusChange` function, we need a much clearer name that is has more than 2 characters differing from `OnNodeStateChange`

mlguerrero12: The main idea of the pr is to start the reconciliation when an user externally changes a parameter of a VF which was set via the operator with a sriov policy. My argument is that the operator should not only act when the user changes the spec but also when the user externally modifies something that affects what was set in the spec.

For instance, we have a customer that was setting the mtu to 9000 with a sriov policy. After that, they were manually changing the mtu of some vfs. Nothing happened. The status of the `SriovNetworkNodeState`  object detected the change but the reconciliation was not triggered because we only do when the generation changes (spec changes). However, this became an issue when the resource version of the device config map, which is part of that spec, weekly changed due to a bug on the kube-api server. The reconciliation loop started and we could see 2 issues.

1.  `OnNodeStateChange` of the generic plugin returned that a drain was needed because the spec and status didn't match. The customer was complaining of this unexpected draining. Unexpected from their side.
2. `Apply` of the generic plugin never reconcile this discrepancy because its logic compares the current spec with the spec of the previously applied `SriovNetworkNodeState` object. Due to this, user was experiencing a weekly draining of nodes (the api-server bug happened every 7 days). We shouldn't rely on a save object to decide when to apply changes. This goes against the logic of `OnNodeStateChange` and behaves differently when the node restarts (saved data is lost).

This pr is addressing all this. Perhaps, `CheckStatusChanges` is better. What do you think @bn222 ?

bn222: if the main idea is to detect when a parameter changes, then I think we need to properly address this instead of only checking when reconciling. I'm in favour of snooping netlink messages to intercept whenever mtu (or other things) change and react immediately. Waiting until reconcile happens makes things a lot more difficult to debug.

The new proposed name is in any case better, but my above comment should be address too I think.

mlguerrero12: I don't agree. That would add more complexity for something that we don't expect users do often. Besides, we don't have the need to react immediately.

Just to be clear, the idea is not act only when reconciliation happens. I'm relying on the `Run` method of `NodeStateStatusWriter` which runs `pollNicStatus` and updates the status of the `SriovNetworkNodeState` every 30 seconds. So, at the worst case, the reconciliation would happen after 30 seconds of making a change.

I think this is good enough for our use cases. No need to add more complexity. This is a good start. Perhaps if some day a user wants their externally changes be reverted immediately, then we can explore other alternatives, such as snooping netlink messages as you said.

bn222: How do you address debuggability today? Having snooping in place would allow us to put a message like "Detected user changing xyz but policy is in place to mange xyz so changing it back", with a timestamp.

mlguerrero12: First of all, we don't refer to policies at all in the daemon. Everything goes around the `SriovNetworkNodeState` object. Regarding debuggability, note that the current logic is to only start the reconciliation loop when the generation field of `SriovNetworkNodeState` has changed. So, when it hasn't changed, the idea is to run `OnNodeStatusChange` (soon to be `CheckStatusChanges`). If changes are detected, there will be logs like so:

log.Log.Info("OnNodeStatusChange(): status changed for interface", "address", iface.PciAddress)

So, there are info for debuggability.

mlguerrero12: changed the name to `CheckStatusChanges`

pkg/daemon/daemon.go
```diff
@@ -724,6 +701,58 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 	return nil
 }
 
+func (dn *Daemon) skipReconciliation(latestState *sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
```
adrianchiris, line 704: nit: shouldSkipReconciliation ?

pkg/daemon/daemon.go
```diff
@@ -724,6 +701,58 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+			}
+			// wait for writer to refresh status
+			<-dn.syncCh
+		}
+		return true, nil
+	}
+
+	// Verify changes in the spec of the SriovNetworkNodeState CR.
+	if dn.nodeState.GetGeneration() == latestState.GetGeneration() {
+		genericPlugin, ok := dn.enabledPlugins[GenericPluginName]
```
adrianchiris, line 731: any reason not to execute this for all plugins?

pkg/daemon/daemon.go
```diff
@@ -724,6 +701,58 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+		if ok {
+			// Verify changes in the status of the SriovNetworkNodeState CR.
+			if genericPlugin.CheckStatusChanges(latestState) {
+				return false, nil
+			}
+		}
+
+		log.Log.V(0).Info("nodeStateSyncHandler(): Interface not changed")
```
adrianchiris, line 739: log msg is wrong here

pkg/plugins/plugin.go
```diff
@@ -5,12 +5,14 @@ import (
...
+	// Spec returns the SpecVersion followed by plugin
 	Spec() string
-	// Invoked when SriovNetworkNodeState CR is created or updated, return if need dain and/or reboot node
-	OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeState) (bool, bool, error)
+	// OnNodeStateChange is invoked when SriovNetworkNodeState CR is created or updated, return if need dain and/or reboot node
+	OnNodeStateChange(*sriovnetworkv1.SriovNetworkNodeState) (bool, bool, error)
 	// Apply config change
 	Apply() error
+	// CheckStatusChanges checks status changes on the SriovNetworkNodeState CR for configured VFs.
+	CheckStatusChanges(*sriovnetworkv1.SriovNetworkNodeState) bool
```
adrianchiris, line 17: cant we use `OnNodeStateChange` to detect that changes are needed ?
perhaps that method need to be extended with another boolean to indicate that changes are required to be performed on the node.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -126,6 +124,31 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
...
+func (p *GenericPlugin) CheckStatusChanges(new *sriovnetworkv1.SriovNetworkNodeState) bool {
+	log.Log.Info("generic-plugin CheckStatusChanges()")
+
+	changed := false
+	for _, iface := range new.Spec.Interfaces {
+		found := false
+		for _, ifaceStatus := range new.Status.Interfaces {
+			if iface.PciAddress == ifaceStatus.PciAddress {
+				found = true
+				if utils.NeedUpdate(&iface, &ifaceStatus) {
```
adrianchiris, line 137: This does not cover all cases e,g kernel args. (all the more reason to not have yet another API to check changes)

pkg/daemon/daemon.go
mlguerrero12: ack

pkg/daemon/daemon.go
mlguerrero12: The only reason is that I don't know well other plugins. For instance, I don't know what to check for the Mellanox plugin. That would be a TODO.

pkg/daemon/daemon.go
mlguerrero12: That was already there. But I don't think it is completely wrong, perhaps incomplete. It logs that the Interface parameter of the spec didn't change.

pkg/plugins/plugin.go
mlguerrero12: that would be convoluted in my opinion. Besides, the logic is different, `OnNodeStateChange` starts from the interfaces of the Status whereas `CheckStatusChange` starts from the interfaces of the Spec. The latter is much simpler and faster.

pkg/plugins/generic/generic_plugin.go
mlguerrero12: Indeed, it is missing that part. Thanks. I will add it.

However, I do believe we need another api. The logic of `needDrainNode` is different to what I have here. The check here starts from the interfaces of the Spec. It's simpler and faster. Same for the logic in `needRebootNode`, we don't to set the kernel args, but we need to check if the desiredKernelArgs parameter still matches. It' simpler.

mlguerrero12 (commented):
Thanks @adrianchiris for the comments.

pkg/plugins/generic/generic_plugin.go
adrianchiris: OnNodeStateChange should not "Apply" anything during the call.
current implementation has drifted from this, we should fix imo.

pkg/daemon/daemon.go
adrianchiris: calling method is different...

pkg/plugins/plugin.go
adrianchiris: signature is the same, checking logic should be the same IMO
why essentially duplicate the logic and have a bug introduced the minute a PR lands that changes one and not the other ?

pkg/daemon/daemon.go
adrianchiris: ack we should create an issue on it eventually

pkg/plugins/plugin.go
adrianchiris: thx for the detailed information @mlguerrero12, and appologies for delay in reviewing this one.

pkg/daemon/daemon.go
adrianchiris: could this bit iterate over all loaded plugins ?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -124,6 +122,42 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
 	return
 }
 
+// CheckStatusChanges verify whether SriovNetworkNodeState CR status present changes on configured VFs.
+func (p *GenericPlugin) CheckStatusChanges(new *sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
```
adrianchiris, line 126: nit: new -> current ?

pkg/daemon/daemon.go
```diff
@@ -685,6 +662,63 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 	return nil
 }
 
+func (dn *Daemon) shouldSkipReconciliation(latestState *sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
```
adrianchiris, line 665: perhaps extend daemon UT with some test cases for this.

e.g when plugin reports that status changed ?

adrianchiris:
gave this one another round. overall looks OK, added a couple of comments.

pkg/plugins/generic/generic_plugin.go
mlguerrero12: right, thanks

mlguerrero12:
> gave this one another round. overall looks OK, added a couple of comments.

@adrianchiris, thanks for the comments. I addressed them. Please have a look again.

Failing test is not related to this PR. `Debug logging should be visible in multus pod` is failing in both jobs.

mlguerrero12:
@zeeke, please have a look

pkg/daemon/daemon.go
```diff
@@ -413,16 +413,26 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+	// load plugins if it has not loaded
+	if len(dn.enabledPlugins) == 0 {
+		dn.enabledPlugins, err = enablePlugins(latestState, dn.HostHelpers)
+		if err != nil {
+			log.Log.Error(err, "nodeStateSyncHandler(): failed to enable vendor plugins")
+			return err
+		}
+	}
+
+	// TODO: include this logic specific of systemd into shouldSkipReconciliation method. Previous workflow hasn't been modified. This part is executed first.
```
adrianchiris, line 434: any reason not to move it ?

OR if you prefer to keep this here, remove the TODO.

I prefer not to have this comment in code.

pkg/plugins/intel/intel_plugin.go
```diff
@@ -35,11 +35,17 @@ func (p *IntelPlugin) Spec() string {
...
 // OnNodeStateChange Invoked when SriovNetworkNodeState CR is created or updated, return if need dain and/or reboot node
-func (p *IntelPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
+func (p *IntelPlugin) OnNodeStateChange(*sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
 	log.Log.Info("intel-plugin OnNodeStateChange()")
 	return false, false, nil
 }
 
+// OnNodeStatusChange verify whether SriovNetworkNodeState CR status present changes on configured VFs.
+func (p *IntelPlugin) CheckStatusChanges(*sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
+	log.Log.Info("intel-plugin CheckStatusChanges()")
```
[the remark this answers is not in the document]
mlguerrero12: ack

pkg/daemon/daemon.go
mlguerrero12: I will remove the TODO. The logic around systemd is easier to understand with that part outside I think

mlguerrero12:
@zeeke @adrianchiris, PTAL again. Thanks

mlguerrero12:
failing test not related

zeeke (approved):
LGTM

SchSeba (commented):
Great work!

I just left some small comments.

two main points:
1. please check the MTU stuff
if the pod start using the nic we lose the MTU in the vf info will that not recall a re-check?
```
before pod:
- Vfs:
    - deviceID: 10ca
      driver: igbvf
      mac: 0a:16:38:97:01:6e
      mtu: 9000
      name: eth2
      pciAddress: "0000:29:10.0"
      vendor: "8086"
      vfID: 0

after pod:
- Vfs:
    - deviceID: 10ca
      driver: igbvf
      pciAddress: "0000:29:10.0"
      vendor: "8086"
      vfID: 0

so both the nic name and the MTU are not visable anymore
```

another point is https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/423350cc283bf1ec3fc97be419d48b914d046cf2/api/v1/helper.go#L276 this will re-call a reconfiguration again and again no?

pkg/daemon/daemon.go
```diff
@@ -413,16 +413,25 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+	if len(dn.enabledPlugins) == 0 {
+		dn.enabledPlugins, err = enablePlugins(latestState, dn.HostHelpers)
+		if err != nil {
+			log.Log.Error(err, "nodeStateSyncHandler(): failed to enable vendor plugins")
+			return err
+		}
+	}
+
+	if dn.nodeState.GetGeneration() == latestState.GetGeneration() {
 		if vars.UsingSystemdMode {
```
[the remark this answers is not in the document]
mlguerrero12: ack

pkg/daemon/daemon.go
```diff
@@ -683,6 +659,62 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+			}
+			// wait for writer to refresh status
+			<-dn.syncCh
+		}
+		return true, nil
+	}
+
+	// Verify changes in the spec of the SriovNetworkNodeState CR.
+	if dn.nodeState.GetGeneration() == latestState.GetGeneration() {
+		changed := false
```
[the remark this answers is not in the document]
mlguerrero12: ack

pkg/plugins/generic/generic_plugin.go
```diff
@@ -124,6 +122,42 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
...
+				if sriovnetworkv1.NeedToUpdateSriov(&iface, &ifaceStatus) {
+					changed = true
+					log.Log.Info("CheckStatusChanges(): status changed for interface", "address", iface.PciAddress)
+				}
+				break
+			}
+		}
+
+		if !found {
+			log.Log.Info("CheckStatusChanges(): no status found for interface", "address", iface.PciAddress)
```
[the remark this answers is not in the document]
mlguerrero12: No, I don't think we should return an error for that here. It doesn't affect the execution of the operator. On the other hand, if the mutating webhook is disabled, the `deviceType` is not set to `netdevice` when empty. For this case, `NeedToUpdateSriov`fails here `if groupSpec.DeviceType != consts.DeviceTypeNetDevice`. I'm creating an issue for it.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -333,18 +368,30 @@ func (p *GenericPlugin) addVfioDesiredKernelArg(state *sriovnetworkv1.SriovNetwo
 	}
 }
 
-func (p *GenericPlugin) needRebootNode(state *sriovnetworkv1.SriovNetworkNodeState) (needReboot bool, err error) {
-	needReboot = false
+func (p *GenericPlugin) needRebootNode(state *sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
+	var err error
+	needReboot := false
+	updateNode := false
```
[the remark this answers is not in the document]
mlguerrero12: ack

pkg/plugins/generic/generic_plugin_test.go
```diff
@@ -141,6 +141,106 @@ var _ = Describe("Generic plugin", func() {
 			Expect(needDrain).To(BeTrue())
 		})
 
+		It("should not detect changes on status", func() {
```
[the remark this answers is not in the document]
mlguerrero12: ack

pkg/plugins/intel/intel_plugin.go
```diff
@@ -35,11 +35,16 @@ func (p *IntelPlugin) Spec() string {
 }
 
 // OnNodeStateChange Invoked when SriovNetworkNodeState CR is created or updated, return if need dain and/or reboot node
-func (p *IntelPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
+func (p *IntelPlugin) OnNodeStateChange(*sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
```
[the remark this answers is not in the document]
mlguerrero12: sure

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -160,6 +160,11 @@ func (p *K8sPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeState)
 	return
 }
 
+// OnNodeStatusChange verify whether SriovNetworkNodeState CR status present changes on configured VFs.
+func (p *K8sPlugin) CheckStatusChanges(*sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
```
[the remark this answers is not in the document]
mlguerrero12: it will be addressed in a follow-up pr. Will create a ticket.

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -171,6 +171,11 @@ func (p *MellanoxPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeS
 	return
 }
 
+// OnNodeStatusChange verify whether SriovNetworkNodeState CR status present changes on configured VFs.
+func (p *MellanoxPlugin) CheckStatusChanges(*sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
```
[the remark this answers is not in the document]
mlguerrero12: it will be addressed in a follow-up pr. Will create a ticket

pkg/plugins/plugin.go
```diff
@@ -6,12 +6,14 @@ import (
 
 //go:generate ../../bin/mockgen -destination mock/mock_plugin.go -source plugin.go
```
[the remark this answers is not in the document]
mlguerrero12: sure, thanks

test/conformance/tests/test_sriov_operator.go
```diff
@@ -2539,6 +2595,7 @@ func createVanillaNetworkPolicy(node string, sriovInfos *cluster.EnabledNodes, n
 			},
 			NumVfs:       numVfs,
 			ResourceName: resourceName,
+			Mtu:          1500,
```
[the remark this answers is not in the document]
mlguerrero12: no, I don't think it is an issue. We don't have mtu related tests within the generic SriovNetworkNodePolicy node.

mlguerrero12:
> Great work!
> 
> I just left some small comments.
> 
> two main points:
> 
> 1. please check the MTU stuff
>    if the pod start using the nic we lose the MTU in the vf info will that not recall a re-check?
> 
> ```
> before pod:
> - Vfs:
>     - deviceID: 10ca
>       driver: igbvf
>       mac: 0a:16:38:97:01:6e
>       mtu: 9000
>       name: eth2
>       pciAddress: "0000:29:10.0"
>       vendor: "8086"
>       vfID: 0
> 
> after pod:
> - Vfs:
>     - deviceID: 10ca
>       driver: igbvf
>       pciAddress: "0000:29:10.0"
>       vendor: "8086"
>       vfID: 0
> 
> so both the nic name and the MTU are not visable anymore
> ```
> 
> another point is
> 
> https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/423350cc283bf1ec3fc97be419d48b914d046cf2/api/v1/helper.go#L276
> 
> this will re-call a reconfiguration again and again no?

Thanks @SchSeba for your comments.

About your points, the first one is not an issue since that mtu check is for the PF. 

The second one is a good catch. Thanks. For the moment, I will skip checking the status for externally managed interfaces. I will create a ticket to investigate how we could avoid returning always true for netdev externally managed interfaces since it is not only affecting the status check but also forces an unneeded drain when only the mtu changes for this type of interfaces. Additionally, this causes issues as well https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/536. That's something we still need to discuss.

pkg/plugins/generic/generic_plugin.go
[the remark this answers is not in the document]
mlguerrero12: https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/629

pkg/plugins/k8s/k8s_plugin.go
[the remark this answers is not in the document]
mlguerrero12: https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/630

pkg/plugins/mellanox/mellanox_plugin.go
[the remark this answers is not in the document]
mlguerrero12: https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/631

mlguerrero12:
created https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/632 as mentioned above.

@SchSeba, @adrianchiris, please have a look again. I've addressed your comments.

mlguerrero12:
@SchSeba, @zeeke, @adrianchiris, please have a look at this PR again. Comments have been addressed.

SchSeba (commented):
nice work

please check the externally manage I feel that we are going to have a look here moving the daemon from inProgress to Success every reconcile loop no?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -145,6 +143,42 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
...
+				found = true
+				if sriovnetworkv1.NeedToUpdateSriov(&iface, &ifaceStatus) {
+					changed = true
+					log.Log.Info("CheckStatusChanges(): status changed for interface", "address", iface.PciAddress)
+				}
+				break
+			}
+		}
+		if !found {
+			log.Log.Info("CheckStatusChanges(): no status found for interface", "address", iface.PciAddress)
```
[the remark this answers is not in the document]
mlguerrero12: see answer in previous comment

pkg/daemon/daemon.go
```diff
@@ -700,6 +674,61 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+			<-dn.syncCh
+		}
+		return true, nil
+	}
+
+	// Verify changes in the spec of the SriovNetworkNodeState CR.
+	if dn.nodeState.GetGeneration() == latestState.GetGeneration() {
+		for _, p := range dn.loadedPlugins {
+			// Verify changes in the status of the SriovNetworkNodeState CR.
+			changed, err := p.CheckStatusChanges(latestState)
```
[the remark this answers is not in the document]
mlguerrero12: not longer an issue

mlguerrero12:
@e0ne, @Eoghan1232, could you please have a look at this? Thanks! (failing test is not related to this PR)

mlguerrero12:
failing test is not related

pkg/plugins/generic/generic_plugin.go
```diff
@@ -141,6 +139,36 @@ func (p *GenericPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
...
 }
 
+// CheckStatusChanges verify whether SriovNetworkNodeState CR status present changes on configured VFs.
+func (p *GenericPlugin) CheckStatusChanges(current *sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
+	log.Log.Info("generic-plugin CheckStatusChanges()")
+
+	for _, iface := range current.Spec.Interfaces {
+		found := false
+		for _, ifaceStatus := range current.Status.Interfaces {
+			if iface.PciAddress == ifaceStatus.PciAddress && !iface.ExternallyManaged {
```
SchSeba, line 149: Hi please a TODO and comment here that we skip the externallyManage with a link to the issue we open on the repo please

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -154,6 +154,11 @@ func (p *K8sPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeState)
...
+	return false, nil
```
SchSeba, line 159: Hi please a TODO and comment here that we skip the externallyManage with a link to the issue we open on the repo please

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -171,6 +171,11 @@ func (p *MellanoxPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeS
...
+	return false, nil
```
SchSeba, line 176: Hi please a TODO and comment here that we skip the externallyManage with a link to the issue we open on the repo please

SchSeba (approved):
LGTM just small nits to add comments and we can merge this one.

nice work!

pkg/plugins/generic/generic_plugin.go
mlguerrero12: done

pkg/plugins/k8s/k8s_plugin.go
mlguerrero12: the problem is with the method NeedToUpdateSriov which is not used by this plugin

pkg/plugins/virtual/virtual_plugin.go
```diff
@@ -67,6 +67,11 @@ func (p *VirtualPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeSt
 	return
 }
 
+// OnNodeStatusChange verify whether SriovNetworkNodeState CR status present changes on configured VFs.
+func (p *VirtualPlugin) CheckStatusChanges(*sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
+	return false, nil
```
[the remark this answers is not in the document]
mlguerrero12: same as above

mlguerrero12:
> LGTM just small nits to add comments and we can merge this one.
> 
> nice work!

Thanks @SchSeba. I added the TODO only for the generic plugin since other plugins have a different logic.

Please reopen https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/632. it was mistakenly closed.

SchSeba (approved):
LGTM thanks!

mlguerrero12:
@Eoghan1232 @e0ne, please have a look at this when you have time. Thanks!

SchSeba (approved):
LGTM!

SchSeba:
merging this one thanks for all the help folks!

PR #531, opened 2023-10-26
Minor improvments for systemd mode implementation

ykulazhenkov:
Couple improvements for systemd mode implementation to cover edge cases:
- check that service enabled in addition to check which validates that service exist
- remove old result file when we update configuration for systemd service to make sure that there is no chance to read outdated info

The original implementation can't detect situation when the service exists on the node, but for some reason was disabled by the user. In this case daemon was not able to detect this misconfiguration.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -466,7 +466,7 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 
 	if dn.nodeState.GetGeneration() == latest {
 		if dn.useSystemdService {
-			serviceExist, err := dn.serviceManager.IsServiceExist(systemd.SriovServicePath)
+			serviceEnabled, err := dn.serviceManager.IsServiceEnabled(systemd.SriovServicePath)
 			if err != nil {
 				log.Log.Error(err, "nodeStateSyncHandler(): failed to check if sriov-config service exist on host")
 				return err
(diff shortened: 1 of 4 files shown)


ykulazhenkov:
@adrianchiris @e0ne @SchSeba  please check this PR. Thanks

pkg/service/service_manager.go
```diff
@@ -40,6 +41,24 @@ func (sm *serviceManager) IsServiceExist(servicePath string) (bool, error) {
 	return true, nil
 }
 
+// IsServiceEnabled check if service is enabled
```
e0ne, line 44: nit: "check if service is enabled" -> "check if service is exists and enabled"

e0ne (approved):

pkg/service/service_manager.go
ykulazhenkov: done

pkg/daemon/daemon.go
```diff
@@ -572,14 +572,23 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 	// or there is a new config we need to apply
 	// When using systemd configuration we write the file
 	if dn.useSystemdService {
-		r, err := systemd.WriteConfFile(latestState, dn.devMode, dn.platform)
+		systemdConfModified, err := systemd.WriteConfFile(latestState, dn.devMode, dn.platform)
```
bn222, line 575: This is a great opportunity to clean this up using https://github.com/coreos/go-systemd

Can you refactor and get rid of all the manual systemd calls? The above library is the official one. That would mean we can clean up pkg/systemd/systemd.go by using that lib.

SchSeba: Hi @bn222 that package is great does it only work for RHEL/COREOS or also other operating systems?

adrianchiris: @bn222  can you open an issue for this ? i do not want to turn this bug fix into a refactor.

adrianchiris (approved):

pkg/daemon/daemon.go
ykulazhenkov: I agree that this is a good idea to switch to the library, but let's handle this as a separate PR with refactoring.

bn222: @SchSeba I believe it works in general. 
@adrianchiris The downside of having a separate PR is that it tends to get forgotten. I've opened https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/533, so let's not let it linger for too long.

adrianchiris:
@ykulazhenkov can you rebase this one ?

zeeke (commented):
LGTM
@ykulazhenkov please rebase

ykulazhenkov:
PR rebased

wizhaoredhat:
lgtm

PR #532, opened 2023-10-30
fix: sriov config service unit-file dependencies

adrianchiris:
run after initial networking has been configured
but before NetworkManager and kubelet.

that way we ensre NICs have a netdev to use during configurations.


SchSeba (approved):
/lgtm

PR #536, opened 2023-11-02
Remove inner comparison for numVfs in externally managed

mlguerrero12:


pkg/utils/utils.go
```diff
@@ -414,11 +414,9 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 	// set numVFs
 	if iface.NumVfs != ifaceStatus.NumVfs {
 		if iface.ExternallyManaged {
```
adrianchiris, line 416: Why have you decided to remove the case where M Vfs are configured externally but only N VFs are to be used by the policy ? (where M >= N)

maybe the err msg need to be fixed instead ?

also i see the webhook also allows this:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/e35bc9ec131984d8ed331bb573b5fb7d239fa5b5/pkg/webhook/validate.go#L313

mlguerrero12: @adrianchiris,  because it doesn't work that way. Even if you have numVfs less than what it was configured. The code will fetch the pci address of all its vfs and attempt to configure them (driver and mtu). https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L450

I believe the case you're referring can be achieved with:

  nicSelector:
    pfNames: ["ens3f0#0-3"]

but I think I have a bug here as well (for all cases, not only externally managed). Even if we configure the policy this way, vfs which do not belong to any vfGroup of the interface in the spec will be configured. That's because we are missing a "found" var to not continue if the vf is not in any group.
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L466

I'll create a different pr for this.

Another reason is because there will be always a inconsistency with the value of numVfs reflected in the status.

Hadn't seen the protection in the webhook. It seems I need to change it there as well.

adrianchiris: > because it doesn't work that way. 

The question is how we want this to work, there  are parts of the code who takes this case into consideration (M > N)

> Hadn't seen the protection in the webhook. It seems I need to change it there as well.

lets get some feedback from @SchSeba as he is the author of this feature (design and impl) before deciding to change things...

what id like to get agreement upon is the desired behaviour for such case (where M > N)

do we:
1. not allow it, and fail both in webhook and daemon
2. allow it and adjust daemon code to handle up to N VFs, taking into account vfgroup.
3. other ?

mlguerrero12: @adrianchiris, @SchSeba, I wrote a test (https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/547) to show that our code doesn't support to have numVfs less to what was externally configured. As I said above, the code gets all vfs of the interface and proceeds to configure them all. You can see that I manually created 2 Vfs and a policy with numVfs equal to 1. All two Vfs were modified (mtu equal to 2000).

The use case that @adrianchiris is mentioning is supported with pf partitioning. There is even an example on @SchSeba design document (https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/doc/design/externally-manage-pf.md#policy-example). However, there is a bug there because there is something missing in the for loop for Vfs which are not part of a Vf group. I'm fixing that in https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/539 . PR also includes a refactoring and a test as requested by some reviewers.

Another thing I noticed which is also a contradiction is that the resource capacity reported on the status of the node. I'm not familiar with the device plugin yet, but I guess it works like here. When there is no pf partition, it gets all Vfs of the interface and reports them as available capacity of the resource. The test I created has capacity equal to 2 even though the policy was created with numVfs equal to 1. "openshift.io/testexternally": "2".

I believe all these reasons justify what I'm proposing on this PR. We are not reducing the available use cases. There is support for the scenario we thought we were covering here by using pf partitioning.

adrianchiris: @SchSeba any inputs on this one ?

adrianchiris: so IIUC this PR needs to be closed and a different one needs to be created as we do want to allow to have more VFs than in policy for externally managed ?

if yes, @SchSeba/ @mlguerrero12 mind creating an issue on this for tracking ?

adrianchiris:
@mlguerrero12 @SchSeba is this PR still relevant ?

PR #539, opened 2023-11-07
Skip vf configuration when not found in VF group

mlguerrero12:


pkg/utils/utils.go
```diff
@@ -476,6 +478,11 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 				}
 			}
 
+			// Do not configure VF which is not part of any VF Group.
+			if !found {
+				continue
+			}
```
bn222, line 484: This is essentially filtering the list of addresses to process. Now, the body of the loop does much too much imo. Let's take the opportunity to refactor this whole function to have a more proper way to filter the list

mlguerrero12: done, please have a look again

mlguerrero12:
created a test in https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/541 to show the issue. I added here as well. It should pass now.

mlguerrero12:
OCP job passed (with new added test). There is a problem setting up the k8s job.

pkg/utils/utils.go
```diff
@@ -458,24 +458,26 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
-			for i, group = range iface.VfGroups {
-				if err != nil {
-					log.Log.Error(err, "configSriovDevice(): unable to get VF id", "device", iface.PciAddress)
-				}
-				if sriovnetworkv1.IndexInRange(vfID, group.VfRange) {
-					isRdma = group.IsRdma
-					if sriovnetworkv1.StringInArray(group.DeviceType, DpdkDrivers) {
-						dpdkDriver = group.DeviceType
-					}
+			if err != nil {
```
adrianchiris, line 464: thx for moving this check !

adrianchiris (approved):

test/conformance/tests/test_sriov_operator.go
```diff
@@ -1071,6 +1071,83 @@ var _ = Describe("[sriov] operator", func() {
 					}))
 				})
 
+				It("Should configure the mtu only for vfs which are part of the partition", func() {
```
zeeke, line 1074: Thanks for moving the test into this PR!

zeeke (approved):
LGTM

Thanks @mlguerrero12

PR #541, opened 2023-11-08
Verify policy is only applied to partions' VFs

mlguerrero12:


mlguerrero12:
DNM - failed OCP job shows that there is a bug. It will be fixed in https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/539. Test will also be included there.

PR #542, opened 2023-11-09
Fix Mellanox plugin logging

e0ne:
This commit fixed an issue introduced during logging refactoring [1].

[1] c155c82c0fc9a3e2db91ea8887a0d3b4ae359bc1


e0ne:
fixed in #543

PR #544, opened 2023-11-09
Reduce operator's controller logs via `OperatorConfig.LogLevel`

zeeke:
Field `SriovOperatorConfig.Spec.LogLevel` controls the verbosity
of the operator logging system. This PR adjusts the operator
controller's log level to that field, as it happens in the
config-daemon.

Log the current and new values for the logger according to what
the user had set in the LogLevel field. Before this commit, a value of
-2 would produce a misleading line:
```
Set log verbose level	{"new-level": 0, "current-level": -2}
```

Also:
- Remove log.Printf(...) instances
- Reduce the number of `syncAllSriovNetworkNodeStates` log lines
- Add end2end test to make sure the LogLevel logic has no regression. Feedback is particularly welcome on this topic, as I'm not sure the test is robust enough to pay back its hypothetical flakiness.

### DIFF
--- a/pkg/apply/apply.go
+++ b/pkg/apply/apply.go
@@ -3,7 +3,6 @@ package apply
 import (
 	"context"
 	"fmt"
-	"log"
 
 	"github.com/pkg/errors"
 
@@ -24,7 +23,6 @@ func DeleteObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruc
 	gvk := obj.GroupVersionKind()
 	// used for logging and errors
 	objDesc := fmt.Sprintf("(%s) %s/%s", gvk.String(), namespace, name)
-	log.Printf("reconciling %s", objDesc)
 
 	if err := IsObjectSupported(obj); err != nil {
 		return errors.Wrapf(err, "object %s unsupported", objDesc)
@@ -36,7 +34,6 @@ func DeleteObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruc
 	err := client.Get(ctx, types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, existing)
 
 	if err != nil && apierrors.IsNotFound(err) {
-		log.Printf("does not exist, do nothing %s", objDesc)
 		return nil
 	}
 	if err != nil {
@@ -45,8 +42,6 @@ func DeleteObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruc
 
 	if err = client.Delete(ctx, existing); err != nil {
 		return errors.Wrapf(err, "could not delete object %s", objDesc)
-	} else {
-		log.Printf("delete was successful")
 	}
 	return nil
 }
@@ -62,7 +57,6 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
 	gvk := obj.GroupVersionKind()
 	// used for logging and errors
 	objDesc := fmt.Sprintf("(%s) %s/%s", gvk.String(), namespace, name)
-	log.Printf("reconciling %s", objDesc)
 
 	if err := IsObjectSupported(obj); err != nil {
 		return errors.Wrapf(err, "object %s unsupported", objDesc)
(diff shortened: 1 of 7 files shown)


pkg/log/log_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+		out, err := os.ReadFile(tempLogFile.Name())
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Expect(string(out)).Should(o.ContainSubstring("test level 0"))
+		o.Expect(string(out)).Should(o.ContainSubstring("test level 1"))
+		o.Expect(string(out)).ShouldNot(o.ContainSubstring("test level 2"))
+	})
+
+	g.It("LogLevel 1", func() {
```
adrianchiris, line 54: nit: LogLevel 2 ?

test/conformance/tests/test_sriov_operator.go
```diff
@@ -2389,6 +2465,67 @@ func setSriovOperatorSpecFlag(flagName string, flagValue bool) {
 	}
 }
 
+func setOperatorConfigLogLevel(level int) {
+	cfg := sriovv1.SriovOperatorConfig{}
+	err := clients.Get(context.TODO(), runtimeclient.ObjectKey{
```
adrianchiris, line 2470: i think both get and update calls should be in the same "Eventually" call
if an update occurred (resource version change) then update call will fail no matter how many times u attempt to set it.
for it to work an update call needs to be passed with the latest resource version IIRC

test/conformance/tests/test_sriov_operator.go
```diff
@@ -171,6 +171,82 @@ var _ = Describe("[sriov] operator", func() {
 			})
 		})
 
+		Context("LogLevel affects operator's logs", func() {
```
adrianchiris, line 174: do we really want this ? 

or we just want to make sure that Options.Level.(zzap.AtomicLevel).Level() has changed following a change in sriovoperator config ?

test/conformance/tests/test_sriov_operator.go
zeeke: good point! Moving get and update to the same Eventually block

test/conformance/tests/test_sriov_operator.go
zeeke: How can we check that? Do you refer to the unit tests in `pkg/log/log_test.go`? I'm torn because I think this e2e test is likely to become flaky. Too complicated to be robust

adrianchiris: ah this one is a different binary 🤦 

i dont have a better idea except: "lets not test it"

zeeke: Let me see if I'm able to test it at [`controllers/sriovoperatorconfig_controller_test.go`](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/bf30003a3d9db7431864b7c8b2ad0a4b2335808b/controllers/sriovoperatorconfig_controller_test.go) level. It would be much faster and more robust. Otherwise, I'm ok with dropping this.

pkg/apply/apply.go
```diff
@@ -74,12 +68,10 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
 	err := client.Get(ctx, types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, existing)
 
 	if err != nil && apierrors.IsNotFound(err) {
-		log.Printf("does not exist, creating %s", objDesc)
```
wizhaoredhat, line 77: In my opinion, These logs seem useful in debugging issues. Shouldn't we try to keep some of them around? Perhaps we can construct a string and pass it back to the caller to log.

wizhaoredhat (commented):
Looks Good! Just a small comment. But I am not too strongly insistent of it.

adrianchiris (approved):

pkg/apply/apply.go
adrianchiris: I believe we mainly care if Apply succeeded or not (and why it failed in case of error)
the rest is noise. current logs are already pretty verbose.

PR #545, opened 2023-11-10
Logs `sriov-network-config-daemon` by default

zeeke:
Container `sriov-network-config-daemon` in the daemonset is the most interesting in the list. Make `kubectl logs ...` target that by default.

### DIFF
--- a/bindata/manifests/daemon/daemonset.yaml
+++ b/bindata/manifests/daemon/daemonset.yaml
@@ -21,6 +21,8 @@ spec:
         component: network
         type: infra
         openshift.io/component: network
+      annotations:
+        kubectl.kubernetes.io/default-container: sriov-network-config-daemon
     spec:
       hostNetwork: true
       hostPID: true


SchSeba (approved):
we only have one running container so it should be the default but I don't see a problem adding this annotation

adrianchiris (approved):

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -21,6 +21,8 @@ spec:
         component: network
         type: infra
         openshift.io/component: network
+      annotations:
+        kubectl.kubernetes.io/default-logs-container: sriov-network-config-daemon
```
adrianchiris, line 25: This one is deprecated: https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-logs-container-deprecated

please use : `kubectl.kubernetes.io/default-container`

adrianchiris (changes requested):
current annot is deprecated

bindata/manifests/daemon/daemonset.yaml
zeeke: thanks for catching that. fixed

zeeke:
> we only have one running container so it should be the default but I don't see a problem adding this annotation

@SchSeba you're right, we have only one non-init container running and kubectl takes the first Spec field:

https://github.com/kubernetes/kubectl/blob/6a876632fa1b55560c6efd3d946de7a21f0d504d/pkg/polymorphichelpers/logsforobject.go#L98

Since the introduction of systemd mode, [sriov-cni and sriov-infiniband-cni containers have been moved to Spec.InitContainers](https://github.com/k8snetworkplumbingwg/sriov-network-operator/commit/f72f2a017df4ad90c4b1bd3b33e2ca02b5bfc67b), which are not taken into consideration when getting the default container to log.

This PR has no actual impact on the upstream operator's behavior.
I'm fine with closing it

SchSeba (approved):

PR #546, opened 2023-11-15
Refactor render of SR-IOV Device plugin

e0ne:
This PR fixes device plugin logic to handle changes on SriovOperatorConfig including `useCDI` flag.

### DIFF
--- a/controllers/helper.go
+++ b/controllers/helper.go
@@ -18,11 +18,27 @@ package controllers
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"strings"
 
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	kscheme "k8s.io/client-go/kubernetes/scheme"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/apply"
 	constants "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/render"
 )
 
 var webhooks = map[string](string){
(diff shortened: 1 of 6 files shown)


controllers/helper.go
```diff
@@ -61,3 +77,156 @@ func GetDefaultNodeSelector() map[string]string {
 	return map[string]string{"node-role.kubernetes.io/worker": "",
 		"kubernetes.io/os": "linux"}
 }
+
+func SyncPluginDaemonObjs(ctx context.Context, client k8sclient.Client, scheme *runtime.Scheme, dp *sriovnetworkv1.SriovNetworkNodePolicy, pl *sriovnetworkv1.SriovNetworkNodePolicyList) error {
```
adrianchiris, line 81: This needs to be public ?

controllers/sriovoperatorconfig_controller.go
```diff
@@ -106,6 +106,44 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 		return reconcile.Result{}, err
 	}
 
+	// Fetch the SriovNetworkNodePolicyList
+	policyList := &sriovnetworkv1.SriovNetworkNodePolicyList{}
```
adrianchiris, line 110: The creation of policy is handled by sriovnetworknodepolicy controller i think we can rely on that controller to create the policy.

perhaps we need to fail if policy is not found in sriovnetworknodepolicy controller Reconcile call

SchSeba:
Hi @e0ne can you please rebase this PR

e0ne:
> Hi @e0ne can you please rebase this PR

done

controllers/helper.go
e0ne: no, let's make it private

controllers/sriovoperatorconfig_controller.go
e0ne: we do it at line ##120

adrianchiris: returning no error and empty reconcile will not requeue, we should requeue here imo to let node policy controller create the default policy.

controllers/sriovoperatorconfig_controller.go
```diff
@@ -109,6 +109,44 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	defaultPolicy := &sriovnetworkv1.SriovNetworkNodePolicy{}
+	err = r.Get(ctx, types.NamespacedName{Name: consts.DefaultPolicyName, Namespace: namespace}, defaultPolicy)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// Default policy object not found, create it.
```
adrianchiris, line 130: do we want to create the policy here ? shouldnt the node policy controller create it ?

e0ne: fixed

adrianchiris (approved):
LGTM

controllers/helper.go
```diff
@@ -62,3 +78,156 @@ func GetDefaultNodeSelector() map[string]string {
 	return map[string]string{"node-role.kubernetes.io/worker": "",
 		"kubernetes.io/os": "linux"}
 }
+
+func syncPluginDaemonObjs(ctx context.Context, client k8sclient.Client, scheme *runtime.Scheme, dp *sriovnetworkv1.SriovNetworkNodePolicy, pl *sriovnetworkv1.SriovNetworkNodePolicyList) error {
+	logger := log.Log.WithName("syncPluginDaemonObjs")
+	logger.Info("Start to sync sriov daemons objects")
```
zeeke, line 84: please, reduce verbosity
```suggestion
	logger.V(1).Info("Start to sync sriov daemons objects")
```

also on L155 and L184

zeeke (commented):
Left some comments about logging verbosity. Also, I think failing unit tests relate to these changes.

controllers/helper.go
e0ne: done

controllers/sriovoperatorconfig_controller_test.go
```diff
@@ -11,11 +11,29 @@ import (
...
+	BeforeEach(func() {
+		defaultPolicy := &sriovnetworkv1.SriovNetworkNodePolicy{}
+		defaultPolicy.SetNamespace(namespace)
+		defaultPolicy.SetName(constants.DefaultPolicyName)
+		defaultPolicy.Spec = sriovnetworkv1.SriovNetworkNodePolicySpec{
+			NumVfs:       0,
+			NodeSelector: make(map[string]string),
+			NicSelector:  sriovnetworkv1.SriovNetworkNicSelector{},
+		}
+		Expect(k8sClient.Create(goctx.TODO(), defaultPolicy)).Should(Succeed())
```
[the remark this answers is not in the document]
e0ne: I prefer such approach to make test independent of the test case configuration

PR #547, opened 2023-11-16
DNM: Test ext managed with requested vfs less than configured

mlguerrero12:


SchSeba:
Hi @mlguerrero12 can we close this one?

PR #548, opened 2023-11-17
Fix error overwrite for setSriovNumVfs

mlguerrero12:
When setSriovNumVfs fails, the error is overwritten by the return error of RemoveUdevRule. If the latter function doesn't return an error (most cases), the reconciliation loop will report success even if vfs were not created/configured.

Issue found while trying to run the operator on a node that didn't have sriov enabled on the bios.


adrianchiris (approved):

zeeke (approved):
LGTM
Thanks for fixing this!

PR #549, opened 2023-11-17
Hash contents of cm instead of using resource version

bn222:


SalDaniele:
LGTM

SalDaniele:
@adrianchiris @SchSeba @zeeke @wizhaoredhat @zshi-redhat @e0ne PTAL we are trying to expedite getting this backported for a cu issue.

adrianchiris:
can you say what exactly the issue is in the commit msg and why this change is needed ??

zshi-redhat:
> can you say what exactly the issue is in the commit msg and why this change is needed ?

About why this change is needed:

Short version: It was suggested by our API team that we should not use ResourceVersion of configmap objects to track its content change since the ResourceVersion may change for various reasons (e.g. etcd key rotation, metadata field update etc), not that we assumed it will only change when sriov policy changes.

Long version: We observed that the ResourceVersion of some k8s objects (e.g. configmap, secret) changes as a consequence of etcd key rotation (because apiserver updates these encrypted k8s objects upon etcd key rotation). In sriov operator case, the ResourceVersion of device-plugin-config configmap changes when etcd key rotates, which results in sriovNetworkNodeState generation update, it then causes unnecessary reconcilication of sriov node state in sriov-config-daemon. Given that nodeStateSyncHandler always restarts sriov-device-plugin pod and the etcd key rotation may happen anytime depending on it's rotating intervals, we have seen customer issue that their pods cannot be started due to unexpected admission error from sriov-device-plugin while it gets restarted.

bn222:
@e0ne PTAL

pkg/utils/utils.go
```diff
@@ -921,6 +927,25 @@ func RunCommand(command string, args ...string) (string, error) {
...
 }
 
+func HashConfigMap(cm *corev1.ConfigMap) string {
+	var keys []string
+	for k := range cm.Data {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	var concatenatedString string
```
mlguerrero12, line 937: Using bytes.Buffer is more efficient than allocating a string in every iteration. It looks better as well. However, the performance gain should be negligible for the amount of data that this config map usually contains. So, ignore this comment if you don't want to change it.

mlguerrero12 (commented):
My concern is that the hash of the data doesn't really go with a parameter named `DpConfigVersion`.  I don't think changing the name of the parameter is an option here. So, could you please add a docstring for this parameter in the api to explain what the content of this parameter will be? This way the custom resource will have a description for this parameter. 

Also, please add a body to the commit explaining why this change is needed.

pkg/utils/utils.go
bn222: Pushed a change to optimize this even further.

bn222:
> My concern is that the hash of the data doesn't really go with a parameter named `DpConfigVersion`. I don't think changing the name of the parameter is an option here. So, could you please add a docstring for this parameter in the api to explain what the content of this parameter will be? This way the custom resource will have a description for this parameter.
> 
> Also, please add a body to the commit explaining why this change is needed.

I've added a commit message. @mlguerrero12  PTAL

mlguerrero12:
Have you thought about setting an empty string for this parameter? What's the purpose of it? Is it to have a quick way to know if networknodestate objects and the DP config map are in sync?

bn222:
> Have you thought about setting an empty string for this parameter? What's the purpose of it? Is it to have a quick way to know if networknodestate objects and the DP config map are in sync? A hash wouldn't be something to use as a quick comparison I think. Hence, we could stop using it.

An empty string is a perfectly valid thing to hash. It will be used when the cm data field is empty. 
The effort I'm pushing for breaks down into two pieces.

1. Let's focus on fixing the issue at hand (which we will backport to earlier versions) which is "the way we use resource version is incorrect".
2. We can rewrite parts of operator to finally get rid of the separate CM altogether.

This PR addresses only number 1.

mlguerrero12:
LGTM

pkg/utils/utils.go
```diff
@@ -921,6 +927,22 @@ func RunCommand(command string, args ...string) (string, error) {
 	return stdout.String(), err
 }
 
+func HashConfigMap(cm *corev1.ConfigMap) string {
```
adrianchiris, line 930: add UT to make sure this behaves as expected ?

adrianchiris:
after this change gets deployed, will sriov network operator sync sriov network node states and update DpConfigVersion with the new hash ?if so this will cause config-daemon to reconcile once.

wizhaoredhat:
LGTM. My only concern is upgrades to sriov-network-operator to this new cksum calculation will cause DpConfigVersion to change. Consequently a reboot of the device plugin and possibly a drain of the node.

pkg/utils/utils.go
wizhaoredhat: Some kind of UT like this
```
	configData := make(map[string]string)
	jsonData1 := []byte(`{
	  "resourceList": [
	    {
	      "resourceName": "mgmtvf",
	      "selectors": {
	        "pfNames": [
	          "ens7f0#0-0"
	        ],
	        "IsRdma": false,
	        "NeedVhostNet": false
	      },
	      "SelectorObj": null
	    },
	    {
	      "resourceName": "mlxnics",
	      "selectors": {
	        "pfNames": [
	          "ens7f0#1-11"
	        ],
	        "IsRdma": false,
	        "NeedVhostNet": false
	      },
	      "SelectorObj": null
	    }
	  ]
	}`)
	jsonData2 := []byte(`{
	  "resourceList": [
	    {
	      "resourceName": "mgmtvf",
	      "selectors": {
	        "pfNames": [
	          "ens8f0#0-0"
	        ],
	        "IsRdma": false,
	        "NeedVhostNet": false
	      },
	      "SelectorObj": null
	    },
	    {
	      "resourceName": "mlxnics",
	      "selectors": {
	        "pfNames": [
	          "ens9f0#1-11"
	        ],
	        "IsRdma": false,
	        "NeedVhostNet": false
	      },
	      "SelectorObj": null
	    }
	  ]
	}`)

	configData["worker-225"] = string(jsonData1)
	configData["worker-226"] = string(jsonData2)

	cm := &corev1.ConfigMap{
		TypeMeta: metav1.TypeMeta{
			APIVersion: "v1",
			Kind:       "ConfigMap",
		},
		ObjectMeta: metav1.ObjectMeta{
			Name:      "my-config-map",
			Namespace: "my-namespace",
		},
		Data: configData,
	}

	hash := HashConfigMap(cm)
```

vrindle:
LGTM

bn222:
> after this change gets deployed, will sriov network operator sync sriov network node states and update DpConfigVersion with the new hash ?if so this will cause config-daemon to reconcile once. (which will restart device plugins ?)

It will not forcefully update it during an upgrade. However, if there is any change to the resource version after that, there will be one final restart to update the dp version to the hash.

adrianchiris:
> It will not forcefully update it during an upgrade. However, if there is any change to the resource version after that, there will be one final restart to update the dp version to the hash.

please confirm you have tested it.

pkg/utils/utils.go
wizhaoredhat: Unit tests looks good Balazs!

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+		hash1 := utils.HashConfigMap(cm1)
+		hash2 := utils.HashConfigMap(cm2)
+
+		Expect(hash1).To(Equal(hash2))
+	})
+
+	It("should not change hash for different key orderings", func() {
+		data1 := map[string]string{}
+		data1["key1"] = "value1"
```
wizhaoredhat, line 56: Maybe another test case to test if the data array or empty or each array element is zero.

bn222: Can you elaborate?

adrianchiris:
> I confirm I have swapped out the image of my cluster with this new version and I've seen the hash change when the cm was reconciled.

did device plugin restart after that ?

bn222:
> > I confirm I have swapped out the image of my cluster with this new version and I've seen the hash change when the cm was reconciled.
> 
> did device plugin restart after that ?

yes, there was 1 more final restart for the dp.

adrianchiris:
>yes, there was 1 more final restart for the dp.

personally im OK with it so LGTM,

id like to get confirmation from @SchSeba / @zeeke  that they are ok with it.

@wizhaoredhat i assume that a device plugin restart is tolerable after upgrade ?

pkg/utils/utils_test.go
wizhaoredhat: How does the hash function work if it finds 0 keys. And if it does find some keys. What if the contents are nil.

bn222: hash of empty thing is still a hash. Can a config map contain nil values? I don't think so. The function will still work, but it just ignores all nil values so no need to handle it in a special way imo.

wizhaoredhat: Me and @bn222 talked offline. The first test case should cover hashing consistency. This comment is resolved.

wizhaoredhat:
LGTM

zeeke (approved):
LGTM

Given the @zshi-redhat explanation, using an hash function to check Device Plugin ConfigMap variations makes sense. 

> yes, there was 1 more final restart for the dp.

The final device-plugin restart is a fair price to pay for this improvement IMO.

PR #551, opened 2023-11-24
envtest: Stick to a specific setup-envtest version

zeeke:
Using `@latest` for tools dependecies produces unrepeatable builds which might break at any point in time.

This is particularly important when maintaining released versions which need a backport fix.

Ref:
- https://github.com/openshift/sriov-network-operator/pull/849


wizhaoredhat:
LGTM

SchSeba (approved):
LGTM

PR #552, opened 2023-11-25
Drain controller

SchSeba:
Add a drain controller to move the logic from the daemon to the operator.

this is the first stage to achieve parallel draining based on https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/479

this was done base on the amazing initial work from @e0ne on https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/488


controllers/drain_controller.go
```diff
@@ -0,0 +1,463 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if drainingNodes >= maxParallelNodeConfiguration {
+		reqLogger.Info("MaxParallelNodeConfiguration limit reached for draining nodes re-enqueue the request")
+		// TODO: make this time configurable
+		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
+	}
+
+	// the node request to drain, but we are at the limit so we re-enqueue the request
```
zeeke, line 193: nit: this comment should go a few lines up

pkg/daemon/daemon.go
```diff
@@ -610,38 +500,23 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
-		if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
-			log.Log.Info("nodeStateSyncHandler(): pause MCP")
-			if err := dn.pauseMCP(); err != nil {
-				return err
+	if reqDrain ||
+		(utils.ObjectHasAnnotationKey(dn.desiredNodeState, consts.NodeStateDrainAnnotationCurrent) &&
+			!utils.ObjectHasAnnotation(dn.desiredNodeState, consts.NodeStateDrainAnnotationCurrent, consts.DrainIdle)) {
+		if utils.ObjectHasAnnotation(dn.desiredNodeState, consts.NodeStateDrainAnnotationCurrent, consts.DrainComplete) {
+			log.Log.Info("nodeStateSyncHandler(): the node complete the draining")
+		} else if !dn.isNodeDraining() {
```
zeeke, line 509: I think with these changes "isNodeDraining" is not very speaking anymore. It's about the value of two annotations, which could be 5 different values each.

controllers/drain_controller.go
```diff
@@ -0,0 +1,463 @@
...
+			nodePoolName, err := dr.openshiftContext.GetNodeMachinePoolName(node)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+
+			if err := dr.pauseMCP(nodeNetworkState, nodePoolName); err != nil {
+				return reconcile.Result{}, err
+			}
+
+			nodeStateDrainAnnotationCurrent = nodeNetworkState.Annotations[constants.NodeStateDrainAnnotationCurrent]
```
zeeke, line 213: This value is always `constants.DrainIdle`, as the `if` contains `nodeStateDrainAnnotationCurrent == constants.DrainIdle` and that value comes from L139. isn't it?

zeeke (commented):
Left a few comments. I personally find it pretty hard to understand the DrainController Reconcile method. Maybe we can write it as a FiniteStateMachine?

controllers/drain_controller.go
zeeke: I just noticed `nodeNetworkState.Annotations` is changed to `Draining_MCP_Paused` by `pauseMCP()`

controllers/drain_controller.go
```diff
@@ -0,0 +1,463 @@
...
+func (dr *DrainReconcile) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	req.Namespace = namespace
+	reqLogger := log.FromContext(ctx)
+	reqLogger.Info("Reconciling Drain")
+
+	node := &corev1.Node{}
+	err := dr.Get(ctx, req.NamespacedName, node)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
SchSeba: sure done.

it can be a case of the node was removed the reconcile loop in the sriovOperatorConfig will remove the nodestate

controllers/drain_controller.go
SchSeba: right!

controllers/drain_controller.go
SchSeba: right, I don't like it but this was how it was working I just copy the function from the daemon.

In the parallel PR I will change that but in this one I am trying to reduce the amount of changes

controllers/drain_controller.go
```diff
@@ -0,0 +1,463 @@
...
+			q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+				Namespace: namespace,
+				Name:      e.ObjectNew.GetName(),
+			}})
+		},
+	}
+
+	// Watch for spec and annotation changes
+	nodePredicates := builder.WithPredicates(DrainAnnotationPredicate{})
+	nodeStatePredicates := builder.WithPredicates(DrainAnnotationPredicate{})
```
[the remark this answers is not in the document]
SchSeba: omg you are right nice catch!

controllers/drain_controller.go
```diff
@@ -0,0 +1,463 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (dr *DrainReconcile) SetupWithManager(mgr ctrl.Manager) error {
+	createUpdateEnqueue := handler.Funcs{
```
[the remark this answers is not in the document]
SchSeba: it's the same I am no problem to update it

controllers/drain_controller.go
```diff
@@ -0,0 +1,465 @@
...
+		err = utils.AnnotateObject(nodeNetworkState, constants.NodeDrainAnnotation, constants.DrainIdle, dr.Client)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		nodeDrainAnnotation = constants.DrainIdle
+	}
+
+	//TODO: change this to save it on runtime
+	mcpPauseDrainNodeList := &corev1.NodeList{}
+	err = dr.List(ctx, mcpPauseDrainNodeList, &client.ListOptions{LabelSelector: dr.mcpPauseDrainSelector})
```
zeeke, line 152: `LabelSelector` doesn't work with Annotations AFAIK. I think we have to list all nodes and loop over them

controllers/drain_controller.go
```diff
@@ -0,0 +1,465 @@
...
+	drainingSelector      labels.Selector
+}
+
+func NewDrainReconcileController(client client.Client, Scheme *runtime.Scheme, kubeClient kubernetes.Interface, openshiftContext *utils.OpenshiftContext) (*DrainReconcile, error) {
+	mcpPauseDrainSelector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: map[string]string{constants.NodeDrainAnnotation: constants.DrainMcpPaused}})
+	if err != nil {
+		return nil, err
+	}
+
+	drainingSelector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: map[string]string{constants.NodeDrainAnnotation: constants.Draining}})
```
zeeke, line 82: we should also consider nodes in `DrainComplete` state for the draining node calculation, as at that stage the node is still cordoned and and will go to DrainIdle after the configdaemon will set nodeState to `DrainIdle`.

```mermaid
flowchart TD
    idle_idle("`node:**Idle** nodeState:**Idle**`") -- 
        daemon wants to drain daemon.go L512 --> 
        drainRequired_idle("`node:**DrainRequired** nodeState:**Idle**`")

    drainRequired_idle -- "<< OpenShift >>" --> pauseMCP{"DrainController.pauseMCP()"}
        pauseMCP -- OK --> drainRequired_MCPPaused("`node:**DrainRequired** nodeState:**Draining_MCP_Paused**`")
        pauseMCP -- ERROR --> TBD1
    
    drainRequired_idle -- "<< Vanilla >>" --> drainNode

    drainRequired_MCPPaused --> drainNode{"DrainController.drainNode()"}
        drainNode -- OK --> drainRequired_drainComplete("`node:**DrainRequired** nodeState:**DrainComplete**`")
        drainNode -- ERROR --> TBD2

    drainRequired_drainComplete --
        daemon.go L561 -->
        idle_drainComplete("`node:**Idle** nodeState:**DrainComplete**`")

    idle_drainComplete --> completeDrain{"DrainController.completeDrain()"}
        completeDrain -- OK --> idle_idle
        completeDrain -- ERROR --> TBD3
```

Found this while coding some unit tests:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/commit/7f0b6e3366a230eb755834f47cf532283d736e50

pkg/utils/cluster.go
```diff
@@ -108,3 +112,82 @@ func openshiftControlPlaneTopologyStatus(c client.Client) (configv1.TopologyMode
...
+
+func ObjectHasAnnotation(obj metav1.Object, annoKey string, value string) bool {
+	// Check if node already contains annotation
+	if anno, ok := obj.GetAnnotations()[annoKey]; ok && (anno == value) {
+		return true
+	}
+	return false
+}
+
+func AnnotateNode(node, key, value string, kubeClient kubernetes.Interface) error {
```
adrianchiris, line 132: you could just create a raw patch and avoid copying and manipulating obj:

```
patchString := []byte(fmt.Sprintf(`{"metadata":{"annotations":{%q: %q}}}`, key, value))
// value "null" will remove the annotation key
if value == "null" {
    patchString = []byte(fmt.Sprintf(`{"metadata":{"annotations":{%q: null}}}`, key))
}
patch := client.RawPatch(types.MergePatchType, patchString)
err := p.K8sClient.Patch(ctx, node, patch)
if err != nil {
    return err
}
```

adrianchiris: same logic can be used for both this and below.

main.go
```diff
@@ -137,6 +135,8 @@ func main() {
 		os.Exit(1)
 	}
 
+	kubeclient := kubernetes.NewForConfigOrDie(ctrl.GetConfigOrDie())
```
adrianchiris, line 138: we already have a kubeclient. see L85

adrianchiris, line 138: nit, we already have kubeClient above, lets call this one kClient ? also its only used in L178 so might as well move this to L177 ? or just call `kubernetes.NewForConfigOrDie(ctrl.GetConfigOrDie())` directly in NewDrainReconcileController method

pkg/consts/constants.go
```diff
@@ -23,6 +23,17 @@ const (
 	DPConfigFileName                   = "config.json"
 	OVSHWOLMachineConfigNameSuffix     = "ovs-hw-offload"
 
+	NodeDrainAnnotation             = "sriovnetwork.openshift.io/state"
```
adrianchiris, line 26: sriovnetwork.openshift.io/daemon-state ?

pkg/consts/constants.go
```diff
@@ -23,6 +23,17 @@ const (
...
+	NodeStateDrainAnnotationCurrent = "sriovnetwork.openshift.io/current-state"
```
adrianchiris, line 27: sriovnetwork.openshift.io/controller-state ?

controllers/drain_controller.go
```diff
@@ -0,0 +1,465 @@
...
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		err = utils.AnnotateObject(nodeNetworkState, constants.NodeStateDrainAnnotationCurrent, constants.DrainComplete, dr.Client)
+		return ctrl.Result{}, err
+	}
+
+	if nodeDrainAnnotation == constants.DrainIdle &&
+		nodeStateDrainAnnotationCurrent == constants.DrainMcpPaused {
```
adrianchiris, line 233: what if nodeStateDrainAnnotationCurrent == constants.Draining ?
i think we should also completeDrain

controllers/drain_controller.go
adrianchiris: +1

controllers/drain_controller.go
```diff
@@ -0,0 +1,465 @@
...
+	mcpPauseDrainNodeList := &corev1.NodeList{}
+	err = dr.List(ctx, mcpPauseDrainNodeList, &client.ListOptions{LabelSelector: dr.mcpPauseDrainSelector})
+	if err != nil {
+		// Failed to get node list
+		reqLogger.Error(err, "Error occurred on LIST nodes request from API server")
+		return reconcile.Result{}, err
+	}
+
+	DrainingNodeList := &corev1.NodeList{}
+	err = dr.List(ctx, DrainingNodeList, &client.ListOptions{LabelSelector: dr.mcpPauseDrainSelector})
```
adrianchiris, line 160: who is labeling the nodes ?

also shouldnt this be  a different selector (drainingSelector) ?

i think for both you need to list nodes and filter.
e.g return a map of nodes keyed by their "current-state"

controllers/drain_controller.go
```diff
@@ -0,0 +1,465 @@
...
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		err = utils.AnnotateObject(nodeNetworkState, constants.NodeStateDrainAnnotationCurrent, constants.DrainIdle, dr.Client)
+		return ctrl.Result{}, err
+	}
+
+	// the node request to drain, but we are at the limit so we re-enqueue the request
+	if drainingNodes >= maxParallelNodeConfiguration {
```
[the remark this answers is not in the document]
adrianchiris: i would inverse the condition and within the if statement schedule new node for drain

SchSeba:
Hi @adrianchiris @zeeke @e0ne  sorry for the noise but can I please ask to first merge https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/553 and then review https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/555

The rebases between them are a nightmare so I will prefer to go it that way. Also using https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/555 makes the code much more cleaner and readable.

Thanks folks!

PR #553, opened 2023-11-25
Interface redesign

SchSeba:
This PR is mainly moving stuff around and creating interfaces so we can have much better coverage in our unit tests

This is the second part needed to have a parallel configuration after https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/552


e0ne (changes requested):

pkg/global/vars/vars.go
```diff
@@ -0,0 +1,72 @@
...
+	"regexp"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/global/consts"
+)
+
+var (
+	ClusterType string
```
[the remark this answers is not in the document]
SchSeba: These are variables that don't change in all the whole container life cycle having a getter that will do for example `os.getEnv` again and again is a waste of syscall and resources if you ask me

pkg/global/consts/constants.go
```diff
@@ -0,0 +1,114 @@
+package consts
```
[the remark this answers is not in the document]
SchSeba: just to have a separation between the constants and the variables

pkg/global/vars/vars.go
```diff
@@ -0,0 +1,72 @@
...
+	SupportedVfIds []string
+
+	DpdkDrivers = []string{"igb_uio", "vfio-pci", "uio_pci_generic"}
+
+	//Cluster variables
+	Config *rest.Config
+	Scheme *runtime.Scheme
+)
+
+func init() {
```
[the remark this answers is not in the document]
zeeke: I agree with Sebastian that global vars will make creating test easier. @e0ne can you point out a case of a test that would become harder?

On Ivan's side, I guess we can simplify `init()` by moving any assignment default to the variable declarations and omitting default values. This way the function becomes:

```go
func init() {

	mode := os.Getenv("DEV_MODE")
	if mode == "TRUE" {
		DevMode = true
	}

	Destdir = "/host/tmp"
	destdir := os.Getenv("DEST_DIR")
	if destdir != "" {
		Destdir = destdir
	}
}
``` 

What do you (both) think?

pkg/utils/cluster.go
```diff
@@ -53,15 +55,15 @@ func IsSingleNodeCluster(c client.Client) (bool, error) {
 // On kubernetes, it is determined by which node the sriov operator is scheduled on. If operator
 // pod is schedule on worker node, it is considered as external control plane.
 func IsExternalControlPlaneCluster(c client.Client) (bool, error) {
-	if os.Getenv("CLUSTER_TYPE") == ClusterTypeOpenshift {
+	if os.Getenv("CLUSTER_TYPE") == consts.ClusterTypeOpenshift {
```
zeeke, line 58: As we created `vars.ClusterType`, we can use that instead of `os.Getenv("CLUSTER_TYPE")`

Same on L40 and L66

cmd/sriov-network-config-daemon/start.go
```diff
@@ -70,8 +73,14 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
...
 	setupLog := log.Log.WithName("sriov-network-config-daemon")
 
+	// Mark that we are running inside a container
+	vars.UsingSystemdMode = false
+	if startOpts.systemd {
+		vars.UsingSystemdMode = true
+	}
 	// To help debugging, immediately log version
-	setupLog.V(2).Info("sriov-network-config-daemon", "version", version.Version)
+	// TODO: we don't need this anymore as we don't update it
```
zeeke, line 82: Can we remove this TODO?

zeeke:
Left some comments.
Though there are a lot of changed files for a single PR, I like the approach of moving toward a more testable operator.
Looking forward to propose PRs based on this with even more cleanups and improvements

SchSeba:
@zeeke yep agree with you there is more stuff to do here. 
my next work on cleanups and improvements will be to move also the systemd stuff we have to interfaces so we can have unit-tests in that area

pkg/utils/cluster.go
SchSeba: I have it on the next PR https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/555/commits/f464e8acbe1151cb684ac106b96f745ae85f1c18

this one is already pretty big :P

zeeke: fine with that!

pkg/global/consts/constants.go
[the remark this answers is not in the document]
adrianchiris: can we drop global folder ?

both consts and vars pkgs can be under pkg/ directly which IMO means they are global.

pkg/global/vars/vars.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/global/consts"
+)
+
+var (
+	ClusterType string
+
+	// PlatformsMap contains supported platforms for virtual VF
+	PlatformsMap map[string]consts.PlatformTypes
+
+	DevMode bool
```
adrianchiris, line 19: can you add short description for each var ?

pkg/global/vars/vars.go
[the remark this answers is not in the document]
adrianchiris: we should initialize whats needed and omit the rest imo.
also no need to explicitly set a var if its default value as initialized by go matches (e.g bools are always false...)

pkg/global/consts/constants.go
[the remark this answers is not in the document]
SchSeba: sure no problem

cmd/sriov-network-config-daemon/start.go
SchSeba: done

api/v1/helper.go
```diff
@@ -211,6 +215,80 @@ func GetVfDeviceID(deviceID string) string {
...
+func FindInterface(interfaces Interfaces, name string) (iface Interface, err error) {
+	for _, i := range interfaces {
+		if i.Name == name {
+			return i, nil
+		}
+	}
+	return Interface{}, fmt.Errorf("unable to find interface: %v", name)
+}
+
+func NeedToUpdateSriov(iface *Interface, ifaceStatus *InterfaceExt) bool {
```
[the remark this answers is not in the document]
SchSeba: done

SchSeba:
Hi @e0ne @zeeke @adrianchiris please take another look if you can I think I fixed everything

adrianchiris:
@SchSeba im reviewing commits 1 by 1 so expect several comment batches from me  :)

pkg/host/sriov.go
```diff
@@ -0,0 +1,443 @@
...
+		if is != nil {
+			mtu = is.Mtu
+		} else {
+			mtu = 1500
+		}
+		log.Log.V(2).Info("ResetSriovDevice(): reset mtu", "value", mtu)
+		if err := h.SetNetdevMTU(ifaceStatus.PciAddress, mtu); err != nil {
+			return err
+		}
+	} else if ifaceStatus.LinkType == consts.LinkTypeIB {
```
adrianchiris, line 60: for IB we dont rely on initial state ?

pkg/host/store.go
```diff
@@ -24,27 +21,29 @@ type StoreManagerInterface interface {
 	ClearPCIAddressFolder() error
 	SaveLastPfAppliedStatus(PfInfo *sriovnetworkv1.Interface) error
 	LoadPfsStatus(pciAddress string) (*sriovnetworkv1.Interface, bool, error)
+
+	GetCheckPointNodeState() (*sriovnetworkv1.SriovNetworkNodeState, error)
+	WriteCheckpointFile(*sriovnetworkv1.SriovNetworkNodeState) error
 }
 
 type StoreManager struct {
```
adrianchiris, line 29: make this one internal ?

pkg/host/host.go
```diff
@@ -47,6 +53,55 @@ type HostManagerInterface interface {
 	TryEnableTun()
 	TryEnableVhostNet()
 	TryEnableRdma() (bool, error)
+	TryToGetVirtualInterfaceName(string) string
```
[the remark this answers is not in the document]
adrianchiris: ok going further through the commits, maybe its worth splitting by type of interface.

as an example different plugins dont need to use everything from this interface.

pkg/plugins/plugin.go
```diff
@@ -14,3 +20,42 @@ type VendorPlugin interface {
 	// Apply config change
 	Apply() error
 }
+
+type HostHelpersInterface interface {
```
adrianchiris, line 24: i think each plugin should require the interfaces that are relevant.

api/v1/helper.go
```diff
@@ -47,6 +49,8 @@ var log = logf.Log.WithName("sriovnetwork")
 // Vendor ID, Physical Function Device ID, Virtual Function Device ID
 var NicIDMap = []string{}
 
+var InitialState SriovNetworkNodeState
```
adrianchiris, line 52: shouldnt this be part of "vars" package ?

also NicIDMap ?

test/conformance/tests/test_sriov_operator.go
```diff
@@ -1188,7 +1188,7 @@ var _ = Describe("[sriov] operator", func() {
 						capacity, _ = resNum.AsInt64()
 						res["openshift.io/testresource1"] = capacity
 						return res
-					}, 2*time.Minute, time.Second).Should(Equal(map[string]int64{
+					}, 15*time.Minute, time.Second).Should(Equal(map[string]int64{
```
adrianchiris, line 1191: why this change needed ? arent we just moving things around and defining interfaces?

pkg/daemon/event_recorder.go
```diff
@@ -12,34 +12,33 @@ import (
...
-		node:             n,
 		eventRecorder:    eventRecorder,
 		eventBroadcaster: eventBroadcaster,
 	}
 }
 
 // SendEvent Send an Event on the NodeState object
 func (e *EventRecorder) SendEvent(eventType string, msg string) {
-	nodeState, err := e.client.SriovnetworkV1().SriovNetworkNodeStates(namespace).Get(context.Background(), e.node, metav1.GetOptions{})
+	nodeState, err := e.client.SriovnetworkV1().SriovNetworkNodeStates(namespace).Get(context.Background(), vars.NodeName, metav1.GetOptions{})
```
adrianchiris, line 39: do we really want this ? or should each component that cares about the node name have it as a struct member ?
that is its New*() method to still accept nodeName  as parameter. the initializer would then pass in vars.NodeName

pkg/daemon/daemon_test.go
```diff
@@ -1,313 +1,316 @@
...
-	return err
-}
-
-func updateSriovNetworkNodeState(c snclientset.Interface, nodeState *sriovnetworkv1.SriovNetworkNodeState) error {
-	_, err := c.SriovnetworkV1().
-		SriovNetworkNodeStates(namespace).
-		Update(context.Background(), nodeState, metav1.UpdateOptions{})
-	return err
-}
+//
```
adrianchiris: why do we have this commented out ? (if its handled in subsequent commit, disregard the comment)

pkg/daemon/daemon.go
```diff
@@ -195,38 +172,27 @@ func New(
...
-		log.Log.V(0).Info("Run(): start daemon.", "openshiftFlavor", dn.openshiftContext.OpenshiftFlavor)
+	if vars.ClusterType == consts.ClusterTypeOpenshift {
+		log.Log.V(0).Info("Run(): start daemon.", "openshiftFlavor", dn.platformHelpers.GetFlavor())
 	} else {
 		log.Log.V(0).Info("Run(): start daemon.")
 	}
 
-	if dn.useSystemdService {
+	if !vars.UsingSystemdMode {
 		log.Log.V(0).Info("Run(): daemon running in systemd mode")
```
adrianchiris, line 184: log is now wrong. maybe change to : "daemon running in daemon mode" and add an else statement with a log that says "daemon running in systemd mode"

pkg/daemon/daemon.go
```diff
@@ -571,8 +537,9 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 	// When running using systemd check if the applied configuration is the latest one
 	// or there is a new config we need to apply
 	// When using systemd configuration we write the file
-	if dn.useSystemdService {
-		systemdConfModified, err := systemd.WriteConfFile(latestState, dn.devMode, dn.platform)
+	if vars.UsingSystemdMode {
```
adrianchiris, line 540: shouldnt this be part of the daemon ?

tbh its not clear to me what we want to put in vars

pkg/daemon/daemon.go
```diff
@@ -628,7 +595,7 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 			}
 		}
 
-		if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
+		if dn.platformHelpers.IsOpenshiftCluster() && !dn.platformHelpers.IsHypershift() {
```
adrianchiris, line 598: maybe platformHelpers interface should have a : GetClusterType() instead of using vars.ClusterType

pkg/daemon/daemon.go
```diff
@@ -739,16 +706,16 @@ func (dn *Daemon) completeDrain() error {
 		}
 	}
 
-	if dn.openshiftContext.IsOpenshiftCluster() && !dn.openshiftContext.IsHypershift() {
+	if dn.platformHelpers.IsOpenshiftCluster() && !dn.platformHelpers.IsHypershift() {
 		log.Log.Info("completeDrain(): resume MCP", "mcp-name", dn.mcpName)
 		pausePatch := []byte("{\"spec\":{\"paused\":false}}")
-		if _, err := dn.openshiftContext.McClient.MachineconfigurationV1().MachineConfigPools().Patch(context.Background(), dn.mcpName, types.MergePatchType, pausePatch, metav1.PatchOptions{}); err != nil {
+		if _, err := dn.platformHelpers.GetMcClient().MachineconfigurationV1().MachineConfigPools().Patch(context.Background(), dn.mcpName, types.MergePatchType, pausePatch, metav1.PatchOptions{}); err != nil {
```
adrianchiris, line 712: it would be nice to handle all MCP stuff in platformHelpers (openshift interface)
that said, it can be handled separately.

cmd/sriov-network-config-daemon/start.go
```diff
@@ -102,24 +112,30 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
 	var config *rest.Config
 	var err error
 
-	if os.Getenv("CLUSTER_TYPE") == utils.ClusterTypeOpenshift {
+	if vars.ClusterType == consts.ClusterTypeOpenshift {
 		kubeconfig, err := clientcmd.LoadFromFile("/host/etc/kubernetes/kubeconfig")
```
adrianchiris, line 113: not related to this PR, but why do we need to do this in openshift ?
in-cluster-config using service account is not enough ?

pkg/daemon/daemon_test.go
```diff
@@ -1,316 +1,273 @@
...
-//	return err
-//}
-//
-//func updateSriovNetworkNodeState(c snclientset.Interface, nodeState *sriovnetworkv1.SriovNetworkNodeState) error {
-//	_, err := c.SriovnetworkV1().
-//		SriovNetworkNodeStates(namespace).
-//		Update(context.Background(), nodeState, metav1.UpdateOptions{})
-//	return err
-//}
+import (
```
adrianchiris: yep, the commented UT is handled here :)

pkg/global/vars/vars.go
```diff
@@ -0,0 +1,72 @@
...
+	"os"
+	"regexp"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/global/consts"
+)
+
+var (
```
adrianchiris: after reviewing all commits (its a huge PR so i trust we just moved things around and defined interfaces and not applied any logic changes) i wonder if using vars is the correct way to go.

personally i think it makes individual components hard to understand as their behaviour depends on some external package.

i believe the correct way to go is to try to eliminate the use of global variables and either have the appropriate interface expose the information (set during interface instance creation) and also pass in relevant attributes (e.g nodeName) to relevant components.

that said, one could argue that it was like that before and this commit just moved it to a single place.
so im not sure what to think of it.

pkg/host/host.go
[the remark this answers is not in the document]
SchSeba: sure I can do it.
do you have a nice idea what to split?
in general I put everything that touches the host under the same interface but I can do another split if needed

pkg/global/vars/vars.go
SchSeba: done

pkg/platforms/openshift/openshift.go
```diff
@@ -17,6 +19,14 @@ const (
 	OpenshiftFlavorDefault OpenshiftFlavor = "default"
 )
 
+//go:generate ../../../bin/mockgen -destination mock/mock_openshift.go -source openshift.go
+type OpenshiftContextInterface interface {
```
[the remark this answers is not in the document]
SchSeba: I can't because we use them in the platform one as interface for both openshift and openstack

pkg/platforms/openshift/openshift.go
```diff
@@ -17,6 +19,14 @@ const (
...
+	GetFlavor() OpenshiftFlavor
+	GetMcClient() mcclientset.Interface
+	IsOpenshiftCluster() bool
+	IsHypershift() bool
+}
+
 // OpenshiftContext contains metadata and structs utilized to interact with Openshift clusters
 type OpenshiftContext struct {
```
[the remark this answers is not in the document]
SchSeba: sure did for both

pkg/platforms/openstack/openstack.go
```diff
@@ -64,6 +35,18 @@ var (
 	ospMetaDataFile    = ospMetaDataDir + "/meta_data.json"
 )
 
+//go:generate ../../../bin/mockgen -destination mock/mock_openstack.go -source openstack.go
+type OpenstackInterface interface {
```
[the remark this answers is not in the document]
SchSeba: I can't because we use them in the platform one as interface for both openshift and openstack

pkg/platforms/openstack/openstack.go
```diff
@@ -64,6 +35,18 @@ var (
...
+	CreateOpenstackDevicesInfo() error
+	CreateOpenstackDevicesInfoFromNodeStatus(*sriovnetworkv1.SriovNetworkNodeState)
+	DiscoverSriovDevicesVirtual() ([]sriovnetworkv1.InterfaceExt, error)
+}
+
+type OpenstackContext struct {
```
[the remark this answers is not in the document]
SchSeba: done

pkg/platforms/platforms.go
```diff
@@ -0,0 +1,31 @@
...
+	openshift.OpenshiftContextInterface
+	openstack.OpenstackInterface
+}
+
+type PlatformHelper struct {
+	openshift.OpenshiftContextInterface
+	openstack.OpenstackInterface
+}
+
+func NewDefaultPlatformHelper() (*PlatformHelper, error) {
```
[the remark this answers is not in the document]
SchSeba: yep I miss that one

pkg/platforms/platforms.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms/openstack"
+)
+
+//go:generate ../../bin/mockgen -destination mock/mock_platforms.go -source platforms.go
+type PlatformHelperInterface interface {
+	openshift.OpenshiftContextInterface
+	openstack.OpenstackInterface
+}
+
+type PlatformHelper struct {
```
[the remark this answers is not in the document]
SchSeba: done

pkg/platforms/platforms.go
```diff
@@ -0,0 +1,31 @@
+package platforms
+
+import (
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms/openshift"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms/openstack"
+)
+
+//go:generate ../../bin/mockgen -destination mock/mock_platforms.go -source platforms.go
+type PlatformHelperInterface interface {
```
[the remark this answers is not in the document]
SchSeba: this one I can because it's a top level interface

pkg/vendors/mellanox/mellanox.go
```diff
@@ -0,0 +1,419 @@
...
+	PreconfiguredLinkType = "Preconfigured"
+	UknownLinkType        = "Uknown"
+	TotalVfs              = "NUM_OF_VFS"
+	EnableSriov           = "SRIOV_EN"
+	LinkTypeP1            = "LINK_TYPE_P1"
+	LinkTypeP2            = "LINK_TYPE_P2"
+	MellanoxVendorID      = "15b3"
+)
+
+type MLXNic struct {
```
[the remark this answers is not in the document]
SchSeba: it was like that before if I remember but sure I can change it :P

pkg/vendors/mellanox/mellanox.go
```diff
@@ -0,0 +1,419 @@
...
+	EnableSriov bool
+	TotalVfs    int
+	LinkTypeP1  string
+	LinkTypeP2  string
+}
+
+//go:generate ../../../bin/mockgen -destination mock/mock_mellanox.go -source mellanox.go
+type MellanoxInterface interface {
```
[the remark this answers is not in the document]
SchSeba: I can't because they are all part of the same interface 

```
type HostHelpersInterface interface {
	utils.UtilsInterface
	host.HostManagerInterface
	host.StoreManagerInterface
	mlx.MellanoxInterface
}
```

pkg/vendors/mellanox/mellanox.go
```diff
@@ -0,0 +1,419 @@
...
+	MstConfigReadData(string) (string, string, error)
+	GetMellanoxBlueFieldMode(string) (BlueFieldMode, error)
+	GetMlnxNicFwData(pciAddress string) (current, next *MLXNic, err error)
```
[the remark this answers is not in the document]
SchSeba: sure it was like that already it was under my rule that I only change places for stuff but sure no problem to rename it

pkg/vendors/mellanox/mellanox.go
```diff
@@ -0,0 +1,419 @@
...
+
+	MLXConfigFW(attributesToChange map[string]MLXNic) error
```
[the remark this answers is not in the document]
SchSeba: sure it was like that already it was under my rule that I only change places for stuff but sure no problem to rename it

pkg/vendors/mellanox/mellanox.go
```diff
@@ -0,0 +1,419 @@
...
+}
+
+type MellanoxHelper struct {
```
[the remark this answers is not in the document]
SchSeba: sure

pkg/host/service.go
```diff
@@ -98,7 +205,7 @@ OUTER:
 }
 
 // ReadServiceInjectionManifestFile reads service injection file
-func ReadServiceInjectionManifestFile(path string) (*Service, error) {
+func (h *HostManager) ReadServiceInjectionManifestFile(path string) (*Service, error) {
```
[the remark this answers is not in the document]
SchSeba: you are right I move it out to the k8s plugin that is the only one that uses it

pkg/host/sriov.go
SchSeba: not today https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils.go#L697

pkg/host/store.go
SchSeba: done

pkg/plugins/plugin.go
SchSeba: I was thinking to do it but here we just past a pointer to the object with all the interfaces I think it makes the code much more cleaner and if the plugin doesn't need all the function that is fine it's there

test/conformance/tests/test_sriov_operator.go
SchSeba: yep that was just because there was some flakes in the virtual CI not related to this PR

api/v1/helper.go
SchSeba: can't do it because they you will have a circular import

pkg/daemon/event_recorder.go
SchSeba: I don't need we need to duplicate data there is no point of that if you ask me.
everything that is in the vars folder is a long-live configuration that is specific for the all run of the daemon or the operator.

it's not possible to have one node name in one package/struct and a different one in another that is why I wanted to have the vars folder to be global for all application

pkg/daemon/daemon_test.go
SchSeba: yep sorry I fix it :P

pkg/daemon/daemon.go
SchSeba: right!

pkg/daemon/daemon.go
SchSeba: I put in vars everything that is used by more than one package so I don't need to start passing that around all the functions

for example this one to not pass it to the daemon and the hostInterfaces and the plugins

pkg/daemon/daemon.go
SchSeba: yep I did that already in the parallel PR :)

cmd/sriov-network-config-daemon/start.go
SchSeba: that was a request from the security team in RedHat as a way to reduce access from the daemon to nodes that he doesn't runs on

pkg/daemon/daemon_test.go
SchSeba: yep :)

pkg/global/vars/vars.go
SchSeba: as I commented above about this one, the global vars is for configuration that is used in different part of the code but they are global meaning they don't change in the run for example cluster type or development flag.

I really try to reduce passing the same info that will never change down the interfaces.
for example the systemdMode it was like this before the change
main -> daemon -> plugins -> utils -> chroot

and I think it's really not useful to pass that boolean if it's really global if we are in systemd it will be the case for the all daemon run

pkg/utils/utils.go
```diff
@@ -2,741 +2,37 @@ package utils
...
-	speedFilePath := filepath.Join(sysClassNet, ifaceName, "speed")
-	data, err := os.ReadFile(speedFilePath)
-	if err != nil {
-		log.Log.Error(err, "getNetDevLinkSpeed(): fail to read Link Speed file", "path", speedFilePath)
-		return ""
-	}
-
-	return fmt.Sprintf("%s Mb/s", strings.TrimSpace(string(data)))
+//go:generate ../../bin/mockgen -destination mock/mock_utils.go -source utils.go
+type UtilsInterface interface {
```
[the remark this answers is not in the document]
SchSeba: done

pkg/utils/utils.go
```diff
@@ -2,741 +2,37 @@ package utils
...
-		if err := setNetdevMTU(ifaceStatus.PciAddress, mtu); err != nil {
-			return err
-		}
-	} else if ifaceStatus.LinkType == constants.LinkTypeIB {
-		if err := setNetdevMTU(ifaceStatus.PciAddress, 2048); err != nil {
-			return err
-		}
-	}
-	return nil
+type UtilsHelper struct {
```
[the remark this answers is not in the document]
SchSeba: done

pkg/vendors/mellanox/mellanox.go
```diff
@@ -0,0 +1,419 @@
+package mellanox
```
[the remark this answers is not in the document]
SchSeba: done

SchSeba:
Hi @adrianchiris thanks for the great review I think I answer everything. please make another round of review to only the new commits when you have time.

I will rebase the commits before we merge this to have a good history but for now I will add new ones so it's easier for everyone to review this PR as it's pretty big

pkg/host/host.go
```diff
@@ -65,20 +120,18 @@ type HostManagerInterface interface {
 }
 
 type HostManager struct {
```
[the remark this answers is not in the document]
SchSeba: done

cmd/sriov-network-config-daemon/start.go
adrianchiris: ack, can you add a comment ?

pkg/host/udev.go
```diff
@@ -0,0 +1,171 @@
...
+	_, err = os.Stat(consts.SriovHostSwitchDevConfPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			if len(cfg.Interfaces) == 0 {
+				err = nil
+				return
+			}
+
+			// Create the sriov-operator folder on the host if it doesn't exist
+			if _, err := os.Stat("/host" + consts.SriovConfBasePath); os.IsNotExist(err) {
```
Eoghan1232, line 95: this is from container right? 
so /host is the host itself.
so then it falls under /etc/sriov..... ?

Can we make "/host" a const somewhere? this will be easier later in cleaning up host imports.

pkg/host/udev.go
```diff
@@ -0,0 +1,171 @@
...
+				err = os.Mkdir("/host"+consts.SriovConfBasePath, os.ModeDir)
```
Eoghan1232, line 96: same as above.

pkg/systemd/systemd.go
```diff
@@ -25,13 +25,14 @@ import (
...
 const (
-	SriovSystemdConfigPath        = utils.SriovConfBasePath + "/sriov-interface-config.yaml"
-	SriovSystemdResultPath        = utils.SriovConfBasePath + "/sriov-interface-result.yaml"
-	sriovSystemdSupportedNicPath  = utils.SriovConfBasePath + "/sriov-supported-nics-ids.yaml"
+	SriovSystemdConfigPath        = consts.SriovConfBasePath + "/sriov-interface-config.yaml"
+	SriovSystemdResultPath        = consts.SriovConfBasePath + "/sriov-interface-result.yaml"
+	sriovSystemdSupportedNicPath  = consts.SriovConfBasePath + "/sriov-supported-nics-ids.yaml"
 	sriovSystemdServiceBinaryPath = "/var/lib/sriov/sriov-network-config-daemon"
 
 	SriovHostSystemdConfigPath        = "/host" + SriovSystemdConfigPath
```
Eoghan1232, line 38: anywhere "/host" is used, let's make a var for it and use this instead. much easier for bigger changes in future

pkg/plugins/virtual/virtual_plugin.go
```diff
@@ -122,7 +119,62 @@ func (p *VirtualPlugin) IsSystemService() bool {
...
+			if iface.VfGroups[i].DeviceType == consts.DeviceTypeVfioPci {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// SyncNodeStateVirtual attempt to update the node state to match the desired state in virtual platforms
+func SyncNodeStateVirtual(newState *sriovnetworkv1.SriovNetworkNodeState, helpers plugin.HostHelpersInterface) error {
```
Eoghan1232, line 131: nit, does it need to be exported now? since it's no longer called from utils pkg?

pkg/plugins/k8s/k8s_plugin.go
```diff
@@ -495,18 +494,39 @@ func (p *K8sPlugin) updateSystemService(serviceObj *service.Service) error {
...
-	p.useSystemdService = true
-}
+// appendToService appends given fields to service
+func appendToService(service *host.Service, options ...*unit.UnitOption) (*host.Service, error) {
+	serviceOptions, err := unit.Deserialize(strings.NewReader(service.Content))
+	if err != nil {
+		return nil, err
+	}
+
+OUTER:
```
Eoghan1232, line 512: what is this?

Eoghan1232: ignore this, I understand.

pkg/platforms/openstack/openstack.go
```diff
@@ -117,8 +100,12 @@ type OSPDeviceInfo struct {
 	NetworkID  string
 }
 
+func NewOpenstackContext() OpenstackInterface {
```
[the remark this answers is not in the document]
SchSeba: we can't because we have multiple interfaces

pkg/platforms/openshift/openshift.go
```diff
@@ -29,25 +39,25 @@ type OpenshiftContext struct {
 	OpenshiftFlavor OpenshiftFlavor
 }
 
-func NewOpenshiftContext(config *rest.Config, scheme *runtime.Scheme) (*OpenshiftContext, error) {
-	if ClusterType != ClusterTypeOpenshift {
+func NewOpenshiftContext() (OpenshiftContextInterface, error) {
```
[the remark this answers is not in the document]
SchSeba: we can't because we have multiple interfaces

pkg/host/udev.go
SchSeba: I didn't change this function at all I put a TODO on a following PR this one is already huge :)

pkg/host/host.go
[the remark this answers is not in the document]
adrianchiris: here is a proposal:

split hostManager interfaces to interfaces by the following categories:

- Driver/Kernel/dpdk : handles driver modules 
- host network : netdev related functionality
- sriov : VF related functionality (can also be part of host network interface)
-  service : handles system service functionality (incl writing switchdev config file
-  udev : handles udev related functionality

WDYT ?

pkg/platforms/openshift/openshift.go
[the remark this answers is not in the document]
adrianchiris: ack.

pkg/platforms/openstack/openstack.go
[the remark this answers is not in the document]
adrianchiris: ack

pkg/platforms/openstack/openstack.go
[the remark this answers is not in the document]
adrianchiris: you can rename it to "New()" there is no name collision (openstack and openshift are different packages).

im fine either way or maybe just rename this one `NewOpenstack()`  since the "context" part doesnt bring much value to the method name.

up2u

pkg/vendors/mellanox/mellanox.go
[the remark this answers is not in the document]
adrianchiris: ack, interface embedding in struct will yield a "redeclaration" error.

pkg/vendors/mellanox/mellanox.go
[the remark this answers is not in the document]
adrianchiris: thx :)

pkg/vendors/mellanox/mellanox.go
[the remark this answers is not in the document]
adrianchiris: thx :)

pkg/host/service.go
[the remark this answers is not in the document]
adrianchiris: how about having an `Append(options ...unit.UintOptions) error` and `Update() error` part of `host.Service` struct ? 
then use it in k8s plugin.

ideally k8s plugin should interact with host service interface and not with unit package directly (serializing deserializing etc...)

pkg/daemon/event_recorder.go
adrianchiris: ack, lets go with this approach.

pkg/global/vars/vars.go
adrianchiris: not really a fan of the approach, but i get it.
in any case this PR improves the code base.

pkg/plugins/plugin.go
adrianchiris: i think each plugin would eventually need 2 or 3 (some zero) interfaces.
and it would be clearer from code POV what each plugin ingests as interfaces
(e.g mlx plugin just needs mellanox interface and kernel/driver interface.

pkg/daemon/daemon.go
adrianchiris: going back to our discussion about globals. lets keep this one as is.

adrianchiris:
provided another round of reviews, 

once remaining comments are addressed im LGTM.

bn222 (commented):
Only 4 comments. If those are addressed, I'm ok with the PR.

pkg/platforms/openstack/openstack.go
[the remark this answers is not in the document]
SchSeba: oh you are right sorry

pkg/host/service.go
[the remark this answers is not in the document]
SchSeba: I was thinking about that I have one problem is that I don't have an env to test the changes so I was really careful on that package.
if I make the changes will you be able to test the hw-offload on vanilla k8s for me?

cmd/sriov-network-config-daemon/start.go
SchSeba: done

pkg/global/consts/constants.go
```diff
@@ -0,0 +1,114 @@
...
+func (e PlatformTypes) String() string {
+	switch e {
+	case Baremetal:
+		return "Baremetal"
+	case VirtualOpenStack:
+		return "Virtual/Openstack"
+	default:
+		return fmt.Sprintf("%d", int(e))
+	}
+}
```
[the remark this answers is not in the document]
SchSeba: I am going to rebase all the commits just before we merge so we should be ok

pkg/global/vars/vars.go
```diff
@@ -0,0 +1,72 @@
...
+	mode := os.Getenv("DEV_MODE")
+	if mode == "TRUE" {
+		DevMode = true
+	}
+
+	Destdir = "/host/tmp"
+	destdir := os.Getenv("DEST_DIR")
+	if destdir != "" {
+		Destdir = destdir
+	}
```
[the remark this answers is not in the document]
SchSeba: you can't because one if them is string and the other one is boolean

pkg/host/udev.go
```diff
@@ -0,0 +1,171 @@
...
+		for _, ifaceStatus := range newState.Status.Interfaces {
+			if iface.PciAddress != ifaceStatus.PciAddress {
+				continue
+			}
+
+			if skip := pfsToSkip[iface.PciAddress]; !skip {
+				continue
+			}
+
+			i := sriovnetworkv1.Interface{}
```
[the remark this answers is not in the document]
SchSeba: It was like that before I didn't change the content of the functions :P 

but you are right I removed it thanks! :)

pkg/host/udev.go
```diff
@@ -0,0 +1,171 @@
...
+		return err
+	}
+	log.Log.V(2).Info("PrepareNMUdevRule()", "stdout", stdout)
+
+	//save the device list to use for udev rules
+	vars.SupportedVfIds = supportedVfIds
+	return nil
+}
+
+func (h *HostManager) WriteSwitchdevConfFile(newState *sriovnetworkv1.SriovNetworkNodeState, pfsToSkip map[string]bool) (update bool, err error) {
```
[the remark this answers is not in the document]
SchSeba: done

pkg/plugins/plugin.go
SchSeba: the idea for this interface was to not check the ability or create objects for the vendor plugins

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/e95012cb1907ddf0ff1225f091e6dd1962cda676/pkg/daemon/plugin.go#L19

if every vendor plugin requires different interfaces we are not going to be able to use the map for the vendors no?

I can send the hostHelper to all the vendors and inside every vendor package will take a subset (I think it will work) is that a good solution?

pkg/host/service.go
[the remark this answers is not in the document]
adrianchiris: yes sure

pkg/plugins/plugin.go
adrianchiris: yes its ok.

pkg/plugins/plugin.go
```diff
@@ -14,3 +20,42 @@ type VendorPlugin interface {
...
+	utils.CmdInterface
+	host.HostManagerInterface
+	host.StoreManagerInterface
+	mlx.MellanoxInterface
+}
+
+type HostHelpers struct {
```
adrianchiris, line 31: can you make this one private ?

pkg/plugins/plugin.go
```diff
@@ -14,3 +20,42 @@ type VendorPlugin interface {
...
+	host.HostManagerInterface
+	host.StoreManagerInterface
+	mlx.MellanoxInterface
+}
+
+// Use for unit tests
+func NewVendorPluginHelpers(utilsHelper utils.CmdInterface,
+	hostManager host.HostManagerInterface,
+	storeManager host.StoreManagerInterface,
+	mlxHelper mlx.MellanoxInterface) *HostHelpers {
```
adrianchiris, line 42: can you return interface type here and below ?

pkg/plugins/plugin.go
adrianchiris, line 24: should you call this interface VendorPluginHelpersInterface ? (and matching struct)

or rename the New methods ?

usually we have `New{Type}()` naming convention for such functions

pkg/host/host.go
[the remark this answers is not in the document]
adrianchiris: Thx for splitting to different files, now just need to define sub interfaces for each.
im OK with having an aggregated interface as well.

that way each domain has its own interface and struct that implements it.
(and we dont keep adding functionallity to one huge interface with one huge struct)

and we aggregate it all under a an interface to be used by the daemon

thoughts ?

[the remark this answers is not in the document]
zeeke: +1 on having smaller interfaces. not blocking this

pkg/host/service.go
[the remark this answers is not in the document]
SchSeba: done

pkg/host/host.go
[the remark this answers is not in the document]
SchSeba: done please take a look

adrianchiris (approved):
LGTM !

please rebase this one as discussed. then once rebased ill pull, and deploy on switchdev setup as discussed.

SchSeba:
Thanks!
Done please have a look :)

zeeke (approved):
LGTM

api/v1/helper.go
```diff
@@ -211,6 +215,80 @@ func GetVfDeviceID(deviceID string) string {
...
+		}
+	}
+	return Interface{}, fmt.Errorf("unable to find interface: %v", name)
+}
+
+func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
+	if ifaceSpec.Mtu > 0 {
+		mtu := ifaceSpec.Mtu
+		if mtu != ifaceStatus.Mtu {
+			log.V(2).Info("NeedUpdate(): MTU needs update", "desired", mtu, "current", ifaceStatus.Mtu)
```
mlguerrero12, line 240: function is now NeedToUpdateSriov instead of NeedUpdate for the log

api/v1/helper.go
```diff
@@ -211,6 +215,80 @@ func GetVfDeviceID(deviceID string) string {
...
+func FindInterface(interfaces Interfaces, name string) (iface Interface, err error) {
+	for _, i := range interfaces {
+		if i.Name == name {
+			return i, nil
+		}
+	}
+	return Interface{}, fmt.Errorf("unable to find interface: %v", name)
+}
+
+func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
```
mlguerrero12, line 236: why can't this be part of pkg/host/sriov.go? It's used there by ConfigSriovInterfaces and both of them are only used by the generic plugin. Both can be called using the helpers. What am I missing?

adrianchiris (approved):
LGTM,

Tested this out on my setup both with switchdev and legacy sriov, seems to behave as expected.

@e0ne PTAL i see you requested changes, please approve if they were resolved.

id like to merge this one tommorow. as several PRs depend on it.

if we have non-critical comments remaining they can be addressed as follloup IMO

pkg/daemon/event_recorder.go
e0ne: I really don't like this approach with global variables but since team agrees on it, I won't block PR for this

pkg/host/store.go
```diff
@@ -137,9 +136,45 @@
...
+}
+
+func (s *StoreManager) WriteCheckpointFile(ns *sriovnetworkv1.SriovNetworkNodeState) error {
+	configdir := filepath.Join(vars.Destdir, consts.CheckpointFileName)
+	file, err := os.OpenFile(configdir, os.O_RDWR|os.O_CREATE, 0644)
+	if err != nil {
+		return err
 	}
-	return FilesystemRoot
+	defer file.Close()
```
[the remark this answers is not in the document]
e0ne: @SchSeba  do we need to address this alert?

e0ne (approved):
Thank for addressing my comments! No blockers from my side now

cmd/sriov-network-config-daemon/service.go
```diff
@@ -91,67 +97,64 @@ func runServiceCmd(cmd *cobra.Command, args []string) error {
...
 	}
 
 	setupLog.V(2).Info("sriov-config-service", "config", nodeStateSpec)
-
-	storeManager, err := utils.NewStoreManager(true)
+	hostHelpers, err := helper.NewDefaultHostHelpers()
 	if err != nil {
-		setupLog.Error(err, "failed to create store manager")
-		return err
+		setupLog.Error(err, "failed to create vendorHelpers")
```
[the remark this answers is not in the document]
SchSeba: right thanks!

api/v1/helper.go
SchSeba: right thanks!

api/v1/helper.go
SchSeba: the reason I didn't put it in the host package is because that function doesn't do anything on the host. this function just make a API verification that is with it should be in the API helpers and not in the host package

SchSeba:
Thanks everyone great work!!!

PR #555, opened 2023-12-05
Parallel drain

SchSeba:
The PR moves the draining from the daemon to the main controller it's based on the work from @e0ne and https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/552.

Also, the PR introduced a new feature to allow multiple nodes to be drained in parallel

base on https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/479


SchSeba:
please first review https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/553

controllers/drain_controller.go
```diff
@@ -0,0 +1,464 @@
...
+//+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;update;patch
+//+kubebuilder:rbac:groups=sriovnetwork.openshift.io,resources=sriovnodestates,verbs=get;list;watch
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
+func (dr *DrainReconcile) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// try to lock the node to this reconcile loop.
+	// if we are not able this means there is another loop already handling this node, so we just exist
```
bn222, line 114: When would that happen? When would 2 reconciles be running for the same node?

controllers/drain_controller.go
```diff
@@ -0,0 +1,464 @@
...
+		//  doesn't need to drain anymore, so we can stop the drain
+		if nodeStateDrainAnnotationCurrent == constants.DrainComplete ||
+			nodeStateDrainAnnotationCurrent == constants.Draining {
+			completed, err := dr.drainer.CompleteDrainNode(ctx, node)
+			if err != nil {
+				reqLogger.Error(err, "failed to complete drain on node")
+				return ctrl.Result{}, err
+			}
+
+			// if we didn't manage to complete the un drain of the node we retry
```
bn222, line 214: This would happen if something else changes on the node annotation we get called while we were still draining?

controllers/drain_controller.go
```diff
@@ -0,0 +1,464 @@
...
+			err = utils.AnnotateObject(nodeNetworkState, constants.NodeStateDrainAnnotationCurrent, constants.DrainIdle, dr.Client)
+			if err != nil {
+				reqLogger.Error(err, "failed to annotate node with annotation", "annotation", constants.DrainIdle)
+				return ctrl.Result{}, err
+			}
+
+			reqLogger.Info("completed the un drain for node")
+			return ctrl.Result{}, nil
+		}
+	} else {
```
bn222, line 238: make it explicit what we expect the annotation to be. It can be this, that or something completely different that we didn't expect at all.

bn222: Maybe a switch case top level.

bn222:
This PR is doing all kinds of things: refactoring/cleaning up/adding tests for missing things/ making interfaces so we can mock things.

I recommend splitting it up into smaller PRs to ease review, and also to get the refactoring PRs in first. The Parallel drain PR should focus on adding the drain_controller. I'd like to make each PR as small as possible. Then I will review in more detail.

A few notes to start:
- We are creating drain annotation if it doesn't exist. I think we should simplify the code to always have a drain annotation so that we don't need to check if it exists or not.
- We are again adding huge functions that are more than 150 lines. I prefer to now take the opportunity and refactor more first (as part of a PR that goes in before we start working on the Parallel drain). The same goes for the drain controller. I prefer that we use more functions with good names. For example, you write `this cover the case the node is on idle`. The blob that you put under that comment can all go into a function with name `reconcileIdleNode` or some other clear name.
- We need to take care of the situation where a node gets deleted while we are draining. I want the drain code to be fool-proof.
- I'm pretty sure there are still races inside this code but I want to first break up the PR such that reviewing it becomes palatable.

controllers/drain_controller.go
```diff
@@ -0,0 +1,464 @@
...
+	dr.nodesInReconcileMutex.Lock()
+	defer dr.nodesInReconcileMutex.Unlock()
+
+	_, exist := dr.nodesInReconcile[nodeName]
+	if exist {
+		return false
+	}
+
+	dr.nodesInReconcile[nodeName] = nil
+	return true
```
bn222, line 94: You've implemented a thread safe set. Maybe use https://github.com/deckarep/golang-set instead?

SchSeba: Hi thanks for pointing out this one!
if possible I would like to avoid adding more packages for something really small here

SchSeba:
Hi @bn222 this was already split please review first https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/553

SchSeba:
Hi @bn222 @zeeke @adrianchiris @e0ne now that we are done with the interface redesign PR this one is ready for review.
so when you have time please take a look please :)

controllers/drain_controller.go
SchSeba: right, and also by the operator itself as we apply changes to the annotation we will get a new reconcile call so we just drop them

pkg/daemon/daemon.go
```diff
@@ -581,38 +475,45 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
 	}
-	if dn.platformHelpers.IsOpenshiftCluster() && !dn.platformHelpers.IsHypershift() {
-		if err = dn.getNodeMachinePool(); err != nil {
-			return err
-		}
-	}
-	if reqDrain {
-		if !dn.isNodeDraining() {
+
+	if reqDrain ||
```
zeeke, line 479: I had to refactor this to a different function to understand this logic:

```go
if reqDrain {
		returnFromSyncHandler, err := dn.handleDrain(reqReboot)
		if returnFromSyncHandler {
			return err
		}
	}
...

func (dn *Daemon) handleDrain(reqReboot bool) (bool, error) {
	if !utils.ObjectHasAnnotationKey(dn.desiredNodeState, consts.NodeStateDrainAnnotationCurrent) {
		return false, nil
	}

	if utils.ObjectHasAnnotation(dn.desiredNodeState, consts.NodeStateDrainAnnotationCurrent, consts.DrainIdle) {
		return false, nil
	}

	if utils.ObjectHasAnnotation(dn.desiredNodeState, consts.NodeStateDrainAnnotationCurrent, consts.DrainComplete) {
		log.Log.Info("nodeStateSyncHandler(): the node complete the draining")
		return false, nil
	}

	if dn.isNodeDraining() {
		log.Log.Info("nodeStateSyncHandler(): the node is still draining waiting")
		return true, nil
	}

	if dn.disableDrain {
		return false, nil
	}

	if reqReboot {
		log.Log.Info("nodeStateSyncHandler(): apply 'Reboot_Required' label for node")
		if err := dn.applyRequirement(consts.RebootRequired); err != nil {
			return true, err
		}

		log.Log.Info("nodeStateSyncHandler(): apply 'Reboot_Required' label for nodeState")
		if err := utils.AnnotateObject(dn.desiredNodeState,
			consts.NodeStateDrainAnnotation,
			consts.RebootRequired, dn.client); err != nil {
			return true, err
		}

		return true, nil
	}

	log.Log.Info("nodeStateSyncHandler(): apply 'Drain_Required' label for node")
	if err := dn.applyRequirement(consts.DrainRequired); err != nil {
		return true, err
	}

	log.Log.Info("nodeStateSyncHandler(): apply 'Drain_Required' label for nodeState")
	if err := utils.AnnotateObject(dn.desiredNodeState,
		consts.NodeStateDrainAnnotation,
		consts.DrainRequired, dn.client); err != nil {
		return true, err
	}

	return true, nil
}
```

It should be equivalent, unless I introduced coding errors. If it's correct, when the nodeState is DrainComplete, it will continue logging `"nodeStateSyncHandler(): the node complete the draining"`, I'm not getting who moves the nodestate from tha value.

Also, conder to add this refactor or a similar one to improve readability

api/v1/sriovnetworkpoolconfig_types.go
```diff
@@ -2,12 +2,25 @@ package v1
...
+	NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
+
+	// maxUnavailable defines either an integer number or percentage
+	// of nodes in the pool that can go Unavailable during an update.
+	//
+	// A value larger than 1 will mean multiple nodes going unavailable during
+	// the update, which may affect your workload stress on the remaining nodes.
+	// Drain will respect Pod Disruption Budgets (PDBs) such as etcd quorum guards,
+	// even if maxUnavailable is greater than one.
+	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
```
e0ne, line 23: Why do we neet `IntOrString` here? Can we use just `int` and:
```
// +kubebuilder:validation:Minimum=1
// +kubebuilder:validation:Maximum=100
```
?

controllers/drain_controller.go
```diff
@@ -0,0 +1,492 @@
...
+			}
+
+			// if we didn't manage to complete the un drain of the node we retry
+			if !completed {
+				reqLogger.Info("complete drain was not completed re queueing the request")
+				dr.recorder.Event(nodeNetworkState,
+					corev1.EventTypeWarning,
+					"DrainController",
+					"node complete drain was not completed")
+				return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
```
e0ne, line 221: Do we want to make reconcile timeout configurable?

controllers/drain_controller.go
```diff
@@ -0,0 +1,492 @@
...
+	reqLogger.Info("Max node allowed to be draining at the same time", "MaxParallelNodeConfiguration", maxUnv)
+	reqLogger.Info("Count of draining", "drainingNodes", current)
+
+	// if maxUnv is zero this means we drain all the nodes in parallel without a limit
+	if maxUnv == -1 {
+		reqLogger.Info("draining all the nodes in parallel")
+	} else if current >= maxUnv {
+		// the node requested to be drained, but we are at the limit so we re-enqueue the request
+		reqLogger.Info("MaxParallelNodeConfiguration limit reached for draining nodes re-enqueue the request")
+		// TODO: make this time configurable
```
e0ne, line 356: this ansers the question above. Please, add TODO note on line 221 too

e0ne (changes requested):

hack/run-e2e-conformance-virtual-ocp.sh
```diff
@@ -176,6 +176,7 @@ spec:
 EOF
 
 kubectl patch configs.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true,"storage":{"emptyDir": null,"pvc":{"claim":"registry-pv-claim"}},"topologySpreadConstraints":[],"rolloutStrategy":"Recreate","tolerations":[{"effect":"NoSchedule","key":"node-role.kubernetes.io/master","operator":"Exists"},{"effect":"NoSchedule","key":"node-role.kubernetes.io/control-plane","operator":"Exists"}]}}' --type=merge
+kubectl patch ingresscontrollers.operator.openshift.io/default -n openshift-ingress-operator --patch '{"spec":{"replicas": 1}}' --type=merge
```
bn222, line 179: How does this allow us to drain multiple nodes. Add the "why" to the "what" that you already in the commit message.

api/v1/helper.go
```diff
@@ -801,3 +802,39 @@ func NetFilterMatch(netFilter string, netValue string) (isMatch bool) {
 
 	return netFilterResult[0][1] == netValueResult[0][1] && netFilterResult[0][2] == netValueResult[0][2]
 }
+
+// MaxUnavailable calculate the max number of unavailable nodes to represent the number of nodes
+// we can drain in parallel
+func (s *SriovNetworkPoolConfig) MaxUnavailable(numOfNodes int) (int, error) {
+	// this means we want to drain all the nodes in parallel
```
bn222, line 809: Can that ever work? I think it would cause issues if it can't find any node to drain to. It may even lock up?

api/v1/helper.go
```diff
@@ -801,3 +802,39 @@ func NetFilterMatch(netFilter string, netValue string) (isMatch bool) {
...
+	if s.Spec.MaxUnavailable == nil {
```
bn222, line 810: I think the default should be 100%. We should not treat -1 as being the same as 100%.

api/v1/helper.go
```diff
@@ -801,3 +802,39 @@ func NetFilterMatch(netFilter string, netValue string) (isMatch bool) {
...
+			i := strings.TrimSuffix(intOrPercent.StrVal, "%")
+			v, err := strconv.Atoi(i)
+			if err != nil {
+				return 0, fmt.Errorf("invalid value %q: %v", intOrPercent.StrVal, err)
+			}
+			if v > 100 || v < 1 {
+				return 0, fmt.Errorf("invalid value: percentage needs to be between 1 and 100")
+			}
+		} else {
+			return 0, fmt.Errorf("invalid type: strings needs to be a percentage")
```
bn222, line 826: `HasSuffix` is not enough to detect percentage.

```
func isPercentage(input string) bool {
	pattern := `^(\d+(\.\d+)?|\.\d+)%$`

	regex, err := regexp.Compile(pattern)
	if err != nil {
		fmt.Println("Error compiling regular expression:", err)
		return false
	}

	return regex.MatchString(input)
}
```

bn222: Looking below, I think we shouldn't even do all kinds of logic on check if it is between 0 and 100 percent. I think we should instead just call GetScaledValueFromIntOrPercent and let it deal with erroring out if it's incorrect. After that function call, we can check if the value is between 1 and max nodes. Note also that your code is using 1% which can be less than 1 node.

controllers/drain_controller.go
```diff
@@ -0,0 +1,492 @@
...
+	err := dr.Get(ctx, req.NamespacedName, node)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			reqLogger.Error(err, "node object doesn't exist", "nodeName", req.Name)
+			return ctrl.Result{}, nil
+		}
+
+		reqLogger.Error(err, "failed to get node object from api re-queue the request", "nodeName", req.Name)
+		return ctrl.Result{}, err
+	}
```
bn222, line 148: Refactor into function

controllers/drain_controller.go
```diff
@@ -0,0 +1,492 @@
...
+	err = dr.Get(ctx, req.NamespacedName, nodeNetworkState)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			reqLogger.Error(err, "sriovNetworkNodeState object doesn't exist", "nodeName", req.Name)
+			return ctrl.Result{}, nil
+		}
+
+		reqLogger.Error(err, "failed to get sriovNetworkNodeState object from api re-queue the request", "nodeName", req.Name)
+		return ctrl.Result{}, err
+	}
```
bn222, line 161: Refactor into function

controllers/drain_controller.go
```diff
@@ -0,0 +1,492 @@
...
+
+	// create the drain state annotation if it doesn't exist in the sriovNetworkNodeState object
+	nodeStateDrainAnnotationCurrent, NodeStateDrainAnnotationCurrentExist := nodeNetworkState.Annotations[constants.NodeStateDrainAnnotationCurrent]
+	if !NodeStateDrainAnnotationCurrentExist {
+		err = utils.AnnotateObject(nodeNetworkState, constants.NodeStateDrainAnnotationCurrent, constants.DrainIdle, dr.Client)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		nodeStateDrainAnnotationCurrent = constants.DrainIdle
+	}
```
bn222, line 171: Refactor into a function, ensureDrainStateAnnotationExists

controllers/drain_controller.go
```diff
@@ -0,0 +1,492 @@
...
+
+	// create the drain state annotation if it doesn't exist in the node object
+	nodeDrainAnnotation, nodeDrainAnnotationExist := node.Annotations[constants.NodeDrainAnnotation]
+	if !nodeDrainAnnotationExist {
+		err = utils.AnnotateObject(node, constants.NodeDrainAnnotation, constants.DrainIdle, dr.Client)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		nodeDrainAnnotation = constants.DrainIdle
+	}
```
bn222, line 181: Why do we have the annotation twice? Once on the node and once on the node state?

controllers/drain_controller.go
```diff
@@ -0,0 +1,492 @@
...
+
+	reqLogger.V(2).Info("Drain annotations", "nodeAnnotation", nodeDrainAnnotation, "nodeStateAnnotation", nodeStateDrainAnnotationCurrent)
+
+	// Check the node request
+	if nodeDrainAnnotation == constants.DrainIdle {
```
bn222, line 186: All in all, it's quite confusing to see why we are tracking two node drain annotations.

bn222 (commented):
All in all, this logic seems very similar to the one in machine config operator; they also have something to manage the number of parallel drains essentially. I'm wondering, why don't we extract that piece of code form the machine config operator in an upstream project, and make both mco and sriov operator use that to avoid duplication of ccode.

pkg/daemon/daemon.go
SchSeba: I like the idea of moving this to a different function.
there is one problem that we had before your change
```golang
reqDrain || <code..>
```
after your change we will not access that function on the OR

I will rework this one

api/v1/sriovnetworkpoolconfig_types.go
SchSeba: this allow us to have both number and percentage that is nice so you don't need to change this if the pool scale you just put something like "33%" and it will scale base on the number of nodes you have in the pool

controllers/drain_controller.go
SchSeba: maybe in the future should this be part of the sriovOperatorConfig or where will we configure this number?

controllers/drain_controller.go
SchSeba: sure

api/v1/helper.go
SchSeba: sure this can work if you have a pool of worker nodes don't doesn't have any PDB pod running on them you can reboot all of them

controllers/drain_controller.go
SchSeba: that is right. one is the req from the daemon and the other one is the state from the operator point of view

controllers/drain_controller.go
SchSeba: it's part of the design proposal. We need two because two different components are updating so every component should know the current status of is flow

controllers/drain_controller.go
SchSeba: good point done

controllers/drain_controller.go
bn222: Can you elaborate on why we can't have one instead of two annotations?

SchSeba: sure :)

so example: 
1. daemon move the annotation to reqDrain
2. then the operator moves to draining the annotation
3. user remove the policy
4. daemon moves the annotation to idle

you are stuck in a middle of a drain because the operator doesn't know where he was

in the case we have 2 annotations
1. daemon move the annotation to reqDrain
2. then the operator mark  draining is own annotation
3. user remove the policy
4. daemon moves the annotation to idle
5. operator see the change and run the un-drain 

let me know if that is clear.

pkg/consts/constants.go
```diff
@@ -53,24 +53,27 @@ const (
...
-	DrainMcpPaused   DrainState = "Draining_MCP_Paused"
-	Draining         DrainState = "Draining"
-	DrainingComplete DrainState = "Draining_Complete"
-	RebootComplete   DrainState = "Reboot_Complete"
+	MachineConfigPoolPausedAnnotation       = "sriovnetwork.openshift.io/state"
+	MachineConfigPoolPausedAnnotationIdle   = "Idle"
+	MachineConfigPoolPausedAnnotationPaused = "Paused"
+
+	NodeDrainAnnotation             = "sriovnetwork.openshift.io/state"
+	NodeStateDrainAnnotation        = "sriovnetwork.openshift.io/desired-state"
```
zeeke, line 61: this annotation is referred in `daemon.go` only, always with write operations. So it looks like it's never read. Can we remove it?

controllers/helper.go
```diff
@@ -39,7 +44,91 @@ const (
...
 
-var namespace = os.Getenv("NAMESPACE")
+type DrainAnnotationPredicate struct {
+	predicate.Funcs
+}
+
+func (DrainAnnotationPredicate) Create(e event.CreateEvent) bool {
+	logger := log.FromContext(context.TODO())
+	if e.Object == nil {
+		logger.Info("Create event: node has no drain annotation", "node", e.Object.GetName())
```
zeeke, line 54: this log lies. In general, I would avoid logging in these predicate functions.

Same applies for L59, L68, L72 and for `DrainStateAnnotationPredicate`

zeeke (commented):
Left few comments. My main concern is about node and nodestate annotations, which might have some inconsistency.

pkg/consts/constants.go
SchSeba: that is to prepare for the next step that will move the operator to look into that annotation

controllers/drain_controller_test.go
```diff
@@ -0,0 +1,292 @@
...
+					"test": "",
+				},
+			}}
+			Expect(k8sClient.Create(context.TODO(), poolConfig)).Should(Succeed())
+
+			// Two nodes require to drain at the same time
+			simulateDaemonSetAnnotation(node1, constants.DrainRequired)
+			simulateDaemonSetAnnotation(node2, constants.DrainRequired)
+
+			// Only the first node drains
```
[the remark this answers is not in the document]
SchSeba: done

controllers/drain_controller.go
```diff
@@ -0,0 +1,492 @@
...
+		client,
+		Scheme,
+		recorder,
+		drainer,
+		resourcePrefix,
+		map[string]interface{}{},
+		sync.Mutex{},
+		sync.Mutex{}}, nil
+}
+func (dr *DrainReconcile) TryLockNode(nodeName string) bool {
```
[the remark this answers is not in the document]
SchSeba: done

deployment/sriov-network-operator/templates/clusterrole.yaml
```diff
@@ -11,17 +11,20 @@ rules:
...
     verbs: ["*"]
+  - apiGroups: [""]
+    resources: ["pods/eviction"]
+    verbs: ["create"]
   - apiGroups: ["apps"]
     resources: ["daemonsets"]
     verbs: ["get"]
   - apiGroups: [""]
-    resources: ["namespaces", "serviceaccounts"]
+    resources: [namespaces, serviceaccounts]
```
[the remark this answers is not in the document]
SchSeba: done

controllers/drain_controller.go
```diff
@@ -0,0 +1,492 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &DrainReconcile{
+		client,
+		Scheme,
+		recorder,
+		drainer,
+		resourcePrefix,
```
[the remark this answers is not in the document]
SchSeba: nice idea!

bn222:
@SchSeba I'm keeping an eye on this, because I want to do a final review when things have been cleaned up.

SchSeba:
Hi @bn222 the PR is in a ready state so if you can give this one another round please do :)
I think I answered all the comments from your side

deployment/sriov-network-operator/templates/clusterrole.yaml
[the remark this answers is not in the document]
zeeke: still see them with quotes

pkg/daemon/daemon_test.go
```diff
@@ -105,15 +117,49 @@ var _ = Describe("Config Daemon", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		kClient := kclient.NewClientBuilder().WithScheme(scheme.Scheme).WithRuntimeObjects(&corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{Name: "test-node"}},
+			&sriovnetworkv1.SriovNetworkNodeState{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-node",
+					Namespace: vars.Namespace,
+				}}).Build()
+
+		FakeSupportedNicIDs := corev1.ConfigMap{
```
zeeke, line 133: `FakeSupportedNicIDs` and `SriovDevicePluginPod` are already defined on lines L35 and L46. I think we can keep them or these, no preference.

SchSeba: you are right I miss that thanks!

deployment/sriov-network-operator/templates/clusterrole.yaml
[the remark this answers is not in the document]
SchSeba: oh you want me to add it sure no problem

zeeke (approved):
LGTM

api/v1/helper.go
ykulazhenkov: can we return value from numOfNodes in this case?

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -47,10 +47,6 @@ func (p *MellanoxPlugin) Spec() string {
 // OnNodeStateChange Invoked when SriovNetworkNodeState CR is created or updated, return if need dain and/or reboot node
 func (p *MellanoxPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
 	log.Log.Info("mellanox plugin OnNodeStateChange()")
-
```
ykulazhenkov, line 50: nit: out of scope for the PR

pkg/platforms/openshift/openshift.go
```diff
@@ -22,29 +32,36 @@ const (
 //go:generate ../../../bin/mockgen -destination mock/mock_openshift.go -source openshift.go
 type OpenshiftContextInterface interface {
 	GetFlavor() OpenshiftFlavor
-	GetMcClient() mcclientset.Interface
 	IsOpenshiftCluster() bool
 	IsHypershift() bool
+
+	OpenshiftDrainNode(context.Context, *corev1.Node) (bool, error)
```
ykulazhenkov, line 38: nit: names for the drain* functions are a bit confusing. My original thought was that these functions will do complete drain operation for Openshift cluster. Maybe we can rename functions to `OpenshiftBeforeDrainNode` and OpenshiftAfterDrainNode` to make it clear that these functions provide  extra logic for the drain, but doesn't do drain by their own. WDYT?

controllers/drain_controller.go
ykulazhenkov: I think  that implementation of the queue in the client-go package [1] will protect you from this case. You will have only 1 reconcile request for one key(node name in our case) even if max parallel reconciles is 50.

[1] https://github.com/adrianchiris/sriov-network-operator/blob/56530b9c0e1c893d686f545df6876ddce807950f/vendor/k8s.io/client-go/util/workqueue/queue.go#L163

controllers/drain_controller.go
```diff
@@ -0,0 +1,501 @@
...
+//+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
+func (dr *DrainReconcile) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// try to lock the node to this reconcile loop.
+	// if we are not able this means there is another loop already handling this node, so we just exist
+	if !dr.tryLockNode(req.Name) {
```
ykulazhenkov, line 107: I think that implementation of the queue in the client-go package [1] will protect you from this case. You will have only 1 reconcile request for one key(node name in our case) even if max parallel reconciles is 50.

[1] https://github.com/adrianchiris/sriov-network-operator/blob/56530b9c0e1c893d686f545df6876ddce807950f/vendor/k8s.io/client-go/util/workqueue/queue.go#L163

SchSeba: done thanks!

pkg/platforms/openshift/openshift.go
SchSeba: no problem sounds good to me

api/v1/helper.go
SchSeba: can we have a race if we do that and a new node is added?

pkg/daemon/daemon.go
```diff
@@ -699,43 +572,68 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+	}
+
+	return nil
+}
+
+func (dn *Daemon) applyRequirement(label string) error {
+	log.Log.Info("applyDrainRequired(): no other node is draining")
+	err := utils.AnnotateNode(vars.NodeName, consts.NodeDrainAnnotation, label, dn.client)
+	if err != nil {
+		log.Log.Error(err, "applyDrainRequired(): Failed to annotate node")
```
adrianchiris, line 636: logs need adjusting

controllers/drain_controller.go
```diff
@@ -0,0 +1,461 @@
...
+	constants "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/drain"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+type DrainReconcile struct {
+	client.Client
+	Scheme   *runtime.Scheme
```
adrianchiris, line 52: i dont see Scheme being used.

controllers/drain_controller.go
```diff
@@ -0,0 +1,461 @@
...
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	constants "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/drain"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+type DrainReconcile struct {
+	client.Client
```
adrianchiris, line 51: is there a particular reason to embed this struct ?  (this adds a bunch of pubic methods to it which i dont think is needed)

controllers/drain_controller.go
```diff
@@ -0,0 +1,461 @@
...
+			continue
+		}
+
+		selector, err := metav1.LabelSelectorAsSelector(npc.Spec.NodeSelector)
+		if err != nil {
+			logger.Error(err, "failed to create label selector from nodeSelector", "nodeSelector", npc.Spec.NodeSelector)
+			return nil, nil, err
+		}
+
+		// If a pool with a nil or empty selector creeps in, it should match nothing, not everything.
```
adrianchiris, line 379: empty selector selects everything usually.

moreover for the default object it technically select everything.

what if i want to override the default obj ?

pkg/daemon/daemon.go
```diff
@@ -668,21 +540,22 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
-			return err
-		}
-	} else {
-		if !dn.nodeHasAnnotation(annoKey, annoIdle) {
-			if err := dn.annotateNode(vars.NodeName, annoIdle); err != nil {
-				log.Log.Error(err, "nodeStateSyncHandler(): failed to annotate node")
-				return err
-			}
-		}
+
```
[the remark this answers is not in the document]
SchSeba: it's above if reboot is required this means also drain is require

pkg/daemon/daemon.go
```diff
@@ -668,21 +540,22 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+		return err
 	}
+
+	log.Log.Info("nodeStateSyncHandler(): apply 'Idle' label for nodeState")
+	if err := utils.AnnotateObject(dn.desiredNodeState,
+		consts.NodeStateDrainAnnotation,
+		consts.DrainIdle, dn.client); err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
SchSeba: what you mean?
when we move to idle it will call the controller again and that one will finish the work

controllers/drain_controller.go
SchSeba: this is done by the kubebuilder when we create a new controller all of them have the same state

controllers/drain_controller.go
SchSeba: this is done by the kubebuilder when we create a new controller all of them have the same state

controllers/drain_controller.go
SchSeba: I removed the default object as requested :)

pkg/daemon/daemon.go
SchSeba: removed the functions

pkg/daemon/daemon.go
```diff
@@ -699,43 +572,68 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+	if err := utils.AnnotateObject(dn.desiredNodeState,
+		consts.NodeStateDrainAnnotation,
+		consts.DrainRequired, dn.client); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (dn *Daemon) applyRequirement(label string) error {
```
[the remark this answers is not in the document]
SchSeba: done

pkg/daemon/daemon.go
```diff
@@ -668,21 +540,22 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
-		}
-	} else {
-		if !dn.nodeHasAnnotation(annoKey, annoIdle) {
-			if err := dn.annotateNode(vars.NodeName, annoIdle); err != nil {
-				log.Log.Error(err, "nodeStateSyncHandler(): failed to annotate node")
-				return err
-			}
-		}
+
+	log.Log.Info("nodeStateSyncHandler(): apply 'Idle' label for node")
```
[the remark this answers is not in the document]
SchSeba: right thanks!

pkg/daemon/daemon.go
```diff
@@ -668,21 +540,22 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+	if err := dn.applyRequirement(consts.DrainIdle); err != nil {
+		log.Log.Error(err, "nodeStateSyncHandler(): failed to annotate node")
+		return err
 	}
+
+	log.Log.Info("nodeStateSyncHandler(): apply 'Idle' label for nodeState")
```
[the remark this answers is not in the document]
SchSeba: right thanks!

pkg/daemon/daemon.go
```diff
@@ -699,43 +572,68 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
-		return true
-	}
-	return false
-}
-
 // isNodeDraining: check if the node is draining
 // both Draining and MCP paused labels will return true
 func (dn *Daemon) isNodeDraining() bool {
-	anno, ok := dn.node.Annotations[annoKey]
+	anno, ok := dn.desiredNodeState.Labels[consts.NodeStateDrainAnnotationCurrent]
```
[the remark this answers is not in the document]
SchSeba: right thanks!

pkg/utils/cluster.go
```diff
@@ -110,3 +111,54 @@ func openshiftControlPlaneTopologyStatus(c client.Client) (configv1.TopologyMode
...
+	log.Log.V(2).Info("AnnotateObject(): Annotate object",
+		"objectName", obj.GetName(),
+		"objectKind", obj.GetObjectKind(),
+		"annotation", value)
+	newObj := obj.DeepCopyObject().(client.Object)
+	if newObj.GetAnnotations() == nil {
+		newObj.SetAnnotations(map[string]string{})
+	}
+
+	patch := client.MergeFrom(obj)
```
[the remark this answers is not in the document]
SchSeba: done

main.go
```diff
@@ -207,11 +264,36 @@ func main() {
...
 	kubeclient := kubernetes.NewForConfigOrDie(ctrl.GetConfigOrDie())
-	if err := sriovnetworkv1.InitNicIDMapFromConfigMap(kubeclient, namespace); err != nil {
+	if err := sriovnetworkv1.InitNicIDMapFromConfigMap(kubeclient, vars.Namespace); err != nil {
 		return err
 	}
 
 	return nil
 }
+
+func createDefaultPoolConfig(c client.Client) error {
```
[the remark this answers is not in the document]
SchSeba: I removed the default creation

SchSeba:
I think I answered all the comments. please take another look when you have time

controllers/drain_controller.go
ykulazhenkov: You removed creation of the default Pool from the main.go, but you still have logic to recreated default pool in the https://github.com/SchSeba/sriov-network-operator-1/blob/8d80f3ea5db2536bb7bebf3773a50181279b8dc7/controllers/sriovnetworkpoolconfig_controller.go#L69. Do we want to remove this as well?

controllers/drain_controller.go
```diff
@@ -0,0 +1,458 @@
...
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
+func (dr *DrainReconcile) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// configure logs
+	reqLogger := log.FromContext(ctx)
+	reqLogger.Info("Reconciling Drain")
+	reqLogger.V(2).Info("node locked for drain controller", "nodeName", req.Name)
```
ykulazhenkov, line 93: I think we can drop this log message.

controllers/drain_controller.go
```diff
@@ -0,0 +1,458 @@
...
+		if err != nil {
+			return "", err
+		}
+		return constants.DrainIdle, nil
+	}
+
+	return value, nil
+}
+
+func (dr *DrainReconcile) checkForNodeDrain(ctx context.Context, node *corev1.Node) (*reconcile.Result, error) {
```
ykulazhenkov, line 278: The function name is a bit confusing for me. Rename to `tryDrainNode`?

controllers/drain_controller.go
```diff
@@ -0,0 +1,458 @@
...
+		nodeList := &corev1.NodeList{}
+		err = dr.List(ctx, nodeList, &client.ListOptions{LabelSelector: selector})
+		if err != nil {
+			logger.Error(err, "failed to list nodes using with label selector", "labelSelector", selector)
+			return nil, nil, err
+		}
+
+		return selectedNpcl[0], nodeList.Items, nil
+	} else {
+		logger.V(2).Info("using \"default\" sriovNetworkPool with MaxUnavailable of one", "pool", *defaultNpcl)
```
ykulazhenkov, line 412: Do we want to mention default pool in the log message?  Maybe we should change the message to say something like: node doesn't belong to any pool, using default drain configuration

deploy/clusterrole.yaml
```diff
@@ -48,12 +51,6 @@ rules:
 - apiGroups: ["apps"]
   resources: ["daemonsets"]
   verbs: ["get"]
-- apiGroups: [""]
-  resources: ["pods/eviction"]
-  verbs: ["create"]
-- apiGroups: ["machineconfiguration.openshift.io"]
```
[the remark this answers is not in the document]
SchSeba: for nodes yes please don't need it anymore as it's part of the operator now

SchSeba:
Done @ykulazhenkov with the comments thanks!

be aware that in the daemon I use the background context as we don't have context there.
that will change on my next PR where we change the daemon to use controller-runtime manager

controllers/drain_controller.go
ykulazhenkov: @SchSeba please address this comment

ykulazhenkov:
> Done @ykulazhenkov with the comments thanks!
> 
> be aware that in the daemon I use the background context as we don't have context there. that will change on my next PR where we change the daemon to use controller-runtime manager

Thanks @SchSeba  for addressing my comments.
Please, check the last one https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/555#discussion_r1514058090

controllers/drain_controller.go
SchSeba: you are right thanks!

SchSeba:
Hi @ykulazhenkov done thanks!

controllers/drain_controller.go
adrianchiris: Thanks @SchSeba 

regarding

> empty selector selects everything usually.

can we make this behaviour aligned with how kubernetes handles empty selectors ? (match all)
currently the opposite happens.

SchSeba:
Hi @zeeke @e0ne @ykulazhenkov when you can please take another look on the PR I will like to try and get this one merge soon.

Thanks!

ykulazhenkov:
Thanks, @SchSeba.
I think it is important to have agreement about this comment https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/555#discussion_r1519320816 before we merge.
Can you comment?

controllers/drain_controller.go
ykulazhenkov: +1 to change the behavior to "match all" for an empty selector unless we have a good reason not to do so

SchSeba: done :)

SchSeba:
Hi @ykulazhenkov I did the requested change for the empty node selector let me know if that works for you :)

ykulazhenkov (approved):
Thanks @SchSeba,
LGTM

e0ne (approved):
LGTM

SchSeba:
Thanks for all the help reviewing this complicated feature!

merging after 3 approvals

PR #556, opened 2023-12-05
operator: LeaderElectionReleaseOnCancel 

zeeke:
When manually restarting the operator, the leader election
takes 5+ minutes to acquire the lease on startup:

```
I1205 16:06:02.101302       1 leaderelection.go:245] attempting to acquire leader lease openshift-sriov-network-operator/a56def2a.openshift.io...
...
I1205 16:08:40.133558       1 leaderelection.go:255] successfully acquired lease openshift-sriov-network-operator/a56def2a.openshift.io
```

This PR makes sure the lease is released when the operator shutdown


zeeke:
I validated this change with a [test commit](https://github.com/k8snetworkplumbingwg/sriov-network-operator/commit/cca761b61d70b300f6fab169fa98b6341ff8bb48). This comes from the related [CI logs](https://github.com/k8snetworkplumbingwg/sriov-network-operator/actions/runs/7113041370):

```
// operator logs
I1206 10:27:04.430837       1 leaderelection.go:250] attempting to acquire leader lease sriov-network-operator/a56def2a.openshift.io...
I1206 10:27:04.435885       1 leaderelection.go:260] successfully acquired lease sriov-network-operator/a56def2a.openshift.io

// events.json
{
            "metadata": {
                "name": "a56def2a.openshift.io.179e37328a9a0420",
                "namespace": "sriov-network-operator",
                ...
            },
            "involvedObject": {
                "kind": "Lease",
                "namespace": "sriov-network-operator",
                "name": "a56def2a.openshift.io",
                ...
            },
            "reason": "LeaderElection",
            "message": "sriov-network-operator-787ddd7794-xl66c_d5427070-808c-4c52-ae03-c53bf37c6e9f became leader",
            "source": {
                "component": "sriov-network-operator-787ddd7794-xl66c_d5427070-808c-4c52-ae03-c53bf37c6e9f"
            },
            "firstTimestamp": "2023-12-06T10:26:39Z",
            ...
            "reportingComponent": "sriov-network-operator-787ddd7794-xl66c_d5427070-808c-4c52-ae03-c53bf37c6e9f",
            "reportingInstance": ""
        },
        {
            "metadata": {
                "name": "a56def2a.openshift.io.179e37382af3762a",
                "namespace": "sriov-network-operator",
                ...
            },
            "involvedObject": {
                "kind": "Lease",
                "namespace": "sriov-network-operator",
                "name": "a56def2a.openshift.io",
                ...
            },
            "reason": "LeaderElection",
            "message": "sriov-network-operator-787ddd7794-xl66c_d5427070-808c-4c52-ae03-c53bf37c6e9f stopped leading",
            "source": {
                "component": "sriov-network-operator-787ddd7794-xl66c_d5427070-808c-4c52-ae03-c53bf37c6e9f"
            },
            "firstTimestamp": "2023-12-06T10:27:03Z",
            ...
            "reportingComponent": "sriov-network-operator-787ddd7794-xl66c_d5427070-808c-4c52-ae03-c53bf37c6e9f",
            "reportingInstance": ""
        },
        {
            "metadata": {
                "name": "a56def2a.openshift.io.179e37385e817eda",
                "namespace": "sriov-network-operator",
                ...
            },
            "involvedObject": {
                "kind": "Lease",
                "namespace": "sriov-network-operator",
                "name": "a56def2a.openshift.io",
                ...
            },
            "reason": "LeaderElection",
            "message": "sriov-network-operator-787ddd7794-vmsck_a6f02db2-11d5-4dc2-b4b6-8c615ba7f452 became leader",
            "source": {
                "component": "sriov-network-operator-787ddd7794-vmsck_a6f02db2-11d5-4dc2-b4b6-8c615ba7f452"
            },
            "firstTimestamp": "2023-12-06T10:27:04Z",
            ...
            "reportingComponent": "sriov-network-operator-787ddd7794-vmsck_a6f02db2-11d5-4dc2-b4b6-8c615ba7f452",
        },
```

adrianchiris:
controller runtime defaults seem much shorter.

i see we override with openshift inspired defaults.

lease is 137s and renew 26s, so it should take no longer than 2.7min
in your logs its 2.6min

given its not 5+ min, do we really need this ?

main.go
```diff
@@ -190,17 +211,20 @@ func main() {
...
+	leaderElectionContext, cancelLeaderElection := context.WithCancel(context.Background())
+	go func() {
+		setupLog.Info("starting leader election manager")
+		if err := leaderElectionMgr.Start(leaderElectionContext); err != nil {
+			setupLog.Error(err, "Leader Election Manager exited non-zero")
+			os.Exit(1)
+		}
+	}()
+
+	<-leaderElectionMgr.Elected()
```
adrianchiris, line 223: we want to wait before setting up signal handler?

main.go
```diff
@@ -190,17 +211,20 @@ func main() {
...
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
-		setupLog.Error(err, "unable to set up ready check")
-		os.Exit(1)
-	}
+	leaderElectionContext, cancelLeaderElection := context.WithCancel(context.Background())
+	go func() {
+		setupLog.Info("starting leader election manager")
+		if err := leaderElectionMgr.Start(leaderElectionContext); err != nil {
+			setupLog.Error(err, "Leader Election Manager exited non-zero")
+			os.Exit(1)
```
adrianchiris, line 219: something seems off to me with the use of os.Exit() here and below

maybe just need to use waitgroup to wait for these managers to complete.

what if we os.Exit() before any defer statements are executed. it may not really be an issue IDK.

main.go
```diff
@@ -99,22 +99,43 @@ func main() {
...
-		LeaderElection:         enableLeaderElection,
-		LeaseDuration:          &le.LeaseDuration,
-		RenewDeadline:          &le.RenewDeadline,
-		RetryPeriod:            &le.RetryPeriod,
-		LeaderElectionID:       "a56def2a.openshift.io",
-		Cache:                  cache.Options{DefaultNamespaces: map[string]cache.Config{namespace: {}}},
+		Scheme:        scheme,
+		Metrics:       server.Options{BindAddress: metricsAddr},
+		WebhookServer: webhook.NewServer(webhook.Options{Port: 9443}),
+
```
adrianchiris, line 131: we need this newline ?

main.go
```diff
@@ -190,17 +211,20 @@ func main() {
...
+	}()
+
+	<-leaderElectionMgr.Elected()
 
 	stopCh := ctrl.SetupSignalHandler()
 	go func() {
+		setupLog.Info("starting global manager")
 		if err := mgrGlobal.Start(stopCh); err != nil {
 			setupLog.Error(err, "Manager Global exited non-zero")
 			os.Exit(1)
```
mlguerrero12, line 230: shouldn't we cancel the context here as well before exiting?

main.go
```diff
@@ -210,7 +234,10 @@ func main() {
 	// Remove all finalizers after controller is shut down
 	defer utils.Shutdown()
 
-	setupLog.Info("starting manager")
+	// LeaderElection must be canceled after resource cleanups
+	defer cancelLeaderElection()
```
mlguerrero12, line 238: then, the defer should be before the one for utils.Shutdown, right?

main.go
mlguerrero12: ignore this if defer goes before the shutdown, just after it is elected

main.go
mlguerrero12: +1

main.go
zeeke: not at all!

zeeke:
> controller runtime defaults seem much shorter.
> 
> i see we override with openshift inspired defaults.
> 
> lease is 137s and renew 26s, so it should take no longer than 2.7min in your logs its 2.6min
> 
> given its not 5+ min, do we really need this ?

You're right: regular clusters may get a delay of 2.7min at most. With Single Node Openshift (SNO in this context) it can grow to 5 minutes. 

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/572dc27acde15859872a34d5ef75279b658e165b/pkg/leaderelection/leaderelection.go#L45
https://github.com/openshift/library-go/commit/2612981f3019479805ac8448b997266fc07a236a#diff-61dd95c7fd45fa18038e825205fbfab8a803f1970068157608b6b1e9e6c27248R127

I'm taking that math for granted, and one of our customers is experiencing that 5 minutes restart during upgrades, on SNO.

```
I1123 13:20:44.792821       1 leaderelection.go:245] attempting to acquire leader lease openshift-sriov-network-operator/a56def2a.openshift.io...
I1123 13:25:56.033843       1 leaderelection.go:255] successfully acquired lease openshift-sriov-network-operator/a56def2a.openshift.io
```

main.go
zeeke: TBH, I don't know: if one manager returns an error while the others are running, it should signal a stop and wait. But it could become a little bit complicated and there is a risk of getting deadlock as far as I can see.
I can explore a little more if you think it's worth

main.go
```diff
@@ -190,17 +211,20 @@ func main() {
...
 
-	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
-		setupLog.Error(err, "unable to set up health check")
-		os.Exit(1)
-	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
-		setupLog.Error(err, "unable to set up ready check")
-		os.Exit(1)
-	}
+	leaderElectionContext, cancelLeaderElection := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
zeeke: That way `leaderElectionContext` would stop as soon as the stop context is Done. Which is what I'm trying to achieve with this PR.

I tried using nested contexts here
https://go.dev/play/p/6dFfBQyXlW1

main.go
zeeke: do you mean `leaderElectionContext`?

main.go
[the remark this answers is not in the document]
mlguerrero12: by using the stop context as parent of this one, the leaderElectionMgr.Start function will be cancelled if a singint or sigterm is received.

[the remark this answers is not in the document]
zeeke: Yes, that's what I don't want to happen. Here's the ending sequence

- sigint arrive
- stopCh is Done
- mgr and mgrGlobals finish their work and return
- as mgr is not on a go routing, when it returns it complete the function, triggering all the `defer`s
- `utils.Shutdown()` defer goes first. do the cleanup (the leader election is still running here, so we have the lock lease)
- `cancelLeaderElection()` defer goes second, stopping the manager which in turn release the lock

Am I missing something here?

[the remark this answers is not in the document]
mlguerrero12: Right, it would be the same as using the `ReleaseOnCancel` option.

My concern here is that after calling `cancelLeaderElection()` (with the defer), the program exits, so nothing ensures that `leaderElectionMgr` really finishes. There is a race condition.

The internal implementation of the `ReleaseOnCancel` option uses a channel to ensure that the routine has stopped before continuing with the shutdown. Perhaps you can do the same here.

zeeke:
> Hi @zeeke @adrianchiris not related to this PR directly but question do you know why we have two manges?? and now 3

AFAIU, mgr watches resource in the operator's namespace. mgrGlobal watches every namespace, as it runs Sriov[IB}Network reconcilers which react on NetworkAttachmentDefinitions.

I guess we can unify those managers with one that watches all namespaces, but maybe we get some performance degradation.

Also, I'm investigating about the shutdown logic to see if we can get rid of some bit and pieces

main.go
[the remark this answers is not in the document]
zeeke: > My concern here is that after calling cancelLeaderElection() (with the defer), the program exits, so nothing ensures that leaderElectionMgr really finishes. There is a race condition.

Added a wait group to ensure leaderElectionMgr is correctly stopped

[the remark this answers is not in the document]
mlguerrero12: it seems to me that this never exits. The leader election manager is stopped with a defer but before that, you are waiting for this to stop, so, it won't happen?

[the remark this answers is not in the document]
zeeke: fixed

main.go
adrianchiris: +1 on looking how to streamline this

adrianchiris:
> You're right: regular clusters may get a delay of 2.7min at most. With Single Node Openshift (SNO in this context) it can grow to 5 minutes.

Ack thx for the clarification.


> Also, I'm investigating about the shutdown logic to see if we can get rid of some bits and pieces, as I'm not very happy with the current election-proxy-manager proposal

LMK when you are done so we can give this one another round of reviews.

zeeke:
@adrianchiris , @mlguerrero12 I streamlined the logic for spinning up the managers and waiting for them to finish.
Can you take another look?

Regarding the shutdown procedure, we have 3 main duties:

a. `Set ResourfceInjectorWebhook.FailurePolicy = Ignore` (mutating webhook for pods)
this can be safely removed, as the resource injector failure policy is Ignore by default.

b. `Set ValidationWebhook.FailurePolicy = Ignore` for `sriovoperatorconfigs` and `sriovnetworknodepolicies`.
I think this is to avoid problem when deleting the operator's namespace. If the operator pod stops before the deletion of the webhook, the API server will reject the operation. 

c. `Remove finalizers from SriovNetworks`
Finalizers are used to delete NetworkAttachmentDefinition when a SriovNetwork is deleted. We can't use OwnerReference because net-attach-def can be in a different namespace than the SriovNetwork. BTW, I'm not getting exactly which scenario we are trying to save with them.

I think it's worth starting a discussion about the shutdown topic and trying to list all the use cases we want to support.

For this PR, I'd prefer to go with this stop-gap solution.

main.go
```diff
@@ -183,6 +203,20 @@ func main() {
...
+
+	select {
+	case <-leaderElectionMgr.Elected():
+	case err := <-leaderElectionErr:
+		setupLog.Error(err, "Leader Election Manager error")
+		os.Exit(1)
+	}
+
 	// Create a default SriovNetworkNodePolicy
 	err = createDefaultPolicy(kubeClient)
```
mlguerrero12, line 221: cancelLeaderElection before exit(1) in createDefaultPolicy?

zeeke: done

main.go
```diff
@@ -183,43 +203,97 @@ func main() {
...
+	case err := <-leaderElectionErr:
+		setupLog.Error(err, "Leader Election Manager error")
+		globalManagerCancel()
+		namespacedManagerCancel()
+		<-globalManagerErr
+		<-namespacedManagerErr
+
+		os.Exit(1)
+
+	case err := <-globalManagerErr:
```
mlguerrero12, line 277: I think a shutdown should be included here as well since the namespace manager might be up, and perhaps also in the case above. Not sure since the operator won't be elected anymore for that case.

zeeke: I'm fine with including `utils.Shutdown()` in case of globalManager and namespacedManager error. I wouldn't do it in case of leaderelection error

mlguerrero12:
LGTM

zeeke:
Shutdown logic for the long run will be discussed in
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/discussions/608

@SchSeba, @e0ne , @adrianchiris Please take another look at this.

I added an end-to-end test to verify the restart of the operator is fast enough

deploy/operator.yaml
```diff
@@ -41,6 +41,8 @@ spec:
           image: $SRIOV_NETWORK_OPERATOR_IMAGE
           command:
           - sriov-network-operator
+          args:
+          - --leader-elect=$OPERATOR_LEADER_ELECTION_ENABLE
```
SchSeba, line 45: do we need the also in the config folder,helm folder and for the openshift bundle?

hack/run-e2e-conformance-virtual-ocp.sh
```diff
@@ -188,6 +188,7 @@ export OPERATOR_EXEC=kubectl
 export CLUSTER_TYPE=openshift
 export DEV_MODE=TRUE
 export CLUSTER_HAS_EMULATED_PF=TRUE
+export OPERATOR_LEADER_ELECTION_ENABLE=true
```
SchSeba, line 191: do we want it to be true also in the k8s lane?

SchSeba (commented):
I left some small comments nice work :)

deploy/operator.yaml
zeeke: openshift bundle already has the `--leader-elect=true` parameter

https://github.com/openshift/sriov-network-operator/blob/master/manifests/stable/sriov-network-operator.clusterserviceversion.yaml#L335

I don't know if we want to enable this feature for Helm, too. Maybe we can discuss and implement it in a different PR?

hack/run-e2e-conformance-virtual-ocp.sh
zeeke: IMO no. We have two lanes, and for me, it's good to have this flag tested with both values. 
The alternative might be to add other lanes with all possible combinations: `ocp|leader-elect=true`, `ocp|leader-elect=false`, `k8s|leader-elect=true`, `k8s|leader-elect=false`. But it would be a waste of resources, in my opinion.

test/conformance/tests/test_sriov_operator.go
```diff
@@ -276,6 +277,26 @@ var _ = Describe("[sriov] operator", func() {
...
+			By("Delete the operator's pod")
+			deleteOperatorPod()
+
+			By("Assert the new operator's pod acquire the lease before 30 seconds")
+			Eventually(func(g Gomega) {
+				newLease, err := clients.CoordinationV1Interface.Leases(operatorNamespace).Get(context.Background(), consts.LeaderElectionID, metav1.GetOptions{})
+				g.Expect(err).ToNot(HaveOccurred())
+
+				g.Expect(newLease.Spec.HolderIdentity).ToNot(Equal(oldLease.Spec.HolderIdentity))
+			}, 45*time.Second, 5*time.Second).Should(Succeed())
```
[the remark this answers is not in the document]
zeeke: good point! This will save us a lot of flakes. 
Fixing

hack/run-e2e-conformance-virtual-ocp.sh
SchSeba: yep agree we don't have many CI resources to test all the combinations

SchSeba (approved):
LGTM

deploy/operator.yaml
SchSeba: sounds good can you please open an issue so we don't forget :)

SchSeba:
please review this one an open the issue so we don't forget :)

zeeke:
> please rebase this one an open the issue so we don't forget :)

rebased! The discussion about shutdown is here:
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/discussions/608

SchSeba:
@e0ne @ykulazhenkov if you have time please take a look on this PR please

ykulazhenkov:
> > Hi @zeeke @adrianchiris not related to this PR directly but question do you know why we have two manges?? and now 3
> 
> AFAIU, mgr watches resource in the operator's namespace. mgrGlobal watches every namespace, as it runs Sriov[IB}Network reconcilers which react on NetworkAttachmentDefinitions.
> 
> I guess we can unify those managers with one that watches all namespaces, but maybe we get some performance degradation.
> 
> Also, I'm investigating about the shutdown logic to see if we can get rid of some bits and pieces, as I'm not very happy with the current election-proxy-manager proposal

@zeeke 

TBH, I think we should try to use the single manager and relie on cancellation logic from the controller-runtime package. 
Recent versions of the controller-runtime package provides a way to configure cache/watches in a very granular way. It is possible to watch some resources in all namespace and other only in specific namespace.

Here is example:
https://github.com/Mellanox/nvidia-k8s-ipam/blob/a34b4a2547815b7c016df7d563a2e04000f0add3/cmd/ipam-node/app/app.go#L146

ykulazhenkov (commented):
> > Hi @zeeke @adrianchiris not related to this PR directly but question do you know why we have two manges?? and now 3
> 
> AFAIU, mgr watches resource in the operator's namespace. mgrGlobal watches every namespace, as it runs Sriov[IB}Network reconcilers which react on NetworkAttachmentDefinitions.
> 
> I guess we can unify those managers with one that watches all namespaces, but maybe we get some performance degradation.
> 
> Also, I'm investigating about the shutdown logic to see if we can get rid of some bits and pieces, as I'm not very happy with the current election-proxy-manager proposal

@zeeke 

TBH, I think we should try to use the single manager and rely on cancellation logic from the controller-runtime package. 
Recent versions of the controller-runtime package provides a way to configure cache/watches in a very granular way. It is possible to watch some resources in all namespace and other only in specific namespace.

Here is example:
https://github.com/Mellanox/nvidia-k8s-ipam/blob/a34b4a2547815b7c016df7d563a2e04000f0add3/cmd/ipam-node/app/app.go#L146 

WDYT?

zeeke:
> > > Hi @zeeke @adrianchiris not related to this PR directly but question do you know why we have two manges?? and now 3
> > 
> > 
> > AFAIU, mgr watches resource in the operator's namespace. mgrGlobal watches every namespace, as it runs Sriov[IB}Network reconcilers which react on NetworkAttachmentDefinitions.
> > I guess we can unify those managers with one that watches all namespaces, but maybe we get some performance degradation.
> > Also, I'm investigating about the shutdown logic to see if we can get rid of some bits and pieces, as I'm not very happy with the current election-proxy-manager proposal
> 
> @zeeke
> 
> TBH, I think we should try to use the single manager and rely on cancellation logic from the controller-runtime package. Recent versions of the controller-runtime package provides a way to configure cache/watches in a very granular way. It is possible to watch some resources in all namespace and other only in specific namespace.
> 
> Here is example: https://github.com/Mellanox/nvidia-k8s-ipam/blob/a34b4a2547815b7c016df7d563a2e04000f0add3/cmd/ipam-node/app/app.go#L146
> 
> WDYT?

Hi @ykulazhenkov , Thank you for your feedback.

I agree with you we should simplify as much as we can. With your suggestion, we can merge the namespace and the non-namespaced manager to a single one, but it's not the problem this PR is trying to solve. 

The problem here is that the operator has a shutdown logic with two constraints:
a. it must run with the controllers turned off
b. it must run as a cluster leader, i.e. no other sriov-network-operators are supposed to do any logic during that shutdown logic

The cleaner solution I found (apart from getting rid of the shutdown logic, see [discussions/608](https://github.com/k8snetworkplumbingwg/sriov-network-operator/discussions/608)) is to add a new manager that controls only the leader election.

I admit it's probably not the best solution, but it solves an issue found in a production environment ([OCPBUGS-23795](https://issues.redhat.com/browse/OCPBUGS-23795))

ykulazhenkov:
> Hi @ykulazhenkov , Thank you for your feedback.
> 
> I agree with you we should simplify as much as we can. With your suggestion, we can merge the namespace and the non-namespaced manager to a single one, but it's not the problem this PR is trying to solve.
> 
> The problem here is that the operator has a shutdown logic with two constraints: a. it must run with the controllers turned off b. it must run as a cluster leader, i.e. no other sriov-network-operators are supposed to do any logic during that shutdown logic
> 
> The cleaner solution I found (apart from getting rid of the shutdown logic, see [discussions/608](https://github.com/k8snetworkplumbingwg/sriov-network-operator/discussions/608)) is to add a new manager that controls only the leader election.
> 
> I admit it's probably not the best solution, but it solves an issue found in a production environment ([OCPBUGS-23795](https://issues.redhat.com/browse/OCPBUGS-23795))

thx, for the clarification

ykulazhenkov (approved):
LGTM

SchSeba (approved):
LGTM

Thanks!

PR #560, opened 2023-12-15
Ignore nil objects occured when rendering a file

vasrem:
This commits addresses an edge case where a nil object can occur while rendering a file. In that situation, in one of the places the renderer is used, it will try to add ownership to a nil object and fail.

### Testing

Before:
```
$ go test ./pkg/render/ -run TestTemplateWithEmptyObject
--- FAIL: TestTemplateWithEmptyObject (0.00s)
    render_test.go:101:
        Expected
            <int>: 3
        to equal
            <int>: 2
FAIL
FAIL    github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/render       0.019s
FAIL
```

### DIFF
--- a/pkg/render/render.go
+++ b/pkg/render/render.go
@@ -103,6 +103,11 @@ func RenderTemplate(path string, d *RenderData) ([]*unstructured.Unstructured, e
 			}
 			return nil, errors.Wrapf(err, "failed to unmarshal manifest %s", path)
 		}
+
+		if u.Object == nil {
+			continue
+		}
+
 		out = append(out, &u)
 	}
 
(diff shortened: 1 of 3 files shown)


zeeke (approved):
nit: The problem this PR fixes occurs even without the dash (`-`):
i.e.:
```yaml
{{ if .Enable }}
---
...
```

as the problem is related to the `\n` before `---`, which ends up in a nil unstructured object.

BTW, the fix would remain the same. thanks for tackling this.
LGTM

adrianchiris (approved):
LGTM!

PR #565, opened 2023-12-27
Update golang to 1.21

bn222:
Bump golang to 1.21

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,10 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.20
+    - name: Set up Go 1.21
       uses: actions/setup-go@v3
       with:
-        go-version: 1.20.x
+        go-version: 1.21.x
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
@@ -32,10 +32,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.20
+    - name: Set up Go 1.21
       uses: actions/setup-go@v2
       with:
-        go-version: 1.20.x
+        go-version: 1.21.x
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
@@ -53,10 +53,10 @@ jobs:
     name: Golangci-lint
     runs-on: ubuntu-latest
     steps:
-    - name: Set up Go 1.20
+    - name: Set up Go 1.21
       uses: actions/setup-go@v2
       with:
-        go-version: 1.20.x
+        go-version: 1.21.x
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
     - name: golangci-lint
@@ -70,10 +70,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.20
+    - name: Set up Go 1.21
       uses: actions/setup-go@v2
       with:
-        go-version: 1.20.x
+        go-version: 1.21.x
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
@@ -107,10 +107,10 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
 
-      - name: Set up Go 1.20
+      - name: Set up Go 1.21
         uses: actions/setup-go@v3
         with:
-          go-version: 1.20.x
+          go-version: 1.21.x
 
       - name: run test
         run: make test-e2e-conformance-virtual-k8s-cluster-ci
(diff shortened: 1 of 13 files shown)


bn222:
@SchSeba @adrianchiris @e0ne PTAL

go.mod
```diff
@@ -25,18 +25,18 @@ require (
...
-	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
-	sigs.k8s.io/controller-runtime v0.16.3
+	k8s.io/api v0.29.0
+	k8s.io/apiextensions-apiserver v0.29.0
+	k8s.io/apimachinery v0.29.0
+	k8s.io/client-go v0.29.0
+	k8s.io/code-generator v0.29.0
+	k8s.io/kubectl v0.29.0
+	k8s.io/utils v0.0.0-20231127182322-b307cd553661
+	sigs.k8s.io/controller-runtime v0.16.1-0.20231215020716-1b80b9629af8
```
adrianchiris, line 39: i dont like this. controller runtime is a pretty central pkg we should use release versions.

can this wait until proper release of controller runtime is made ?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/k8snetworkplumbingwg/sriov-network-operator
 
-go 1.20
+go 1.21
```
adrianchiris, line 3: do you need to update all relevant github action workflows as well ?

go.mod
bn222: I understand your concern but there is a deadline to get this in pretty soon. We don't control the release cadence of the controller-runtime. I'm sure they will tag a commit with a release version at some point in the future, but there is no way for us to control that easily.

I prefer to bump the controller runtime, and we can bump again when there is a release. If you want, I will take a note to watch for a release and then I'll bump it again when there is an official tag.

go.mod
SchSeba: yes we need

bn222: Agreed. I've bumped the version in GitHub actions and also in all Dockerfiles

go.mod
adrianchiris: alternative is to replace otel pkg, maybe less risky ?

bn222: Can you elaborate? Do you mean a replace section in go.mod?

SchSeba: yep that can be a good idea but before we go that wait I was looking in the code did you find a place where we use the opentelemetry?

even on the vendor folder I was not able to find a place where we call the function that is effected by the CVE

bn222: `go mod graph | grep otel@`

You can see that the library is pulled in indirectly. It's not that we use otel, rather one of our dependencies uses it

SchSeba: right but I searched in the code and was not able to find anything...
can you try to find it and see if we even use it.

maybe it's about metrics and we disable that I the operator

bn222: Since we don't have a stable version for controller-runtime, we have decided to not bump controller-runtime just yet. I will make this PR instead focus on bumping golang only.

adrianchiris:
@bn222 you will need to bump golangci-lint version as well , they added support for go 1.21 in ver 1.54

bn222:
Bumped to 1.55.2 (latest as of today)

.github/workflows/test.yml
```diff
@@ -53,10 +53,10 @@ jobs:
...
     steps:
-    - name: Set up Go 1.20
+    - name: Set up Go 1.21
       uses: actions/setup-go@v2
       with:
-        go-version: 1.20.x
+        go-version: 1.21.x
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
     - name: golangci-lint
```
adrianchiris, line 62: @bn222  need to bump golangci lint ver here as welll

bn222:
I think the new linter is more pedantic. I will wait until https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/553 merges, and then rebase on top to fix the remaining linter errors. 

@adrianchiris Unless I missed something fundamental.

adrianchiris:
@bn222 #553 merged ! we can proceed with updating this one 🎊

api/v1/helper.go
```diff
@@ -37,7 +37,8 @@ const (
 	SriovCniStateAuto    = "auto"
 	SriovCniStateOff     = "off"
 	SriovCniStateOn      = "on"
-	SriovCniIpamEmpty    = "\"ipam\":{}"
+	SriovCniIpam         = "\"ipam\""
+	SriovCniIpamEmpty    = SriovCniIpam + "{}"
```
adrianchiris, line 41: @bn222 i believe this is whats failing unit tests ...
you need `+ ":{}"`

bn222: ty for the catch.

bn222:
@adrianchiris ptal

adrianchiris (approved):

zeeke (approved):
LGTM

PR #566, opened 2023-12-27
Change behavior when deleting default SriovNetworkNodePolicy and SriovOperatorConfig

vasrem:
Opening this PR for discussion as it looks to me like the simplest way forward. I'm not sure how bad is it to delete the default CRs and if warning the user instead of blocking the user is sufficient.

---

This commit changes the behavior of the webhook when deleting the default `SriovNetworkNodePolicy` and `SriovOperatorConfig`. This change is needed so that we can smoothly uninstall Helm releases that have the webhooks enabled.

Without this change, when running `helm uninstall`, it fails because the webhook doesn't allow deletion of such resources mentioned above.

Since these Webhook Configurations resources are deployed via the controller itself and not Helm, it's much more difficult to handle the lifecycle of them via Helm in the current state. Instead, it's easier to send a warning that these resources should not be deleted.

### Testing

Used Case 5 of https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/561 and managed to install and uninstall without any issue. `helm uninstall` log looks like this:
```
...
client.go:310: [debug] Starting delete for "default" SriovNetworkNodePolicy
W1227 13:20:14.885788  123736 warnings.go:70] default SriovNetworkNodePolicy shouldn't be deleted
client.go:310: [debug] Starting delete for "default" SriovOperatorConfig
W1227 13:20:15.078219  123736 warnings.go:70] default SriovOperatorConfig shouldn't be deleted
...
```

### DIFF
--- a/pkg/webhook/validate.go
+++ b/pkg/webhook/validate.go
@@ -40,7 +40,7 @@ func validateSriovOperatorConfig(cr *sriovnetworkv1.SriovOperatorConfig, operati
 	}
 
 	if operation == v1.Delete {
-		return false, warnings, fmt.Errorf("default SriovOperatorConfig shouldn't be deleted")
+		warnings = append(warnings, "default SriovOperatorConfig shouldn't be deleted")
 	}
 
 	if cr.Spec.DisableDrain {
@@ -97,12 +97,11 @@ func validateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePolicy, o
 
 	if cr.GetName() == constants.DefaultPolicyName && cr.GetNamespace() == os.Getenv("NAMESPACE") {
 		if operation == v1.Delete {
-			// reject deletion of default policy
-			return false, warnings, fmt.Errorf("default SriovNetworkNodePolicy shouldn't be deleted")
-		} else {
-			// skip validating default policy
-			return true, warnings, nil
+			warnings = append(warnings, "default SriovNetworkNodePolicy shouldn't be deleted")
 		}
+
+		// skip validating default policy
+		return true, warnings, nil
 	}
 
 	if cr.GetNamespace() != os.Getenv("NAMESPACE") {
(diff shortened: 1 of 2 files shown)


adrianchiris:
in helm i see we only create the default policy so during un-install i expect just that to fail. cluster would remain with sriovOperatorConfig obj.

is that not the case ?

generally for both obj's the relevant controller will first re-create the default obj before proceeding with reconcile.
so i think its ok to just return warning and not block deletion.

as an alternative, we can create default obj via helm as post install and remove in pre delete hooks.

SchSeba:
Adding @zeeke as he is also working on some issues with the namespace been removed and objects gets stuck because the webhook pod is already deleted

zeeke (approved):
Not an helm expert, but I'm ok with deleting the default policy, as it gets recreated by the controller.

In theory, this problem (and others I'm having by deleting the operator's namespace) should have been addressed by the shutdown procedure:

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/shutdown.go

But it seems to not work for reasons that aren't very clear to me yet.

LGTM

vasrem:
@adrianchiris 

> in helm i see we only create the default policy so during un-install i expect just that to fail. cluster would remain with sriovOperatorConfig obj.
> 
> is that not the case ?

Yes that's the case. If Helm faces an error when deleting an object it manages, the helm release will keep lingering in the cluster in errored state. So, the uninstall is not clean and may leave objects lingering in the cluster.

> generally for both obj's the relevant controller will first re-create the default obj before proceeding with reconcile. so i think its ok to just return warning and not block deletion. (thus avoiding the problem in helm)

:+1:

> as an alternative, we can create default obj via helm as post install hook and remove in pre delete hooks.

I think the removal won't work in that case. We will need to bypass the webhook to do so, and this means either of the following (and there might be more):
* Delete webhook configuration on helm uninstall (not so easy because the webhook config is deployed by the operator itself, it's not managed by helm)
* Set `failurePolicy: Ignore` on the webhook configuration (again not so easy because the configuration is deployed by the operator itself)
* Extend logic to bypass webhook when custom annotation is set (too complicated IMO)

---

I think what this PR does is the simplest we can do today without too much of an impact. Especially if the operator itself will recreate the objects anyway.

adrianchiris:
> I think what this PR does is the simplest we can do today without too much of an impact. Especially if the operator itself will recreate the objects anyway.

I agree, i think the creation of these default objects is something we would need to discuss, but for now lets not tie between the two.

Im OK with this change.

adrianchiris (approved):

PR #567, opened 2023-12-27
Externally managed fixes

adrianchiris:
- block mtu change on PF when externally managed
- fix handling in mellanox plugin for externally managed interfaces

### DIFF
--- a/pkg/daemon/plugin.go
+++ b/pkg/daemon/plugin.go
@@ -82,7 +82,9 @@ func registerVendorPlugins(ns *sriovnetworkv1.SriovNetworkNodeState, helpers hel
 				log.Log.Error(err, "registerVendorPlugins(): failed to load plugin", "plugin-name", plug.Name())
 				return vendorPlugins, fmt.Errorf("registerVendorPlugins(): failed to load the %s plugin error: %v", plug.Name(), err)
 			}
-			vendorPlugins[plug.Name()] = plug
+			if _, ok := vendorPlugins[plug.Name()]; !ok {
+				vendorPlugins[plug.Name()] = plug
+			}
 		}
 	}
 
--- a/pkg/host/sriov.go
+++ b/pkg/host/sriov.go
@@ -343,6 +343,12 @@ func (s *sriov) ConfigSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *
 	}
 	// set PF mtu
 	if iface.Mtu > 0 && iface.Mtu > ifaceStatus.Mtu {
+		if iface.ExternallyManaged {
+			err := fmt.Errorf("ConfigSriovDevice(): requested MTU(%d) is greater than configured MTU(%d) for device %s. cannot change MTU as policy is configured as ExternallyManaged",
+				iface.Mtu, ifaceStatus.Mtu, iface.PciAddress)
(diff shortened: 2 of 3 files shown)


adrianchiris:
@SchSeba  PTAL

pkg/daemon/plugin.go
```diff
@@ -71,12 +71,20 @@ func enablePlugins(platform utils.PlatformType, useSystemdService bool, ns *srio
 	return enabledPlugins, nil
 }
 
-func registerVendorPlugins(ns *sriovnetworkv1.SriovNetworkNodeState) (map[string]plugin.VendorPlugin, error) {
+func registerVendorPlugins(ns *sriovnetworkv1.SriovNetworkNodeState, storeManager utils.StoreManagerInterface) (map[string]plugin.VendorPlugin, error) {
 	vendorPlugins := map[string]plugin.VendorPlugin{}
+	vendorSet := map[string]struct{}{}
 
+	// populate Set of vendors for plugin creation
```
adrianchiris, line 78: while at it, i added this to instantiate plugin only once per vendor, hope its ok

adrianchiris:
@SchSeba can we merge this one ? these fixes are much needed

pkg/daemon/plugin.go
ykulazhenkov: Why we need a separate set here? Is it possible to simply check if existing `vendorPlugins` map already contains a vendor name and skip plugin initialization in this case?

pkg/utils/utils.go
```diff
@@ -445,6 +445,12 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 	}
 	// set PF mtu
 	if iface.Mtu > 0 && iface.Mtu > ifaceStatus.Mtu {
+		if iface.ExternallyManaged {
```
mlguerrero12, line 448: Thanks for adding this. We won't rely only on the webhook anymore.

Just one question, this will prevent the configuration but how can we avoid `NeedUpdate` to return `true` and force a node drain when this is the only change? We will need a check there to prevent the draining.

adrianchiris: good question,

it seems we always need to drain in case of externally managed since even if VF are there we still need to set admin mac / node&port guid

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/b393f307f037b3bced2392e7dfbab3c85d18b372/pkg/utils/utils.go#L393

mlguerrero12: right, we could do it for the dpdk case though. Maybe in a future pr.

pkg/daemon/plugin.go
adrianchiris: going back to this change, 

vendorPlugins is keyed by plugin name and not by vendor.

to instantiate once we then need this set.
if we want just to set a plugin once in vendorPlugins map then we can do as u say.

seeing its not a really big deal, i will just set plugin in map once.

adrianchiris:
@ykulazhenkov  PTAL
@SchSeba  PTAL,  PR rebased and ready for review.

adrianchiris:
@zeeke  could you take a look ? :)

zeeke (approved):
LGTM

PR #563, opened 2023-12-25
image build target additions

adrianchiris:
- Add webhook to image build target
- use APP_NAME var in image name for consistency

PR #568, opened 2023-12-27
Extend Makefile to build webhook image

vasrem:
We already build the webhook under the `build` make target but we are not creating an image for it under the `image` make target. This PR introduces that.


adrianchiris:
covered in: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/563

PR #569, opened 2023-12-31
Add disable plugins

adrianchiris:
Add option to disable sriov network config daemon plugins

some plugins perform vendor specific firmware configuration
to enable SR-IOV (e.g `mellanox` plugin). certain deployment environments may prefer to perform such configuration
once during node provisioning, while ensuring the configuration will be compatible with any sriov network node policy
defined for the particular environment. This will reduce or completely eliminate the need for reboot of nodes during SR-IOV
configurations by the operator.

This can be done by setting SriovOperatorConfig `default` CR `spec.disablePlugins` with the list of desired plugins
to disable.


SchSeba:
general question do we want to have this as part of SriovOperatorConfig CR instead of another env variable in the operator? :)

adrianchiris:
> general question do we want to have this as part of SriovOperatorConfig CR instead of another env variable in the operator?

im not sure, i see this decision (deciding if/which plugins to disable) is made during deployment time and should not change.

moreover deleting sriov operator config CR will trigger creation of the default config which will have no plugins disabled by default which may be problematic.

SchSeba:
I am not fully familiar with helm but can't we request a CR also when we do the helm install?

bn222:
> im not sure, i see this decision (deciding if/which plugins to disable) is made during deployment time and should not change.

Even if the decision is not made, I also prefer to have it be part of the CR since that would cover more use cases. It feels cleaner.

adrianchiris:
>  I also prefer to have it be part of the CR since that would cover more use cases.

what other use-cases ?

bn222:
> > I also prefer to have it be part of the CR since that would cover more use cases.
> 
> what other use-cases ?

Suppose the customer decides that they want to disable some plugin, then at a later point in time, they change their mind.

adrianchiris:
> Suppose the customer decides that they want to disable some plugin, then at a later point in time, they change their mind.

then IMO they need to redeploy. our (NVIDIA) use case that it should not change, moreover accidently changing it may have catastrophic effect to workers (firmware parameters resetting).

my concern is creating the default operator config obj in the operator. if you are ok with dropping its creation, we can set this in operator config CR.

we discussed this in community meeting. Sebastian needs to double check internally.

deployment/sriov-network-operator/README.md
```diff
@@ -69,6 +69,12 @@ We have introduced the following Chart parameters.
 | `operator.cniBinPath` | string | `/opt/cni/bin` | Path for CNI binary |
 | `operator.clusterType` | string | `kubernetes` | Cluster environment type |
 
+### SR-IOV Config Daemon parameters
+
+| Name | Type | Default | description |
+| ---- | ---- | ------- | ----------- |
+| `sriovConfigDaemon.disablePlugins` | []string | `nil` | List of plugins to disable, currently only `mellanox` plugin is allowed |
```
mlguerrero12, line 76: shouldn't we add "use-systemd" now that you're adding this?

deployment/sriov-network-operator/values.yaml
```diff
@@ -29,6 +29,11 @@ operator:
   cniBinPath: "/opt/cni/bin"
   clusterType: "kubernetes"
 
+#sriovConfigDaemon:
+#  disablePlugins:
+#    - mellanox
+#    - intel
```
mlguerrero12, line 35: remove intel?

deployment/sriov-network-operator/templates/operator.yaml
```diff
@@ -90,3 +90,7 @@ spec:
               value: {{ .Values.operator.cniBinPath }}
             - name: CLUSTER_TYPE
               value: {{ .Values.operator.clusterType }}
+            {{- if and .Values.sriovConfigDaemon .Values.sriovConfigDaemon.disablePlugins }}
```
mlguerrero12, line 93: shouldn't the "and" be between the values? or does this work?

pkg/daemon/plugin.go
```diff
@@ -29,61 +29,91 @@ var (
...
 			return nil, err
 		}
-		enabledPlugins[virtualPlugin.Name()] = virtualPlugin
+		pluginName := virtualPlugin.Name()
+		if !isPluginDisabled(pluginName, disabledPlugins) {
+			loadedPlugins[pluginName] = virtualPlugin
+		}
 	} else {
-		enabledVendorPlugins, err := registerVendorPlugins(ns)
+		loadedVendorPlugins, err := loadVendorPlugins(ns, disabledPlugins)
```
mlguerrero12, line 53: I think we should only allow to disable vendor plugins (the ones we loaded with the `loadVendorPlugins`). The others are essential for the daemon to work. Virtual for openstack and generic for baremetal. It doesn't make sense to allow users to disable them, unless you see a valid use case.

deployment/sriov-network-operator/README.md
adrianchiris: im not sure, that one can be configured during runtime.
this one is during deployment time.

cmd/sriov-network-config-daemon/start.go
```diff
@@ -81,6 +106,14 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
 		startOpts.nodeName = name
 	}
 
+	// TODO(adrianc): move allowed disable plugins to globals pkg
+	allowedDisablePlugins := map[string]struct{}{"mellanox": {}}
+	for _, p := range startOpts.disabledPlugins {
+		if _, ok := allowedDisablePlugins[p]; !ok {
+			return fmt.Errorf("%s plugin is cannot be disabled", p)
```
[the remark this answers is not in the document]
adrianchiris: ack

deployment/sriov-network-operator/values.yaml
adrianchiris: ack

pkg/daemon/plugin.go
adrianchiris: > It doesn't make sense to allow users to disable them, unless you see a valid use case.

ATM this is blocked at API level (user can only disable a single plugin atm), from code perspective, it seems easier to make reason about. 

i.e daemon calls "loadPlugins" and it has an arg with a list of which plugins are disabled.

possible use-case for disabling generic plugin - all sriov configurations are preconfigured and we just want the operator to handle "the rest"
i.e device plugin and its configuration,  cni deployment, net-attach-def creation

that said i dont have such a use-case atm. my main argument is less cognitive complexity.

mlguerrero12: fair enough

deployment/sriov-network-operator/README.md
mlguerrero12: Right, didn't notice that this readme is for the helm chart

cmd/sriov-network-config-daemon/start.go
```diff
@@ -81,6 +106,14 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
 		startOpts.nodeName = name
 	}
 
+	// TODO(adrianc): move allowed disable plugins to globals pkg
+	allowedDisablePlugins := map[string]struct{}{"mellanox": {}}
```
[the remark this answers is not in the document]
bn222: What about `deactivatablePlugins` or `activatablePlugins` if we switch to a positive instead (which is maybe better to avoid double negatives and have shorter names throughout).

README.md
```diff
@@ -228,6 +228,23 @@ This feature was created to support deployments where the user want to use some
...
+is not needed or un-desirable.
+
+As an example, some plugins perform vendor specific firmware configuration
+to enable SR-IOV (e.g `mellanox` plugin). certain deployment environments may prefer to perform such configuration
+once during node provisioning, while ensuring the configuration will be compatible with any sriov network node policy
+defined for the particular environment. This will reduce or completely eliminate the need for reboot of nodes during SR-IOV
+configurations by the operator.
+
+This can be done by providing the names of plugins to disable as a comma separated list of strings via
+`SRIOV_NETWORK_CONFIG_DAEMON_DISABLE_PLUGINS` environment variable to the operator deployment or using
```
e0ne, line 243: nit: I don't know what will be better, maybe `_DISABLED_PLUGINS`

README.md
```diff
@@ -228,6 +228,23 @@ This feature was created to support deployments where the user want to use some
...
+It is possible to disable SR-IOV network operator config daemon plugins in case their operation
+is not needed or un-desirable.
+
+As an example, some plugins perform vendor specific firmware configuration
+to enable SR-IOV (e.g `mellanox` plugin). certain deployment environments may prefer to perform such configuration
+once during node provisioning, while ensuring the configuration will be compatible with any sriov network node policy
+defined for the particular environment. This will reduce or completely eliminate the need for reboot of nodes during SR-IOV
+configurations by the operator.
+
+This can be done by providing the names of plugins to disable as a comma separated list of strings via
```
e0ne, line 242: nit: "comma-separated" here and below

cmd/sriov-network-config-daemon/start.go
```diff
@@ -66,6 +90,7 @@ func init() {
 	startCmd.PersistentFlags().StringVar(&startOpts.kubeconfig, "kubeconfig", "", "Kubeconfig file to access a remote cluster (testing only)")
 	startCmd.PersistentFlags().StringVar(&startOpts.nodeName, "node-name", "", "kubernetes node name daemon is managing")
 	startCmd.PersistentFlags().BoolVar(&startOpts.systemd, "use-systemd-service", false, "use config daemon in systemd mode")
+	startCmd.PersistentFlags().VarP(&startOpts.disabledPlugins, "disable-plugins", "", "comma separated list of plugins")
```
e0ne, line 93: comma separated => comma-separated

e0ne:
> my concern is creating the default operator config obj in the operator. if you are ok with dropping its creation, we can set this in operator config CR.

+1 on this. this option isn't supposed to be changed during runtime

cmd/sriov-network-config-daemon/start.go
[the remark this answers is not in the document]
adrianchiris: addressed, will upload PR soon

README.md
adrianchiris: i prefer to keep it the same name as the flag (disable-plugins)

cmd/sriov-network-config-daemon/start.go
adrianchiris: done

adrianchiris:
@SchSeba PTAL

mlguerrero12:
please rebase

LGTM

api/v1/sriovoperatorconfig_types.go
```diff
@@ -45,6 +45,8 @@ type SriovOperatorConfigSpec struct {
 	ConfigurationMode ConfigurationModeType `json:"configurationMode,omitempty"`
 	// Flag to enable Container Device Interface mode for SR-IOV Network Device Plugin
 	UseCDI bool `json:"useCDI,omitempty"`
+	// DisablePlugins is a list of sriov-network-config-daemon plugins to disable
+	DisablePlugins []string `json:"disablePlugins,omitempty"`
```
mlguerrero12, line 49: you can add a validation at the api level

	// +kubebuilder:validation:Enum={"mellanox"}

README.md
adrianchiris: done.

pkg/daemon/plugin_test.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+var _ = Describe("config daemon plugin loading tests", func() {
+	Context("loadPlugins", func() {
+		var gmockController *gomock.Controller
+		var helperMock *helperMocks.MockHostHelpersInterface
+		prevK8sPluginFn := K8sPlugin
+
+		BeforeEach(func() {
+			vars.ClusterType = consts.ClusterTypeKubernetes
```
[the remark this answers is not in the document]
adrianchiris: ack, let me do it in AfterEach statement.

api/v1/sriovoperatorconfig_types.go
adrianchiris: ack. thx

adrianchiris: tried this out, it doesnt work for lists just for scalar values

`spec.disablePlugins: Unsupported value: []interface {}{"mellanox"}: supported values: "mellanox"`

mlguerrero12: I see, perhaps this https://github.com/kubernetes-sigs/kubebuilder/issues/2812?

adrianchiris: works. PR updated.

mlguerrero12 (approved):
LGTM

Failing tests are not related. They are also failing in other PRs. 
This seems to help https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/600

adrianchiris:
rebased. @zeeke PTAL :)

zeeke (approved):
LGTM

PR #574, opened 2024-01-04
Add multi-arch docker image support

cyclinder:
support amd64, arm64 and ppc64le platform for sriov-network-operator, sriov-network-operator-config-daemon and sriov-network-operator-webhook.

Fixes https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/430

PR #577, opened 2024-01-08
Enable multi-arch image builds

e0ne:


e0ne:
> 

yes, I missed #574 :(

PR #578, opened 2024-01-09
Use go-install-tool instead of go-get-tool for gomock

ykulazhenkov:
`go-get-tool` is not available in the Makefile


ykulazhenkov:
@adrianchiris  @SchSeba @e0ne PTAL

zeeke (approved):
LGTM

adrianchiris (approved):

PR #581, opened 2024-01-10
Use default `{ib-}sriov-cni` entrypoint script

zeeke:
Avoid overriding entrypoint script for sriov-cni and ib-sriov-cni. It is more robust to rely on the image logic, as the source image can change the path where the binary is stored.

Move the copy logic back to the `containers:` section, as the image's entrypoint script ends with a "sleep forever" [1].

Mount `/etc/os-release` file to let the copy logic to make decision on which binary it should copy depending on the operating system.

refs:
[1] https://github.com/k8snetworkplumbingwg/sriov-cni/blob/07ecbf9b6ee867054df47423e51a800034ef4e05/images/entrypoint.sh#L63

cc @SchSeba


mlguerrero12:
LGTM

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -123,10 +93,42 @@ spec:
...
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            cpu: 10m
+            memory: 10Mi
+        volumeMounts:
+        - name: cnibin
+          mountPath: /host/opt/cni/bin
+        - name: os-release
```
adrianchiris, line 107: who is using this mount ? sriov cni entrypoint does not use it AFAIK

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -123,10 +93,42 @@ spec:
...
```
adrianchiris, line 121: same Q as above :)

adrianchiris:
> Move the copy logic back to the containers: section, as the image's entrypoint script ends with a "sleep forever" [1].

we could extend sriov-cni and ib-sriov-cni with a `--oneshot` flag which will avoid the sleep. WDYT ?
then we can keep in init containers

zeeke:
> we could extend sriov-cni and ib-sriov-cni with a `--oneshot` flag which will avoid the sleep. WDYT ?
> then we can keep in init containers

I like that! Going to file a PR in sriov-cni and ib-sriov-cni.

Regarding the `/etc/os-release`, I need it to implement an openshift only change to support running the cni on multiple RHEL versions. I'm trying to keep the community changes as low as possible: the idea is to pass that file to the cni image entrypoint, where there can be multiple sriov-cni binaries. Then a script will copy the right binary to the host.

This logic is also present in other multus cni plugin:
https://github.com/openshift/cluster-network-operator/blob/873b3290a2de28fb7fb41ad498fc9872dcb1a93b/bindata/network/multus/multus.yaml#L13

zeeke:
/hold

this PR depends on:
- https://github.com/k8snetworkplumbingwg/sriov-cni/pull/289
- https://github.com/k8snetworkplumbingwg/ib-sriov-cni/pull/83

zeeke:
There's still a problem with `ib-sriov-cni`. Waiting for 
- https://github.com/k8snetworkplumbingwg/ib-sriov-cni/pull/84

zeeke:
/hold cancel

related work on sriov-cni and ib-sriov-cni has been merged

SchSeba (approved):
nice work!

LGTM

adrianchiris (commented):
overall LGTM,

can you amend the commit msg.

i think this part is no longer true:
```
Move the copy logic back to the `containers:` section, as the image's
entrypoint script ends with a "sleep forever" [1].
```

we keep copy of cni bin in initContainer still, the cni image entrypoint now supports --no-sleep flag which allows it to function as init container.

adrianchiris (approved):

adrianchiris:
@zeeke feel free to remove the hold label and merge this one, thx for working on it !

PR #583, opened 2024-01-10
[switchdev 2/x] API: Add vdpaType field to the node SriovNetworkNodeState CRD

ykulazhenkov:
This field should be used to report information about type of the VDPA device that is configured for VF.
Empty string means that there is no VDPA device.

Valid values are: `virtio`, `vhost` (same as in `SriovNetworkNodePolicySpec`)

Context:

- part of [switchdev refactoring](https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/559)

### DIFF
--- a/api/v1/sriovnetworknodestate_types.go
+++ b/api/v1/sriovnetworknodestate_types.go
@@ -82,6 +82,7 @@ type VirtualFunction struct {
 	Vlan       int    `json:"Vlan,omitempty"`
 	Mtu        int    `json:"mtu,omitempty"`
 	VfID       int    `json:"vfID"`
+	VdpaType   string `json:"vdpaType,omitempty"`
 }
 
 // SriovNetworkNodeStateStatus defines the observed state of SriovNetworkNodeState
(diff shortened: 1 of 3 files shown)


api/v1/sriovnetworknodestate_types.go
```diff
@@ -82,6 +82,7 @@ type VirtualFunction struct {
 	Vlan       int    `json:"Vlan,omitempty"`
 	Mtu        int    `json:"mtu,omitempty"`
 	VfID       int    `json:"vfID"`
+	VdpaType   string `json:"vdpaType,omitempty"`
```
adrianchiris, line 85: i guess we should hold off merging this until we got some logic to populate this field ?

ykulazhenkov: I will submit commits which relies on this change soon. Also this field is a part of internal API which is used only by the operator itself.

adrianchiris: ack.

adrianchiris (approved):

zeeke (approved):

PR #586, opened 2024-01-16
Change layout of the pkg/host to simplify testing

ykulazhenkov:
The commit includes the following changes:
* move interfaces and public types(types that used in interfaces)
to a separate package
* move implementation of the interfaces to internal pkgs to avoid
unwanted imports by users
* move StoreManager to a separate package

These changes are required to generate a mock for HostManagerInterface in the host package, which can be used to test other interfaces in the host package.

Without this change, it may be problematic to use mocks for interfaces from the host package to test other interface implementations in the host package because of the cyclic imports.

Other benefits:
* code for interfaces is more isolated
* less shared variables
* we have a strict contract - everything in the "internal" package is an implementation detail of the package and can't be imported by the user ("internal" package will protect from this)


ykulazhenkov:
@SchSeba  @adrianchiris @ykulazhenkov PTAL at this proposal

adrianchiris (approved):

adrianchiris:
Thx @ykulazhenkov !

@zeeke could you take a look on this one ?

pkg/host/types/interfaces.go
```diff
@@ -0,0 +1,161 @@
+package types
```
zeeke, line 1: not an expert on MockGen: does it make things any worse to keep these interfaces in `pkg/host` package? Whether in host.go or in a different file for each interface

pkg/host/store/store.go
```diff
@@ -1,4 +1,4 @@
-package host
+package store
```
zeeke, line 1: nit: either put the keyworkd "store" in the package name or in the interface name.

zeeke (commented):
Left a few minor comments. Though this PR left the code tidier than it was, I put a strong +1 on the way the StoreManagerInterface has been refactored rather than using and `internal`  package.

IDK if it can create any trouble (hence not blocking this PR for this), but a layout like:

```
host/
  manager.go  
            with HostManagerInterface superinterface, maybe without the Host prefix
            NewManager func
  
  kernel/
      kernel.go
            public `Interface` type
            private `kernel` struct (as already it is)
      kernel_test.go
            test for the implementation using other actor's mocks
      mock/
            mock_kernel.go

     // same for network, udev, sriov, service
```
    
WDYT?

ykulazhenkov:
> Left a few minor comments. Though this PR left the code tidier than it was, I put a strong +1 on the way the StoreManagerInterface has been refactored rather than using and `internal` package.

With the layout you propose it will be possible to create an instance of the Kernel interface by the code outside of the host pkg because the kernel.go - New() function will be public. If this is not a problem, then I believe we can use the layout which you proposed. WDYT?

ykulazhenkov:
> > Left a few minor comments. Though this PR left the code tidier than it was, I put a strong +1 on the way the StoreManagerInterface has been refactored rather than using and `internal` package.
> 
> @zeeke With the layout you propose it will be possible to create an instance of the Kernel interface by the code outside of the host pkg because the kernel.go - New() function will be public. If this is not a problem, then I believe we can use the layout which you proposed. WDYT?

+ with your layout we will not be able to use "self mock", e.g. use mock for KernelInterface inside the kernel package. "self mock" can be very useful to tests stucts which have many methods which call each other. e.g. if we want to mock `sriov.SetSriovNumVfs` and `sriov.VFIsReady` while testing  `sriov.ConfigSriovDevice`

pkg/host/store/store.go
ykulazhenkov: renamed store.StoreManagerInterface > store.ManagerInterface

pkg/host/types/interfaces.go
ykulazhenkov: In this case package with the interface implementation (e.g. kernel pkg) will not be able to import interface because of cyclic import.

manager (part of the host pkg) -> imports implementations
implementations -> imports interfaces

We have two options:
- move types + interfaces out of the `host` pkg
- move manager out of the `host` pkg

I decided to move types + interfaces.

ykulazhenkov:
I pushed the update that renames `store.StoreManagerInterface` to `store.ManagerInterface`.

pkg/host/types/interfaces.go
zeeke: ok, thanks for elaborating that. Makes sense to leave the `types/interfaces.go` file

zeeke:
> With the layout you propose it will be possible to create an instance of the Kernel interface by the code outside of the host pkg because the kernel.go - New() function will be public. If this is not a problem, then I believe we can use the layout which you proposed. WDYT?

I think it's ok. That actor would depend on an interface, and in the end, it would be better to have components to depend on the specific interface (like the Kernel one) rather than the super interface (interface segregation principle).

> ++ with your layout we will not be able to use "self mock", e.g. use mock for KernelInterface inside the kernel package. "self mock" can be very useful to tests stucts which have many methods which call each other. e.g. if we want to mock sriov.SetSriovNumVfs and sriov.VFIsReady while testing sriov.ConfigSriovDevice. I would prefer to have this option

Ok, that's a good reason to go with your proposed layout. 

LGTM, thanks for addressing my comments

zeeke (approved):

PR #587, opened 2024-01-17
Add more checks on generic plugin to discover discrepancies from the desired state

vasrem:
## Summary 

This PR does the following things:
* Adds additional test coverage for `OnNodeStateChange()` function to indicate when we are draining the node
* Adjusts `NeedToUpdateSriov()` helper function to ensure that:
  * VF MAC Address is set when we are in Ethernet mode
  * VF GUID is set when we are in Ethernet mode and RDMA is enabled
  * VF GUID is set when we are in Infiniband mode
  * PF Link is up
* (API Change) Adds necessary fields in the `SriovNetworkNodeState` struct to facilitate the above

## Partial Reconciliation

Today we reconcile on `.spec` change since we skip reconciliation if the generation of the object is the same as the object that was last reconciled successfully. Therefore, changes in `.status`, like what we try to detect with that PR, are not reconciled unless the `.spec` changes when the config daemon runs continuously (i.e. no restart). On daemon restart, the `.status` field will be reconciled the first time until the last successfully reconciled generation is saved in memory.

## Decisions

* Decided to add tests on the `OnNodeStateChange()` instead of `NeedToUpdateSriov()` because that one looks to be the most impactful for the whole system (i.e. decides whether to drain). I can add on both `Apply()` and `OnNodeStateChange()` if we choose to split the `NeedToUpdateSriov()` (see open questions below).
* Decided to just look for Node GUID and not the Port GUID since we expect both of them to be populated together on bind/unbind. This is the case because we parse the GUID from the RDMA link, and the node GUID is populated there on bind/unbind.

## Open questions

* [ ] Is it enough to partially reconcile or do we need to change the logic to take into account changes on `.status` (i.e. full reconciliation of changes that the controller does to the system)? (relatively big change in the operator behaviour I suppose)
* [ ] Do we need to drain on any of the discrepancies we find? (e.g. link is not up). If not, we will need to adjust the function that is used in `Apply()` and `OnNodeStateChange()` to be something different than `NeedToUpdateSriov()`.
* [ ] Are there additional end to end tests to be added? I would appreciate if you can point out the place I should be putting those (if we want to test end to end).
* [ ] Do we prefer getting info by reading directly `/sys/class/net/*` or use netlink? I see both approaches:
  * https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/09c987aaa8e5b1913f3306b3f889209e06adb8a1/pkg/host/network.go#L195-L205
  * https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/09c987aaa8e5b1913f3306b3f889209e06adb8a1/pkg/host/sriov.go#L618-L635
  Based on the answer I will use or discard https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/587/commits/d18aae1e1e975cdf9b5fca18172a1d0cafc9ca0e

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -273,7 +273,7 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 		return true
 	}
 
-	if ifaceStatus.LinkAdminState == "down" {
+	if ifaceStatus.LinkAdminState == consts.LinkAdminStateDown {
 		log.V(2).Info("NeedToUpdateSriov(): PF link status needs update", "desired to include", "up", "current", ifaceStatus.LinkAdminState)
 		return true
 	}
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -272,6 +272,12 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 		log.V(2).Info("NeedToUpdateSriov(): NumVfs needs update", "desired", ifaceSpec.NumVfs, "current", ifaceStatus.NumVfs)
 		return true
 	}
+
+	if ifaceStatus.LinkAdminState == "down" {
+		log.V(2).Info("NeedToUpdateSriov(): PF link status needs update", "desired to include", "up", "current", ifaceStatus.LinkAdminState)
+		return true
+	}
+
 	if ifaceSpec.NumVfs > 0 {
 		for _, vfStatus := range ifaceStatus.VFs {
 			ingroup := false
@@ -301,6 +307,16 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 							return true
 						}
 
+						if (strings.EqualFold(ifaceStatus.LinkType, consts.LinkTypeETH) && groupSpec.IsRdma) || strings.EqualFold(ifaceStatus.LinkType, consts.LinkTypeIB) {
+							// We do this check only if a Node GUID is set to ensure that we were able to read the
+							// Node GUID. We intentionally skip empty Node GUID in vfStatus because this may happen
+							// when the VF is allocated to a workload.
+							if vfStatus.GUID == consts.UninitializedNodeGUID {
+								log.V(2).Info("NeedToUpdateSriov(): VF GUID needs update",
+									"vf", vfStatus.VfID, "current", vfStatus.GUID)
+								return true
+							}
+						}
 						// this is needed to be sure the admin mac address is configured as expected
 						if ifaceSpec.ExternallyManaged {
 							log.V(2).Info("NeedToUpdateSriov(): need to update the device as it's externally manage",
--- a/api/v1/sriovnetworknodestate_types.go
+++ b/api/v1/sriovnetworknodestate_types.go
@@ -64,6 +64,7 @@ type InterfaceExt struct {
 	NumVfs            int               `json:"numVfs,omitempty"`
 	LinkSpeed         string            `json:"linkSpeed,omitempty"`
 	LinkType          string            `json:"linkType,omitempty"`
+	LinkAdminState    string            `json:"linkAdminState,omitempty"`
 	EswitchMode       string            `json:"eSwitchMode,omitempty"`
 	ExternallyManaged bool              `json:"externallyManaged,omitempty"`
 	TotalVfs          int               `json:"totalvfs,omitempty"`
@@ -84,6 +85,7 @@ type VirtualFunction struct {
 	VfID            int    `json:"vfID"`
 	VdpaType        string `json:"vdpaType,omitempty"`
 	RepresentorName string `json:"representorName,omitempty"`
+	GUID            string `json:"guid,omitempty"`
 }
 
 // Bridges contains list of bridges
--- a/pkg/consts/constants.go
+++ b/pkg/consts/constants.go
@@ -43,6 +43,9 @@ const (
 	LinkTypeIB  = "IB"
 	LinkTypeETH = "ETH"
 
+	LinkAdminStateUp   = "up"
+	LinkAdminStateDown = "down"
+
 	UninitializedNodeGUID = "0000:0000:0000:0000"
 
 	DeviceTypeVfioPci   = "vfio-pci"
--- a/pkg/consts/constants.go
+++ b/pkg/consts/constants.go
@@ -43,6 +43,8 @@ const (
 	LinkTypeIB  = "IB"
 	LinkTypeETH = "ETH"
 
+	UninitializedNodeGUID = "0000:0000:0000:0000"
+
 	DeviceTypeVfioPci   = "vfio-pci"
 	DeviceTypeNetDevice = "netdevice"
 	VdpaTypeVirtio      = "virtio"
--- a/pkg/host/internal/network/network.go
+++ b/pkg/host/internal/network/network.go
@@ -1,7 +1,10 @@
 package network
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
+	"net"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -187,6 +190,35 @@ func (n *network) GetNetDevMac(ifaceName string) string {
 	return link.Attrs().HardwareAddr.String()
 }
 
+// GetNetDevNodeGUID returns the network interface node GUID if device is RDMA capable otherwise returns empty string
+func (n *network) GetNetDevNodeGUID(pciAddr string) string {
+	if len(pciAddr) == 0 {
+		return ""
+	}
+
+	rdmaDevicesPath := filepath.Join(vars.FilesystemRoot, consts.SysBusPciDevices, pciAddr, "infiniband")
+	rdmaDevices, err := os.ReadDir(rdmaDevicesPath)
+	if err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			log.Log.Error(err, "GetNetDevNodeGUID(): failed to read RDMA related directory", "pciAddr", pciAddr)
+		}
+		return ""
+	}
+
+	if len(rdmaDevices) != 1 {
+		log.Log.Error(err, "GetNetDevNodeGUID(): expected just one RDMA device", "pciAddr", pciAddr, "numOfDevices", len(rdmaDevices))
+		return ""
+	}
+
+	rdmaLink, err := n.netlinkLib.RdmaLinkByName(rdmaDevices[0].Name())
+	if err != nil {
+		log.Log.Error(err, "GetNetDevNodeGUID(): failed to get RDMA link", "pciAddr", pciAddr)
+		return ""
+	}
+
+	return rdmaLink.Attrs.NodeGuid
+}
+
 func (n *network) GetNetDevLinkSpeed(ifaceName string) string {
 	log.Log.V(2).Info("GetNetDevLinkSpeed(): get LinkSpeed", "device", ifaceName)
 	speedFilePath := filepath.Join(vars.FilesystemRoot, consts.SysClassNet, ifaceName, "speed")
@@ -328,4 +360,24 @@ func (n *network) EnableHwTcOffload(ifaceName string) error {
 	}
 	log.Log.V(0).Info("EnableHwTcOffload(): feature is still disabled, not supported by device", "device", ifaceName)
 	return nil
+}
+
+// GetNetDevLinkAdminState returns the admin state of the interface.
+func (n *network) GetNetDevLinkAdminState(ifaceName string) string {
+	log.Log.V(2).Info("GetNetDevLinkAdminState(): get LinkAdminState", "device", ifaceName)
+	if len(ifaceName) == 0 {
+		return ""
+	}
+
+	link, err := n.netlinkLib.LinkByName(ifaceName)
+	if err != nil {
+		log.Log.Error(err, "GetNetDevLinkAdminState(): failed to get link", "device", ifaceName)
+		return ""
+	}
+
+	if link.Attrs().Flags&net.FlagUp == 0 {
+		return "down"
+	}
+
+	return "up"
 }
--- a/pkg/host/internal/network/network.go
+++ b/pkg/host/internal/network/network.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
-	"net"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -375,9 +374,9 @@ func (n *network) GetNetDevLinkAdminState(ifaceName string) string {
 		return ""
 	}
 
-	if link.Attrs().Flags&net.FlagUp == 0 {
-		return "down"
+	if n.netlinkLib.IsLinkAdminStateUp(link) {
+		return consts.LinkAdminStateUp
 	}
 
-	return "up"
+	return consts.LinkAdminStateDown
 }
--- a/pkg/host/internal/sriov/sriov.go
+++ b/pkg/host/internal/sriov/sriov.go
@@ -3,6 +3,7 @@ package sriov
 import (
 	"errors"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -134,6 +135,7 @@ func (s *sriov) GetVfInfo(pciAddr string, devices []*ghw.PCIDevice) sriovnetwork
 			vf.Mac = link.Attrs().HardwareAddr.String()
 		}
 	}
+	vf.GUID = s.networkHelper.GetNetDevNodeGUID(pciAddr)
 
 	for _, device := range devices {
 		if pciAddr == device.Address {
@@ -259,15 +261,16 @@ func (s *sriov) DiscoverSriovDevices(storeManager store.ManagerInterface) ([]sri
 		}
 
 		iface := sriovnetworkv1.InterfaceExt{
-			Name:       pfNetName,
-			PciAddress: device.Address,
-			Driver:     driver,
-			Vendor:     device.Vendor.ID,
-			DeviceID:   device.Product.ID,
-			Mtu:        link.Attrs().MTU,
-			Mac:        link.Attrs().HardwareAddr.String(),
-			LinkType:   s.encapTypeToLinkType(link.Attrs().EncapType),
-			LinkSpeed:  s.networkHelper.GetNetDevLinkSpeed(pfNetName),
+			Name:           pfNetName,
+			PciAddress:     device.Address,
+			Driver:         driver,
+			Vendor:         device.Vendor.ID,
+			DeviceID:       device.Product.ID,
+			Mtu:            link.Attrs().MTU,
+			Mac:            link.Attrs().HardwareAddr.String(),
+			LinkType:       s.encapTypeToLinkType(link.Attrs().EncapType),
+			LinkSpeed:      s.networkHelper.GetNetDevLinkSpeed(pfNetName),
+			LinkAdminState: s.networkHelper.GetNetDevLinkAdminState(pfNetName),
 		}
 
 		pfStatus, exist, err := storeManager.LoadPfsStatus(iface.PciAddress)
@@ -571,7 +574,7 @@ func (s *sriov) configSriovDevice(iface *sriovnetworkv1.Interface, skipVFConfigu
 	if err != nil {
 		return err
 	}
-	if pfLink.Attrs().OperState != netlink.OperUp {
+	if pfLink.Attrs().Flags&net.FlagUp == 0 {
 		err = s.netlinkLib.LinkSetUp(pfLink)
 		if err != nil {
 			return err
--- a/pkg/host/internal/sriov/sriov.go
+++ b/pkg/host/internal/sriov/sriov.go
@@ -3,7 +3,6 @@ package sriov
 import (
 	"errors"
 	"fmt"
-	"net"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -574,7 +573,7 @@ func (s *sriov) configSriovDevice(iface *sriovnetworkv1.Interface, skipVFConfigu
 	if err != nil {
 		return err
 	}
-	if pfLink.Attrs().Flags&net.FlagUp == 0 {
+	if !s.netlinkLib.IsLinkAdminStateUp(pfLink) {
 		err = s.netlinkLib.LinkSetUp(pfLink)
 		if err != nil {
 			return err
--- a/pkg/host/types/interfaces.go
+++ b/pkg/host/types/interfaces.go
@@ -94,6 +94,8 @@ type NetworkInterface interface {
 	SetNetdevMTU(pciAddr string, mtu int) error
 	// GetNetDevMac returns the network interface mac address
 	GetNetDevMac(name string) string
+	// GetNetDevNodeGUID returns the network interface node GUID if device is RDMA capable otherwise returns empty string
+	GetNetDevNodeGUID(pciAddr string) string
 	// GetNetDevLinkSpeed returns the network interface link speed
 	GetNetDevLinkSpeed(name string) string
 	// GetDevlinkDeviceParam returns devlink parameter for the device as a string, if the parameter has multiple values
@@ -105,6 +107,8 @@ type NetworkInterface interface {
 	SetDevlinkDeviceParam(pciAddr, paramName, value string) error
 	// EnableHwTcOffload make sure that hw-tc-offload feature is enabled if device supports it
 	EnableHwTcOffload(ifaceName string) error
+	// GetNetDevLinkAdminState returns the admin state of the interface.
+	GetNetDevLinkAdminState(ifaceName string) string
 }
 
 type ServiceInterface interface {
--- a/pkg/plugins/generic/generic_plugin_test.go
+++ b/pkg/plugins/generic/generic_plugin_test.go
@@ -53,18 +53,19 @@ var _ = Describe("Generic plugin", func() {
 				},
 				Status: sriovnetworkv1.SriovNetworkNodeStateStatus{
 					Interfaces: sriovnetworkv1.InterfaceExts{{
-						PciAddress:  "0000:00:00.0",
-						NumVfs:      1,
-						TotalVfs:    1,
-						DeviceID:    "1015",
-						Vendor:      "15b3",
-						Name:        "sriovif1",
-						Mtu:         1500,
-						Mac:         "0c:42:a1:55:ee:46",
-						Driver:      "mlx5_core",
-						EswitchMode: "legacy",
-						LinkSpeed:   "25000 Mb/s",
-						LinkType:    "ETH",
+						PciAddress:     "0000:00:00.0",
+						NumVfs:         1,
+						TotalVfs:       1,
+						DeviceID:       "1015",
+						Vendor:         "15b3",
+						Name:           "sriovif1",
+						Mtu:            1500,
+						Mac:            "0c:42:a1:55:ee:46",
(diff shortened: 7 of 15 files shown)


api/v1/helper.go
```diff
@@ -271,6 +279,20 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 							return true
 						}
 
+						if strings.EqualFold(ifaceStatus.LinkType, consts.LinkTypeETH) {
+							if len(vfStatus.Mac) == 0 {
+								log.V(2).Info("NeedToUpdateSriov(): VF MAC Address needs update",
+									"vf", vfStatus.VfID, "current", vfStatus.Mac)
+								return true
```
adrianchiris, line 286: we also need to reconcile if node/port guid is zeroes if device is rdma capable.

you may have VF admin mac set BUT node/port guid is unset (since driver was not unbind/bind)

root@cloud-dev-45:~# cat /sys/bus/pci/devices/0000\:03\:00.2/infiniband/mlx5_4/node_guid
0000:0000:0000:0000
root@cloud-dev-45:~# echo 0000:03:00.2 > /sys/bus/pci/drivers/mlx5_core/unbind
root@cloud-dev-45:~# echo 0000:03:00.2 > /sys/bus/pci/drivers/mlx5_core/bind
root@cloud-dev-45:~# cat /sys/bus/pci/devices/0000\:03\:00.2/infiniband/mlx5_4/node_guid
0012:34ff:fe56:7891

api/v1/sriovnetworknodestate_types.go
```diff
@@ -83,6 +84,7 @@ type VirtualFunction struct {
 	Mtu        int    `json:"mtu,omitempty"`
 	VfID       int    `json:"vfID"`
 	VdpaType   string `json:"vdpaType,omitempty"`
+	GUID       string `json:"guid,omitempty"`
```
adrianchiris, line 87: generally need to deal with 3 cases here:

device is IB: GUID is present
device is eth but is not RDMA capable: GUID is empty/nil
device is eth and is RDMA capable: GUID is present, GUID represents the device "node guid"

adrianchiris:
>Today we reconcile on .spec change since we skip reconciliation if the generation of the object is the same as the object that was last reconciled successfully. Therefore, changes in .status, like what we try to detect with that PR, are not reconciled unless the .spec changes when the config daemon runs continuously (i.e. no restart). On daemon restart, the .status field will be reconciled the first time until the last successfully reconciled generation is saved in memory.

PR #530  aims to tackle that

pkg/host/internal/network/network.go
```diff
@@ -172,10 +173,19 @@ func (n *network) SetNetdevMTU(pciAddr string, mtu int) error {
 
 func (n *network) GetNetDevMac(ifaceName string) string {
 	log.Log.V(2).Info("GetNetDevMac(): get Mac", "device", ifaceName)
-	macFilePath := filepath.Join(vars.FilesystemRoot, consts.SysClassNet, ifaceName, "address")
-	data, err := os.ReadFile(macFilePath)
+	return readNetDevInterfaceAddress(ifaceName)
+}
+
+func (n *network) GetNetDevGUID(ifaceName string) string {
```
adrianchiris, line 179: what are you getting for ipoib interface ? i think you are getting GID (128bit) and not the node GUID (64 bit)

i think we should fetch GUID via netlink: https://github.com/vishvananda/netlink/blob/06219cde3e811f757f263ae54e848891c8ce6cbc/rdma_link_linux.go#L18

pkg/host/internal/network/network.go
```diff
@@ -193,3 +203,17 @@ func (n *network) GetNetDevLinkSpeed(ifaceName string) string {
 
 	return fmt.Sprintf("%s Mb/s", strings.TrimSpace(string(data)))
 }
+
+func (n *network) GetNetDevLinkStatus(ifaceName string) string {
+	log.Log.V(2).Info("GetNetDevLinkStatus(): get LinkStatus", "device", ifaceName)
+	if len(ifaceName) > 0 {
+		link, err := netlink.LinkByName(ifaceName)
```
adrianchiris, line 210: please use NetlinkLib from host/internal/lib/netlink

pkg/host/internal/sriov/sriov.go
```diff
@@ -93,7 +93,7 @@ func (s *sriov) ResetSriovDevice(ifaceStatus sriovnetworkv1.InterfaceExt) error
 	return nil
 }
 
-func (s *sriov) GetVfInfo(pciAddr string, devices []*ghw.PCIDevice) sriovnetworkv1.VirtualFunction {
+func (s *sriov) GetVfInfo(pciAddr string, ifaceType string, devices []*ghw.PCIDevice) sriovnetworkv1.VirtualFunction {
```
adrianchiris, line 96: i think you will not need ifaceType as we will need to get guid of rdma device regardless of the link.

pkg/host/types/interfaces.go
```diff
@@ -90,8 +90,12 @@ type NetworkInterface interface {
 	SetNetdevMTU(pciAddr string, mtu int) error
 	// GetNetDevMac returns the network interface mac address
 	GetNetDevMac(name string) string
+	// GetNetDevGUID returns the network interface GUID (Infiniband)
```
adrianchiris, line 93: nit:
```
// GetNetDevGUID returns the network interface GUID if device is RDMA capable, returns empy string otherwise.
```

pkg/host/internal/network/network.go
vasrem: You are right:
```
# cat /sys/class/net/ibp75s0f0v1/address
00:00:01:e8:fe:80:00:00:00:00:00:00:00:00:00:00:00:00:00:00
```

Will adjust the code.

pkg/host/internal/network/network.go
```diff
@@ -172,16 +178,42 @@ func (n *network) SetNetdevMTU(pciAddr string, mtu int) error {
...
+		rdmaDevices, err := os.ReadDir(rdmaDevicesPath)
+		if err != nil {
+			log.Log.Error(err, "GetNetDevNodeGUID(): failed to read RDMA related directory", "pciAddr", pciAddr)
+			return ""
+		}
+
+		if len(rdmaDevices) != 1 {
+			log.Log.Error(err, "GetNetDevNodeGUID(): expected just one RDMA device", "pciAddr", pciAddr, "numOfDevices", len(rdmaDevices))
+			return ""
+		}
```
vasrem, line 204: Is this assumption right?

pkg/host/internal/sriov/sriov.go
vasrem: The outcome of this is that MAC address is populated with the GID for IB. Is that acceptable?

```
    - Vfs:
      - deviceID: 101c
        driver: mlx5_core
        guid: "01:00:00:00:00:00:00:00"
        mac: 00:00:00:e8:fe:80:00:00:00:00:00:00:00:00:00:00:00:00:00:02
        mtu: 4092
        name: ibp75s0f0v0
        pciAddress: 0000:4b:00.2
        vendor: 15b3
        vfID: 0
      - deviceID: 101c
        driver: mlx5_core
        guid: "05:00:00:00:00:00:00:00"
        mac: 00:00:01:e8:fe:80:00:00:00:00:00:00:00:00:00:00:00:00:00:00
        mtu: 4092
        name: ibp75s0f0v1
        pciAddress: 0000:4b:00.3
        vendor: 15b3
        vfID: 1
      deviceID: 101b
      driver: mlx5_core
      linkStatus: down
      linkType: IB
      mac: 00:00:05:91:fe:80:00:00:00:00:00:00:08:c0:eb:03:00:c1:33:66
      mtu: 4092
      name: ibp75s0f0
      numVfs: 2
      pciAddress: 0000:4b:00.0
      totalvfs: 2
      vendor: 15b3
```

api/v1/helper.go
```diff
@@ -271,6 +279,28 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 							return true
 						}
 
+						if strings.EqualFold(ifaceStatus.LinkType, consts.LinkTypeETH) {
+							if len(vfStatus.Mac) == 0 || vfStatus.Mac == consts.DefaultMAC {
```
adrianchiris, line 283: I have been thinking on some possible scenarios which might make checking  mac/guid tricky.

for MAC: if VF netdev is used by a container then you will not be able to discover its netdev hence its mac will be `""`
which will trigger re-configuration.

for GUID basically the same but this will happen only when RDMA namespace mode is set to exclusive.

adrianchiris: we also could do some merge logic where if we fail to read a value, assume the previous one is still valid. but that also has its drawbacks.

vasrem: Makes sense. I will change the condition to just do checks based on default MAC and Node GUID.

api/v1/helper.go
```diff
@@ -247,6 +248,13 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 		log.V(2).Info("NeedToUpdateSriov(): NumVfs needs update", "desired", ifaceSpec.NumVfs, "current", ifaceStatus.NumVfs)
 		return true
 	}
+
+	var linkUp netlink.LinkOperState = netlink.OperUp
+	if ifaceStatus.LinkStatus != linkUp.String() {
```
SchSeba, line 253: we should not have this check here.
there are cases where users use a down PF for communication between to VFs on the same PF.
they can configure `LinkState on` and just use the nic as a backbone for 2 DPDK applications passing traffic

api/v1/sriovnetworknodestate_types.go
```diff
@@ -64,6 +64,7 @@ type InterfaceExt struct {
 	NumVfs            int               `json:"numVfs,omitempty"`
 	LinkSpeed         string            `json:"linkSpeed,omitempty"`
 	LinkType          string            `json:"linkType,omitempty"`
+	LinkStatus        string            `json:"linkStatus,omitempty"`
```
SchSeba, line 67: I don't think we need to track the PF linkStatus

pkg/host/internal/network/network.go
```diff
@@ -175,16 +179,42 @@ func (n *network) SetNetdevMTU(pciAddr string, mtu int) error {
 
 func (n *network) GetNetDevMac(ifaceName string) string {
 	log.Log.V(2).Info("GetNetDevMac(): get Mac", "device", ifaceName)
-	macFilePath := filepath.Join(vars.FilesystemRoot, consts.SysClassNet, ifaceName, "address")
-	data, err := os.ReadFile(macFilePath)
+	addressFilePath := filepath.Join(vars.FilesystemRoot, consts.SysClassNet, ifaceName, "address")
+	data, err := os.ReadFile(addressFilePath)
 	if err != nil {
-		log.Log.Error(err, "GetNetDevMac(): fail to read Mac file", "path", macFilePath)
+		log.Log.Error(err, "readInterfaceAddress(): fail to read address file", "path", addressFilePath)
```
SchSeba, line 185: `readInterfaceAddress` is not the name of the functions

pkg/host/internal/network/network.go
SchSeba: @adrianchiris I am not an expert in RDMA

pkg/host/internal/sriov/sriov.go
```diff
@@ -117,6 +117,7 @@ func (s *sriov) GetVfInfo(pciAddr string, devices []*ghw.PCIDevice) sriovnetwork
 		vf.Name = name
 		vf.Mac = s.networkHelper.GetNetDevMac(name)
 	}
+	vf.GUID = s.networkHelper.GetNetDevNodeGUID(pciAddr)
```
SchSeba, line 120: should we check here if the string is empty we don't allocate it?
this way it will not be showed in the nodeState

pkg/host/internal/network/network.go
```diff
@@ -175,16 +179,42 @@ func (n *network) SetNetdevMTU(pciAddr string, mtu int) error {
...
 		return ""
 	}
 
 	return strings.TrimSpace(string(data))
 }
 
+func (n *network) GetNetDevNodeGUID(pciAddr string) string {
+	log.Log.V(2).Info("GetNetDevNodeGUID(): get node GUID", "pciAddr", pciAddr)
+	if len(pciAddr) > 0 {
+		rdmaDevicesPath := filepath.Join(vars.FilesystemRoot, consts.SysBusPciDevices, pciAddr, "infiniband")
```
SchSeba, line 195: will this file exist even if we move the network device into the pod namespace?

api/v1/helper.go
vasrem: Hmm.. so basically users will overwrite what the operator does when it configures VFs [(i.e. set PF link up)](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/6e8a86125ffbe1e3539cf1a7451c86e90b229cbb/pkg/host/sriov.go#L454-L464) and rely on the operator to not move the PF link up?

This sounds hacky. Will bringing the PF up break their use case?

api/v1/sriovnetworknodestate_types.go
vasrem: This is based on the above comment right? If we converge on the comment above, I will remove this change.

pkg/host/internal/network/network.go
vasrem: True, will update that. Leftover while refactoring.

pkg/host/internal/sriov/sriov.go
vasrem: GUID is configured with the json tag `omitempty`. https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/d8ddad1adfa06e34409cfa47ad268b1781671952/api/v1/sriovnetworknodestate_types.go#L87

For a field of type `string`, this means that if the string you pass in is empty, the field will be omitted. https://pkg.go.dev/encoding/json#:~:text=The%20%22omitempty%22%20option%20specifies%20that%20the%20field%20should%20be%20omitted%20from%20the%20encoding%20if%20the%20field%20has%20an%20empty%20value%2C%20defined%20as%20false%2C%200%2C%20a%20nil%20pointer%2C%20a%20nil%20interface%20value%2C%20and%20any%20empty%20array%2C%20slice%2C%20map%2C%20or%20string.

pkg/host/internal/network/network.go
vasrem: I believe this depends on the RDMA mode we have configured on the system. Take a look at [this](https://man7.org/linux/man-pages/man8/rdma-system.8.html#:~:text=NEWMODE%20%2D%20specifies%20the,can%20be%0A%20%20%20%20%20%20%20used.) doc.

There is a case where the file won't exist. If it doesn't exist, then we return empty string. Based on this comment https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/587#discussion_r1477863309, I changed the implementation of the check to take action only when the GUID (and MAC address in other cases) can be read. Otherwise do nothing. It won't catch 100% of manual actions done by the user on the system, but should be good enough.

WDYT?

api/v1/sriovnetworknodestate_types.go
SchSeba: I think we have two different stuff here
1. Operational link state
2. Actual link state (if the card is not connected to anything for example)

if you ask me I will be interesting in see the actual link state on the nodeState the Opr one can always be up I think as we have today right?

SchSeba (commented):
left two small comments and one important part that we need to understand the difference between operational nic state and actual nic state

pkg/host/internal/network/network.go
```diff
@@ -175,16 +179,42 @@ func (n *network) SetNetdevMTU(pciAddr string, mtu int) error {
...
+		log.Log.Error(err, "GetNetDevMac(): fail to read address file", "path", addressFilePath)
 		return ""
 	}
 
 	return strings.TrimSpace(string(data))
 }
 
+func (n *network) GetNetDevNodeGUID(pciAddr string) string {
+	log.Log.V(2).Info("GetNetDevNodeGUID(): get node GUID", "pciAddr", pciAddr)
+	if len(pciAddr) > 0 {
```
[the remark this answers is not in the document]
vasrem: Good point, addressed.

pkg/host/internal/network/network.go
```diff
@@ -196,3 +226,17 @@ func (n *network) GetNetDevLinkSpeed(ifaceName string) string {
 
 	return fmt.Sprintf("%s Mb/s", strings.TrimSpace(string(data)))
 }
+
+func (n *network) GetNetDevLinkStatus(ifaceName string) string {
+	log.Log.V(2).Info("GetNetDevLinkStatus(): get LinkStatus", "device", ifaceName)
+	if len(ifaceName) > 0 {
```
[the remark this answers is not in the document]
vasrem: Addressed

pkg/host/manager.go
```diff
@@ -38,11 +38,12 @@ type hostManager struct {
...
 func NewHostManager(utilsInterface utils.CmdInterface) HostManagerInterface {
 	dpUtils := dputils.New()
+	nl := netlink.New()
 	k := kernel.New(utilsInterface)
-	n := network.New(utilsInterface, dpUtils)
+	n := network.New(utilsInterface, dpUtils, nl)
 	sv := service.New(utilsInterface)
 	u := udev.New(utilsInterface)
-	sr := sriov.New(utilsInterface, k, n, u, netlink.New(), dpUtils)
+	sr := sriov.New(utilsInterface, k, n, u, nl, dpUtils)
```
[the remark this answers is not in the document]
vasrem: I think a couple of methods using the netlink as well https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/host/internal/sriov/sriov.go#L137-L156.

vasrem:
I can make the commits more compact if you feel strong about it. 14 commits are way too many for this change.

vasrem:
@SchSeba could you please make another pass here?

api/v1/sriovnetworknodestate_types.go
```diff
@@ -64,6 +66,7 @@ type InterfaceExt struct {
 	NumVfs            int               `json:"numVfs,omitempty"`
 	LinkSpeed         string            `json:"linkSpeed,omitempty"`
 	LinkType          string            `json:"linkType,omitempty"`
+	LinkFlags         net.Flags         `json:"linkFlags,omitempty"`
```
SchSeba, line 69: I don't know if I will like to expose this into the interfaceExt struct this is a uint number that doesn't mean much
maybe we have something like LinkOprState string - up/down

SchSeba (commented):
everything looks good from my side (but I am not an RDMA expert here) beside the LinkFlags I don't think we should expose something like that in the CRD 

please check my comment
Thanks!

api/v1/helper.go
```diff
@@ -292,6 +299,28 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 							return true
 						}
 
+						if strings.EqualFold(ifaceStatus.LinkType, consts.LinkTypeETH) {
+							if vfStatus.Mac == consts.DefaultMAC {
```
adrianchiris, line 303: can you add a comment that we only do these checks if mac/guid contains a value ? (so that is clear it is intentional)

these fields in vf status may be empty because device is allocated to some workload.

pkg/host/internal/network/network.go
```diff
@@ -187,6 +188,34 @@ func (n *network) GetNetDevMac(ifaceName string) string {
...
 
+// GetNetDevNodeGUID returns the network interface node GUID if device is RDMA capable otherwise returns empty string
+func (n *network) GetNetDevNodeGUID(pciAddr string) string {
+	log.Log.V(2).Info("GetNetDevNodeGUID(): get node GUID", "pciAddr", pciAddr)
+	if len(pciAddr) == 0 {
+		return ""
+	}
+
+	rdmaDevicesPath := filepath.Join(vars.FilesystemRoot, consts.SysBusPciDevices, pciAddr, "infiniband")
+	rdmaDevices, err := os.ReadDir(rdmaDevicesPath)
```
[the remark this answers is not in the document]
vasrem: Added error check with conditional logging instead.

api/v1/helper.go
vasrem: Added.

pkg/plugins/generic/generic_plugin_test.go
```diff
@@ -78,14 +181,165 @@ var _ = Describe("Generic plugin", func() {
...
+			}
+
+			hostHelper.EXPECT().WriteSwitchdevConfFile(networkNodeState, map[string]bool{"0000:00:00.0": false}).Return(false, nil)
+			needDrain, needReboot, err := genericPlugin.OnNodeStateChange(networkNodeState)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(needReboot).To(BeFalse())
+			Expect(needDrain).To(BeTrue())
+		})
+
+		It("should not drain if PF link is not down", func() {
```
SchSeba, line 241: can you check what it will do for the PF link is empty? and the same for the GUID as in the code we don't return an error if we are not able to read the information

SchSeba (commented):
just one small comment and we should be good

pkg/plugins/generic/generic_plugin_test.go
vasrem: > can you check what it will do for the PF link is empty?

That's what is tested in that check. We only take action if the link is down.

> and the same for the GUID as in the code we don't return an error if we are not able to read the information

I believe I have covered the cases below.🤔 Could you give me an example of a struct you would like me to test?
e.g. 
```
			networkNodeState := &sriovnetworkv1.SriovNetworkNodeState{
				Spec: sriovnetworkv1.SriovNetworkNodeStateSpec{
					Interfaces: sriovnetworkv1.Interfaces{{
						PciAddress: "0000:00:00.0",
						NumVfs:     1,
						Mtu:        1500,
						VfGroups: []sriovnetworkv1.VfGroup{{
							DeviceType:   "netdevice",
							PolicyName:   "policy-1",
							ResourceName: "resource-1",
							VfRange:      "0-1",
							Mtu:          1500,
							IsRdma:       true,
						}}}},
				},
				Status: sriovnetworkv1.SriovNetworkNodeStateStatus{
					Interfaces: sriovnetworkv1.InterfaceExts{{
						PciAddress:   "0000:00:00.0",
						NumVfs:       1,
						TotalVfs:     1,
						DeviceID:     "1015",
						Vendor:       "15b3",
						Name:         "sriovif1",
						Mtu:          1500,
						Mac:          "0c:42:a1:55:ee:46",
						Driver:       "mlx5_core",
						EswitchMode:  "legacy",
						LinkSpeed:    "25000 Mb/s",
						LinkType:     "ETH",
						LinkAdmState: "up",
						VFs: []sriovnetworkv1.VirtualFunction{{
							PciAddress: "0000:00:00.1",
							DeviceID:   "1016",
							Vendor:     "15b3",
							VfID:       0,
							Name:       "sriovif1v0",
							Mtu:        1500,
							Driver:     "mlx5_core",
							Mac:        "0c:42:a1:55:ee:46",
						}},
					}},
				},
			}
```

pkg/consts/constants.go
```diff
@@ -42,6 +42,9 @@ const (
 	LinkTypeIB  = "IB"
 	LinkTypeETH = "ETH"
 
+	DefaultNodeGUID = "0000:0000:0000:0000"
```
adrianchiris, line 45: nit: can you rename to: UninitializedNodeGUID ?

since you cannot set it back to "default "

api/v1/helper.go
adrianchiris: also been thinking about it, i dont think MAC is never zero for a netdevice since its always set by the kernel.
@SchSeba do you agree ?

vasrem: Dropping this check. Sebastian let us know if you think otherwise.

api/v1/helper.go
```diff
@@ -300,6 +306,28 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 							return true
 						}
 
+						if strings.EqualFold(ifaceStatus.LinkType, consts.LinkTypeETH) {
```
adrianchiris, line 309: nit: any chance to have one check to avoid duplication ?

e.g if (ETH && RDMA) || IB { ....}

adrianchiris (commented):
overall LGTM, one minor nit.

api/v1/helper.go
vasrem: Yeap, makes sense. Addressing.

adrianchiris (approved):

pkg/host/internal/network/network.go
```diff
@@ -187,6 +190,36 @@ func (n *network) GetNetDevMac(ifaceName string) string {
 	return link.Attrs().HardwareAddr.String()
 }
 
+// GetNetDevNodeGUID returns the network interface node GUID if device is RDMA capable otherwise returns empty string
+func (n *network) GetNetDevNodeGUID(pciAddr string) string {
+	log.Log.V(2).Info("GetNetDevNodeGUID(): get node GUID", "pciAddr", pciAddr)
```
zeeke, line 195: This function gets called during the polling of NIC status, for each configured VF. 
I would avoid this logging statement, as it increases config-daemon logs without adding specific value.

You can have a look at the end2end job artifact
https://github.com/k8snetworkplumbingwg/sriov-network-operator/actions/runs/9005014290/artifacts/1484967983

vasrem: Agreed, removed.

adrianchiris (approved):

vasrem:
@zeeke and @SchSeba bump on that one if you could take a look.

pkg/host/internal/network/network.go
```diff
@@ -329,3 +361,23 @@ func (n *network) EnableHwTcOffload(ifaceName string) error {
...
+	}
+
+	link, err := n.netlinkLib.LinkByName(ifaceName)
+	if err != nil {
+		log.Log.Error(err, "GetNetDevLinkAdminState(): failed to get link", "device", ifaceName)
+		return ""
+	}
+
+	if link.Attrs().Flags&net.FlagUp == 0 {
+		return "down"
```
SchSeba, line 379: please put this one as a const and use it also in https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/587/files#diff-371d95e79c61d8cc1e6791d27e009289552efcaada38f9b2fc75ebb487c90ea7R275

pkg/host/internal/network/network.go
```diff
@@ -329,3 +361,23 @@ func (n *network) EnableHwTcOffload(ifaceName string) error {
...
+	}
+
+	return "up"
```
SchSeba, line 382: please put this one as a const

pkg/host/internal/sriov/sriov.go
```diff
@@ -571,7 +574,7 @@ func (s *sriov) configSriovDevice(iface *sriovnetworkv1.Interface, skipVFConfigu
 	if err != nil {
 		return err
 	}
-	if pfLink.Attrs().OperState != netlink.OperUp {
+	if pfLink.Attrs().Flags&net.FlagUp == 0 {
```
SchSeba, line 577: please put this one as part of the `netlinkLib` lets use it here and on the other place in the code

SchSeba (commented):
sorry it took soo much time to merge this one.

It's almost remove just small comments and we can merge this PR

vasrem:
@SchSeba please have a look, I addressed your comments.

adrianchiris (approved):

SchSeba (approved):
nice work!

sorry about the amount of time the review took

PR #589, opened 2024-01-22
CI: remove images after we push to cluster

SchSeba:
removing the images after we push as part of the CI, we need this to avoid exploding the CI nodes


zeeke (approved):
LGTM

adrianchiris (approved):

PR #594, opened 2024-01-25
re-enable mtu test on virtual cluster

SchSeba:
QEMU 8.2.0 introduce a fix to the kernel panic
when allocation jumbo MTU so we can re-enable the test


zeeke (approved):
LGTM

adrianchiris (approved):

PR #599, opened 2024-01-30
Optimize a request to find device plugin pods.

alan-kut:
Add a resourceVersion=0 to the reuqest to find device plugin pods on a given node from the latest cache from the kube-apiserver. Without a resourceVersion the request needs to reach etcd first and then kube-apiserver will filter out the pods.

Due to this etcd can get overloaded. Since it runs in daemonset, they get created on every node. If a cluster has a lot of nodes the list requests will put a cluster down.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -732,8 +732,9 @@ func (dn *Daemon) restartDevicePluginPod() error {
 
 	var podToDelete string
 	pods, err := dn.kubeClient.CoreV1().Pods(vars.Namespace).List(context.Background(), metav1.ListOptions{
-		LabelSelector: "app=sriov-device-plugin",
-		FieldSelector: "spec.nodeName=" + vars.NodeName,
+		LabelSelector:   "app=sriov-device-plugin",
+		FieldSelector:   "spec.nodeName=" + vars.NodeName,
+		ResourceVersion: "0",
 	})
 	if err != nil {
 		if errors.IsNotFound(err) {


pkg/daemon/daemon.go
```diff
@@ -734,8 +734,9 @@ func (dn *Daemon) restartDevicePluginPod() error {
 
 	var podToDelete string
 	pods, err := dn.kubeClient.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{
-		LabelSelector: "app=sriov-device-plugin",
-		FieldSelector: "spec.nodeName=" + vars.NodeName,
+		LabelSelector:   "app=sriov-device-plugin",
+		FieldSelector:   "spec.nodeName=" + vars.NodeName,
+		ResourceVersion: "0",
```
adrianchiris, line 739: as we just care for the existance of such pod i believe we are OK with specifying this.

some additional reading on the semantics of ResourceVersion field for other reviewers[1]

[1] https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-get-and-list

adrianchiris (approved):
@alan-kut
at what number of nodes did you hit this issue ?

LGTM.

alan-kut:
I hit the issue with 4k nodes running the plugin

alan-kut:
The failures are in the controllers, not daemon.

adrianchiris:
@alan-kut could you rebase ?

@SchSeba @zeeke PTAL on this one when you get a chance

pkg/daemon/daemon.go
zeeke: @adrianchiris thanks for linking the documentation.

zeeke (approved):
LGTM

alan-kut:
Rebased

adrianchiris (approved):

PR #600, opened 2024-01-30
Use APITimeout and RetryInterval constant in tests

e0ne:


mlguerrero12:
LGTM

controllers/sriovnetworkpoolconfig_controller_test.go
```diff
@@ -65,7 +66,7 @@ var _ = Describe("Operator", func() {
 					return err
 				}
 				return nil
-			}, timeout*3, interval).Should(Succeed())
+			}, util.APITimeout*3, util.RetryInterval).Should(Succeed())
```
adrianchiris, line 69: this is odd that we need it. probably related to where this test is executed.

i mean we just spin up apiserver +etcd with envtest then a simple reconcile

adrianchiris (approved):

controllers/sriovnetworkpoolconfig_controller_test.go
e0ne: IMO, that's something we have to fix in the future in a sepatare PR

zeeke (approved):
I can't reproduce the error this PR is fixing on my local environment, but looking at the CI, it fixes the job.

LGTM

PR #601, opened 2024-01-30
move namespace to vars pkg

adrianchiris:
Namespace variable is used in both controllers and daemon. Move it to vars pkg and use in both.


zeeke (approved):
LGTM

adrianchiris:
failure does not seem to be reelated. merging this one.

PR #602, opened 2024-01-30
[switchdev 4/x] split systemd service to two parts

ykulazhenkov:
split systemd service to two parts:
The first part is executed before the
NeworkManager or systemd-networkd
starts on the host.
The second part is triggered after the
network configuration is complete.

This change is required to support hw offload
use-case with systemd service.

cc @adrianchiris @zeeke @SchSeba

### DIFF
--- a/bindata/manifests/sriov-config-service/kubernetes/sriov-config-service.yaml
+++ b/bindata/manifests/sriov-config-service/kubernetes/sriov-config-service.yaml
@@ -1,14 +1,15 @@
 contents: |
   [Unit]
-  Description=Configures SRIOV NIC
-  After=network-pre.target
-  Before=NetworkManager.service kubelet.service
-  
+  Description=Configures SRIOV NIC - pre network configuration
+  DefaultDependencies=no
+  After=network-pre.target systemd-udev-settle.service systemd-sysusers.service systemd-sysctl.service
+  Before=network.target NetworkManager.service systemd-networkd.service ovs-vswitchd.service ovsdb-server.service
+
   [Service]
   Type=oneshot
-  ExecStart=/var/lib/sriov/sriov-network-config-daemon -v 2 service
+  ExecStart=/var/lib/sriov/sriov-network-config-daemon -v 2 --zap-log-level 2 service --phase pre
   StandardOutput=journal+console
-  
+
   [Install]
   WantedBy=multi-user.target
 enabled: true
--- a/cmd/sriov-network-config-daemon/service.go
+++ b/cmd/sriov-network-config-daemon/service.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -36,6 +37,11 @@ import (
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/version"
 )
 
+const (
+	PhasePre  = "pre"
+	PhasePost = "post"
+)
+
 var (
 	serviceCmd = &cobra.Command{
 		Use:   "service",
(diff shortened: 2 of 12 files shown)


bindata/manifests/sriov-config-service/kubernetes/sriov-config-service.yaml
```diff
@@ -1,14 +1,15 @@
 contents: |
   [Unit]
-  Description=Configures SRIOV NIC
-  After=network-pre.target
-  Before=NetworkManager.service kubelet.service
-  
+  Description=Configures SRIOV NIC - pre network configuration
+  DefaultDependencies=no
+  After=network-pre.target systemd-udev-settle.service systemd-sysusers.service systemd-sysctl.service
+  Before=network.target NetworkManager.service systemd-networkd.service ovs-vswitchd.service ovsdb-server.service
```
ykulazhenkov, line 6: Note: If a service defined in After or Before section is not exist it will be simply ignored.

ykulazhenkov:
PR updated to add tests for systemd service

cmd/sriov-network-config-daemon/service.go
```diff
@@ -102,105 +102,155 @@ func runServiceCmd(cmd *cobra.Command, args []string) error {
...
+
+func getPluginAndState(setupLog logr.Logger, hostHelpers helper.HostHelpersInterface, conf *systemd.SriovConfig) (
+	plugin.VendorPlugin, *sriovv1.SriovNetworkNodeState, error) {
+	var (
+		configPlugin  plugin.VendorPlugin
+		ifaceStatuses []sriovv1.InterfaceExt
+		err           error
+	)
+	switch conf.PlatformType {
+	case consts.Baremetal:
```
adrianchiris, line 166: should we have a "default" ? just to be on safe side ?

pkg/daemon/daemon.go
```diff
@@ -429,13 +429,19 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+			if err != nil {
+				log.Log.Error(err, "nodeStateSyncHandler(): failed to check if sriov-config-post-network service exist on host")
+				return err
+			}
 
 			// if the service doesn't exist we should continue to let the k8s plugin to create the service files
 			// this is only for k8s base environments, for openshift the sriov-operator creates a machine config to will apply
 			// the system service and reboot the node the config-daemon doesn't need to do anything.
-			if !serviceEnabled {
+			if !(serviceEnabled && postNetworkServiceEnabled) {
```
adrianchiris, line 441: This systemd handling part of the daemon starting to get long, we should split it to separate "handleSystemdMode" method.

that said, its not related to this PR....

cmd/sriov-network-config-daemon/service.go
ykulazhenkov: If the switch doesn't match, then the `getPluginAndState` function return nil for the configPlugin. We handle this on line 138 and log message that there is no plugin for this platform.

pkg/daemon/daemon.go
ykulazhenkov: yes, in follow up PRs I will move things a bit in the function.

cmd/sriov-network-config-daemon/service.go
```diff
@@ -102,105 +102,155 @@ func runServiceCmd(cmd *cobra.Command, args []string) error {
...
+		syncStatus = consts.SyncStatusInProgress
 	}
-
-	setupLog.V(0).Info("shutting down sriov-config-service")
-	return nil
+	return updateResult(setupLog, syncStatus, "")
 }
 
-func updateSriovResultErr(errMsg string) error {
+func updateResult(setupLog logr.Logger, result, msg string) error {
```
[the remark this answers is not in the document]
ykulazhenkov: > that way we dont need that in progress on pre phase at L238.

I think this is a nice thing that we set SyncStatus to InProgress in the "pre" phase because:
1. We are really in the process of the configuration at this point, so the status reflects the actual state.
2. The logic in the daemon and in the service is very simple and you can describe it with the single sentence. The daemon is not aware about phases at all.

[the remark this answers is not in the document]
adrianchiris: ack, if it will get more complex down the road and find we need more information about the outcome of each state (e.g for config daemon), we can extend.

adrianchiris (approved):
LGTM!

zeeke (commented):
Left a few comments. I'm finding a little bit difficult to read the `service.go` file with that `phase` variable read in many places.
Have you considered the option of creating two functions `runServiceCmdPre()` and `runServiceCmdPost()` called at the very beginning of `runServiceCmd()`? That way we have a bunch of duplicated code,  but we'd get rid of many `if/else` statments. WDYT?

bindata/manifests/sriov-config-service/kubernetes/sriov-config-service.yaml
```diff
@@ -1,14 +1,15 @@
...
+
   [Service]
   Type=oneshot
-  ExecStart=/var/lib/sriov/sriov-network-config-daemon -v 2 service
+  ExecStart=/var/lib/sriov/sriov-network-config-daemon -v 2 service -p pre
```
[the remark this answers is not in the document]
ykulazhenkov: fixed

cmd/sriov-network-config-daemon/service.go
```diff
@@ -102,105 +107,155 @@ func runServiceCmd(cmd *cobra.Command, args []string) error {
...
+		err = updateSriovResultErr(setupLog, fmt.Errorf("failed to apply configuration: %v", err))
+	} else {
+		err = updateSriovResultSucceed(setupLog)
+	}
+
+	setupLog.V(0).Info("shutting down sriov-config-service")
+	return err
+}
+
+func getPluginAndState(setupLog logr.Logger, hostHelpers helper.HostHelpersInterface, conf *systemd.SriovConfig) (
```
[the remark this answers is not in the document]
ykulazhenkov: done

zeeke (approved):
LGTM

PR #604, opened 2024-01-31
Move default policy creation in controller tests to BeforeSuite.

ykulazhenkov:
This is required to fix the following problem:
SriovOperatorConfigReconciler starts in the BeforeSuite handler, the default OperatorConfig is also created in BeforeSuite. That means the SriovOperatorConfigReconciler will be triggered with the first test in the controller suite.

SriovOperatorConfigReconciler will fail (return an error) until DefaultPolicy is created by the BeforeEach function in sriovoperatorconfig_controller_test.go.

The problem is that tests in sriovoperatorconfig_controller_test.go starts after the test in some other *_test.go files in the controller package.  At the time when tests from sriovoperatorconfig_controller_test.go are eventually executed, SriovOperatorConfigReconciler controller is already failed multiple times in a row (because there was no default policy) and is waiting with a backoff delay (already quite large) before reconcilation retry.

### DIFF
--- a/controllers/sriovoperatorconfig_controller_test.go
+++ b/controllers/sriovoperatorconfig_controller_test.go
@@ -11,52 +11,13 @@ import (
 	. "github.com/onsi/gomega"
 
 	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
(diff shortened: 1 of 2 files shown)


zeeke (approved):
LGTM

thanks for looking at this

e0ne (approved):

PR #610, opened 2024-02-01
openstack: Fix `nil` HostManager in `openstackContext`

zeeke:
Refactoring in [1] left the openstackContext uninitialized, making the config-daemon crash on OpenStack deployments.

This error has been caught in Openshift downstream CI [2] 

[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/553
[2] https://github.com/openshift/sriov-network-operator/pull/882#issuecomment-1921340066

cc @SchSeba, @EmilienM

### DIFF
--- a/pkg/platforms/openstack/openstack.go
+++ b/pkg/platforms/openstack/openstack.go
@@ -102,8 +102,10 @@ type OSPDeviceInfo struct {
 	NetworkID  string
 }
 
-func New() OpenstackInterface {
-	return &openstackContext{}
+func New(hostManager host.HostManagerInterface) OpenstackInterface {
+	return &openstackContext{
+		hostManager: hostManager,
+	}
 }
 
 // GetOpenstackData gets the metadata and network_data
--- a/pkg/platforms/platforms.go
+++ b/pkg/platforms/platforms.go
@@ -1,8 +1,10 @@
 package platforms
 
 import (
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/host"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms/openshift"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms/openstack"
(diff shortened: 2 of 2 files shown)


pkg/platforms/openstack/openstack.go
```diff
@@ -103,7 +104,11 @@ type OSPDeviceInfo struct {
 }
 
 func New() OpenstackInterface {
```
adrianchiris, line 106: New should accept hostManager interface IMO.

similarly to what we do with interfaces and their impl in host pkg

SchSeba: +1 it will also help us if we need to pass a mock for unit-tests

you will only need to add the hostManager creation here https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/5887e7522ca8e1ebb13587bce6dafe43d000a992/main.go#L141

the other two places where we call the platform NewDefaultPlatformHelper() already have hostHelper created

zeeke: In service.go and start.go we create the HostHelper, not the HostManager. 

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/5887e7522ca8e1ebb13587bce6dafe43d000a992/cmd/sriov-network-config-daemon/service.go#L105

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/5887e7522ca8e1ebb13587bce6dafe43d000a992/cmd/sriov-network-config-daemon/start.go#L181

I'm not 100% getting the difference between these actors, but I'd leave the helpers's `New` function without arguments, and all the other interfaces get created by injecting the dependencies. WDYT?

adrianchiris: > and all the other interfaces get created by injecting the dependencies. WDYT?

you will not be able to do so if the test is in a different package. e.g `platform_test`.
that said, this is not the case ATM. so im fine with either.
we can always change later.

adrianchiris (approved):

PR #614, opened 2024-02-02
[switchdev 5/x] Add skipVFconfiguration option for generic_plugin

ykulazhenkov:
Add skipVFconfiguration option for generic_plugin

If `WithSkipVFConfiguration()` option is passed during the plugin initialization, then the plugin will configure PFs and create VFs, but will skip VF configuration phase.

This option is used by the systemd service during the `pre` phase. At this point we only need to create VFs but not configure them(e.g. bind to driver, create VDPA device, etc)

In the `post` phase the plugin will be called without this option and as a result will execute full plugin flow, meaning it will re-check that VFs are create and will continue configurations of VFs.

cc @adrianchiris  @zeeke  @SchSeba

### DIFF
--- a/cmd/sriov-network-config-daemon/service.go
+++ b/cmd/sriov-network-config-daemon/service.go
@@ -206,14 +206,12 @@ func getPlugin(setupLog logr.Logger, phase string,
 	case consts.Baremetal:
 		switch phase {
 		case PhasePre:
-			configPlugin, err = newGenericPluginFunc(hostHelpers)
-			if err != nil {
-				return nil, fmt.Errorf("failed to create generic plugin for %v", err)
-			}
+			configPlugin, err = newGenericPluginFunc(hostHelpers, generic.WithSkipVFConfiguration())
 		case PhasePost:
-			// TODO add initialization for the generic plugin for the post phase
-			setupLog.Info("post phase is not implemented for generic plugin yet, skip")
-			return nil, nil
+			configPlugin, err = newGenericPluginFunc(hostHelpers)
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to create generic plugin for %v", err)
 		}
 	case consts.VirtualOpenStack:
 		switch phase {
--- a/cmd/sriov-network-config-daemon/service_test.go
+++ b/cmd/sriov-network-config-daemon/service_test.go
@@ -236,6 +236,11 @@ var _ = Describe("Service", func() {
 				"/etc/sriov-operator/sriov-interface-result.yaml":   getTestResultFileContent("InProgress", ""),
 			},
 		})
+		hostHelpers.EXPECT().DiscoverSriovDevices(hostHelpers).Return([]sriovnetworkv1.InterfaceExt{{
+			Name: "enp216s0f0np0",
+		}}, nil)
+		genericPlugin.EXPECT().OnNodeStateChange(newNodeStateContainsDeviceMatcher("enp216s0f0np0")).Return(true, false, nil)
+		genericPlugin.EXPECT().Apply().Return(nil)
 		Expect(runServiceCmd(&cobra.Command{}, []string{})).NotTo(HaveOccurred())
 		testHelpers.GinkgoAssertFileContentsEquals("/etc/sriov-operator/sriov-interface-result.yaml",
 			string(getTestResultFileContent("Succeeded", "")))


pkg/host/internal/sriov/sriov.go
```diff
@@ -417,25 +430,53 @@ func (s *sriov) ConfigSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *
...
 						"driver", group.DeviceType, "device", addr)
 					return err
 				}
 			}
 		}
 	}
+	return nil
+}
+
+func (s *sriov) ConfigSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt, skipVFConfiguration bool) error {
```
zeeke, line 455: can we make this function private and remove it from the interface?

zeeke (commented):
Left one minor comment. Overall PR LGTM

pkg/host/internal/sriov/sriov.go
ykulazhenkov: Sounds good. I made the function private.

ykulazhenkov:
@zeeke thx for the review, I addressed your comment.

api/v1/helper.go
```diff
@@ -253,7 +253,12 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 			for _, groupSpec := range ifaceSpec.VfGroups {
 				if IndexInRange(vfStatus.VfID, groupSpec.VfRange) {
 					ingroup = true
-					if groupSpec.DeviceType != consts.DeviceTypeNetDevice {
+					if vfStatus.Driver == "" {
+						log.V(2).Info("NeedToUpdateSriov(): Driver needs update - has no driver",
+							"desired", groupSpec.DeviceType)
+						return true
+					}
+					if groupSpec.DeviceType != "" && groupSpec.DeviceType != consts.DeviceTypeNetDevice {
```
adrianchiris, line 261: what is the reason for changing this check ?

pkg/host/internal/sriov/sriov.go
```diff
@@ -417,25 +430,53 @@ func (s *sriov) ConfigSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *
...
 						"driver", group.DeviceType, "device", addr)
 					return err
 				}
 			}
 		}
 	}
+	return nil
+}
+
+func (s *sriov) configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt, skipVFConfiguration bool) error {
```
adrianchiris, line 455: still need to give this one a bit more thought ...

api/v1/helper.go
ykulazhenkov: I think this check originally has a bug. This branch should run in case if device is not a network device. The problem is that empty  groupSpec.DeviceType also means network device.

adrianchiris: ack, i see.

empty DeviceType means the default which is "netdevice", mutating webhook sets this if not set by user.

pkg/host/internal/sriov/sriov.go
adrianchiris: We had an offline discussion about it. latest iteration is OK imo.
(had some concerns about how this plays out with Externally managed)

adrianchiris:
overall im LGTM.  lets have a final pass after rebase.

zeeke (approved):
LGTM

pkg/host/internal/sriov/sriov.go
```diff
@@ -258,8 +258,8 @@ func (s *sriov) DiscoverSriovDevices(storeManager store.ManagerInterface) ([]sri
 			iface.Name = name
 			iface.Mac = s.networkHelper.GetNetDevMac(name)
 			iface.LinkSpeed = s.networkHelper.GetNetDevLinkSpeed(name)
+			iface.LinkType = s.GetLinkType(name)
```
adrianchiris, line 261: we can get mtu, mac, link type from netlink Link obj and avoid a bunch of file operations.

not related to this PR.  i will submit a small refactor to DiscoverSriovDevices.

adrianchiris (approved):
LGTM!

Issue #592, opened 2024-01-23
Stop default creation of SriovOperatorConfig and Policy


adrianchiris:
Summarizing what was discussed in recent community meetings:

- [x] Stop the creation of SriovOperatorConfig by default
- [x] Stop the creation of SriovPolicy by default
- [x] Support only one SriovOperatorConfig named "default"
- [x] Move ownerRef of all components the operator creates to the "default" OperatorConfig CR instead of "default" SriovPolicy CR
- [x] Block creation of non "default" SriovOperatorConfig CR in webhook

adrianchiris:
Note on the last bullet : we already do so today

zeeke:
@adrianchiris can we close this? I see no other todo item left

adrianchiris:
Yes we can. 

closing as dependent PRs merged. Thx for the reviews !

PR #617, opened 2024-02-05
[Remove Default Objs  1/4] Dont create default SriovOperatorConfig

adrianchiris:
Different deployment may have different configurations for SriovOperatorConfig.

if for example `SriovOperatorConfig.spec.disablePlugins` contains a list of plugins to disable, the creation of the default SriovOperatorConfig object may have undesirable effects on the cluster (as a plugin that was suppose to be disabled, would be enabled by default until configuration has updated).

This PR removes the creation of SriovOperatorConfig object by the operator and leaves the end user to decide the desired configuration.

To preserve a similar UX, we move the creation of SriovOperatorConfig CR to helm with same defaults as created by the operator.
upgrade of existing deployments would simply use the existing SriovOperatorConfig CR.
Deployments that dont install the operator via Helm (e.g if installed via OLM) the end user is expected to create SriovOperatorConfig CR post installation.

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -106,6 +106,16 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(ctx context.Context, req ct
 		return reconcile.Result{}, err
 	}
 
+	// Fetch the default SriovOperatorConfig
+	defaultOpConf := &sriovnetworkv1.SriovOperatorConfig{}
+	if err := r.Get(ctx, types.NamespacedName{Namespace: vars.Namespace, Name: constants.DefaultConfigName}, defaultOpConf); err != nil {
+		if errors.IsNotFound(err) {
+			reqLogger.Info("default SriovOperatorConfig object not found, cannot reconcile SriovNetworkNodePolicies. Requeue.")
+			return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
 	// Fetch the SriovNetworkNodePolicyList
 	policyList := &sriovnetworkv1.SriovNetworkNodePolicyList{}
 	err = r.List(ctx, policyList, &client.ListOptions{})
@@ -124,10 +134,6 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(ctx context.Context, req ct
 	lo := &client.MatchingLabels{
 		"node-role.kubernetes.io/worker": "",
 		"kubernetes.io/os":               "linux",
-	}
-	defaultOpConf := &sriovnetworkv1.SriovOperatorConfig{}
-	if err := r.Get(ctx, types.NamespacedName{Namespace: vars.Namespace, Name: constants.DefaultConfigName}, defaultOpConf); err != nil {
-		return reconcile.Result{}, err
 	}
 	if len(defaultOpConf.Spec.ConfigDaemonNodeSelector) > 0 {
 		labels := client.MatchingLabels(defaultOpConf.Spec.ConfigDaemonNodeSelector)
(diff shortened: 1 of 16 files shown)


controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -106,6 +106,12 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(ctx context.Context, req ct
 		return reconcile.Result{}, err
 	}
 
+	// Fetch the default SriovOperatorConfig
```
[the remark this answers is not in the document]
adrianchiris: i mean its still named "default"  :) so i think the name is ok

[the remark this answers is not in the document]
adrianchiris: plus we dont want to effect existing deployments that already have this created.

adrianchiris:
Related to: #592

zeeke (commented):
Only one minor comment from my side.

Also, can't validate the helm configuration change, as I'm not very familiar with that. Maybe we can change how we deploy the operator during e2e tests (hack/run-e2e-conformance-virtual-cluster.sh) to helm in a future PR.

Once rebased, LGTM

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -106,6 +106,12 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(ctx context.Context, req ct
...
+	defaultOpConf := &sriovnetworkv1.SriovOperatorConfig{}
+	if err := r.Get(ctx, types.NamespacedName{Namespace: vars.Namespace, Name: constants.DefaultConfigName}, defaultOpConf); err != nil {
+		return reconcile.Result{}, err
```
ykulazhenkov, line 112: nit: Maybe we can log here something like "no operator config found, waiting for creation" and explicitly requeue with some delay, e.g. 5 seconds?

controllers/sriovoperatorconfig_controller.go
```diff
@@ -163,7 +139,12 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 // SetupWithManager sets up the controller with the Manager.
 func (r *SriovOperatorConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&sriovnetworkv1.SriovOperatorConfig{}).
+		For(&sriovnetworkv1.SriovOperatorConfig{}, ctrl_builder.WithPredicates(predicate.NewPredicateFuncs(func(object client.Object) bool {
```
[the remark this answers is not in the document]
adrianchiris: sure, will do that.

controllers/sriovnetworknodepolicy_controller.go
adrianchiris: i dont think we need to requeue, we will be called as soon as the object is created.

im thinking just to add log and return a successful reconcile WDYT ?

adrianchiris:
@zeeke @ykulazhenkov PTAL :)

controllers/sriovnetworknodepolicy_controller.go
ykulazhenkov: We are inside policy controller. This Get call is for OperatorConfig object. Maybe I'm missing something, but I don't see that policy controller watches for OperatorConfig object.

zeeke: @ykulazhenkov the scenario you're referring to is:
a. The user deploys the operator
b. the user creates an SriovNetworkNodePolicy
c. The user creates the SriovOperatorConfig

right?

In this case, the policy would not get reconciled until another SriovNetworkNodePolicy is created/modified, I think.

ykulazhenkov: @zeeke 

Yes, I'm referring to this case. 
My concern here is that if the SriovNetworkNodePolicy controller fails multiple times in a row because there is no Operator config, then retry with delay will be used. If there were many failures, then the retry interval would be quite large. I propose explicitly sending a message to the queue again with a non-error exit to ensure we will retry reconciliation quite soon. This means that we will do a periodic polling in the  SriovNetworkNodePolicy controller until operator's config is created. Current implementation will work just fine, but in some edge-cases it may introduce unneeded delay.

adrianchiris: >We are inside policy controller. This Get call is for OperatorConfig object. Maybe I'm missing something, but I don't see that policy controller watches for OperatorConfig object.

yes, you are right, i thought you are referring to the operator config controller  :\

adrianchiris: Added a check here. requeue after 5 seconds

ykulazhenkov (approved):

zeeke (approved):

PR #621, opened 2024-02-06
[Remove Default Objs  2/4] Set OwnerReference to `default` SriovOperatorConfig

adrianchiris:
We are deprecating the use of the default SriovNetworkNodePolicy.
This PR changes the OwnerReference of sriov-network-device-plugin and SriovNetworkNodePolicy which are created
by the operator to point to the default SriovOperatorConfig CR.

This PR depends on #617  Only the last 2 commits are related (the rest will be rebased away once #617 merges)

PR #622, opened 2024-02-06
Remove default policy

adrianchiris:
There is no need to create it anymore

This PR depends on #621 , only the last commit is relevant, the rest will be rebased away


adrianchiris:
Related to: #592

ykulazhenkov (commented):
I have no comments, LGTM after rebase

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -192,11 +169,23 @@ func (r *SriovNetworkNodePolicyReconciler) SetupWithManager(mgr ctrl.Manager) er
...
 		},
+		GenericFunc: func(ctx context.Context, e event.GenericEvent, q workqueue.RateLimitingInterface) {
+			log.Log.WithName("SriovNetworkNodePolicy").
+				Info("Enqueuing sync for generic event", "resource", e.Object.GetName())
+			qHandler(q)
+		},
 	}
 
+	// send initial sync event to trigger reconcile when controller is started
+	var eventChan = make(chan event.GenericEvent, 1)
```
SchSeba, line 180: I think a better solution here will be to move the logic for the nodeState creation into the sriovOperatorConfig controller this way we know the nodeSelector for the config daemons

SchSeba (commented):
just one comment.

controllers/sriovnetworknodepolicy_controller.go
adrianchiris: Shouldn't same controller handle the create update and removal of the obj ?

If we go this way, IMO we should handle this in a separate PR IMO.

PR #623, opened 2024-02-06
Set OwnerReference to sriov-device-plugin configmap

adrianchiris:
The configmap is created by the operator but never deleted. Set OwnerReference to SriovOperatorConfig default CR to ensure its deletion.

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -147,7 +147,7 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(ctx context.Context, req ct
 		return reconcile.Result{}, err
 	}
 	// Sync Sriov device plugin ConfigMap object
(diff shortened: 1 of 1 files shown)


controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -221,6 +222,8 @@ func (r *SriovNetworkNodePolicyReconciler) syncDevicePluginConfigMap(ctx context
 		},
 		Data: configData,
 	}
+	controllerutil.SetControllerReference(dc, cm, r.Scheme)
```
zeeke, line 225: Are we ignoring the returned error on purpose? Not sure if it's worth to return the error, bit a log message would be useful.

adrianchiris: Will add a check as we do in other places for consitency.

technically there is no chance for errors to occur here.

e0ne (approved):

PR #624, opened 2024-02-06
[Remove Default Objs  4/4] Update Webhook

adrianchiris:
As the operator no longer creates default
SriovOperatorConfig and SriovNetworkNodePolicy
the webhook is updated in the following Manner:

Validating:
- Allow deletion of default config/policy.
- Block create/update of non default config CR

Mutating:
- keep skipping default but add a comment to mark as deprecated.

NOTE for reviewers: should be merged after #622

### DIFF
--- a/pkg/webhook/validate.go
+++ b/pkg/webhook/validate.go
@@ -35,12 +35,12 @@ func validateSriovOperatorConfig(cr *sriovnetworkv1.SriovOperatorConfig, operati
 	log.Log.V(2).Info("validateSriovOperatorConfig", "object", cr)
 	var warnings []string
 
-	if cr.GetName() != consts.DefaultConfigName {
-		return false, warnings, fmt.Errorf("only default SriovOperatorConfig is used")
+	if operation == v1.Delete {
+		return true, warnings, nil
 	}
 
-	if operation == v1.Delete {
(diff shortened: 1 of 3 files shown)


adrianchiris:
Related to #592

pkg/webhook/validate.go
```diff
@@ -35,12 +35,12 @@ func validateSriovOperatorConfig(cr *sriovnetworkv1.SriovOperatorConfig, operati
 	log.Log.V(2).Info("validateSriovOperatorConfig", "object", cr)
 	var warnings []string
 
-	if cr.GetName() != consts.DefaultConfigName {
-		return false, warnings, fmt.Errorf("only default SriovOperatorConfig is used")
+	if operation == v1.Delete {
```
e0ne, line 38: I would like to get more feedback here: from one side, it should be OK to re-create default policy, from other, I would like to allow only update once we've got default policy to not introduce possible race conditions and unnecessary reconcile loops.

I'll press `Request changes` button just to get more attention to this from other maintainers, feel free to ignore my vote

e0ne (changes requested):

e0ne (approved):

ykulazhenkov (approved):

adrianchiris:
@zeeke @SchSeba could you take a quick look at this one ?  the last of the series :)

PR #625, opened 2024-02-07
Remove bool ptr

adrianchiris:
Change boolean attribute in SriovOperatorConfig from pointer to concrete.
Main motivation is to simplify code a bit


e0ne (approved):
LGTM. Thanks for making our code more simple

adrianchiris:
@SchSeba could you take a look, this is a short one.

SchSeba (commented):
general question this will mean that it will be false by default so if someone create the sriovOperatorConfig without this fields it will be false and in case of a pointer it will not exist no?

adrianchiris:
> general question this will mean that it will be false by default so if someone create the sriovOperatorConfig without this fields it will be false and in case of a pointer it will not exist no?

yes, but we have no use-case for a "missing" value.
moreover we always create the default config with values set today.

SchSeba (approved):
yep you are right I have no problem moving this one out of a pointer and if in the future this is needed we can revert it. from the CRD point of view there is no change so it's transparent to the user

PR #626, opened 2024-02-07
Add 'Parallel SR-IOV configuration' design document

SchSeba:


doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+### Workflow Description
+A new Drain controller will be introduced to manage node drain and cordon procedures. That means we don't need to do
+drain and use `drain lock` in config daemon anymore. The overall drain process will be covered by the following states:
+
+```golang
+NodeDrainAnnotation             = "sriovnetwork.openshift.io/state"
+NodeStateDrainAnnotation        = "sriovnetwork.openshift.io/desired-state"
+NodeStateDrainAnnotationCurrent = "sriovnetwork.openshift.io/current-state"
```
zeeke, line 48: Can you state which annotation is written by which actor? 
config-daemon writes Node["sriovnetwork.openshift.io/state"] and reads both node states?
the drain controller writes SriovNetworkNodeState only?

zeeke (commented):
Left few comments

doc/design/parallel-node-config.md
SchSeba: done :)

SchSeba:
Hi @e0ne @adrianchiris @ykulazhenkov please review this PR I will like to merge this one before the parallel drain implementation is done

e0ne (approved):
LGTM. Thanks for this PR, @SchSeba

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+Once this change will be implemented `SriovNetworkPoolConfig` configuration will be applied both to vanilla Kubernetes
+and OpenShift clusters.
+
+### Implementation Constraints
+
+Node can only be part of one pool beside the default pool that catch all the nodes.
+
+### Upgrade & Downgrade considerations
```
adrianchiris, line 142: do we want to cover upgrade / downgrade in the middle of node sriov config ?

e.g config daemon started cordon drain and we upgrade the operator ?

i guess we should say "dont do that"

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,214 @@
...
+and write the `sriovnetwork.openshift.io/current-state` annotation in the SriovNetworkNodeState.
+
+Config Daemon will read SriovNetworkNodeState annotation `sriovnetwork.openshift.io/desired-state` and write both
+node annotation, `sriovnetwork.openshift.io/state` and SriovNetworkNodeState annotation `sriovnetwork.openshift.io/desired-state`.
+
+*NOTE:* In the future we are going to drop the node annotation and only use the SriovNetworkNodeState
+
+Draining procedure:
+
+1. config daemon mark the node as `Drain_Required` or `Reboot_Required`
```
adrianchiris, line 67: nit could you add the specific way it does so.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,214 @@
...
+2. operator drain controller reconcile loop find the node SriovNetworkPoolConfig
+   1. if number of `Draining` nodes is great or equal to the `MaxUnavailable` the operator will re-queue the request
+   2. if number of `Draining` nodes is lower than the `MaxUnavailable` the operator will start the draining process
+5. on Openshift platform we will pause the machine config pool related to the node
+6. the operator will start the drain process
+   1. if `Drain_Required` the operator will remove ONLY pods used sriov devices
```
adrianchiris, line 73: Thats not really related to parallel sriov config.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,214 @@
...
+    NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
+
+    // maxUnavailable defines either an integer number or percentage
+    // of nodes in the pool that can go Unavailable during an update.
+    //
+    // A value larger than 1 will mean multiple nodes going unavailable during
+    // the update, which may affect your workload stress on the remaining nodes.
+    // Drain will respect Pod Disruption Budgets (PDBs) such as etcd quorum guards,
+    // even if maxUnavailable is greater than one.
+    MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
```
adrianchiris, line 101: this is relevant for drain/parallel configuration only right ?
we may have additional parameters down the road. 

maybe put this one under "DrainConfig" similarly to the existing "OvsHardwareOffloadConfig"

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,214 @@
...
+OVS Hardware Offloading. We can extend it to add configuration for the drain
+pool. E.g.:
+
+```golang
+// SriovNetworkPoolConfigSpec defines the desired state of SriovNetworkPoolConfig
+type SriovNetworkPoolConfigSpec struct {
+    ...
+	
+    // nodeSelector specifies a label selector for Nodes
+    NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
```
adrianchiris, line 92: will this one affect pool config controller ?

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,214 @@
...
+      - key: some-label
+        operator: In
+        values:
+          - val-2
+    - matchExpressions:
+      - key: other-label
+        operator: "Exists"
+```
+
+Default pool should be always created with the lowest priority:
```
adrianchiris, line 124: where is the priority field ?

adrianchiris, line 124: also i thought we want to get rid of "default" objects created by the operator.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,214 @@
...
+spec:
+  maxUnavailable: 1
+```
+
+Once this change will be implemented `SriovNetworkPoolConfig` configuration will be applied both to vanilla Kubernetes
+and OpenShift clusters.
+
+### Implementation Constraints
+
+Node can only be part of one pool beside the default pool that catch all the nodes.
```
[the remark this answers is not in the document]
SchSeba: For now, I would like to keep this one as simple as possible, there is no priority in the API for now if we see a requirement we can add it in the future for sure.

If a node is part of one or more pools we will see it in the logs (maybe also we can add an event to the nodeState in the future) and the node will not be able to drain waiting for the node to be part of only one pool

doc/design/parallel-node-config.md
SchSeba: it's not possible to say don't do it because there are systems that as part of the upgrade for example OCP we reboot the nodes to install a new version of the operating system so you can have a state where one of the daemons is doing a configuration

doc/design/parallel-node-config.md
SchSeba: but it's part of the changes do you want me to remove it?

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,214 @@
...
+
+```yaml
+apiVersion: v1
+kind: SriovNetworkPoolConfig
+metadata:
+  name: pool-1
+  namespace: network-operator
+spec:
+  maxUnavailable: "20%"
+  nodeSelectorTerms:
```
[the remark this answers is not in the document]
SchSeba: I will like to have this as `nodeSelector` as we have in other places of the operator already

doc/design/parallel-node-config.md
SchSeba: what do you mean?
for now the state is that I decople between the OVSHWoffload and the drain from the CR they can exist together I also have a webhook to block it

doc/design/parallel-node-config.md
SchSeba: check my preview comment do you think we need the priority?
This is different from the reason we add it to the SriovPolicy.

SchSeba: I don't know if we want to drop the default for a pool.
This is to explicitly say `maxUnavailable=1`

also in the future, I would like to add an annotation or field to the NodeState to say what pool the node is part of
and in the pool how many nodes are part of a pool and thinks like that

doc/design/parallel-node-config.md
SchSeba: you are right that is for draining but a configuration of a pool. and I must say I am not familiar with having `DrainConfig` with only field inside it just makes the API more complicated that he needs to be if you ask me :)

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,214 @@
...
+Config Daemon will read SriovNetworkNodeState annotation `sriovnetwork.openshift.io/desired-state` and write both
+node annotation, `sriovnetwork.openshift.io/state` and SriovNetworkNodeState annotation `sriovnetwork.openshift.io/desired-state`.
+
+*NOTE:* In the future we are going to drop the node annotation and only use the SriovNetworkNodeState
+
+Draining procedure:
+
+1. config daemon mark the node as `Drain_Required` or `Reboot_Required`
+2. operator drain controller reconcile loop find the node SriovNetworkPoolConfig
+   1. if number of `Draining` nodes is great or equal to the `MaxUnavailable` the operator will re-queue the request
```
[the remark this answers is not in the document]
SchSeba: done

doc/design/parallel-node-config.md
adrianchiris: ack, so need to add some info about it IMO.

doc/design/parallel-node-config.md
adrianchiris: yes please, and preferably address it in a separate PR on top if it makes sense.

we should either remove from design and implementation or keep in both. i prefer the former and a folllowup PR to fine-tune the draining process.

doc/design/parallel-node-config.md
adrianchiris: what about :

> we may have additional parameters down the road.

doc/design/parallel-node-config.md
adrianchiris: im fine with requiring an explicit pool but i dont want the operator to be the one creating it (which means we need to delete it at some point)

as we do try to avoid creating default objects (e.g sriov policy and operator config)

adrianchiris (commented):
Alternative API proposal:

We may want to support other set of configurations per group of nodes hence i think we should have each one in its own struct.

`OtherNodeLevelConfig` and `GlobalOvsConfig` are for demonstration purposes

```
type SriovNetworkPoolConfigSpec struct {
	// OvsHardwareOffloadConfig describes the OVS HWOL configuration for selected Nodes
	OvsHardwareOffloadConfig *OvsHardwareOffloadConfig `json:"ovsHardwareOffloadConfig,omitempty"`  <-- this is a pointer now

	// NodeSelector only valid for the fields below
	NodeSelector map[string]string
	// Drain configuration parameters
	DrainConfig *DrainConfiguration
	// Other Node level configuration parameters
	OtherNodeLevelConfig *OtherConfiguration
	// OVS configuration (in global table) to be applied to the group of nodes
	GlobalOvsConfig *GlobalOvsConfig
}
```

- node selector, defines the nodes that belong to the pool
- node selector does not apply to `OvsHardwareOffloadConfig` config field (preserve  current behaviour)
- a node can have exactly one configuration of a certain type
  - e.g a node cannot be selected by two pools with `DrainConfig` specified
- a pool may have more than one configuration field set
- for new configuration types, if MCP ref is needed MCP label on node can be used in NodeSelector.

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,216 @@
...
+RebootRequired                  = "Reboot_Required"
+Draining                        = "Draining"
+DrainComplete                   = "DrainComplete"
+```
+
+Drain controller will watch for Node annotation, `sriovnetwork.openshift.io/state` 
+and SriovNetworkNodeState annotation `sriovnetwork.openshift.io/desired-state`
+and write the `sriovnetwork.openshift.io/current-state` annotation in the SriovNetworkNodeState.
+
+Config Daemon will read SriovNetworkNodeState annotation `sriovnetwork.openshift.io/desired-state` and write both
```
adrianchiris, line 60: > sriovnetwork.openshift.io/desired-state

you mean sriovnetwork.openshift.io/current-state ?

in #555 config daemon never reads `sriovnetwork.openshift.io/desired-state`

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,216 @@
...
+2. operator drain controller reconcile loop find the node SriovNetworkPoolConfig
+   1. if number of `Draining` nodes is great or equal to the `MaxUnavailable` the operator will re-queue the request
+   2. if number of `Draining` nodes is lower than the `MaxUnavailable` the operator will start the draining process
+   and annotate the SriovNetworkNodeState annotation `sriovnetwork.openshift.io/current-state` with `Draining`
+5. on Openshift platform we will pause the machine config pool related to the node
+6. the operator will start the drain process
+   1. if `Drain_Required` the operator will remove ONLY pods used sriov devices
+   2. if `Reboot_Required` the operator will remove ALL the pods on the system
+9. operator moves the `sriovnetwork.openshift.io/current-state` annotation to `DrainComplete`
+10. daemon will continue to the configuration when it's done it will move back the `sriovnetwork.openshift.io/state` annotation to `Idle`
```
adrianchiris, line 78: you mean both `sriovnetwork.openshift.io/state` on node and `sriovnetwork.openshift.io/desired-state` on SriovNetworkNodeState

doc/design/parallel-node-config.md
SchSeba: The thing is we are going to deprecate `OvsHardwareOffloadConfig` the pool CRD should never have contained that.
everything related to a pool of nodes should go under this CRD if we decide in the future to have priority or `skipDrainOnReboot` or maybe even the creation of the switches for switchdev part for example should be part of the CRD directly without another layer inside of it.
That is my take if you ask me

doc/design/parallel-node-config.md
SchSeba: sure we can take it to the next meeting I will put it on the agenda

doc/design/parallel-node-config.md
SchSeba: ok I am fine not having the default pool created let me update the doc

doc/design/parallel-node-config.md
SchSeba: if possible I will like to leave it here as the  implementation is already in place and the change is super small we just check if we have `Drain_required` or `Reboot_requried` but if you insist I can remove it :)

doc/design/parallel-node-config.md
SchSeba: yep sorry about that

SchSeba:
Hi @adrianchiris thanks for the comments!

about

> Alternative API proposal:
> 
> We may want to support other set of configurations per group of nodes hence i think we should have each one in its own struct.
> 
> OtherNodeLevelConfig and GlobalOvsConfig are for demonstration purposes
> 
> type SriovNetworkPoolConfigSpec struct {
> 	// OvsHardwareOffloadConfig describes the OVS HWOL configuration for selected Nodes
> 	OvsHardwareOffloadConfig *OvsHardwareOffloadConfig `json:"ovsHardwareOffloadConfig,omitempty"`  <-- this is a pointer now
> 
> 	// NodeSelector only valid for the fields below
> 	NodeSelector map[string]string
> 	// Drain configuration parameters
> 	DrainConfig *DrainConfiguration
> 	// Other Node level configuration parameters
> 	OtherNodeLevelConfig *OtherConfiguration
> 	// OVS configuration (in global table) to be applied to the group of nodes
> 	GlobalOvsConfig *GlobalOvsConfig
> }
> node selector, defines the nodes that belong to the pool
> node selector does not apply to OvsHardwareOffloadConfig config field (preserve current behaviour)
> a node can have exactly one configuration of a certain type
> e.g a node cannot be selected by two pools with DrainConfig specified
> a pool may have more than one configuration field set
> for new configuration types, if MCP ref is needed MCP label on node can be used in NodeSelector.


I don't think it's a good idea to have multiple pool with specific stuff I only did that for the `ovsHardwareOffloadConfig` because it's already in the API BUT we must deprecate that field there is no reason after the great work we are doing with the new switchdev API .

and having something like:

```
type SriovNetworkPoolConfigSpec struct {
	// OvsHardwareOffloadConfig describes the OVS HWOL configuration for selected Nodes
	OvsHardwareOffloadConfig OvsHardwareOffloadConfig `json:"ovsHardwareOffloadConfig,omitempty"`  <-- this one will be deprecated

	// NodeSelector only valid for the fields below
	NodeSelector map[string]string
	// maxUnavailable defines either an integer number or percentage
       SkipDrainOnReboot bool
       OvsSwitches list[<struct for PF : switchName>]
       LinuxBridges list[<struct for PF : switchName>]
       EnabledPlugins list[string]
       logLevel int
       CDIsupport bool
       DaemonMode string
}
```

And any other configuration that a pool of nodes will have in common

That is my general goal for the Pool CRD. having the SriovOperatorConfig is good when you have a single node or a small cluster, but if you install the operator on a cluster with a high number of nodes I would like to allow the user a more granular configuration using pools

let me know what do you think about it :)

doc/design/parallel-node-config.md
SchSeba: do we want in this PR or the next one that will address the transition between the two annotations?

doc/design/parallel-node-config.md
SchSeba: as we spoke in the community meeting we are going to leave everything that is related to a node configure in this CR, and a node will be part of only one pool example of future extensions to the API

```
type SriovNetworkPoolConfigSpec struct {
	// OvsHardwareOffloadConfig describes the OVS HWOL configuration for selected Nodes
	OvsHardwareOffloadConfig OvsHardwareOffloadConfig `json:"ovsHardwareOffloadConfig,omitempty"`  <-- this one will be deprecated

       // NodeSelector only valid for the fields below
       NodeSelector map[string]string
       // maxUnavailable defines either an integer number or percentage
       SkipDrainOnReboot bool
       EnabledPlugins list[string]
       logLevel int
       CDIsupport bool
       DaemonMode string
}
```

doc/design/parallel-node-config.md
adrianchiris: as there is a potential to have additional configuration down the road, i think its worth "partitioning" according to domains now and not later.

if you feel strongly about keeping MaxUnavailable top level im OK.

doc/design/parallel-node-config.md
adrianchiris: controller behaviour WRT handling such case should be mentioned in this PR.
also we may have +1 node being cordoned in this case IMO, we should mention this as well.

moving to the use of a single object (nodestate) is a different topic (and different PR).

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,207 @@
...
+### Implementation Constraints
+
+Node can only be part of one pool. if the node is not part of any node it will be allocated
+to a virtual default pool with `maxUnavailable` of 1
+
+### Upgrade & Downgrade considerations
+After operator upgrade we have to support `sriovnetwork.openshift.io/state` node annotation and `sriovnetwork.openshift.io/desired-state`
+annotation in the `sriovNetworkNodeState`. in the future we are going to migrate to only using the annotation in the `sriovNetworkNodeState`
+
+There is no change in upgrade from the user point of view as we are going to create a default pool with `maxUnavailable` 1
```
adrianchiris, line 139: i Thought we arent going to create this default obj.

SchSeba: yep yep no default pool sorry I miss that line :)

done

SchSeba:
Hi @e0ne @zeeke @ykulazhenkov let me know if I am missing something or we can move forward and merge the proposal.
Thanks! :)

zeeke (approved):
Only one nit. The design looks good to me. If we found something to adjust, we can fix it in the implementation PR IMO

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,209 @@
...
+place which will reduce chances of race conditions and bugs that were encountered before in sriov-network-config-daemon
+with draining.
+
+### Use Cases
+
+### Goals
+* Number of drainable nodes should be 1 by default
+* Number of drainable nodes should be configured by pool
+* Nodes pool should be defined by node selector
+* A node could be included into several pools. In this case pool should be
```
ykulazhenkov, line 29: Let's remove this goal, we decided not to implement priorities.

ykulazhenkov (approved):
added on small comment.

Looks good for me

doc/design/parallel-node-config.md
adrianchiris: as its already there and both the design and implementation already went through several iterations lets keep it.

adrianchiris (commented):
Consider me as LGTM once my remaining comment is addressed

doc/design/parallel-node-config.md
```diff
@@ -0,0 +1,209 @@
...
+  - adrianchiris
+  - e0ne
+creation-date: 18-07-2023
+last-updated: 18-07-2023
+---
+
+# Parallel SR-IOV configuration
+
+## Summary
+Allow SR-IOV Network Operator to configure more than one node at the moment.
```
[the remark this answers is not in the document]
SchSeba: Thanks done

SchSeba:
Hi @e0ne @ykulazhenkov can we merge this one?

ykulazhenkov:
> Hi @e0ne @ykulazhenkov can we merge this one?

yes, looks good for me

doc/design/parallel-node-config.md
SchSeba: the reason we are doing this in two steps is to not have this issue.
meaning that in the future when we drop the node label if it's in an upgrade case the existing operator will already have the code in place to add both annotations in parallel so you can't have a problem that some daemons use the old node annotation and some use the new annotation.

adrianchiris: in our current code, the config daemon will drain if no other node has `sriovnetwork.openshift.io/state: Draining` annotation.

i went over #555 and we never set "Draining" on the node object. so basically i dont understand why we need to set the node annotations (to a value which old config daemon is not aware of).

see: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/82a6d6fdce71bd88a0d9368fb1750488e9a8e4e2/pkg/daemon/daemon.go#L381C49-L381C73

which is why i wanted to have more details about it.

e0ne (approved):
LGTM

PR #628, opened 2024-02-12
[switchdev 7/x] Add function EnableHwTcOffload to the network pkg

ykulazhenkov:
This function ensures that hw-tc-offload feature is enabled for the interface if the interface supports it.
required to be compatible with the [current switchdev implementation ](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/08a79f79b2d281ab835407b84ee058587013c5ec/bindata/manifests/switchdev-config/files/switchdev-configuration-before-nm.sh.yaml#L76)

uses [`github.com/safchain/ethtool`](https://github.com/safchain/ethtool) lib

cc @adrianchiris @SchSeba @zeeke 

depends on #633 

only last commit to review


pkg/host/internal/lib/ethtool/ethtool.go
```diff
@@ -0,0 +1,51 @@
...
+	// FeatureNames shows supported features by their name.
+	FeatureNames(intf string) (map[string]uint, error)
+	// Change requests a change in the given device's features.
+	Change(intf string, config map[string]bool) error
+}
+
+type libWrapper struct{}
+
+// Features retrieves features of the given interface name.
+func (w *libWrapper) Features(intf string) (map[string]bool, error) {
```
e0ne, line 24: let's use the same variable name in all places to specify interface name. E.g. `ifaceName`

pkg/host/internal/network/network.go
```diff
@@ -290,3 +293,43 @@ func (n *network) SetDevlinkDeviceParam(pciAddr, paramName, value string) error
 	}
 	return nil
 }
+
+// EnableHwTcOffload makes sure that hw-tc-offload feature is enabled if device supports it
+func (n *network) EnableHwTcOffload(ifaceName string) error {
+	log.Log.V(2).Info("EnableHwTcOffload(): enable offloading", "device", ifaceName)
+	featureName := "hw-tc-offload"
```
e0ne, line 300: it's used only once here and in tests so I'm OK to have it here instead of constants

e0ne, line 300: nit: featureName -> hwOffloadfeatureName

e0ne (changes requested):

pkg/host/internal/lib/ethtool/ethtool.go
ykulazhenkov: done

pkg/host/internal/network/network.go
ykulazhenkov: yes, I think there is no need to put the feature name to a const

adrianchiris (commented):
LGTM! 

will require rebase once [switchdev 6/x] is merged

pkg/host/internal/network/network.go
```diff
@@ -286,3 +289,43 @@ func (n *network) SetDevlinkDeviceParam(pciAddr, paramName, value string) error
...
+	log.Log.V(2).Info("EnableHwTcOffload(): enable offloading", "device", ifaceName)
+	hwTcOffloadFeatureName := "hw-tc-offload"
+
+	knownFeatures, err := n.ethtoolLib.FeatureNames(ifaceName)
+	if err != nil {
+		log.Log.Error(err, "EnableHwTcOffload(): can't list supported features", "device", ifaceName)
+		return err
+	}
+	if _, isKnown := knownFeatures[hwTcOffloadFeatureName]; !isKnown {
+		log.Log.V(2).Info("EnableHwTcOffload(): can't enable feature, feature is not supported", "device", ifaceName)
```
zeeke, line 304: I think this can log with `V(0)`, or it can even return an error. 

It really depends on how you're going to use this function, but if the user set a flag somewhere to enable HwTcOffload, it is expected to work or to get an error somewhere. WDYT?
Same for L329

ykulazhenkov: agree, I fixed the log level

zeeke (approved):
LGTM

PR #634, opened 2024-02-13
Optmize node watcher in daemon

alan-kut:
Node watcher is very heavy when watching
all the nodes, not only the current one.
Nodes are regularly updated, e.g. by kubelet.

Watch only the current node all the time and
other nodes only when holding the drain lock.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -98,8 +98,6 @@ type Daemon struct {
 
 	disableDrain bool
 
-	nodeLister listerv1.NodeLister
-
 	workqueue workqueue.RateLimitingInterface
 
 	mcpName string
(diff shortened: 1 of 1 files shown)


adrianchiris:
@alan-kut currently there is work around moving the drain logic to a centralized controller[1]

I think that part of this work we can limit the node informer to just the current node. @SchSeba WDYT ?

[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/555

SchSeba:
sure we can do it after we merge https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/555

SchSeba:
Hi @alan-kut thanks for the PR!

so the idea base on https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/555 is that we are not going to watch the node anymore the daemon will only watch the nodeState and in the future also the operator will not watch the nodes as we are going to move the drain request annotation to the nodeState only

SchSeba:
Hi @alan-kut can you take a look on the operator now that we remove the nodes update from the config daemon we have it only on the controller

PR #637, opened 2024-02-16
Fix indentation for --parallel-nic-config flag in the daemon manifest

ykulazhenkov:
Fix indentation for --parallel-nic-config flag in the daemon manifest.
Without the fix template produces invalid YAML if `parallelNicConfig` feature gate is enabled.

cc @e0ne


e0ne (approved):

zeeke (approved):
LGTM

thanks for fixing this

PR #640, opened 2024-02-20
Add "Software bridge management" design doc

ykulazhenkov:
This document contains a proposal to add limited support for software bridges configuration to the **sriov-network-operator**.

This feature assumes integration with [ovs-cni](https://github.com/k8snetworkplumbingwg/ovs-cni) and [accelerated-bridge-cni](https://github.com/k8snetworkplumbingwg/accelerated-bridge-cni).

Depends on [_switchdev and systemd modes refactoring_](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/doc/design/switchdev-refactoring.md) feature

cc @adrianchiris @zeeke @SchSeba

### DIFF
--- a/doc/design/software-bridge-management.md
+++ b/doc/design/software-bridge-management.md
@@ -1,0 +1,518 @@
+---
+title: software bridge management
+authors:
+  - ykulazhenkov
+reviewers:
+creation-date: 15-02-2024
+last-updated: 15-02-2024
+---
+
+# Software bridge management
+
+## Summary
+
+When NIC is configured to switchdev mode, a VF representor net device is created for each VF on it. 
+These representors are used by a software switch (OVS, Linux bridge) to control traffic and configure hardware offloads. 
+The software bridge is an essential part of using NICs in switchdev mode.
+
+**sriov-network-operator** can set switchdev mode for a NIC and create VFs on it, 
+but it doesn't provide any functionality to create and configure software bridges. 
+
+This document contains a proposal to add limited support for software bridges configuration to the **sriov-network-operator**.
+
+This feature assumes integration with [ovs-cni](https://github.com/k8snetworkplumbingwg/ovs-cni) and 
+[accelerated-bridge-cni](https://github.com/k8snetworkplumbingwg/accelerated-bridge-cni).
+
+Depends on [_switchdev and systemd modes refactoring_](switchdev-refactoring.md) feature.
+
+## Motivation
+
+SRIOV Legacy mode is no longer actively developed, and we need to encourage users to migrate to switchdev mode,
+which is actively developed and will continue to receive new features and improvements.
+
+To promote switching to switchdev configurations, we need to provide a nice UX for the end user.
+This requires providing an easy way to configure software switches, which are prerequisites for NICs in switchdev mode.
+
+
+### Use Cases
+
+* As a user, I expect that **sriov-network-operator** will install `ovs-cni` and `accelerated-bridge-cni` to hosts.
+* As a user, I want to create `OVSNetwork` CR, which will result in creation of `NetworkAttachmentDefinition` CR that uses
+`ovs-cni` and contains required resource request.
+* As a user, I want to create `BridgeNetwork` CR which will result in creation of `NetworkAttachmentDefinition` CR that uses
+`accelerated-bridge-cni` and contains the required resource requests.
+* As a user, I want to define configuration for software bridges inside the `SriovNetworkNodePolicy` CR and expect that the
+operator will create required bridges, configure them, and attach uplinks (physical functions).
+
+### Goals
+
+* handle installation of `ovs-cni` and `accelerated-bridge-cni`
+* add `OVSNetwork` and `BridgeNetwork` CRDs as an API for end users to simplify creation of `NetworkAttachmentDefinition` CR
+for `ovs-cni` and `accelerated-bridge-cni`
+* extend `SriovNetworkNodePolicy` CRD to support configuration of software bridges (bridge-level configuration)
+* extend `SriovNetworkNodeState` (spec and status) CRD to support configuration of software bridges (bridge-level configuration)
+* support configuration of software bridge in both modes (operator's `configurationMode` setting): `daemon` and `systemd`
+* implementation should be compatible with [_Externally Managed PF_](externally-manage-pf.md) feature
+
+
+### Non-Goals
+
+* replace `SriovNetworkPoolConfig` CRD
+* change API for host-level settings, e.g. `ovs-hw-offload`
+	
+	_**Note:** we may need to extend this API to support additional options_
+
+* add support for VF-lag use-case
+
+## Proposal
+
+1. deploy `ovs-cni` and `accelerated-bridge-cni` with init containers of `sriov-network-config-daemon` Pod
+2. define `OVSNetwork` and `BridgeNetwork` CRDs and implement controllers for them which will create `NetworkAttachmentDefinition` CRs
+3. extend `SriovNetworkNodePolicy`and `SriovNetworkNodeState` (spec and status) CRDs to support configuration of software bridges (bridge-level configuration)
+
+
+### Workflow Description
+
+Implementation should be compatible with the following workflows:
+
+* [Fully automatic workflow](#fully-automatic-workflow)
+* [NIC configuration only flow](#nic-configuration-only-flow) (Externally managed bridge)
+* [Externally managed NIC flow](#externally-managed-nic-flow)
+
+#### Fully automatic workflow
+
+This workflow assumes that **sriov-network-operator** handles PFs and VFs configuration, creation and configuration of software bridge, announcement of SRIOV resources with device plugin and preparation of `NetworkAttachmentDefinition` CR.
+
+1. User creates `SriovNetworkNodePolicy` where:
+    * eswitch mode set to `switchdev`
+    * configuration for selected software bridge is defined
+
+2. The operator populates `SriovNetworkNodeState` for matching nodes with PF and bridge
+configurations 
+
+3. `sriov-network-config-daemon` applies PF configuration, create and configure bridge, attach PF to the bridge, applies VF configuration
+
+
+    _**Note 1:** if the operator runs in the `systemd` mode then bridge creation should happen in the `pre` phase._
+
+    _**Note 2:** we should create udev rule which will set `NM_UNMANAGED=1` for bridges created by the operator_
+
+4. `sriov-network-config-daemon` should report information about software bridges in the status field of the `SriovNetworkNodeState` CR.
+
+5. SRIOV resources are announced by the Device plugin
+
+6. User creates `OVSNetwork` or `BridgeNetwork` CR to create `NetworkAttachmentDefinition` CR, which relies on resources announced by the Device Plugin
+
+
+_**Note:** created software bridge should be removed during the PF configuration reset_
+
+
+#### NIC configuration only flow
+
+This workflow is kept to support existing HW offloading use-case. 
+
+In this case, **sriov-network-operator** handles PFs and VFs configuration, announcement of SRIOV resources with device plugin and preparation of `NetworkAttachmentDefinition` CR.
+
+In some scenarios, it may be compatible with `configurationMode: daemon`, but it is supposed to be used when the operator runs in `systemd` mode.
+
+1. User creates `SriovNetworkNodePolicy` where:
+    * eswitch mode set to `switchdev`
+
+    _**Note:** `SriovNetworkNodePolicy` CR should not include bridge configuration_
+
+
+2. The operator populates `SriovNetworkNodeState` for matching nodes with PF configurations
+
+3. `pre` systemd service creates VFs and configure PF
+
+4. NetworkManager or systemd-networkd or environment-specific scripts create bridge
+
+5. `post` systemd service binds VFs to required driver and proceed with other configuration steps
+
+6. SRIOV resources are announced by the Device plugin
+
+7. User creates `OVSNetwork` or `BridgeNetwork` CR to create `NetworkAttachmentDefinition` CR which relies on resource announced by the Device Plugin
+
+_**Note:** it is possible to use externally created `NetworkAttachmentDefinition` CR that contains configuration for any CNI plugin that support
+resources from NICs in switchdev mode._
+
+
+_**Note 1:** bridge is not removed during the PF configuration reset_
+
+_**Note 2:** information about bridges is not reported in the status field of the `SriovNetworkNodeState` CR_
+
+#### Externally managed NIC flow
+
+In this case, **sriov-network-operator** handles announcement of SRIOV resources with device plugin and preparation of `NetworkAttachmentDefinition` CR.
+
+1. User configures PFs, creates VFs, creates bridge and attaches PFs to the bridge with custom scripts.
+
+2. User creates `SriovNetworkNodePolicy` where:
+    * eswitch mode set to `switchdev`
+    * numVFs set to be equal or less than amount of precreated VFs
+    * `externallyManaged` set to true for PFs
+
+3. `sriov-network-config-daemon` configure VFs
+
+3. SRIOV resources are announced by the Device plugin
+
+4. User creates `OVSNetwork` or `BridgeNetwork` CR to create `NetworkAttachmentDefinition` CR, which relies on resources announced by the Device Plugin
+
+### API Extensions
+
+#### Environment variables for the operator
+
+| Variable                       | Description                                                       |
+| ---                            | ---                                                               |
+| `OVS_CNI_IMAGE`                | contains full image name for `ovs-cni`                            |
+| `ACCELERATED_BRIDGE_CNI_IMAGE` | contains full image name for `accelerated-bridge-cni`             |
+| `OVSDB_SOCKET_PATH`            | path to the OVSDB socket, used to path to the sriov config daemon |
+
+
+#### Feature flags
+
+`manageSoftwareBridges` - control state of the feature (default: `false`)
+
+#### Daemon command line args
+
+`--ovsdb-socket-path` - path to the OVSDB socket, defaults to `/var/run/openvswitch/db.sock`
+
+`--manage-software-bridges` - enables management of the OVS and linux bridges by the operator
+
+#### OVSNetwork CRD `new`
+
+```golang
+// OVSNetworkSpec defines the desired state of OvsNetwork
+type OVSNetworkSpec struct {
+	// Namespace of the NetworkAttachmentDefinition custom resource
+	NetworkNamespace string `json:"networkNamespace,omitempty"`
+	// OVS Network device plugin endpoint resource name
+	ResourceName string `json:"resourceName"`
+	// Capabilities to be configured for this network.
+	// Capabilities supported: (mac|ips), e.g. '{"mac": true}'
+	Capabilities string `json:"capabilities,omitempty"`
+	// IPAM configuration to be used for this network.
+	IPAM string `json:"ipam,omitempty"`
+	// name of the OVS bridge, if not set OVS will automatically select bridge
+	// based on VF PCI address
+	Bridge string `json:"bridge,omitempty"`
+	// Vlan to assign for the OVS port
+	Vlan uint `json:"vlan,omitempty"`
+	// Mtu for the OVS port
+	MTU uint `json:"mtu",omitempty`
+	// Trunk configuration for the OVS port
+	Trunk []*TrunkConfig `json:"trunk,omitempty"`
+	// The type of interface on ovs.
+	InterfaceType string `json:"interfaceType,omitempty"`
+	// MetaPluginsConfig configuration to be used in order to chain metaplugins
+	MetaPluginsConfig string `json:"metaPlugins,omitempty"`
+}
+
+// OVSTrunkConfig contains configuration for OVS trunk
+type TrunkConfig struct {
+	MinID *uint `json:"minID,omitempty"`
+	MaxID *uint `json:"maxID,omitempty"`
+	ID    *uint `json:"id,omitempty"`
+}
+
+// OvsNetworkStatus defines the observed state of OvsNetwork
+type OvsNetworkStatus struct {
+}
+
+// OvsNetwork is the Schema for the ovsnetworks API
+type OvsNetwork struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   OvsNetworkSpec   `json:"spec,omitempty"`
+	Status OvsNetworkStatus `json:"status,omitempty"`
+}
+
+```
+
+#### BridgeNetwork CRD `new`
+
+```golang
+// BridgeNetworkSpec defines the desired state of BridgeNetwork
+type BridgeNetworkSpec struct {
+	// Namespace of the NetworkAttachmentDefinition custom resource
+	NetworkNamespace string `json:"networkNamespace,omitempty"`
+	// OVS Network device plugin endpoint resource name
+	ResourceName string `json:"resourceName"`
+	// Capabilities to be configured for this network.
+	// Capabilities supported: (mac|ips), e.g. '{"mac": true}'
+	Capabilities string `json:"capabilities,omitempty"`
+	// IPAM configuration to be used for this network.
+	IPAM string `json:"ipam,omitempty"`
+	// name of the Linux bridge, if not set will automatically select bridge
+	// based on VF PCI address
+	Bridge string `json:"bridge,omitempty"`
+	// VLAN ID for VF
+	Vlan uint `json:"vlan,omitempty"`
+	// VLAN Trunk configuration
+	Trunk []TrunkConfig `json:"trunk,omitempty"`
+	// enable setting matching vlan tags on the bridge uplink interface, default is false
+	SetUplinkVlan bool `json:"setUplinkVlan,omitempty"`
+	// MTU for VF and representor
+	MTU uint `json:"mtu,omitempty"`
+	// MetaPluginsConfig configuration to be used in order to chain metaplugins
+	MetaPluginsConfig string `json:"metaPlugins,omitempty"`
+}
+
+// BridgeNetworkStatus defines the observed state of BridgeNetwork
+type BridgeNetworkStatus struct {
+}
+
+// BridgeNetwork is the Schema for the ovsnetworks API
+type BridgeNetwork struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   BridgeNetworkSpec   `json:"spec,omitempty"`
+	Status BridgeNetworkStatus `json:"status,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+
+// BridgeNetworkList contains a list of BridgeNetwork
+type BridgeNetworkList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []BridgeNetwork `json:"items"`
+}
+
+```
+
+#### SriovNetworkNodePolicy CRD
+
+```golang
+// SriovNetworkNodePolicySpec defines the desired state of SriovNetworkNodePolicy
+type SriovNetworkNodePolicySpec struct {
+	// ...existing fields...
+	// contains spec for the software bridge
+	Bridge Bridge `json:"bridge,omitempty"`
+}
+// contains spec for the bridge
+// only one bridge type can be set
+type Bridge struct {
+	// contains optional config for OVS bridge
+	Ovs *OVSConfig `json:"ovs,omitempty"`
+	// contains optional config for Linux bridge
+	Linux *LinuxBridgeConfig `json:"linux,omitempty"`
+}
+
+// OVSConfig optional configuration for OVS bridge and uplink Interface
+type OVSConfig struct {
+	// contains bridge level settings
+	Bridge OVSBridgeConfig `json:"bridge,omitempty"`
+	// contains settings for uplink (PF)
+	Uplink OVSUplinkConfig `json:"uplink,omitempty"`
+}
+
+// OVSBridgeConfig contains some options from the Bridge table in OVSDB
+type OVSBridgeConfig struct {
+	DatapathType string            `json:"datapathType,omitempty"`
+	ExternalIDs  map[string]string `json:"externalIDs,omitempty"`
+	OtherConfig  map[string]string `json:"otherConfig,omitempty"`
+}
+
+// OVSUplinkConfig contains PF interface configuration for the bridge
+type OVSUplinkConfig struct {
+	Interface OVSInterfaceConfig `json:"interface,omitempty"`
+	// can be extended to support OVSPortConfig which will include
+	// settings from the OVS Port table
+}
+
+// OVSInterfaceConfig contains some options from the Interface table of the OVSDB for PF
+type OVSInterfaceConfig struct {
+	Type        string            `json:"type,omitempty"`
+	Options     map[string]string `json:"options,omitempty"`
+	ExternalIDs map[string]string `json:"externalIDs,omitempty"`
+	OtherConfig map[string]string `json:"otherConfig,omitempty"`
+}
+
+// LinuxBridgeConfig optional configuration for Linux bridge and uplink interface
+type LinuxBridgeConfig struct {
+	Bridge BridgeConfig      `json:"bridge,omitempty"`
+	Uplink map[string]string `json:"uplink,omitempty"` // TODO clarify required settings
+}
+
+// BridgeConfig contains some options for linux bridge
+type BridgeConfig struct {
+	VlanFiltering bool `json:"vlanFiltering,omitempty"`
+	// +kubebuilder:validation:Enum=802.1Q;802.1ad
+	VlanProtocol string `json:"vlanProtocol,omitempty"`
+}
+```
+
+_**Note 1**: multiple policies can match single PF (vf range use-case), bridge settings from the policy with
+ the higher priority (which is the one with the lowest `.spec.priority` value) will be applied to PF._
+
+_**Note 2**: multiple NICs can match the same policy on a host. In this case a separate bridge will be created for each NIC._
+
+#### SriovNetworkNodeState CRD
+
+```golang
+
+type SriovNetworkNodeStateSpec struct {
+	// ...existing fields...
+	Interfaces      Interfaces `json:"interfaces,omitempty"`
+	Bridges         Bridges    `json:"bridges,omitempty"`
+}
+
+// SriovNetworkNodeStateStatus defines the observed state of SriovNetworkNodeState
+type SriovNetworkNodeStateStatus struct {
+	Interfaces    InterfaceExts `json:"interfaces,omitempty"`
+	Bridges       Bridges       `json:"bridges,omitempty"`
+	SyncStatus    string        `json:"syncStatus,omitempty"`
+	LastSyncError string        `json:"lastSyncError,omitempty"`
+}
+
+// Bridges contains list of bridges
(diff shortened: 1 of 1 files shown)


doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+### Use Cases
+
+* As a user, I expect that **sriov-network-operator** will install `ovs-cni` and `accelerated-bridge-cni` to hosts.
+* As a user, I want to create `OVSNetwork` CR, which will result in creation of `NetworkAttachmentDefinition` CR that uses
+`ovs-cni` and contains required resource request.
+* As a user, I want to create `BridgeNetwork` CR which will result in creation of `NetworkAttachmentDefinition` CR that uses
+`accelerated-bridge-cni` and contains the required resource requests.
+* As a user, I want to define configuration for software bridges inside the `SriovNetworkNodePolicy` CR and expect that the
+operator will create required bridges, configure them, and attach uplinks (physical functions).
+
```
wizhaoredhat, line 46: What about the OVN-K CNI use case? Will the api to set it remain the same?
Will OVN-K and OVS-CNI have any conflicts if OVN-K is installed in the first place.

ykulazhenkov: I believe that the only difference for OVN-k use case is that sriov-network-operator will need to be configured to `systemd` mode (check [this doc](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/doc/design/switchdev-refactoring.md) for details).  [NIC configuration only flow](https://github.com/ykulazhenkov/sriov-network-operator/blob/3753c5dcc4cf83a46a85bf1334f6ada31d9a22ce/doc/design/software-bridge-management.md?plain=1#L110) is proposed to be compatible with existing OVN-k use-case. In this case NIC will be configured to switchdev mode early on boot by the systemd service, but the operator will not manage software bridges.

ykulazhenkov: The current proposal assumes that "Fully automatic" flow can be used only for dedicated NICs (not used by OVN-k) which are fully controlled by the sriov-network-operator.

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+	NetFilter         string            `json:"netFilter,omitempty"`
+	Mtu               int               `json:"mtu,omitempty"`
+	NumVfs            int               `json:"numVfs,omitempty"`
+	LinkSpeed         string            `json:"linkSpeed,omitempty"`
+	LinkType          string            `json:"linkType,omitempty"`
+	EswitchMode       string            `json:"eSwitchMode,omitempty"`
+	ExternallyManaged bool              `json:"externallyManaged,omitempty"`
+	TotalVfs          int               `json:"totalvfs,omitempty"`
+	VFs               []VirtualFunction `json:"Vfs,omitempty"`
+	Bridge            *Bridge           `json:"bridge,omitempty"`
```
zeeke, line 389: Is this the same type as we have in the `SriovNetworkNodeState.Spec` and in the `SriovNetworkNodeState.Spec`? If this information comes from an inspection on the host, it might reflect the output of a `ovs-vsctl` or `ip link`, right?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+// LinuxBrBridgeConfig contains some options for linux bridge
+type LinuxBrBridgeConfig struct {
+	VlanFiltering bool `json:"vlan_filtering,omitempty"`
+	// +kubebuilder:validation:Enum=802.1Q;802.1ad
+	VlanProtocol string `json:"vlan_protocol,omitempty"`
+}
+```
+
+Open questions:
+- do we want to support each OVS Bridge/Interface option explicitly or should we support
```
zeeke, line 350: I'd personally go with a generic container for options, describing in the docs in which command they will be used.

zeeke (commented):
Left a few comments. Overall design looks good to me

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+ generic map[string]string struct which allows set any field in the Bridge/Interface tables?
+- need to clarify list of required args for OVS and Linux bridge
+
+_**Note**: multiple policies can match single PF (vf range use-case), bridge settings from the policy with the higher priority
+will be applied to PF._
```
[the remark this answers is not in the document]
ykulazhenkov: done

doc/design/software-bridge-management.md
ykulazhenkov: yes, currently I think that it is possible to use the same type here and in the policy/state.

> `ovs-vsctl` or `ip link`
just as a  note, I think we will use direct requests to ovsd  to configure ovs bridges and rely on netlink for linux bridges.
Usage of cmd tools is not expected.

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+	// based on VF PCI address
+	BrName string `json:"bridge,omitempty"`
+	// VlanTag to assign for the OVS port
+	VlanTag uint `json:"vlan,omitempty"`
+	// Mtu for the OVS port
+	MTU int `json:"mtu",omitempty`
+	// Trunk configuration for the OVS port
+	Trunk []*OVSTrunkConfig `json:"trunk,omitempty"`
+	// OpenFlow port number in range 1 to 65,279
+	OfportRequest uint `json:"ofport_request,omitempty"`
```
adrianchiris, line 196: do we want to have this configurable for our usecase ?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+// OVSBridgeConfig contains some options from the Bridge table in OVSDB
+type OVSBridgeConfig struct {
+	DatapathType string            `json:"datapath_type,omitempty"`
+	ExternalIDs  map[string]string `json:"external_ids,omitempty"`
+	OtherConfig  map[string]string `json:"other_config,omitempty"`
+}
+
+// OVSInterfaceConfig contains some options from the Interface table of the OVSDB for PF
+type OVSInterfaceConfig struct {
+	Type        string            `json:"type,omitempty"`
```
adrianchiris, line 329: what values can be provided to type ? (dpdk / netdev ?)

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+
+// contains spec for the bridge
+// only one bridge type can be set
+type Bridge struct {
+	// +kubebuilder:validation:Enum=ovs;linux
+	Type string `json:"type"`
+	// contains optional config for OVS bridge
+	Ovs *OVSConfig `json:"ovs,omitempty"`
+	// contains optional config for LinuxBr bridge
+	LinuxBr *LinuxBrConfig `json:"linux,omitempty"`
```
adrianchiris, line 311: nit: why not name this "Linux" ?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+type OVSInterfaceConfig struct {
+	Type        string            `json:"type,omitempty"`
+	Options     map[string]string `json:"options,omitempty"`
+	ExternalIDs map[string]string `json:"external_ids,omitempty"`
+	OtherConfig map[string]string `json:"other_config,omitempty"`
+}
+
+// LinuxBrConfig optional configuration for Linux bridge and uplink interface
+type LinuxBrConfig struct {
+	Interface map[string]string    // TODO clarify required settings
```
adrianchiris, line 337: can you give some examples of this field ?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+##### Phase 1
+
+Add support for [NIC configuration only flow](#nic-configuration-only-flow) and
+[Externally managed NIC flow](#externally-managed-nic-flow) for Open vSwitch
+
+Requirements: 
+* add bridge auto-selection logic to `ovs-cni`
+* define `OVSNetwork` CRD and implement controller for it
+
+##### Phase 2
```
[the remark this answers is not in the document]
zeeke: +1 though this is an implementation detail. 

To me, in the design doc it's enough to refer to the config-daemon component.

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+configurations 
+
+3. `sriov-network-config-daemon` applies PF configuration, create and configure bridge, attach PF to the bridge, applies VF configuration
+
+
+    _**Note 1:** if the operator runs in the `systemd` mode then bridge creation should happen in the `pre` phase._
+
+    _**Note 2:** we should create udev rule which will set `NM_UNMANAGED=1` for bridges created by the operator_
+
+4. `sriov-network-config-daemon` should report information about software bridges in the status field of the `SriovNetworkNodePolicy` CR.
```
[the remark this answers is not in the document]
ykulazhenkov: If "fully auto" mode is used  we should create a bridge with predictable name for each PF that match a policy that contains a bridge configuration. We can't use interface name to generate a predictable name for the bridge, because the limit for the interface length is only 15 symbols and PF name will already have length around this. 
I propose to use PF's PCI address to generate predictable bridge name, like `br-0000_d8_01.0`.

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+_**Note:** created software bridge should be removed during the PF configuration reset_
+
+
+#### NIC configuration only flow
+
+This workflow is kept to support existing HW offloading use-case. 
+
+In this case, **sriov-network-operator** handles PFs configuration, announcement of SRIOV resources with device plugin and preparation of `NetworkAttachmentDefinition` CR.
+
+In some scenarios, it may be compatible with `configurationMode: daemon`, but it is supposed to be used when the operator runs in `systemd` mode.
```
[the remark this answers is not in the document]
ykulazhenkov: > when would it not be compatible with daemon mode ?

When `NIC configuration only flow` is used with `systemd` configurationMode, then you can apply any network configuration between  steps when VF is created  and when VF is configured(VF binds to the driver). In the `daemon` mode you don't have such option, you can only apply your custom network configuration before the operator or after, but not in the middle. This can be a problem for some configuration, e.g. for VF LAG for NVIDIA NICs

doc/design/software-bridge-management.md
ykulazhenkov: For our use-case we don't need this. But I think we should support all options from ovs-cni. WDYT?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+	// VlanTag to assign for the OVS port
+	VlanTag uint `json:"vlan,omitempty"`
+	// Mtu for the OVS port
+	MTU int `json:"mtu",omitempty`
+	// Trunk configuration for the OVS port
+	Trunk []*OVSTrunkConfig `json:"trunk,omitempty"`
+	// OpenFlow port number in range 1 to 65,279
+	OfportRequest uint `json:"ofport_request,omitempty"`
+	// The type of interface on ovs.
+	InterfaceType string `json:"interface_type,omitempty"`
```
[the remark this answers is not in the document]
ykulazhenkov: this is required for ovs-dpdk

doc/design/software-bridge-management.md
ykulazhenkov: yes. I think we should not filter them on the operator level. User may want to use some vendor-specific values for this field.

doc/design/software-bridge-management.md
ykulazhenkov: > do we have any option which is mandatory for the use-cases in this doc ?

No. I assume that ovs-kernel (with TC based HW offloading) should work without extra options.

User will need to:
a) enable HW offload 
b) create policy that contains Bridge.type = "ovs", without extra OVS options
c) create OVSNetwork CR

doc/design/software-bridge-management.md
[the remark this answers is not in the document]
ykulazhenkov: I removed the mention of the generic_plugin from the doc. Let's discuss this during the implementation.

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+	VlanProtocol string `json:"vlan_protocol,omitempty"`
+}
+```
+
+Open questions:
+- do we want to support each OVS Bridge/Interface option explicitly or should we support
+ generic map[string]string struct which allows set any field in the Bridge/Interface tables?
+- need to clarify list of required args for OVS and Linux bridge
+
+_**Note**: multiple policies can match single PF (vf range use-case), bridge settings from the policy with
```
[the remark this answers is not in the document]
ykulazhenkov: > what will happen when multiple PFs match the policy ?

A separate bridge will be created for each PF. I added additional note about this to the doc

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+
+Implementation should be compatible with the following workflows:
+
+* [Fully automatic workflow](#fully-automatic-workflow)
+* [NIC configuration only flow](#nic-configuration-only-flow) (Externally managed bridge)
+* [Externally managed NIC flow](#externally-managed-nic-flow)
+
+#### Fully automatic workflow
+
+This workflow assumes that **sriov-network-operator** handles PFs configuration, creation and configuration of software bridge, announcement of SRIOV resources with device plugin and preparation of `NetworkAttachmentDefinition` CR.
```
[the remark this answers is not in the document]
ykulazhenkov: done

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+	// OVS Network device plugin endpoint resource name
+	ResourceName string `json:"resourceName"`
+	// Capabilities to be configured for this network.
+	// Capabilities supported: (mac|ips), e.g. '{"mac": true}'
+	Capabilities string `json:"capabilities,omitempty"`
+	// IPAM configuration to be used for this network.
+	IPAM string `json:"ipam,omitempty"`
+	// name of the OVS bridge, if not set OVS will automatically select bridge
+	// based on VF PCI address
+	BrName string `json:"bridge,omitempty"`
```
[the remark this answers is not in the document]
ykulazhenkov: I renamed the field to "Bridge"

doc/design/software-bridge-management.md
ykulazhenkov: renamed

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,460 @@
...
+	VlanTag uint `json:"vlan,omitempty"`
+	// Mtu for the OVS port
+	MTU int `json:"mtu",omitempty`
+	// Trunk configuration for the OVS port
+	Trunk []*OVSTrunkConfig `json:"trunk,omitempty"`
+	// OpenFlow port number in range 1 to 65,279
+	OfportRequest uint `json:"ofport_request,omitempty"`
+	// The type of interface on ovs.
+	InterfaceType string `json:"interface_type,omitempty"`
+	// The type of interface on ovs.
```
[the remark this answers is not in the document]
ykulazhenkov: fixed

doc/design/software-bridge-management.md
ykulazhenkov: This field should hold PF's port-level bridge configuration . 
It is not clear which setting we should put here. 
VLAN configuration for the PF is a good candidate, but I'm not sure that we need it.
By default OVS doesn't require to explicitly allow VLANs on the PF interface. 
Linux bridge requires to explicitly configure VLANs on PF, but accelerated-bridge-cni already has an option to handle this.

adrianchiris (commented):
Had some discussions with @ykulazhenkov regarding possible alternatives and i believe what is currently proposed is a good way forward so overall im OK with the proposed approach.

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	// Mtu for the OVS port
+	MTU int `json:"mtu",omitempty`
+	// Trunk configuration for the OVS port
+	Trunk []*OVSTrunkConfig `json:"trunk,omitempty"`
+	// OpenFlow port number in range 1 to 65,279
+	OfportRequest uint `json:"ofport_request,omitempty"`
+	// The type of interface on ovs.
+	InterfaceType string `json:"interface_type,omitempty"`
+	// Path to the configuration file on the host
+	ConfigurationPath string `json:"configuration_path,omitempty"`
```
SchSeba, line 200: can you give an example for this?
not sure I want to expose something like that to the user

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	// Paths to OVSDB socket
+	SocketFile string `json:"socket_file,omitempty"`
+	// Amount of retries for link state check
+	LinkStateCheckRetries int `json:"link_state_check_retries,omitempty"`
```
SchSeba, line 204: we will have a default for this?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	// Interval between retries for link state check
+	LinkStateCheckInterval int `json:"link_state_check_interval,omitempty"`
```
SchSeba, line 206: we will have a default for this?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	IPAM string `json:"ipam,omitempty"`
+	// name of the Linux bridge, if not set will automatically select bridge
+	// based on VF PCI address
+	Bridge string `json:"bridge,omitempty"`
+	// VLAN ID for VF
+	Vlan int `json:"vlan,omitempty"`
+	// VLAN Trunk configuration
+	Trunk []BridgeTrunkConfig `json:"trunk,omitempty"`
+	// enable setting matching vlan tags on the bridge uplink interface, default is false
+	SetUplinkVlan bool `json:"setUplinkVlan,omitempty"`
```
SchSeba, line 255: do we want something like this also for the ovsBridge API?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	// MTU for VF and representor
+	MTU int `json:"mtu,omitempty"`
+	// MetaPluginsConfig configuration to be used in order to chain metaplugins
+	MetaPluginsConfig string `json:"metaPlugins,omitempty"`
+}
+
+// BridgeTrunkConfig contains configuration for OVS trunk
+type BridgeTrunkConfig struct {
```
SchSeba, line 263: can we use just the same TrunkConfig for both ?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+
+// OVSConfig optional configuration for OVS bridge and uplink Interface
+type OVSConfig struct {
+	Bridge    *OVSBridgeConfig    `json:"bridge,omitempty"`
+	Interface *OVSInterfaceConfig `json:"interface,omitempty"`
+}
+
+// OVSBridgeConfig contains some options from the Bridge table in OVSDB
+type OVSBridgeConfig struct {
+	DatapathType string            `json:"datapath_type,omitempty"`
```
SchSeba, line 322: Is this datapathType can be different from the one in the ovsNetwork?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	ExternalIDs  map[string]string `json:"external_ids,omitempty"`
+	OtherConfig  map[string]string `json:"other_config,omitempty"`
+}
+
+// OVSInterfaceConfig contains some options from the Interface table of the OVSDB for PF
+type OVSInterfaceConfig struct {
```
SchSeba, line 328: can you give some example for this field I can see everything is string and map[string]string :P

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+}
+```
+
+Open questions:
+- do we want to support each OVS Bridge/Interface option explicitly or should we support
+ generic map[string]string struct which allows set any field in the Bridge/Interface tables?
+- need to clarify list of required args for OVS and Linux bridge
+
+_**Note 1**: multiple policies can match single PF (vf range use-case), bridge settings from the policy with
+ the higher priority (which is the one with the lowest `.spec.priority` value) will be applied to PF._
```
SchSeba, line 355: should be just tell the user that only 1 policy per PF can contain bridge section and block that via a webhook?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	DeviceID          string            `json:"deviceID,omitempty"`
+	NetFilter         string            `json:"netFilter,omitempty"`
+	Mtu               int               `json:"mtu,omitempty"`
+	NumVfs            int               `json:"numVfs,omitempty"`
+	LinkSpeed         string            `json:"linkSpeed,omitempty"`
+	LinkType          string            `json:"linkType,omitempty"`
+	EswitchMode       string            `json:"eSwitchMode,omitempty"`
+	ExternallyManaged bool              `json:"externallyManaged,omitempty"`
+	TotalVfs          int               `json:"totalvfs,omitempty"`
+	VFs               []VirtualFunction `json:"Vfs,omitempty"`
```
SchSeba, line 390: in the virtual functions do we want to also add the representorName?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	Bridge            *Bridge           `json:"bridge,omitempty"`
+}
+
+```
+`SriovNetworkNodeState.spec` and `SriovNetworkNodeState.status` should be extended to contain same `Bridge` struct as `SriovNetworkNodePolicySpec` CRD.
```
SchSeba, line 395: I don't think we can use the same object as the Policy as in the policy I can have nicSelector: vendor: 8086 
but in the spec and status for the NodeState I will like to have a list of all the bridges we need to create with the configuration, name and uplink object

SchSeba (commented):
great document!
I left some comments nice work :)

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+* As a user, I want to define configuration for software bridges inside the `SriovNetworkNodePolicy` CR and expect that the
+operator will create required bridges, configure them, and attach uplinks (physical functions).
+
+### Goals
+
+* handle installation of `ovs-cni` and `accelerated-bridge-cni`
+* add `OVSNetwork` and `BridgeNetwork` CRDs as an API for end users to simplify creation of `NetworkAttachmentDefinition` CR
+for `ovs-cni` and `accelerated-bridge-cni`
+* extend `SriovNetworkNodePolicy` CRD to support configuration of software bridges (bridge-level configuration)
+* extend `SriovNetworkNodeState` CRD to support configuration of software bridges (bridge-level configuration)
```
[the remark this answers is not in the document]
ykulazhenkov: yes, I added a note about this

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+
+3. `pre` systemd service creates VFs and configure PF
+
+4. NetworkManager or systemd-networkd or environment-specific scripts create bridge
+
+5. `post` systemd service binds VFs to required driver and proceed with other configuration steps
+
+6. SRIOV resources are announced by the Device plugin
+
+7. User creates `OVSNetwork` or `BridgeNetwork` CR to create `NetworkAttachmentDefinition` CR which relies on resource announced by the Device Plugin
```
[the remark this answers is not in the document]
ykulazhenkov: yes, good idea

doc/design/software-bridge-management.md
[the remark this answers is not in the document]
ykulazhenkov: Yes, we will report this information in the status field of the state object.

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+    * numVFs set to match amount of precreated VFs
+    * `externallyManaged` set to true for PFs
+
+3. `sriov-network-config-daemon` configure VFs
+
+3. SRIOV resources are announced by the Device plugin
+
+4. User creates `OVSNetwork` or `BridgeNetwork` CR to create `NetworkAttachmentDefinition` CR, which relies on resources announced by the Device Plugin
+
+### API Extensions
```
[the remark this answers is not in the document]
ykulazhenkov: Original thought was to control this with environment variable -  skip deployment if `OVS_CNI_IMAGE` is explicitly set to empty string.
WDYT?

doc/design/software-bridge-management.md
ykulazhenkov: no, ovs-cni will handle default value. If `LinkStateCheckRetries` is zero we will not add it to CNI args of the ovs-cni

doc/design/software-bridge-management.md
ykulazhenkov: no, ovs-cni will handle default value. If `LinkStateCheckRetries` is zero we will not add it to CNI args of the ovs-cni

doc/design/software-bridge-management.md
ykulazhenkov: OVSNetwork and BridgeNetwork CRDs are wrappers to create CNI args for ovs-cni and accelerated-bridge-cni. 
OVS-cni doesn't have such option as for now. It is not critical to have it in the OVS-cni, because by default ovs allows all vlans on a trunk port. Linux bridge require to set vlans on a trunk port explicitly.

doc/design/software-bridge-management.md
ykulazhenkov: yes. we ca

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	// ...existing fields...
+	// contains spec for the software bridge
+	Bridge *Bridge `json:"bridge,omitempty"`
+}
+
+// contains spec for the bridge
+// only one bridge type can be set
+type Bridge struct {
+	// +kubebuilder:validation:Enum=ovs;linux
+	Type string `json:"type"`
```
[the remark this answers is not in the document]
ykulazhenkov: I think we can't. Structs contain optional arguments for bridges.
Minimal valid config is 
```
kind: SriovNetworkNodePolicy
metadata:
  name: connectx6dx-switched
  namespace: sriov-network-operator
spec:
  eSwitchMode: switchdev
  nicSelector:
    deviceID: 101d
    vendor: 15b3
  nodeSelector:
    feature.node.kubernetes.io/network-sriov.capable: "true"
  numVfs: 4
  resourceName: connectx6dx-switched
  bridge:
    type: ovs
```

In this case OVS bridge with default configuration will be created for each PF that match the policy. 
By default configuration I mean config which OVS uses if you exec `ovs-vsctl add-br foo` (no extra args).
If HW offloading is enable on the node level (with old ConfigPool API), then such bridge will work nicely with OVS-kernel hw offloading.

doc/design/software-bridge-management.md
ykulazhenkov: yes

doc/design/software-bridge-management.md
ykulazhenkov: This is raw options from Interface TABLE of the OVSDB. You can check available builtin options in the OVS docs https://www.openvswitch.org/support/dist-docs/ovs-vswitchd.conf.db.5.html. This options will be set for PF interface while attaching it to OVS bridge.
These maps can contain options from upstream OVS or can be used to inject vendor specific options if required. Example for NVIDIA https://docs.nvidia.com/doca/sdk/openvswitch+offload/index.html#src-2571330903_OpenvSwitchOffload-OVS-DOCAHardwareOffloads

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+// OVSInterfaceConfig contains some options from the Interface table of the OVSDB for PF
+type OVSInterfaceConfig struct {
+	Type        string            `json:"type,omitempty"`
+	Options     map[string]string `json:"options,omitempty"`
+	ExternalIDs map[string]string `json:"external_ids,omitempty"`
+	OtherConfig map[string]string `json:"other_config,omitempty"`
+}
+
+// LinuxBrConfig optional configuration for Linux bridge and uplink interface
+type LinuxBrConfig struct {
```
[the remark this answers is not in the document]
ykulazhenkov: We have LinuxBrBridgeConfig struct which contains bridge level config for linux bridge. Then we will need to rename it to LinuxBridgeBridgeConfig which is not nice :) Let's discuss this during the implementation

doc/design/software-bridge-management.md
ykulazhenkov: Not sure. I think it is better to preserve existing behavior. Currently we follow this rule for existing PF level options defined in the policy (linkType, eswitchMode, etc)

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+
+#### Externally managed NIC flow
+
+In this case, **sriov-network-operator** handles announcement of SRIOV resources with device plugin and preparation of `NetworkAttachmentDefinition` CR.
+
+1. User configures PFs, creates VFs, creates bridge and attaches PFs to the bridge with custom scripts.
+
+2. User creates `SriovNetworkNodePolicy` where:
+    * eswitch mode set to `switchdev`
+    * numVFs set to match amount of precreated VFs
```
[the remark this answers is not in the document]
ykulazhenkov: fixed

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	// Capabilities to be configured for this network.
+	// Capabilities supported: (mac|ips), e.g. '{"mac": true}'
+	Capabilities string `json:"capabilities,omitempty"`
+	// IPAM configuration to be used for this network.
+	IPAM string `json:"ipam,omitempty"`
+	// name of the OVS bridge, if not set OVS will automatically select bridge
+	// based on VF PCI address
+	Bridge string `json:"bridge,omitempty"`
+	// VlanTag to assign for the OVS port
+	VlanTag uint `json:"vlan,omitempty"`
```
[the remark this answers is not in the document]
ykulazhenkov: fixed

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,495 @@
...
+	// Trunk configuration for the OVS port
+	Trunk []*OVSTrunkConfig `json:"trunk,omitempty"`
+	// OpenFlow port number in range 1 to 65,279
+	OfportRequest uint `json:"ofport_request,omitempty"`
+	// The type of interface on ovs.
+	InterfaceType string `json:"interface_type,omitempty"`
+	// Path to the configuration file on the host
+	ConfigurationPath string `json:"configuration_path,omitempty"`
+	// Paths to OVSDB socket
+	SocketFile string `json:"socket_file,omitempty"`
```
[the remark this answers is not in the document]
ykulazhenkov: Not sure. My assumption was that we want to expose all user-faced options available as STDIN args in ovs-cni.  (except device-id). I think it is better to keep this option, but if it is not set explicitly by user, then we can try to use value from the daemon config.

doc/design/software-bridge-management.md
ykulazhenkov: > agree we should use netlink (if we miss something we can add to the golang netlink) and for ovsdb I think we can use the same package ovn-k use?

yes, I think we have options to avoid usage of cmd tools

doc/design/software-bridge-management.md
ykulazhenkov: I think this is not directly related to the scope of this proposal, but I agree that this is a good improvement.

doc/design/software-bridge-management.md
SchSeba: can you give an example what is datapathType?
is like vdpa dpdk stuff like that?
do we want an enum with supported options?

doc/design/software-bridge-management.md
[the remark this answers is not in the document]
SchSeba: the other one can be just BridgeConfig as we are already inside the LinuxBridgeConfig section no?

doc/design/software-bridge-management.md
SchSeba: can we add the field to represent the change in the API as part of this proposal?

SchSeba (commented):
I did another round of review.

general point about expose everything from the cnis args to the user.
maybe we can remove the ones I add a comment on and add them later if we see a use-case?
adding is easier than removing I think :) 

and please one critical comment if you can take a look
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/640/files#r1518902408

doc/design/software-bridge-management.md
ykulazhenkov: ok

doc/design/software-bridge-management.md
ykulazhenkov: yes: like netdev, dpdk... 
`datapath_type` can contain proprietary value, it is better not to validate it on the API level

doc/design/software-bridge-management.md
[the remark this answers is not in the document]
ykulazhenkov: ok, let's drop the type field

doc/design/software-bridge-management.md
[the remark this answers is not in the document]
ykulazhenkov: ok, fixed

doc/design/software-bridge-management.md
ykulazhenkov: @SchSeba I updated proposed API change for SriovNetworkNodeState object. Now it contains Bridges list.
I propose to use the same Bridges object in the spec and in the status. We may want to extend the one in the status field with additional info when we will have feedback from the users.

ykulazhenkov:
thanks @SchSeba 

> general point about expose everything from the cnis args to the user. maybe we can remove the ones I add a comment on and add them later if we see a use-case? adding is easier than removing I think :)

Ok, let's remove these field for now. I removed them from the design


> and please one critical comment if you can take a look https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/640/files#r1518902408

I replied to this comment and updated the design.

PTAL

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,508 @@
...
+type LinuxBridgeConfigExt struct {
+	Name    string                       `json:"name"`
+	Bridge  BridgeConfig                 `json:"bridge,omitempty"`
+	Uplinks []LinuxBridgeUPlinkConfigExt `json:"uplinks,omitempty"`
+}
+
+type LinuxBridgeUPlinkConfigExt struct {
+	PciAddress string             `json:"pciAddress"`
+	Name       string             `json:"name,omitempty"`
+	Interface  OVSInterfaceConfig `json:"interface,omitempty"`
```
SchSeba, line 398: `OVSInterfaceConfig` -> the link bridge one right?

doc/design/software-bridge-management.md
```diff
@@ -0,0 +1,508 @@
...
+}
+
+type VirtualFunction struct {
+	// ...existing fields...
+	// contains VF representor name for NICs in switchdev mode
+	RepresentorName string `json:"representorName,omitempty"`
+}
+
+```
+`SriovNetworkNodeState.spec` and `SriovNetworkNodeState.status` should be extended to contain same `Bridges`.
```
SchSeba, line 408: is this statement still correct?

SchSeba (commented):
almost ready just two small nits :)

doc/design/software-bridge-management.md
ykulazhenkov: Yes, it is correct. The statement differs from the original one - now is says that we use the same struct in "spec" and "status" of the SriovNetworkNodeState. Also I added note which says that we may want to extend/replace the struct in the future.

e0ne (approved):

ykulazhenkov:
I pushed small update to the document:

* use camelCase for json fields to follow k8s guidlines
* add featureFlag that controls state of the feature

ykulazhenkov:
@SchSeba @zeeke 
Could you please take another look at the document? Let's merge it if we have agreement about proposed APIs.
Thx

zeeke (approved):
LGTM

PR #644, opened 2024-02-27
e2e: fix `CNI Logging level` test case

zeeke:
It can happen that clock of the cluster nodes is not aligned to the clock of test runner. In such cases, asking a pod for logs since a given time might end in an empty response, as the requested time can be in the future (from the cluster perspective).

Use the pod's CreationTimestamp timestamp field to be sure to pass to the API server a meaningful value.


ykulazhenkov (approved):

SchSeba (approved):
nice catch!

/lgtm

adrianchiris (approved):

PR #645, opened 2024-03-04
Require reboot if drain is needed for systemd mode

e0ne:
We need to reboot the node in a systemd mode if any changes are required to be applied to re-stert sriov config service.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -573,7 +573,8 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 			}
 		}
 		reqDrain = reqDrain || systemdConfModified
-		reqReboot = reqReboot || systemdConfModified
+		// require reboot if drain needed for systemd mode
+		reqReboot = reqReboot || systemdConfModified || reqDrain
 		log.Log.V(0).Info("nodeStateSyncHandler(): systemd mode WriteConfFile results",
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -573,7 +573,8 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 			}
 		}
 		reqDrain = reqDrain || systemdConfModified
-		reqReboot = reqReboot || systemdConfModified
+        // require reboot if drain needed for systemd mode
+		reqReboot = reqReboot || systemdConfModified || reqReboot
```
ykulazhenkov, line 577: It looks like the behaviour of the new line is exact same as of the old one. Do we need this change?

e0ne: @ykulazhenkov it's a typo, I'll update it

ykulazhenkov: nit: you can drop `systemdConfModified` from reqReboot condition

ykulazhenkov (approved):
This change should help to fix the scenario when systemd config has up to date configuration, but node configuration is not apllied. This can happen if the operator was redeloyed (systemd config is not removed in this case)

pkg/daemon/daemon.go
e0ne: > nit: you can drop `systemdConfModified` from reqReboot condition

I would like to make it more explicit for this particular case. That's why I left `systemdConfModified`

SchSeba (approved):

PR #647, opened 2024-03-07
Fix sriov.GetNicSriovMode to return legacy mode in case of any failure

ykulazhenkov:
If ENODEV error occurred - silently return legacy mode 
If no error occurred, but eswitch mode is empty - silently return legacy mode
If unknown error occurred - log error and return legacy mode
if eswitch mode is returned from devlink - return eswitch mode

This change is required to fix situation when devlink is supported by kernel/device, but the device doesnt report current eswitch mode via devlink (mode will be an empty string in this case)

To be on a safe side I modified GetNicSriovMode to always return "legacy" in case if it can read the real mode of the NIC for whatever reason. 

@e0ne @zeeke  please check this fix

### DIFF
--- a/pkg/helper/mock/mock_helper.go
+++ b/pkg/helper/mock/mock_helper.go
@@ -445,12 +445,11 @@ func (mr *MockHostHelpersInterfaceMockRecorder) GetNetdevMTU(pciAddr interface{}
 }
 
 // GetNicSriovMode mocks base method.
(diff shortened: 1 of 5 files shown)


e0ne (approved):
LGTM, thanks @ykulazhenkov

SchSeba (approved):
LGTM

PR #648, opened 2024-03-07
Add logic to deploy OVS-cni

ykulazhenkov:
Deploy OVS CNI with initi container of the sriov-config-daemon.

Helm values or OVS_CNI_IMAGE env variable can be used to configure the OVS-cni image.

If OVS_CNI_IMAGE env variable is set to empty string, then deployment of the CNI will be skipped

Part of #640


e0ne (approved):
LGTM

deployment/sriov-network-operator/values.yaml
```diff
@@ -95,6 +95,7 @@ images:
   sriovConfigDaemon: ghcr.io/k8snetworkplumbingwg/sriov-network-operator-config-daemon
   sriovCni: ghcr.io/k8snetworkplumbingwg/sriov-cni
   ibSriovCni: ghcr.io/k8snetworkplumbingwg/ib-sriov-cni
+  ovsCni: quay.io/kubevirt/ovs-cni-plugin
```
Eoghan1232, line 98: ghcr.io/k8snetworkplumbingwg/ovs-cni-plugin

could we migrate to host image here instead of quay?

Eoghan1232: https://github.com/k8snetworkplumbingwg/ovs-cni

or am I mixing these up? this is the proj correct?

ykulazhenkov: @Eoghan1232 
Yes, the repo is in k8snetworkplumbingwg, but it is managed by kubevirt CI. I didn't find ovs-cni image in github registry, I think it is only uploaded kubevirt registry

zeeke (approved):
LGTM

PR #650, opened 2024-03-07
Experiment: Remove stdin from the ExecCommand call in tests

ykulazhenkov:
We don't use stdin to path data, so there is no need to open the stream.
This should improve stability of the Exec call.

On my environment I noticed that sometimes ExecCommand returns empty output during the test.
Detaching of the stdin stream fixed this issue on my setup.


SchSeba (approved):
Thanks!
/lgtm

e0ne (approved):

PR #653, opened 2024-03-11
Add 'IB VF GUID Configuration' design doc

almaslennikov:


doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Implementation Details/Notes/Constraints
+
+There can be less VFs created than GUIDs. To persist the dynamic nature of the SR-IOV Network operator,
+it’s proposed not to return an error in this case but assign as many GUIDs as possible.
+To ensure that nothing breaks when users add/remove VFs, the GUID distribution order should always be the same for each individual host.
+
+Example of the config file:
+
+```json
```
ykulazhenkov, line 53: Can you add some comments about the config  format?
For example, clarifying these things:
Can `pci_adress` and `pf_guid` be set at the same time for the list element?
What will happen if list contains multiple entries for the same device?
`rangeStart` and `rangeEnd` are included or excluded from the resulting range?

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,72 @@
...
+# IB VF GUID Configuration
+
+## Summary
+Allow SR-IOV Network Operator to use a static configuration file from the host filesystem to assign GUIDs to IB VFs
+
+## Motivation
+We have customers using the SR-IOV operator to create IB VFs, and they need a way to automate GUID assignment,
+so that IB VFs are automatically bound to the required PKeys and no additional manual configuration is needed.
+In this use case, the GUID configuration is static and known in advance.
+Now the GUIDs are assigned by the sriov-network-config-daemon randomly.
```
ykulazhenkov, line 21: Can you mention in this section the reason why we need to handle this in the operator and not in the CNI?

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Workflow Description
+
+In the `pkg/host/internal/sriov/sriov.go:configSriovVFDevices` read from the static IB config file
+and assign the GUIDs according to the configuration. Each PF is described by either PCI address
+or the PF GUID and has either a list or a range of VF GUIDs.
+
+### Implementation Details/Notes/Constraints
+
+There can be less VFs created than GUIDs. To persist the dynamic nature of the SR-IOV Network operator,
```
ykulazhenkov, line 47: What will happen if we have less GUIDs then VFs on PF?

ykulazhenkov (commented):
Thx @almaslennikov.

I added few comments

doc/design/ib-vf-configuration.md
almaslennikov: Added new sections on an alternative solution and the comparison between the two

doc/design/ib-vf-configuration.md
almaslennikov: Clarified that with an additional note

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,115 @@
...
+4. Users create a single resource pool / per PF resource pools for IB VFs
+5. Users create an IB network CR (and provide a PKey here)
+6. Users create a pod requesting an IB VF
+7. ib-sriov-cni allocates a next free GUID from the pool depending on the PKey and assigns it to the allocated VF
+
+## Comparison between the two alternatives
+
+The SR-IOV Network Operator approach:
+* Easier to implement and less error-prone
+* Manages the whole lifecycle of the VF (GUID is assigned at creation and never changes throughout the lifecycle)
```
ykulazhenkov, line 109: For me this is the main argument why I prefer the "operator" option.

adrianchiris: +1

plus if in the future there will be other CNI for IB VF then no need to re-implement.

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,115 @@
...
+    "rangeEnd": "02:00:00:00:00:aa:00:0a"
+  }
+]
+```
+
+Requirements for the config file:
+
+* `pci_adress` and `pf_guid` cannot be set at the same time for a single device - should return an error
+* if the list contains multiple entries for the same device, the first one shall be taken
+* `rangeStart` and `rangeEnd` are both included in the range
```
adrianchiris, line 83: i assume we support both "guids" and "rangeStart/End" keys  ?

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,115 @@
...
+4. Users create an IB network CR
+5. Users create pods and request resources from the specific pools
+
+### Implementation Details/Notes/Constraints
+
+There can be fewer VFs created than GUIDs. To persist the dynamic nature of the SR-IOV Network operator,
+it’s proposed not to return an error in this case but assign as many GUIDs as possible.
+To ensure that nothing breaks when users add/remove VFs, the GUID distribution order should always be the same for each individual host.
+
+If there are fewer GUIDs than VFs, then all the GUIDs should be assigned.
```
adrianchiris, line 56: we fail in this case ?
or do we randomly generate for the rest ?

if you have 5 GUIDS and 10 VFs, what will be the "other" 5 GUIDs ?

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,115 @@
...
+so that IB VFs are automatically bound to the required PKeys and no additional manual configuration is needed.
+In this use case, the GUID configuration is static and known in advance.
+Now the GUIDs are assigned by the sriov-network-config-daemon randomly.
+
+### Use Cases
+
+### Goals
+
+* IB GUID configuration can be read from a static json file on the host
+* IB GUID configuration is static and created in advance
```
SchSeba, line 27: please add `per node` right as you need unique pkeys in the cluster right?

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Workflow Description
+
+In the [`pkg/host/internal/sriov/sriov.go:configSriovVFDevices`](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/82a6d6fdce71bd88a0d9368fb1750488e9a8e4e2/pkg/host/internal/sriov/sriov.go#L458) read from the static IB config file
+and assign the GUIDs according to the configuration. Each PF is described by either PCI address
+or the PF GUID and has either a list or a range of VF GUIDs.
+
+1. A script creating the config file is deployed to the host and creates a static GUID config file
+2. SR-IOV network operator reads the file and assign GUIDs when IB VFs are created
+3. Users employ [PKey selector](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/517) in sriov-network-device-plugin to create PKey-specific resource pools
```
SchSeba, line 46: how this will get populated to the device plugin via the sriov-network-operator?

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,115 @@
...
+4. Users create an IB network CR
+5. Users create pods and request resources from the specific pools
+
+### Implementation Details/Notes/Constraints
+
+There can be fewer VFs created than GUIDs. To persist the dynamic nature of the SR-IOV Network operator,
+it’s proposed not to return an error in this case but assign as many GUIDs as possible.
+To ensure that nothing breaks when users add/remove VFs, the GUID distribution order should always be the same for each individual host.
```
SchSeba, line 54: how this will work with the parallel nic configuration that was implemented in the sriov operator?

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,115 @@
...
+
+If there are fewer GUIDs than VFs, then all the GUIDs should be assigned.
+
+### Config file
```
SchSeba, line 58: what will be the know location of this file?

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,115 @@
...
+and assign the GUIDs according to the configuration. Each PF is described by either PCI address
+or the PF GUID and has either a list or a range of VF GUIDs.
+
+1. A script creating the config file is deployed to the host and creates a static GUID config file
+2. SR-IOV network operator reads the file and assign GUIDs when IB VFs are created
+3. Users employ [PKey selector](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/517) in sriov-network-device-plugin to create PKey-specific resource pools
+4. Users create an IB network CR
+5. Users create pods and request resources from the specific pools
+
+### Implementation Details/Notes/Constraints
```
SchSeba, line 50: can you add info that if the file doesn't exist we will fallback to the method we use today in the operator

doc/design/ib-vf-configuration.md
almaslennikov: We support both but not at the same time for a single PF. Noted that in the doc

doc/design/ib-vf-configuration.md
almaslennikov: They will be assigned randomly as it's done now. Noted that in the doc

doc/design/ib-vf-configuration.md
almaslennikov: Made a note on that in the doc. Essentially, once the GUID is assigned, the corresponding PKey is stored in the sysfs. The device plugin takes it from there to populate resouce pools with PKey selector enabled

doc/design/ib-vf-configuration.md
almaslennikov: I believe that if we synchronize the access to the GUID pool, we shouldn't face any issues

doc/design/ib-vf-configuration.md
almaslennikov: I propose to use `/var/opt/infiniband_guids`. It should be writable across different cloud platforms. Noted that in the doc

ykulazhenkov:
thx, looks good to me

doc/design/ib-vf-configuration.md
SchSeba: we already have folder that is in used by the sriov operator maybe we can use that one and not create another one?

doc/design/ib-vf-configuration.md
SchSeba: please add this one to the test or validation section so we remember to test this :)

doc/design/ib-vf-configuration.md
almaslennikov: Let's store the file in `/etc/sriov-operator/infiniband/guids` then

doc/design/ib-vf-configuration.md
almaslennikov: Actually, we have changed the impl so that the pool object is immutable. No additional synchronization measures are needed. Tested that on a baremetal cluster

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Alternative solution
+
+The alternative solution is also based on the GUID configuration file being deployed on the host.
+The difference here is that GUID assignment is done on the cni level when a VF is allocated to a pod.
+ib-sriov-cni manages a host-local per-PF pool of allocated/free GUIDs and dynamically allocates the next free GUID to an allocated VF.
+
+### Workflow:
+
+1. A script is deployed to the host and creates a static GUID config file
```
SchSeba, line 105: this is out of the scope of the operator right? if that is the case please mention that and maybe give an example of how the customer is supposed to configure the file

almaslennikov: Mentioned that the step is out of scope of the operator.

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Config file
+
+Config file can be stored at `/etc/sriov-operator/infiniband/guids`. This location is writable across most cloud platforms.
+
+Example of the config file:
+
+```json
+[
+  {
```
adrianchiris, line 71: can you add here the go type used for an entry ?

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,123 @@
...
+    "pci_address": "<pci_address_1>",
+    "guids": [
+      "02:00:00:00:00:00:00:00",
+      "02:00:00:00:00:00:00:01"
+    ]
+  },
+  {
+    "pf_guid": "<pf_guid_2>",
+    "rangeStart": "02:00:00:00:00:aa:00:02",
```
adrianchiris, line 80: can we put rangeStart/End in a sub struct ? so we have something like:

```
type GUIDAllocation struct {
  PCIAddress string
  PFGUID string
  Guids []string
  GUIDsRange *GUIDRange
}

type GUIDRange struct {
  Start string
  End string
}
```

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Implementation Details/Notes/Constraints
+
+There can be fewer VFs created than GUIDs. To persist the dynamic nature of the SR-IOV Network operator,
+it’s proposed not to return an error in this case but assign as many GUIDs as possible.
+To ensure that nothing breaks when users add/remove VFs, the GUID distribution order should always be the same for each individual host.
+
+Parallel NIC configuration should be supported by making the GUID immutable. VF id should be used to determine the correct GUID for each VF.
+
+If there are fewer GUIDs than VFs, then all the GUIDs should be assigned. For the rest of VFs, apply random GUIDs as it's been done before.
```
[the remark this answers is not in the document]
adrianchiris: also what would be the behavior in case the content is invalid ?

my take : if such a file exists then we need to use it.

adrianchiris (commented):
overall looks OK. added a few small comments, once addressed im LGTM

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,123 @@
...
+---
+
+# IB VF GUID Configuration
+
+## Summary
+Allow SR-IOV Network Operator to use a static configuration file from the host filesystem to assign GUIDs to IB VFs
+
+## Motivation
+We have customers using the SR-IOV operator to create IB VFs, and they need a way to automate GUID assignment,
+so that IB VFs are automatically bound to the required PKeys and no additional manual configuration is needed.
```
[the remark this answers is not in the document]
almaslennikov: thx, added this part

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,123 @@
...
+In this use case, the GUID configuration is static and known in advance.
+Now the GUIDs are assigned by the sriov-network-config-daemon randomly.
+
+### Use Cases
+
+### Goals
+
+* Per node IB GUID configuration can be read from a static json file on the host
+* Per node IB GUID configuration is static and created in advance
```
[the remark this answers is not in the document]
almaslennikov: Moved to non-goals and added another sections on assumptions and pre-requisites

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,123 @@
...
+* Static config file is available at the same path on every host
```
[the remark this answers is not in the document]
almaslennikov: Added that part, thx

doc/design/ib-vf-configuration.md
almaslennikov: Added a go struct declaration

doc/design/ib-vf-configuration.md
almaslennikov: Sure we can, added that

doc/design/ib-vf-configuration.md
adrianchiris: @almaslennikov did you add this clarification to the design doc ?

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,147 @@
...
+  }
+]
+```
+
+```go
+type ibPfGUIDJSONConfig struct {
+    PciAddress string     `json:"pciAddress,omitempty"`
+    PfGUID     string     `json:"pfGuid,omitempty"`
+    GUIDs      []string   `json:"guids,omitempty"`
+	GUIDsRange *GUIDRange `json:"guidsRange,omitempty"`
```
adrianchiris, line 101: nit: indentation

adrianchiris (commented):
one minor nit otherwise lgtm

doc/design/ib-vf-configuration.md
```diff
@@ -0,0 +1,147 @@
...
+Config file can be stored at `/etc/sriov-operator/infiniband/guids`. This location is writable across most cloud platforms.
+
+Example of the config file:
+
+```json
+[
+  {
+    "pci_address": "<pci_address_1>",
+    "guids": [
+      "02:00:00:00:00:00:00:00",
```
zeeke, line 82: Is the VF <-> GUID  assignment implicit in this array? i.e., VF0 takes the first guid, VF1 the second, and so on?  If so, please add a note to the field description.

zeeke (commented):
Left one minor comment. Other than that, LGTM

doc/design/ib-vf-configuration.md
almaslennikov: Thanks, added a note about the ordered assignment

zeeke (approved):
LGTM

adrianchiris (approved):

adrianchiris:
LGTM, merging.

PR #657, opened 2024-03-14
Fix `IsKernelLockdownMode`

zeeke:
Passing `/bin/sh -c xxx` to `RunCommand` requires xxx to be a single string
representing a full command. Multiple arguments shifts the executed command
from:
```
/bin/sh -c "cat file.txt"
```
to:
```
/bin/sh -c cat file.txt
```

This regression has been introduced by:
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/553


SchSeba (approved):
nice catch! thanks!

Issue #632, opened 2024-02-13
Check status changes for externally managed interface - netdev device type


mlguerrero12:
follow-up of https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/530

`NeedToUpdateSriov` always returns true for externally managed interfaces with netdev device type. This causes a loop in the check status logic.

As per the comment, we always return true to be sure the admin mac address is configured as expected. Why do we need to do this? is it not overwritten when mac is set in the sriov-cni config. Can we not do this operation in the cni instead?

Additionally, `NeedToUpdateSriov` shouldn't check mtu or numVfs changes for externally managed interfaces. At least, it will avoid draining the node when the device type is a dpdk one.

c.c. @SchSeba

SchSeba:
Hi @mlguerrero12 thanks for the issue!

So the idea why we don't want to do it in the sriov-cni is from security point of view.
some network drivers like MLX and the Pensando for example don't configure mac address on the VF administrative mac this means the mac will be "00:00:00:00:00:00" if that is the case you can use the VF interface to sniff all the traffic the driver will not filter packets to is mac address only.

PR #660, opened 2024-03-18
Set netdevice as default in the CRD

mlguerrero12:
This sets netdevice as default in the CRD to not depend on the mutating webhook.

Fixes https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/632


zeeke (approved):
LGTM

Thanks @mlguerrero12

PR #646, opened 2024-03-05
Add OVSNetwork CRD and controller for it

ykulazhenkov:
This PR adds OVSNetwork CRD and controller for it.

Context is #640 

Depends on: #663. Last two commits to review

PR #663, opened 2024-03-20
Refactor SriovNetwork and SriovIBNetwork controllers to use same code

ykulazhenkov:
Controllers were refactoring to use genericNetworkReconciler, which contains all logic

This is necessary to make it easier to add new controllers.

### DIFF
--- a/controllers/generic_network_controller.go
+++ b/controllers/generic_network_controller.go
@@ -1,0 +1,253 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


ykulazhenkov:
used by #646

controllers/generic_network_controller.go
```diff
@@ -0,0 +1,253 @@
+/*
+Copyright 2021.
```
e0ne, line 2: nit: please change to 2024 if you do another update

e0ne (approved):
Good cleanup, thanks @ykulazhenkov

ykulazhenkov:
@zeeke @SchSeba 
Could you please review this PR? :)
Thx!

PR #666, opened 2024-03-24
Implement RDMA subsystem mode change

e0ne:
Now it's possible to configure RDMA subsystem mode using SR-IOV Network Operator in systemd mode.

We can't configure RDMA subsystem in a daemon mode because it should be done on host before any network namespace is created.


pkg/systemd/systemd.go
```diff
@@ -64,15 +65,21 @@ func ReadConfFile() (spec *SriovConfig, err error) {
...
 }
 
-func WriteConfFile(newState *sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
+func WriteConfFile(newState *sriovnetworkv1.SriovNetworkNodeState, netPoolConfig *sriovnetworkv1.SriovNetworkPoolConfig) (bool, error) {
 	newFile := false
 	// remove the device plugin revision as we don't need it here
 	newState.Spec.DpConfigVersion = ""
 
+	// shared mode is a default on OS
+	rdmaMode := consts.RdmaSubsystemModeShared
```
ykulazhenkov, line 74: I think we should try to query/change mode only in case if rdmaMode parameter is explicitly set in the poolConfig, to provide a safer behavior for ENVs which doesn't use RDMA.

cmd/sriov-network-config-daemon/service.go
```diff
@@ -152,6 +152,17 @@ func phasePre(setupLog logr.Logger, conf *systemd.SriovConfig, hostHelpers helpe
 	hostHelpers.TryEnableTun()
 	hostHelpers.TryEnableVhostNet()
 
+	rdmaSubsystem, err := hostHelpers.GetRDMASubsystem()
```
ykulazhenkov, line 155: I think we should execute this logic only if mode configuration is explicitly requested by a user.

api/v1/sriovnetworkpoolconfig_types.go
```diff
@@ -21,6 +21,10 @@ type SriovNetworkPoolConfigSpec struct {
 	// Drain will respect Pod Disruption Budgets (PDBs) such as etcd quorum guards,
 	// even if maxUnavailable is greater than one.
 	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
+
+	// +kubebuilder:validation:Enum=shared;exclusive
+	// RDMA subsystem. Allowed value "shared", "exclusive".
+	RdmaMode string `json:"rdmaMode,omitempty"`
```
ykulazhenkov, line 27: This option is only valid for systemd mode? 
Do we want to document this somehow?

cmd/sriov-network-config-daemon/service.go
e0ne: done

api/v1/sriovnetworkpoolconfig_types.go
e0ne: done as log message in a SriovNetworkPoolConfig controller

cmd/sriov-network-config-daemon/service.go
```diff
@@ -152,6 +152,19 @@ func phasePre(setupLog logr.Logger, conf *systemd.SriovConfig, hostHelpers helpe
 	hostHelpers.TryEnableTun()
 	hostHelpers.TryEnableVhostNet()
 
+	if conf.RdmaMode != "" {
+		rdmaSubsystem, err := hostHelpers.GetRDMASubsystem()
+		if err != nil {
+			setupLog.Error(err, "failed to get RDMA subsystem mode")
```
ykulazhenkov, line 158: If conf.RdmaMode is not empty string, then the user explicitly requested RDMA mode configuration. I think we can return error in  this case. WDYT?

cmd/sriov-network-config-daemon/service.go
```diff
@@ -152,6 +152,19 @@ func phasePre(setupLog logr.Logger, conf *systemd.SriovConfig, hostHelpers helpe
...
+		}
+		if rdmaSubsystem != conf.RdmaMode {
+			err = hostHelpers.SetRDMASubsystem(conf.RdmaMode)
+			if err != nil {
+				setupLog.Error(err, "failed to set RDMA subsystem mode")
```
ykulazhenkov, line 163: Do we want to return error here?

pkg/host/internal/kernel/kernel.go
```diff
@@ -522,6 +522,34 @@ func (k *kernel) InstallRDMA(packageManager string) error {
 	return nil
 }
 
+func (k *kernel) GetRDMASubsystem() (string, error) {
+	log.Log.Info("GetRDMASubsystem(): retrieving RDMA subsystem mode")
+	chrootDefinition := utils.GetChrootExtension()
```
ykulazhenkov, line 527: We have helper to enter chroot (part of utilsHelper). Do we want to use it here?

ykulazhenkov (commented):
I added few additional comments

pkg/host/internal/kernel/kernel.go
e0ne: We'd got the same implementation in all `kernel' methods. Let's do it in a scope of a separate PR

cmd/sriov-network-config-daemon/service.go
e0ne: we want, thanks!

cmd/sriov-network-config-daemon/service.go
e0ne: done

e0ne:
@SchSeba  could you please review this PR?

pkg/daemon/daemon.go
```diff
@@ -8,6 +8,8 @@ import (
 	"sync"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
+
```
SchSeba, line 12: you can remove this empty line here

pkg/host/internal/kernel/kernel.go
```diff
@@ -522,6 +522,34 @@ func (k *kernel) InstallRDMA(packageManager string) error {
...
+
+	stdout, stderr, err := k.utilsHelper.RunCommand("/bin/sh", "-c", fmt.Sprintf("%s /usr/bin/rdma system show", chrootDefinition))
+	if err != nil && len(stderr) != 0 {
+		log.Log.Error(err, "GetRDMASubsystem(): failed to get RDMA subsystem mode", "stdout", stdout, "stderr", stderr)
+		return "", err
+	}
+
+	// Example of an =output: netns shared copy-on-fork on
+	subsystem := strings.Fields(stdout)[1]
```
SchSeba, line 536: =output -> output

pkg/host/internal/kernel/kernel.go
```diff
@@ -522,6 +522,34 @@ func (k *kernel) InstallRDMA(packageManager string) error {
...
+	return subsystem, nil
```
SchSeba, line 537: please add a len validation here, it's always hard to work with command outputs

pkg/utils/cluster.go
```diff
@@ -26,6 +28,14 @@ const (
 	controlPlaneNodeLabelKey = "node-role.kubernetes.io/control-plane"
 )
 
+var (
+	oneNode     = intstr.FromInt32(1)
+	defaultNpcl = &sriovnetworkv1.SriovNetworkPoolConfig{Spec: sriovnetworkv1.SriovNetworkPoolConfigSpec{
+		MaxUnavailable: &oneNode,
+		NodeSelector:   &metav1.LabelSelector{},
+		RdmaMode:       "shared"}}
+)
```
SchSeba, line 37: should this one be empty by default so we don't change anything or we want the default to be shared?

controllers/sriovnetworkpoolconfig_controller.go
```diff
@@ -73,6 +73,19 @@ func (r *SriovNetworkPoolConfigReconciler) Reconcile(ctx context.Context, req ct
 		return reconcile.Result{}, err
 	}
 
+	// RdmaMode could be set in systemd mode only
+	if instance.Spec.RdmaMode != "" {
+		operatorConfig := &sriovnetworkv1.SriovOperatorConfig{}
```
SchSeba, line 78: I think we need to implement this also in the webhook

pkg/daemon/daemon.go
```diff
@@ -439,7 +441,19 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
-		systemdConfModified, err := systemd.WriteConfFile(dn.desiredNodeState)
+		// get node object
+		node := &corev1.Node{}
+		err := dn.client.Get(context.TODO(), client.ObjectKey{Name: vars.NodeName}, node)
+		if err != nil {
+			log.Log.Error(err, "nodeStateSyncHandler(): failed to get node object")
+			return err
+		}
+		netPoolConfig, _, err := utils.FindNodePoolConfig(context.Background(), node, dn.client)
+		if err != nil {
```
SchSeba, line 452: I don't think this is the right way to handle it.

we can talk about it in the community meeting but in general my idea will be:

add a new section in the sriovNetworkNodeState.Spec.Config.Rdma Mode
add a new section in the sriovNetworkNodeState.Status.Config.Rdma Mode

so the configure daemon continues to only want to NodeState only

also in case of a change in the nodePool or selecting a different node pool for this node we will need to wait for a policy change so the reconcile will act on the nodePool change

WDYT?

SchSeba (commented):
nice work!

in general I have an issue with the implementation in the config-daemon section.
I don't think the daemon should search for NodePools the operator should allocate the requested configuration in the spec of the node state

pkg/daemon/daemon.go
e0ne: sounds good to me! I'll update my PR

pkg/host/internal/kernel/kernel.go
```diff
@@ -522,6 +522,34 @@ func (k *kernel) InstallRDMA(packageManager string) error {
 	return nil
 }
 
+func (k *kernel) GetRDMASubsystem() (string, error) {
```
SchSeba, line 525: please use https://github.com/vishvananda/netlink/blob/578e95cc3190839cd7ce259a699d147b85687085/rdma_link_linux.go#L235

e0ne: done

api/v1/sriovnetworknodestate_types.go
```diff
@@ -26,6 +26,7 @@ import (
 // SriovNetworkNodeStateSpec defines the desired state of SriovNetworkNodeState
 type SriovNetworkNodeStateSpec struct {
 	Interfaces Interfaces `json:"interfaces,omitempty"`
+	RdmaMode   string     `json:"rdmaMode,omitempty"`
```
SchSeba, line 29: we need to talk about this one in the community I think we should have an higher level object first something like System/Host and we have that inside

api/v1/sriovnetworknodestate_types.go
```diff
@@ -87,6 +88,7 @@ type VirtualFunction struct {
 // SriovNetworkNodeStateStatus defines the observed state of SriovNetworkNodeState
 type SriovNetworkNodeStateStatus struct {
 	Interfaces    InterfaceExts `json:"interfaces,omitempty"`
+	RdmaMode      string        `json:"rdmaMode,omitempty"`
```
SchSeba, line 91: we need to talk about this one in the community I think we should have an higher level object first something like System/Host and we have that inside

pkg/host/internal/kernel/kernel.go
e0ne: no need this anymore since we use netutils

pkg/utils/cluster.go
e0ne: done

api/v1/sriovnetworknodestate_types.go
e0ne: done

api/v1/sriovnetworknodestate_types.go
e0ne: done

SchSeba (commented):
nice work getting close to be ready :)

cmd/sriov-network-config-daemon/service.go
```diff
@@ -152,6 +153,21 @@ func phasePre(setupLog logr.Logger, conf *systemd.SriovConfig, hostHelpers helpe
 	hostHelpers.TryEnableTun()
 	hostHelpers.TryEnableVhostNet()
 
+	if conf.Spec.System.RdmaMode != "" {
+		rdmaSubsystem, err := netlink.RdmaSystemGetNetnsMode()
```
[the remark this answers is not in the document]
e0ne: done

cmd/sriov-network-config-daemon/service.go
```diff
@@ -152,6 +153,21 @@ func phasePre(setupLog logr.Logger, conf *systemd.SriovConfig, hostHelpers helpe
...
+		if err != nil {
+			setupLog.Error(err, "failed to get RDMA subsystem mode")
+			return fmt.Errorf("failed to get RDMA subsystem mode: %v", err)
+		}
+		if rdmaSubsystem != conf.Spec.System.RdmaMode {
+			err = netlink.RdmaSystemSetNetnsMode(conf.Spec.System.RdmaMode)
```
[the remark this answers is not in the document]
e0ne: done

deployment/sriov-network-operator/templates/clusterrole.yaml
```diff
@@ -58,3 +58,6 @@ rules:
   - apiGroups: [ "config.openshift.io" ]
     resources: [ "infrastructures" ]
     verbs: [ "get", "list", "watch" ]
+  - apiGroups: [ "sriovnetwork.openshift.io" ]
```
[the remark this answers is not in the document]
e0ne: done

api/v1/sriovnetworknodestate_types.go
```diff
@@ -23,9 +23,14 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+type System struct {
```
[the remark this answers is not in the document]
e0ne: done

pkg/daemon/writer.go
```diff
@@ -127,7 +128,13 @@ func (w *NodeStateStatusWriter) pollNicStatus() error {
 	if err != nil {
 		return err
 	}
+	rdmaMode, err = w.hostHelper.GetRDMASubsystem()
```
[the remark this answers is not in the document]
e0ne: renamed to `DiscoverRDMASubsystem`. It will always return a string so there is no need to introduce a new type

adrianchiris:
@e0ne can you rebase the pr ?

adrianchiris (dismissed):

api/v1/sriovnetworkpoolconfig_types.go
adrianchiris: @e0ne cant we set this using module parameter ? 

https://github.com/torvalds/linux/blob/55027e689933ba2e64f3d245fb1ff185b3e7fc81/drivers/infiniband/core/device.c#L128C1-L128C19

pkg/utils/cluster.go
```diff
@@ -161,3 +171,94 @@ func AnnotateNode(ctx context.Context, nodeName string, key, value string, c cli
 
 	return AnnotateObject(ctx, node, key, value, c)
 }
+
+func FindNodePoolConfig(ctx context.Context, node *corev1.Node, c client.Client) (*sriovnetworkv1.SriovNetworkPoolConfig, []corev1.Node, error) {
```
adrianchiris, line 175: nit: add docstring

pkg/utils/cluster.go
```diff
@@ -26,6 +28,14 @@ const (
 	controlPlaneNodeLabelKey = "node-role.kubernetes.io/control-plane"
 )
 
+var (
+	oneNode     = intstr.FromInt32(1)
+	defaultNpcl = &sriovnetworkv1.SriovNetworkPoolConfig{Spec: sriovnetworkv1.SriovNetworkPoolConfigSpec{
```
adrianchiris, line 33: nit: can we use full name here ? e.g defaultPoolConfig ?

also the 'l' at the end is not related

pkg/utils/cluster.go
adrianchiris: also im thinking we should have two functions:

1. find node pool for node
2. find nodes for node pool (with special handling for case where default node pool was provided)

WDYT ?


Also please add UT for whatever we end up with

pkg/utils/cluster.go
```diff
@@ -161,3 +171,94 @@ func AnnotateNode(ctx context.Context, nodeName string, key, value string, c cli
...
+	} else if len(selectedNpcl) == 1 {
+		// found one pool for our node
+		logger.V(2).Info("found sriovNetworkPool", "pool", *selectedNpcl[0])
+		selector, err := metav1.LabelSelectorAsSelector(selectedNpcl[0].Spec.NodeSelector)
+		if err != nil {
+			logger.Error(err, "failed to create label selector from nodeSelector", "nodeSelector", selectedNpcl[0].Spec.NodeSelector)
+			return nil, nil, err
+		}
+
+		// list all the nodes that are also part of this pool and return them
```
adrianchiris, line 237: for those nodes why arent we validating they match exactly one ncp ? like in L223

pkg/daemon/daemon.go
```diff
@@ -420,6 +421,13 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 	// When using systemd configuration we write the file
 	if vars.UsingSystemdMode {
 		log.Log.V(0).Info("nodeStateSyncHandler(): writing systemd config file to host")
+		// get node object
+		node := &corev1.Node{}
```
adrianchiris, line 425: i dont see node is being used in this scope.

api/v1/sriovnetworknodestate_types.go
```diff
@@ -114,10 +115,15 @@ type OVSUplinkConfigExt struct {
 	Interface OVSInterfaceConfig `json:"interface,omitempty"`
 }
 
+type System struct {
+	RdmaMode string `json:"rdmaMode,omitempty"`
```
SchSeba, line 119: I think you can add here also 
```
// +kubebuilder:validation:Enum=shared;exclusive
// RDMA subsystem. Allowed value "shared", "exclusive".
```

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -269,6 +270,13 @@ func (r *SriovNetworkNodePolicyReconciler) syncAllSriovNetworkNodeStates(ctx con
 		ns.Name = node.Name
 		ns.Namespace = vars.Namespace
 		j, _ := json.Marshal(ns)
+		netPoolConfig, _, err := utils.FindNodePoolConfig(context.Background(), &node, r.Client)
```
SchSeba, line 273: please use the context from the function don't create a new one

controllers/sriovnetworkpoolconfig_controller.go
```diff
@@ -73,6 +73,19 @@ func (r *SriovNetworkPoolConfigReconciler) Reconcile(ctx context.Context, req ct
 		return reconcile.Result{}, err
 	}
 
+	// RdmaMode could be set in systemd mode only
+	if instance.Spec.RdmaMode != "" {
```
SchSeba, line 77: please remove this one as we support this on both modes

pkg/host/internal/kernel/kernel.go
```diff
@@ -522,6 +523,29 @@ func (k *kernel) InstallRDMA(packageManager string) error {
 	return nil
 }
 
+func (k *kernel) DiscoverRDMASubsystem() (string, error) {
```
SchSeba, line 526: I think we can move this function to the network or sriov package

pkg/host/internal/kernel/kernel.go
```diff
@@ -522,6 +523,29 @@ func (k *kernel) InstallRDMA(packageManager string) error {
...
+	log.Log.Info("DiscoverRDMASubsystem(): retrieving RDMA subsystem mode")
+	subsystem, err := netlink.RdmaSystemGetNetnsMode()
```
SchSeba, line 528: please use the netlink interface in the project so we can have a mock for it on unit tests

pkg/host/internal/kernel/kernel.go
```diff
@@ -522,6 +523,29 @@ func (k *kernel) InstallRDMA(packageManager string) error {
...
+
+	if err != nil {
+		log.Log.Error(err, "DiscoverRDMASubsystem(): failed to get RDMA subsystem mode")
+		return "", err
+	}
+
+	return subsystem, nil
+}
+
+func (k *kernel) SetRDMASubsystem(mode string) error {
```
SchSeba, line 538: I think this function is no needed now that we use the modprobe file

pkg/utils/cluster.go
SchSeba: +1

SchSeba:
Hi @e0ne can you please rebase the PR?

e0ne:
> Hi @e0ne can you please rebase the PR?

done

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -272,6 +273,13 @@ func (r *SriovNetworkNodePolicyReconciler) syncAllSriovNetworkNodeStates(ctx con
 		ns.Name = node.Name
 		ns.Namespace = vars.Namespace
 		j, _ := json.Marshal(ns)
+		netPoolConfig, _, err := utils.FindNodePoolConfig(ctx, &node, r.Client)
```
SchSeba, line 276: just a general todo here we should have in memory map I think for this

pkg/host/internal/network/network.go
```diff
@@ -23,6 +24,8 @@ import (
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
 )
 
+var ManifestsPath = "./bindata/manifests"
```
SchSeba, line 27: lets put this in consts

pkg/host/internal/network/network.go
```diff
@@ -421,3 +424,40 @@ func (n *network) GetPciAddressFromInterfaceName(interfaceName string) (string,
...
+}
+
+func (n *network) SetRDMASubsystem(mode string) error {
+	log.Log.Info("SetRDMASubsystem(): Updating RDMA subsystem mode")
+
+	modeValue := 1
+	if mode == "exclusive" {
+		modeValue = 0
+	}
+	config := fmt.Sprintf("options ib_core netns_mode=%d\n", modeValue)
```
SchSeba, line 447: I think if you use this we can remove the other file in the manifests no?

pkg/host/internal/network/network.go
```diff
@@ -421,3 +424,40 @@ func (n *network) GetPciAddressFromInterfaceName(interfaceName string) (string,
...
+	err := os.WriteFile("/etc/modprobe.d/ib_core.conf", []byte(config), 0644)
```
SchSeba, line 448: please use the getExtention here so we know if we are not inside a chroot

pkg/host/internal/network/network.go
```diff
@@ -421,3 +424,40 @@ func (n *network) GetPciAddressFromInterfaceName(interfaceName string) (string,
...
+
+	if err != nil {
+		log.Log.Error(err, "SetRDMASubsystem(): failed to write ib_core config")
+		return fmt.Errorf("failed to write ib_core config: %v", err)
+	}
+
+	err = os.WriteFile(path.Join(consts.Chroot, "/etc/modprobe.d/ib_core.conf"), []byte(config), 0644)
```
SchSeba, line 455: this looks like a duplicate?

pkg/render/render.go
```diff
@@ -77,6 +77,14 @@ func RenderDir(manifestDir string, d *RenderData) ([]*unstructured.Unstructured,
 	return out, nil
 }
 
+func RenderToString(path string, d *RenderData) (string, error) {
```
SchSeba, line 80: I don't think we need this function where we use it?

pkg/utils/cluster.go
SchSeba, line 175: please move this function to the helpers in controllers better then adding more stuff to utils

SchSeba (commented):
nice work!

I left some small comments

controllers/drain_controller.go
```diff
@@ -346,94 +336,7 @@ func (dr *DrainReconcile) tryDrainNode(ctx context.Context, node *corev1.Node) (
...
-
-		defaultNodeLists := []corev1.Node{}
-		for _, nodeObj := range nodeList.Items {
-			if _, exist := nodesInPools[nodeObj.Name]; !exist {
-				defaultNodeLists = append(defaultNodeLists, nodeObj)
-			}
-		}
-		return defaultNpcl, defaultNodeLists, nil
-	}
+	return utils.FindNodePoolConfig(ctx, node, dr.Client)
```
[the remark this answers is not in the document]
e0ne: it makes sense. done

controllers/sriovnetworknodepolicy_controller.go
e0ne: could you please elaborate on this?

pkg/host/internal/network/network.go
e0ne: It's not needed anymore, so I deleted it

pkg/host/internal/network/network.go
e0ne: it's deleted now

pkg/host/internal/network/network.go
e0ne: rebase issue, it's deleted now

pkg/utils/cluster.go
e0ne: deleted

api/v1/sriovnetworknodestate_types.go
e0ne: done

SchSeba (commented):
in general another point we need to take care here is when we create/update a pool it will not do anything we need to wait for the nodePolicy controller to apply this.

we have two options here:
1. we pass a channel so the pool controller can trigger a policy one
2. we handle the system section in the pool controller directly

I must say I am not sure what is the best option @adrianchiris @zeeke WDYT?

SchSeba:
/hold

this doesn't work on OCP it puts the node in boot loop because the mode didn't change.

adrianchiris:
> in general another point we need to take care here is when we create/update a pool it will not do anything we need to wait for the nodePolicy controller to apply this.
> 
> we have two options here:
> 
> 1. we pass a channel so the pool controller can trigger a policy one
> 2. we handle the system section in the pool controller directly
> 
> I must say I am not sure what is the best option @adrianchiris @zeeke WDYT?

we can watch on pool obj as well and trigger reconcile event.

adrianchiris:
> this doesn't work on OCP it puts the node in boot loop because the mode didn't change.

writing files under `/etc/modprobe.d`  is not persistent in coreOS ?

SchSeba:
> we can watch on pool obj as well and trigger reconcile event.

sure that also can work on we don't expect machine changes on it

SchSeba:
closing this one in favor or https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/799

PR #669, opened 2024-03-31
Remove DPversion from nodeState

SchSeba:
removing this field from the nodeState as we don't need it.

### DIFF
--- a/api/v1/sriovnetworknodestate_types.go
+++ b/api/v1/sriovnetworknodestate_types.go
@@ -25,8 +25,7 @@ import (
 
 // SriovNetworkNodeStateSpec defines the desired state of SriovNetworkNodeState
 type SriovNetworkNodeStateSpec struct {
-	DpConfigVersion string     `json:"dpConfigVersion,omitempty"`
-	Interfaces      Interfaces `json:"interfaces,omitempty"`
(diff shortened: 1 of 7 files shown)


SchSeba:
we don't need this field it just makes more problems than help.

It was also not really working as we first called in the code the NodeState sync and only after that did we call the device plugin update so the hash was not really right. that also caused another reconcile loop after 5 min from the original one...

```
	sort.Sort(sriovnetworkv1.ByPriority(policyList.Items))
	// Sync SriovNetworkNodeState objects
	if err = r.syncAllSriovNetworkNodeStates(ctx, defaultOpConf, policyList, nodeList); err != nil {
	}
	// Sync Sriov device plugin ConfigMap object
	if err = r.syncDevicePluginConfigMap(ctx, defaultOpConf, policyList, nodeList); err != nil {
	}
```

SchSeba:
@zeeke @bn222 @e0ne @ykulazhenkov please have a look

zeeke (approved):
LGTM

zeeke (approved):
LGTM

ykulazhenkov (approved):
LGTM

e0ne (approved):
LGTM

PR #670, opened 2024-04-01
Check for kernel lockdown only in mlx plugin

SchSeba:
We need to check for the kernel lockdown only in
the mlx plugin.

if the firmware in the mlx card is already configured as needed we can allocate virtual functions even if the system is in lockdown


SchSeba:
@e0ne @ykulazhenkov please take a look this can be a way to make the sriov-operator work with mlx cards in secure mode.

in general the idea is
1. the user configure the max number of vfs he wants in the firmware as part of the installation process (not secure boot)
2. switch the system to secure boot
3. install the operator with disablePlugin option for mlx
4. create the sriovNetworkNodePolicy as always

e0ne (approved):
LGTM

zeeke (approved):
LGTM

PR #671, opened 2024-04-01
Fix root device overlapping

SchSeba:


api/v1/helper.go
```diff
@@ -547,6 +547,19 @@ func ParsePFName(name string) (ifName string, rngSt, rngEnd int, err error) {
 	return
 }
 
+// ParseRootDevice: partse the rootDevice with VF range
```
e0ne, line 550: nit: partse => parse

SchSeba: done thanks!

api/v1/helper.go
```diff
@@ -481,7 +481,7 @@ func (p *SriovNetworkNodePolicy) generateVfGroup(iface *InterfaceExt) (*VfGroup,
 	var rngStart, rngEnd int
 	found := false
 	for _, selector := range p.Spec.NicSelector.PfNames {
```
zeeke, line 483: Ranges are still defined on `PfNames` field only. 
You can verify it by adding unit tests like  https://github.com/k8snetworkplumbingwg/sriov-network-operator/commit/a92b76433a5debe9749e82a66ebcb35536b2453e .

SchSeba: you are right!

I think supporting rage in the root device will be much harder than I was expecting so I will first fix the webhook issue and than we can take to the community to talk about the range support for rootDevices.

let me update the PR

zeeke (approved):
LGTM

SchSeba:
Hi @e0ne @Eoghan1232 when you have time please take a look on this small PR

Eoghan1232 (approved):

SchSeba:
Thanks for the review folks!

merging this one

PR #673, opened 2024-04-02
Add featuregate package and use it in controllers

ykulazhenkov:
This change is required to simplify usage of feature gates outside of SriovOperatorConfig controller.

 SriovOperatorConfig controller is responsible for initialization of the featuregate helper.
Other controllers can check configured feature gates by using the interface.

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -48,6 +48,7 @@ import (
 
 	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
 	constants "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/featuregate"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/render"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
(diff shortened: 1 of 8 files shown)


controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -58,7 +59,8 @@ const nodePolicySyncEventName = "node-policy-sync-event"
 // SriovNetworkNodePolicyReconciler reconciles a SriovNetworkNodePolicy object
 type SriovNetworkNodePolicyReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme      *runtime.Scheme
+	FeatureGate featuregate.FeatureGate
```
zeeke, line 63: This field is not used. I guess it will be used in some future PR, right?

pkg/featuregate/featuregate.go
```diff
@@ -0,0 +1,73 @@
...
+// FeatureGate provides methods to check state of the feature
+type FeatureGate interface {
+	// IsEnabled returns state of the feature,
+	// if feature name is unknown will always return false
+	IsEnabled(feature string) bool
+	// Init set state for the features from the provided map.
+	// completely removes the previous state
+	Init(features map[string]bool)
+	// Set state of the specific feature, keeps existing state for other features
+	Set(feature string, state bool)
```
zeeke, line 18: Same here, `Set` seems to be used in tests only. How is it going to be used?

controllers/sriovnetworknodepolicy_controller.go
ykulazhenkov: Yes, this is preparation for the next PR.

pkg/featuregate/featuregate.go
ykulazhenkov: @zeeke I removed `Set` function

ykulazhenkov:
@zeeke Can we merge this PR if it looks good for you? thx

zeeke (approved):
LGTM

I was waiting for the CI to finish

PR #675, opened 2024-04-04
Allow switchdev+externallyManged configs in webhook

ykulazhenkov:
Current implementation supports this combination.
NIC should be already in switchdev mode and have enough VFs to pass checks in the operator.

cc @e0ne @zeeke @SchSeba


e0ne (approved):
Please, add a link to commit which implements `Current implementation supports this combination` if you'll do one more change to this PR

SchSeba (commented):
can you please add one or two unit tests to cover a validation case
switch dev but no vf, the other way around (both failed ) and one that pass the validation

ykulazhenkov:
> can you please add one or two unit tests to cover a validation case switch dev but no vf, the other way around (both failed ) and one that pass the validation

thanks @SchSeba,
I pushed some tests to validate this behavior

ykulazhenkov (commented):
PR rebased

ykulazhenkov:
@SchSeba I rebased the PR and added some tests. PTAL 
Thanks

SchSeba (approved):
LGTM
nice work!

PR #679, opened 2024-04-10
Fix crash panic in operator when policy has empty node selector

SchSeba:
The commit will fix a nil pointer exception in the operator when there is a policy with an empty selector the update of the devicePlugin daemonSet will fail


SchSeba:
@zeeke @Eoghan1232 @e0ne please take a look for this quick fix

controllers/helper.go
```diff
@@ -302,17 +365,7 @@ func syncDaemonSet(ctx context.Context, client k8sclient.Client, scheme *runtime
 
 		if equality.Semantic.DeepEqual(in.OwnerReferences, ds.OwnerReferences) &&
 			equality.Semantic.DeepDerivative(in.Spec, ds.Spec) {
-			// DeepDerivative has issue detecting nodeAffinity change
```
mlguerrero12, line 305: Why is this part not needed anymore?

controllers/helper.go
```diff
@@ -302,17 +365,7 @@ func syncDaemonSet(ctx context.Context, client k8sclient.Client, scheme *runtime
...
-			// https://bugzilla.redhat.com/show_bug.cgi?id=1914066
-			// DeepDerivative doesn't detect changes in containers args section
-			// This code should be fixed both with NodeAffinity comparation
-			if equality.Semantic.DeepEqual(in.Spec.Template.Spec.Affinity.NodeAffinity,
-				ds.Spec.Template.Spec.Affinity.NodeAffinity) &&
-				equality.Semantic.DeepEqual(in.Spec.Template.Spec.Containers[0].Args,
-					ds.Spec.Template.Spec.Containers[0].Args) {
-				logger.V(1).Info("Daemonset spec did not change, not updating")
```
mlguerrero12, line 313: I think the log is still valid before `return nil`

controllers/helper.go
SchSeba: because `equality.Semantic.DeepDerivative` is able to handle that

controllers/helper.go
SchSeba: I didn't remove I just move the functions to the right file.

before this change functions in helpers called functions in the controller file that breaks the point of helpers

mlguerrero12: I meant this line

`logger.V(1).Info("Daemonset spec did not change, not updating")`

SchSeba: ohhhh you are right! sorry stupid me :(

SchSeba:
@mlguerrero12 done thanks! :)

mlguerrero12 (approved):
LGTM

SchSeba:
Hi @zeeke @e0ne @Eoghan1232 please take a look when you have some time this is a small fix

zeeke (approved):
LGTM

SchSeba:
Thanks folks!
merging

PR #680, opened 2024-04-12
Change system dependencies for "post" systemd service

ykulazhenkov:
This is required to be compatible with OVS installations that use SysV init scripts (wrapped with systemd-sysv-generator).
 On such systems, OVS depends on the network-online.target, so the start order of services changes (OVS will start after the "post" service). The proposed solution is to add the following explicit dependencies 

After: openvswitch-switch.service
Before: kubelet.service.

Note:  the dependency is ignored if the openvswitch-switch.service is not found.

cc @e0ne @SchSeba @zeeke


e0ne (approved):
LGTM

SchSeba (approved):
LGTM

PR #683, opened 2024-04-17
Make OVS_CNI_IMAGE truly optional

mandre:
The `hack/env.sh` script expected that `OVS_CNI_IMAGE` is set, and exited with an error when it was not. There is no good reason for erroring out on missing variable for an optional image. Instead, set the variable to empty string when it is unset.

When `SKIP_VAR_SET` is set, the default for `OVS_CNI_IMAGE` remains a valid image.


SchSeba (approved):
LGTM

zeeke (approved):
LGTM

SchSeba:
@ykulazhenkov any objections here?

ykulazhenkov:
> @ykulazhenkov any objections here?

no, LGTM

ykulazhenkov (approved):

e0ne (approved):

PR #686, opened 2024-04-19
Let "depoly-setup.sh" and "undeploy.sh" fail on error

thom311:
Ignoring errors seems not best. It can make it harder to find problems with the script.

If a command in the script is known to fail for valid reasons, then that should be treated specially (for example, with `|| :`). By default, treat any unexpected failure as fatal and propagate the failure to the caller.

Fixes: 88441502023d ('Add RBAC for cluster resource')

CC: @pliurh

### DIFF
--- a/hack/deploy-setup.sh
+++ b/hack/deploy-setup.sh
@@ -3,7 +3,7 @@
 # to deploy.  It assumes it is capable of login as a
 # user who has the cluster-admin role
 
-# set -euxo pipefail
+set -eo pipefail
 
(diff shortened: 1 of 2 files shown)


SchSeba (approved):
LGTM

adrianchiris (approved):
lgtm

SchSeba:
Hi @thom311 can you please do a force push I am not able to trigger the k8s lane for some reason

SchSeba:
all green merging nice work!

PR #695, opened 2024-05-01
don't reset the device plugin twice

SchSeba:


e0ne (approved):

SchSeba:
Thanks for the review!

merging

PR #698, opened 2024-05-10
Use helm values to expand the supported NICs

manuelbuil:
The `supported-nic-ids` ConfigMap is a whitelist that includes details of some NICs. However, the user might want to add other NICs to that list so that the sriov operator works in his env.

The only way to do it right now is deploying the sriov-operator, editing that configmap and then restarting some of the sriov-operator components so that the new config is consumed.

This PR adds a new way to do it which can be easily automated: use a values.yaml parameter in the helm chart to include extra NICs. We can add the extra NICs when installing the sriov-operator, thus not requiring to restart any of the components


adrianchiris:
Hi,
can you add some info in the  README of the chart ?

manuelbuil:
> Hi, can you add some info in the README of the chart ?

Done!

adrianchiris (approved):

PR #700, opened 2024-05-17
Fix issue preventing lowering of PF's MTU

almaslennikov:
When the MTU set in the SRIOV Network Node Policy is lower than the actual MTU of the PF, it cannot be changed due to a condition in question. This triggers the reconcile loop for the Node state indefinitely, preventing the configuration from completing.

### DIFF
--- a/test/conformance/tests/test_sriov_operator.go
+++ b/test/conformance/tests/test_sriov_operator.go
@@ -1035,12 +1035,12 @@ var _ = Describe("[sriov] operator", func() {
 				}, 3*time.Minute, time.Second).Should(Equal(corev1.PodRunning))
 			})
 
-			It("should reconcile managed VF if status changes", func() {
+			It("should reconcile managed VF if status is changed", func() {
 				originalMtu := sriovDevice.Mtu
-				newMtu := 1000
+				lowerMtu := originalMtu - 500
 
-				By("manually changing the MTU")
-				_, errOutput, err := runCommandOnConfigDaemon(node, "/bin/bash", "-c", fmt.Sprintf("echo %d > /sys/bus/pci/devices/%s/net/%s/mtu", newMtu, sriovDevice.PciAddress, sriovDevice.Name))
+				By("manually decreasing the MTU")
+				_, errOutput, err := runCommandOnConfigDaemon(node, "/bin/bash", "-c", fmt.Sprintf("echo %d > /sys/bus/pci/devices/%s/net/%s/mtu", lowerMtu, sriovDevice.PciAddress, sriovDevice.Name))
 				Expect(err).ToNot(HaveOccurred())
 				Expect(errOutput).To(Equal(""))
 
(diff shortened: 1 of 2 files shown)


almaslennikov:
I see that there was a reason for never allowing to lower the PF's MTU: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/489

@SchSeba @adrianchiris What's your take on this?

I see two options here: either allow to lower the MTU of the PF or also update the condition [here](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/api/v1/helper.go#L260)

mlguerrero12:
Thanks @almaslennikov. Another way to hit this issue is when the user manually increases the mtu after having configured a policy with a lower mtu. We have a test (`should reconcile managed VF if status changes`) in `test_sriov_operator.go` for the opposite case (manually decreasing mtu). 

I think we should go for the second option, aligning `NeedToUpdateSriov` with which is actually being applied in order to avoid the loop. Please consider adding a test for this scenario as well.

adrianchiris:
IMO we should update `NeedToUpdateSriov ` so we support the use-case of a PF having a different (higher) MTU than VF as requested in spec.

SchSeba:
agree if the MTU on the PF is higher the `NeedToUpdateSriov` should return false

SchSeba:
Hi @almaslennikov please check the unit-tests

test/conformance/tests/test_sriov_operator.go
```diff
@@ -1071,6 +1071,74 @@ var _ = Describe("[sriov] operator", func() {
...
+						"PciAddress": Equal(sriovDevice.PciAddress),
+						"NumVfs":     Equal(sriovDevice.NumVfs),
+					})))
+
+				By("expecting the mtu to consistently stay at the new higher level")
+				Consistently(func() sriovv1.InterfaceExts {
+					nodeState, err := clients.SriovNetworkNodeStates(operatorNamespace).Get(context.Background(), node, metav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
+					return nodeState.Status.Interfaces
+				}, 3*time.Minute, 15*time.Second).Should(ContainElement(MatchFields(
```
adrianchiris, line 1101: This whole flow extends the this test by quite a bit.

@zeeke ? thoughts ? do we want this back and forth of the MTU in e2e tests

adrianchiris (approved):
LGTM, thx @almaslennikov

e0ne (approved):
LGTM. 

Nvidia e2e test CI seems to be broken now

mlguerrero12:
LGTM

test/conformance/tests/test_sriov_operator.go
SchSeba: I am fine with it for now if we see that it's unstable for some reason we can revisit this

SchSeba (approved):

PR #701, opened 2024-05-20
Create and update SriovNetworkNodeState CRD in a single loop

e0ne:
We don't need another 5 minutes to update SriovNetworkNodeState CRD in case if it was created.

Let's same some time between object creation and spec filling to make SR-IOV configuration more consistant and faster.


adrianchiris:
@e0ne is this one still relevant ?

SchSeba:
Hi @e0ne should we close this one?

PR #709, opened 2024-06-06
Remove webhook validation for InfiniBand and RDMA

e0ne:
Since we always deploy both mutation and validating webhooks there is no need to validate policy for InfiniBand and RDMA enabled because mutating webhook will wet 'isRdma' flag to 'true' for InfiniBand'.


SchSeba:
is it bad if we have this validation here?
the mutation webhook is coming before the validation webhook so we should be ok here no?

e0ne:
> is it bad if we have this validation here? the mutation webhook is coming before the validation webhook so we should be ok here no?

mutation webhook is always executed before validating, so it means we've got dead code: the statement will be always `false`

e0ne:
Here is admission controller documentation: https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#admission-control-phases

SchSeba (approved):
works for me :)

adrianchiris (approved):

SchSeba:
failing CI lane is not related to this change

PR #710, opened 2024-06-10
Avoid reconfiguring unmentioned DPDK VFs

zeeke:
If a Virtual Function is configured with a DPDK driver (e.g. `vfio-pci`) and it is not referred by any SriovNetworkNodePolicy, `NeedToUpdateSriov` function must not trigger a  reconfiguration. This may happen if a PF is configured by multiple policies (via PF partitioning) and a policy is deleted by the user. In these cases, the VF is not reconfigured [1] and a drain loop is started

refs:
[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/5f3c4e903f789aa177fe54686efd6c18576b7ab1/pkg/host/internal/sriov/sriov.go#L457

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -280,10 +280,8 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 
 	if ifaceSpec.NumVfs > 0 {
 		for _, vfStatus := range ifaceStatus.VFs {
-			ingroup := false
 			for _, groupSpec := range ifaceSpec.VfGroups {
 				if IndexInRange(vfStatus.VfID, groupSpec.VfRange) {
-					ingroup = true
 					if vfStatus.Driver == "" {
 						log.V(2).Info("NeedToUpdateSriov(): Driver needs update - has no driver",
 							"desired", groupSpec.DeviceType)
@@ -331,12 +329,6 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 					}
 					break
 				}
-			}
-			if !ingroup && (StringInArray(vfStatus.Driver, vars.DpdkDrivers) || vfStatus.VdpaType != "") {
-				// need to reset VF if it is not a part of a group and:
-				// a. has DPDK driver loaded
-				// b. has VDPA device
-				return true
 			}
 		}
 	}
--- a/README.md
+++ b/README.md
@@ -249,6 +249,10 @@ VF groups** (when #-notation is used in pfName field) are merged, otherwise only
 the highest priority policy is applied. In case of same-priority policies and
 overlapping VF groups, only the last processed policy is applied.
 
+When using #-notation to define VF group, no actions are taken on virtual functions that
+are not mentioned in any policy (e.g. if a policy defines a `vfio-pci` device group for a device, when 
+it is deleted the VF are not reset to the default driver).
+
 #### Externally Manage virtual functions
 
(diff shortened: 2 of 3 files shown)


api/v1/helper.go
```diff
@@ -332,10 +332,8 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 					break
 				}
 			}
-			if !ingroup && (StringInArray(vfStatus.Driver, vars.DpdkDrivers) || vfStatus.VdpaType != "") {
-				// need to reset VF if it is not a part of a group and:
-				// a. has DPDK driver loaded
-				// b. has VDPA device
+			if !ingroup && vfStatus.VdpaType != "" {
```
zeeke, line 335: @lmilleri @wizhaoredhat I'm thinking about removing also this case. Are you aware of any case in which a VF with VdpaType that is not referenced by any policy should be configured/reset by the operator?

lmilleri: @zeeke Vdpa devices should be created/delete with policies, they should not be on their own. IIUC with this change, we won't trigger a reconfiguration if vdpa device is not in a policy? If yes, it looks ok to me

zeeke: Correct, this `if` statement checks if a VF has Vdpa but it is not referenced by any policy. The current behavior is to try to reconcile it (not sure which changes are going to be done). I'm proposing to make the operator leave the device as is, as there is no policy that says it should be managed.

wizhaoredhat: Looks good to me too.

api/v1/helper.go
```diff
@@ -332,12 +330,6 @@ func NeedToUpdateSriov(ifaceSpec *Interface, ifaceStatus *InterfaceExt) bool {
 					break
 				}
 			}
-			if !ingroup && (StringInArray(vfStatus.Driver, vars.DpdkDrivers) || vfStatus.VdpaType != "") {
```
adrianchiris, line 335: shouldnt the fix be in in configureSriovInterface ?

that is, if a VF does not belong to any group -> bind it to default driver ?

zeeke: That can be a different way to solve the problem. I'm inclined to prefer fixing NeetToUpdateSriov because:
a. it's a little simpler. `configureSriovDevices` would have to find every VF that is not part of a group definition and rebind them
b. Suppose a user wants to keep a group of VFs for some host level networking. If he doesn't mention them in a SriovNetworkNodePolicy, the operator should not touch them.

I don't know if (b) is a valid and/or important use case, and if you know any particular reason to rebind the default driver, I'm open to that option.

adrianchiris: @SchSeba @ykulazhenkov thoughts ?

technically these VFs will get configured the next time they are selected.
meanwhile they will remain as is f(driver/vdpa)

i dont have an objection to current approach. more feedback is welcome.

ykulazhenkov: it looks fine for me to keep VFs configured in this case

ykulazhenkov (approved):
LGTM

api/v1/helper.go
zeeke: @adrianchiris , I added a comment to the README. Does it sound clear?
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/710/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R252-R255

adrianchiris (approved):
LGTM !

PR #711, opened 2024-06-13
Avoid reconciling field `Webhook.ClientConfig.CABundle`

zeeke:
Webhook resources (`ValidatingWebhookConfiguration` and `MutatingWebhookConfiguration`) in OpenShift are configured with `service.beta.openshift.io/inject-cabundle` in a way that a third component fills the ClientConfig.CABundle field of the webhook. When reconciling webhooks, do not override the field and avoid a flakiness, as there might be a time slot in which the API server is not configured with a valid client certificate:

```
Error from server (InternalError): error when creating "policies": Internal error occurred: failed calling webhook "operator-webhook.sriovnetwork.openshift.io": failed to call webhook: Post "https://operator-webhook-service.openshift-sriov-network-operator.svc:443/mutating-custom-resource?timeout=10s": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

Refs:
- https://docs.openshift.com/container-platform/4.15/security/certificates/service-serving-certificate.html
- https://issues.redhat.com/browse/OCPBUGS-32139


SchSeba:
Hi @zeeke,

just a general comment we need to take care of this also for the cert-manager in vanilla kubernetes we will have the same issue https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/bindata/manifests/operator-webhook/003-webhook.yaml#L12

zeeke:
> Hi @zeeke,
> 
> just a general comment we need to take care of this also for the cert-manager in vanilla kubernetes we will have the same issue https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/bindata/manifests/operator-webhook/003-webhook.yaml#L12

Good point. Though the unit test I implemented refers to the Openshift CA injector, the fix should be valid for the Cert Manager as well.

pkg/apply/merge.go
```diff
@@ -116,6 +122,103 @@ func MergeServiceAccountForUpdate(current, updated *uns.Unstructured) error {
...
+		return nil
+	}
+
+	current.GetAnnotations()
+	currentAnnotations := current.GetAnnotations()
+	if currentAnnotations == nil {
+		return nil
+	}
+
+	injectCABundle, ok := currentAnnotations["service.beta.openshift.io/inject-cabundle"]
```
zeeke, line 143: @SchSeba I think you refer to this. Removing

ykulazhenkov (approved):

controllers/sriovoperatorconfig_controller_test.go
```diff
@@ -327,5 +327,32 @@ var _ = Describe("SriovOperatorConfig controller", Ordered, func() {
 			})
 			Expect(err).ToNot(HaveOccurred())
 		})
+
+		// This test verifies that the CABundle field in the webhook configuration  added by third party components is not
+		// removed during the reconciliation loop. This is important when dealing with OpenShift certificate mangement:
+		// https://docs.openshift.com/container-platform/4.15/security/certificates/service-serving-certificate.html
+		// and when CertManager is used
+		It("should not remove the field Spec.ClientConfig.CABundle from webhook configuration when reconciling", func() {
```
SchSeba, line 335: can we add a unit test to also cover the case the CABUNDLE ENV variable exists it does gets configured?

zeeke: good point! added

adrianchiris (approved):

SchSeba:
3 approvals

merging

PR #714, opened 2024-06-16
Fix release chart update

adrianchiris:
refer to individual commits for more context.

gist:
* set correct ovs cni image
* fix operator name in chart


SchSeba (approved):
LGTM

PR #720, opened 2024-06-20
introduce a recheck for the VF interface name

SchSeba:
It's possible to have a race in the VFIsReady function. vf netdevice can have a default eth0 device name and be the time we call the netlink syscall to get the device information eth0 can be a different device.

this cause duplicate mac allocation on vf admin mac address


SchSeba:
cc @zeeke @adrianchiris
please take a look on this race condition

pkg/host/internal/sriov/sriov.go
```diff
@@ -176,8 +176,16 @@ func (s *sriov) VFIsReady(pciAddr string) (netlink.Link, error) {
 		vfLink, err = s.netlinkLib.LinkByName(vfName)
 		if err != nil {
 			log.Log.Error(err, "VFIsReady(): unable to get VF link", "device", pciAddr)
+			return false, nil
 		}
-		return err == nil, nil
+
+		// try to get the interface name again and check if the name changed
+		vfName = s.networkHelper.TryGetInterfaceName(pciAddr)
```
zeeke, line 183: The problem is that `TryGetInterfaceName(pciAddr)` might return `eth0` value that is not unique for the system, and `LinkByName('eth0')` might return a different VF.

I'm not sure this would solve the problem :thinking:

SchSeba: so the idea here was
1. getInterface name
2. vfLink object 
3. getInterface name again

and we have two options here:
if the second time it's still eth0 this means the vfLink object we did before contains the right mac address(even if the VF will change name).
if the vf change is name we will try again

zeeke: I think the problem is this:
suppose in the system there are two different VFs named `eth0`  (0000:19:00.5 and 0000:19:00.6) and we are calling `VFIsReady("0000:19:00.5")`

`vfName = s.networkHelper.TryGetInterfaceName("0000:19:00.5")` -> `eth0`
`vfLink, err = s.netlinkLib.LinkByName(vfName)` -> might return `0000:19:00.6`

then 
`vfName = s.networkHelper.TryGetInterfaceName(pciAddr)` still  returns `eth0`

and the function will return the .6 netlink reference, instead of the .5 one. WDYT?

SchSeba: oh ok now I get your point the thing is that is not possible on a given time we can only have one interface with the same name that from the kernel lock system. some more info about that from our kernel developer

```
The interface name is assigned by the kernel during netdevice_register(). It is while holding the lock (rtnl_lock) and the kernel performs a name collision check.
It is not possible to have two interfaces with the same name, but when eth0 is renamed to e.g. enp12s23, another device can use eth0 again, but not at the same time (rtnl_lock). The name of the device can be changed later (e.g. with the help of udev and this is also easily done, but again there is rtnl_lock which prevents collisions in names).
```

adrianchiris:
discussed in community meeting. we should modify the way we get the Link device by relying on interface index instead of name.

interface index is found under sysfs
`/sys/bus/pci/deivces/<address>/net/<ifaceName/ifindex`
if path does not exist we should retry (udev rule renamed interface)

pkg/host/internal/sriov/sriov.go
SchSeba: Right!

SchSeba:
/hold

waiting for https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/569

THIS PR CONTAINS A REPLACE IN GO MODS!

pkg/host/internal/network/network.go
```diff
@@ -95,10 +96,20 @@ func (n *network) TryGetInterfaceName(pciAddr string) string {
...
 	}
 
-	log.Log.V(2).Info("tryGetInterfaceName()", "name", netDevName)
+	log.Log.V(2).Info("TryGetInterfaceName()", "name", netDevName)
 	return netDevName
 }
 
+// TryGetInterfaceName tries to find the SR-IOV virtual interface index base on pci address
+func (n *network) TryGetInterfaceIndex(pciAddr string) int {
+	ifIndex, err := n.dputilsLib.GetNetIndex(pciAddr)
```
adrianchiris, line 105: you may get the wrong index in case of switched since implementation returns the first netdev interface index found.

in TryGetInterfaceNames , if in switchdev we return the "uplink representor" netdev.  since in switchdev all representors will be found under the PF /net folder.

adrianchiris: in device plugin we also have GetPFNames()[1] which also takes switchdev into account

[1] https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/fe9f2f85553267c884dc1e68b94e8561c57e8c50/pkg/utils/utils.go#L72

SchSeba:
I switch the implementation to call the tryGetInterfaceName in the new function that way we don't need any re function in the sriov network device plugin.

pkg/host/internal/sriov/sriov.go
```diff
@@ -189,12 +189,13 @@ func (s *sriov) VFIsReady(pciAddr string) (netlink.Link, error) {
 	var err error
 	var vfLink netlink.Link
 	err = wait.PollImmediate(time.Second, 10*time.Second, func() (bool, error) {
-		vfName := s.networkHelper.TryGetInterfaceName(pciAddr)
-		vfLink, err = s.netlinkLib.LinkByName(vfName)
+		vfIndex := s.networkHelper.TryGetInterfaceIndex(pciAddr)
```
ykulazhenkov, line 192: Do we want to explicitly handle -1 result?

SchSeba: we didn't handle the empty interface name on TryGetInterfaceName so I was thinking it could be the same but I don't really care :)

pkg/host/internal/network/network.go
```diff
@@ -95,10 +96,39 @@ func (n *network) TryGetInterfaceName(pciAddr string) string {
 		return name
 	}
 
-	log.Log.V(2).Info("tryGetInterfaceName()", "name", netDevName)
+	log.Log.V(2).Info("TryGetInterfaceName()", "name", netDevName)
 	return netDevName
 }
 
+// TryGetInterfaceName tries to find the SR-IOV virtual interface index base on pci address
```
adrianchiris, line 103: nit: TryGetInterfaceIndex

pkg/host/internal/network/network.go
```diff
@@ -95,10 +96,39 @@ func (n *network) TryGetInterfaceName(pciAddr string) string {
...
+func (n *network) TryGetInterfaceIndex(pciAddr string) int {
+	ifName := n.TryToGetVirtualInterfaceName(pciAddr)
```
adrianchiris, line 105: not related to this PR but shouldnt we merge the added logic in TryToGetVirtualInterfaceName into TryGetInterfaceName ?

pkg/host/internal/network/network.go
```diff
@@ -95,10 +96,39 @@ func (n *network) TryGetInterfaceName(pciAddr string) string {
...
+	if ifName == "" {
+		return -1
+	}
+
+	netDir := filepath.Join(consts.SysBusPciDevices, pciAddr, "net", ifName)
+	if _, err := os.Lstat(netDir); err != nil {
```
adrianchiris, line 111: do we really need this check ?

between here and L118 the path may no longer be valid when netdev name changes via udev right after this spot.

pkg/host/internal/network/network.go
```diff
@@ -95,10 +96,39 @@ func (n *network) TryGetInterfaceName(pciAddr string) string {
 		return name
 	}
 
-	log.Log.V(2).Info("tryGetInterfaceName()", "name", netDevName)
+	log.Log.V(2).Info("TryGetInterfaceName()", "name", netDevName)
 	return netDevName
 }
 
+// TryGetInterfaceName tries to find the SR-IOV virtual interface index base on pci address
+func (n *network) TryGetInterfaceIndex(pciAddr string) int {
```
adrianchiris, line 104: any chance to add UT for this one ?

pkg/host/internal/network/network.go
SchSeba: sorry I used the wrong function here

pkg/host/internal/network/network.go
SchSeba: done

pkg/host/internal/sriov/sriov.go
SchSeba: done

pkg/host/internal/network/network.go
SchSeba: done both for this function and the vfIsReady function

adrianchiris: ATM we dont use this method in a "Try" manner but rather fail on error (-1 returned)

WDYT about changing this one to: `GetInterfaceIndex(pciAddr string) (int, error)` ?
i think its better checking error obj than checking for -1.

adrianchiris (commented):
final comment from my side otherwise LGTM

pkg/host/internal/network/network.go
SchSeba: sure done :)

zeeke (approved):
LGTM

pkg/host/internal/network/network.go
```diff
@@ -95,10 +96,33 @@ func (n *network) TryGetInterfaceName(pciAddr string) string {
 		return name
 	}
 
-	log.Log.V(2).Info("tryGetInterfaceName()", "name", netDevName)
+	log.Log.V(2).Info("TryGetInterfaceName()", "name", netDevName)
 	return netDevName
 }
 
+// TryGetInterfaceIndex tries to find the SR-IOV virtual interface index base on pci address
```
adrianchiris, line 103: nit: please update comment :)

pkg/host/internal/network/network.go
```diff
@@ -95,10 +96,33 @@ func (n *network) TryGetInterfaceName(pciAddr string) string {
...
+	ifName := n.TryGetInterfaceName(pciAddr)
+	if ifName == "" {
+		return -1, fmt.Errorf("failed to get interface name")
+	}
+
+	// read the ifindex file from the interface folder
+	indexFile := filepath.Join(vars.FilesystemRoot, consts.SysBusPciDevices, pciAddr, "net", ifName, "ifindex")
+	ifIndex, err := os.ReadFile(indexFile)
+	if err != nil {
+		log.Log.Error(err, "TryGetInterfaceIndex(): failed to read ifindex file", "indexFile", indexFile)
```
adrianchiris, line 114: nit: please update log msg

pkg/host/internal/network/network.go
```diff
@@ -95,10 +96,33 @@ func (n *network) TryGetInterfaceName(pciAddr string) string {
...
+		return -1, err
+	}
+
+	intIfIndex, err := strconv.Atoi(strings.TrimSpace(string(ifIndex)))
+	if err != nil {
+		log.Log.Error(err, "TryGetInterfaceIndex(): failed to parse ifindex file content", "ifIndex", string(ifIndex))
```
adrianchiris, line 120: nit: please update log msg

pkg/host/internal/network/network_test.go
```diff
@@ -238,4 +238,37 @@ var _ = Describe("Network", func() {
 			Expect(n.GetNetDevNodeGUID("0000:4b:00.3")).To(Equal("1122:3344:5566:7788"))
 		})
 	})
+	Context("TryGetInterfaceIndex", func() {
```
adrianchiris, line 241: nit: please update name

pkg/host/internal/network/network.go
SchSeba: done :)

pkg/host/internal/network/network.go
SchSeba: done :)

pkg/host/internal/network/network.go
SchSeba: done :)

pkg/host/internal/network/network_test.go
SchSeba: done :)

pkg/host/internal/lib/netlink/netlink.go
```diff
@@ -24,6 +24,8 @@ type NetlinkLib interface {
 	LinkSetVfPortGUID(link Link, vf int, portguid net.HardwareAddr) error
 	// LinkByName finds a link by name and returns a pointer to the object.
 	LinkByName(name string) (Link, error)
+	// LinkByName finds a link by index and returns a pointer to the object.
```
adrianchiris, line 27: just noticed this one :(. can you fix the typo

pkg/host/internal/network/network.go
```diff
@@ -95,10 +96,33 @@ func (n *network) TryGetInterfaceName(pciAddr string) string {
 		return name
 	}
 
-	log.Log.V(2).Info("tryGetInterfaceName()", "name", netDevName)
+	log.Log.V(2).Info("TryGetInterfaceName()", "name", netDevName)
 	return netDevName
 }
 
+// GetInterfaceIndex tries to find the SR-IOV virtual interface index base on pci address
```
adrianchiris, line 103: this function is not limited to SRIOV virtual interfaces as far as i can tell. 

how about:
`// GetInterfaceIndex returns network interface index base on pci address or error if occurred`

SchSeba:
done with all the function names :P 

can you please take another look?

adrianchiris (approved):

PR #725, opened 2024-06-26
fix: regenerate API pkg with controller-gen 0.14.0 and go 1.22

ykulazhenkov:
when `make generate` is called with controller-gen 0.14.0 (specified in the Makefile) and go 1.22 the `zz_generated.deepcopy.go` will have no `// +build !ignore_autogenerated` line.


zeeke (approved):
LGTM

adrianchiris (approved):

PR #728, opened 2024-06-27
feat: support extraDeploy field in helm chart

fengxsong:
This PR made the following changes:

1. Remove duplicate environment variable `METRICS_EXPORTER_SECRET_NAME` ~~and correct the `METRICS_EXPORTER_PORT` environment variable~~.

2. Support adding additional deployment objects in the Helm chart, such as `SriovNetworkNodePolicy` and `SriovNetwork`, along with the release.

### DIFF
--- a/deployment/sriov-network-operator-chart/README.md
+++ b/deployment/sriov-network-operator-chart/README.md
@@ -144,3 +144,13 @@ This section contains general parameters that apply to both the operator and dae
 | `images.webhook` | Operator Webhook image |
 | `images.metricsExporter` | Network Metrics Exporter image |
 | `images.metricsExporterKubeRbacProxy` | Kube RBAC Proxy image used for metrics exporter |
+
+### Extra objects parameters
+
+**Disclaimer**:
+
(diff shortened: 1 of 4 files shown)


adrianchiris:
im not sure we should support deploying arbitrary "things" as part of the chart.

my take is that the chart should deploy the minimal set of components that are required for sriov network operator to work.
other CRs should be deployed as a post install/update step.

fengxsong:
@zeeke I have reverted this part of the modification back.

> my take is that the chart should deploy the minimal set of components that are required for sriov network operator to work.

@adrianchiris yes, I'm agree with that, keeping deployments as simple as possible is a good point, but sometimes it might be better to provide simple functionality extensions.

In the scenario where we manage cluster deployments using GitOps, I prefer a single deployment that includes everything needed rather than applying additional configurations after deployment.

adrianchiris:
Hi @fengxsong we discussed this in today's community meeting. we are OK with this change. please add a disclaimer in helm chart README that whatever is deployed via this value is the user's responsibility.

adrianchiris (approved):

SchSeba (approved):
LGTM

PR #729, opened 2024-07-01
Bump to 1.22.4

bn222:


Eoghan1232 (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/k8snetworkplumbingwg/sriov-network-operator
 
-go 1.22
+go 1.22.4
```
adrianchiris, line 3: @bn222 why do we need to require .4 explicitly ?

bn222: gopls doesn't work with non-explicit z-stream. Also, our images downstream are 1.22.4. Making it explicit brings our env and upstream in alignment.

Side note: A difference between 1.22.1 and 1.22.4 causes code gen to behave differently. Making it explicit makes sure everyone is generating the same code.

e0ne: I don't like this approach. It will force us to. make more updates (1.22.5 is already releases). Could you please provide any real example how this will help?

e0ne (dismissed):

go.mod
bn222: * gopls lsp servers doesn't work when you don't specify 1.22.4. There is an issue between 1.22.0 and 122.4. 
* I'm not recommending always bumping versions to z-stream. If 1.22.5, we don't necessarily want it.

zeeke: Also, As far as I know. version specified in the go.mod file is the _minum required version_ for this project to build.

https://go.dev/doc/modules/gomod-ref

So, there will be no need to bump this in the future in my opinion.

@bn222 do you have any reference (github issue, stackoverflow) to the problem with gopls?

bn222: https://github.com/NixOS/nixpkgs/issues/293180

bn222: That's the closest I can find.

SchSeba:
@bn222 please correct me if I am wrong here but I think this only points to the min go version if I have an higher version I should be able and use it right?

bn222:
> @bn222 please correct me if I am wrong here but I think this only points to the min go version if I have an higher version I should be able and use it right?

Correct. The generated code will be compatible with 1.22.4 though.

SchSeba:
I will take this PR to the sriov-community meeting today.

adrianchiris (approved):

zeeke (approved):

adrianchiris:
@e0ne this change just specified the minimal version of 1.22.x in go.mod so we should be ok.
if you are fine with this. please approve (i cant merge as you requested changes )

SchSeba (approved):

bn222:
@e0ne ptal

PR #730, opened 2024-07-05
Round up MaxUnavailable to 1 node instead of 0

e0ne:
When MaxUnavailable is configured as percent and cluster is too small we need to roundup MaxUnavailable count to 1 instead of 0 to get draining working.


ykulazhenkov (dismissed):

adrianchiris (approved):

e0ne:
After discussion in the meeting we decided that it works as designed. We can't set `maxUnavailable: 10%` in a cluster of 2 nodes because 1 unavailable node is 50%

PR #733, opened 2024-07-11
feat: implement MlxResetFW to reset the FW on VF changes

tobiasgiese:
Todos:
- [x] implement new method to `mstfwreset`
- [x] add feature flag to opt-in/opt-out
- [x] add unit test
- [x] test it manually or via e2e test job (preferred)
- [ ] add docs (will be added with https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/756)

Manual test shows that the implementation is working:

```
sriov-network-config-daemon 2024-07-11T14:39:36.793959572Z    INFO    mellanox/mellanox_plugin.go:190 mellanox-plugin: resetFW()      {"cmd-args": ["-d", "0000:03:00.0", "-l", "3", "-y", "reset"]}
```
```
foo@bar:~$ sudo mlxconfig -d /dev/mst/mt4119_pciconf0 q | grep -e SRIOV_EN -e NUM_OF_VFS
        NUM_OF_VFS                                  4
        SRIOV_EN                                    True(1)
```

### DIFF
--- a/Dockerfile.sriov-network-config-daemon
+++ b/Dockerfile.sriov-network-config-daemon
@@ -5,7 +5,9 @@ RUN make _build-sriov-network-config-daemon BIN_PATH=build/_output/cmd
 
 FROM quay.io/centos/centos:stream9
 ARG MSTFLINT=mstflint
-RUN ARCH_DEP_PKGS=$(if [ "$(uname -m)" != "s390x" ]; then echo -n ${MSTFLINT} ; fi) && yum -y install hwdata $ARCH_DEP_PKGS && yum clean all
+# We have to ensure that pciutils is installed. This package is needed for mstfwreset to succeed.
+# xref pkg/vendors/mellanox/mellanox.go#L150
+RUN ARCH_DEP_PKGS=$(if [ "$(uname -m)" != "s390x" ]; then echo -n ${MSTFLINT} ; fi) && yum -y install hwdata pciutils $ARCH_DEP_PKGS && yum clean all
 LABEL io.k8s.display-name="sriov-network-config-daemon" \
       io.k8s.description="This is a daemon that manage and config sriov network devices in Kubernetes cluster"
 COPY --from=builder /go/src/github.com/k8snetworkplumbingwg/sriov-network-operator/build/_output/cmd/sriov-network-config-daemon /usr/bin/
--- a/cmd/sriov-network-config-daemon/start.go
+++ b/cmd/sriov-network-config-daemon/start.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/spf13/cobra"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
@@ -41,6 +42,7 @@ import (
 	snclientset "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/clientset/versioned"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/daemon"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/featuregate"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/helper"
 	snolog "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/log"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms"
@@ -276,6 +278,18 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
 	}
 	go nodeWriter.Run(stopCh, refreshCh, syncCh)
 
+	// Init feature gates once to prevent race conditions.
+	defaultConfig := &sriovnetworkv1.SriovOperatorConfig{}
+	err = kClient.Get(context.Background(), types.NamespacedName{Namespace: vars.Namespace, Name: consts.DefaultConfigName}, defaultConfig)
+	if err != nil {
+		log.Log.Error(err, "Failed to get default SriovOperatorConfig object")
+		return err
+	}
+	featureGates := featuregate.New()
+	featureGates.Init(defaultConfig.Spec.FeatureGates)
+	vars.MlxPluginFwReset = featureGates.IsEnabled(consts.MellanoxFirmwareResetFeatureGate)
+	log.Log.Info("Enabled featureGates", "featureGates", featureGates.String())
+
 	setupLog.V(0).Info("Starting SriovNetworkConfigDaemon")
 	err = daemon.New(
 		kClient,
@@ -288,6 +302,7 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
 		syncCh,
 		refreshCh,
 		eventRecorder,
+		featureGates,
 		startOpts.disabledPlugins,
 	).Run(stopCh, exitCh)
 	if err != nil {
--- a/pkg/consts/constants.go
+++ b/pkg/consts/constants.go
@@ -139,6 +139,9 @@ const (
 	// ManageSoftwareBridgesFeatureGate: enables management of software bridges by the operator
 	ManageSoftwareBridgesFeatureGate = "manageSoftwareBridges"
 
+	// MellanoxFirmwareResetFeatureGate: enables the firmware reset via mstfwreset before a reboot
+	MellanoxFirmwareResetFeatureGate = "mellanoxFirmwareReset"
+
 	// The path to the file on the host filesystem that contains the IB GUID distribution for IB VFs
 	InfinibandGUIDConfigFilePath = SriovConfBasePath + "/infiniband/guids"
 )
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os/exec"
+	"reflect"
 	"sync"
 	"time"
 
@@ -23,6 +24,7 @@ import (
 	snclientset "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/clientset/versioned"
 	sninformer "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/informers/externalversions"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/featuregate"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/helper"
 	snolog "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/log"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms"
@@ -82,6 +84,8 @@ type Daemon struct {
 	workqueue workqueue.RateLimitingInterface
 
 	eventRecorder *EventRecorder
+
+	featureGate featuregate.FeatureGate
 }
 
 func New(
@@ -95,6 +99,7 @@ func New(
 	syncCh <-chan struct{},
 	refreshCh chan<- Message,
 	er *EventRecorder,
+	featureGates featuregate.FeatureGate,
 	disabledPlugins []string,
 ) *Daemon {
 	return &Daemon{
@@ -113,6 +118,7 @@ func New(
 			&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(updateDelay), 1)},
 			workqueue.NewItemExponentialFailureRateLimiter(1*time.Second, maxUpdateBackoff)), "SriovNetworkNodeState"),
 		eventRecorder:   er,
+		featureGate:     featureGates,
 		disabledPlugins: disabledPlugins,
 	}
 }
@@ -286,6 +292,7 @@ func (dn *Daemon) operatorConfigAddHandler(obj interface{}) {
 }
 
 func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
+	oldCfg := old.(*sriovnetworkv1.SriovOperatorConfig)
 	newCfg := new.(*sriovnetworkv1.SriovOperatorConfig)
 	if newCfg.Namespace != vars.Namespace || newCfg.Name != consts.DefaultConfigName {
 		log.Log.V(2).Info("unsupported SriovOperatorConfig", "namespace", newCfg.Namespace, "name", newCfg.Name)
@@ -299,6 +306,13 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
 		dn.disableDrain = newDisableDrain
 		log.Log.Info("Set Disable Drain", "value", dn.disableDrain)
 	}
+
+	if !reflect.DeepEqual(oldCfg.Spec.FeatureGates, newCfg.Spec.FeatureGates) {
+		dn.featureGate.Init(newCfg.Spec.FeatureGates)
+		log.Log.Info("Updated featureGates", "featureGates", dn.featureGate.String())
+	}
+
+	vars.MlxPluginFwReset = dn.featureGate.IsEnabled(consts.MellanoxFirmwareResetFeatureGate)
 }
 
 func (dn *Daemon) nodeStateSyncHandler() error {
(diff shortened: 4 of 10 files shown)


pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -19,6 +19,7 @@ type MellanoxPlugin struct {
 	helpers     helper.HostHelpersInterface
 }
 
+var pciAddressesToChange []string
```
tobiasgiese, line 22: I'd prefer to have this in the `MellanoxPlugin` struct, but I decided to create a global var to keep the style.
I will think about refactoring the mellanox plugin code

controllers/sriovoperatorconfig_controller.go
```diff
@@ -188,6 +188,7 @@ func (r *SriovOperatorConfigReconciler) syncConfigDaemonSet(ctx context.Context,
 		data.Data["UsedSystemdMode"] = false
 	}
 	data.Data["ParallelNicConfig"] = r.FeatureGate.IsEnabled(consts.ParallelNicConfigFeatureGate)
+	data.Data["ConfigDaemonEnvVars"] = dc.Spec.ConfigDaemonEnvVars
```
adrianchiris, line 191: can you add test for this one ?

api/v1/sriovoperatorconfig_types.go
```diff
@@ -65,6 +65,8 @@ type SriovOperatorConfigSpec struct {
 	DisablePlugins PluginNameSlice `json:"disablePlugins,omitempty"`
 	// FeatureGates to enable experimental features
 	FeatureGates map[string]bool `json:"featureGates,omitempty"`
+	// ConfigDaemonEnvVars specifies a set of environment variables to be passed to the sriov-network-config-daemon.
+	ConfigDaemonEnvVars map[string]string `json:"configDaemonEnvVars,omitempty"`
```
adrianchiris, line 69: maybe make this one [[]v1.EnvVar](https://github.com/kubernetes/api/blob/fc8a03c10db36574cd45fefb4bb6832a5db0236a/core/v1/types.go#L2267) ? WDYT ?

pkg/vars/vars.go
```diff
@@ -51,6 +51,9 @@ var (
 	// ParallelNicConfig global variable to perform NIC configuration in parallel
 	ParallelNicConfig = false
 
+	// MlxFWReset global variable enables mstfwreset before rebooting a node on VF changes
+	MlxFWReset = false
```
adrianchiris, line 55: nit: MlxPluginFwReset ?

Dockerfile.sriov-network-config-daemon
```diff
@@ -5,7 +5,7 @@ RUN make _build-sriov-network-config-daemon BIN_PATH=build/_output/cmd
 
 FROM quay.io/centos/centos:stream9
 ARG MSTFLINT=mstflint
-RUN ARCH_DEP_PKGS=$(if [ "$(uname -m)" != "s390x" ]; then echo -n ${MSTFLINT} ; fi) && yum -y install hwdata $ARCH_DEP_PKGS && yum clean all
+RUN ARCH_DEP_PKGS=$(if [ "$(uname -m)" != "s390x" ]; then echo -n ${MSTFLINT} ; fi) && yum -y install hwdata pciutils $ARCH_DEP_PKGS && yum clean all
```
wizhaoredhat, line 8: I know that pciutils is needed because of mstflint. If it is not installed mstflint fails. Would there be a better way to handle this dependency. Otherwise looking at this you will never know there is such a dependency.

pkg/plugins/mellanox/mellanox_plugin.go
wizhaoredhat: Shouldn't this be closer to pciAddressesToReset?

tobiasgiese: You are right, will name it like this

cmd/sriov-network-config-daemon/start.go
```diff
@@ -109,6 +109,10 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
 
 	vars.ParallelNicConfig = startOpts.parallelNicConfig
 
+	if strings.ToLower(os.Getenv("MLX_PLUGIN_FW_RESET")) == "true" {
```
wizhaoredhat, line 112: Some questions:
1) Why are we using ENV variables? When there are other means of getting configuration data e.g. config map
2) Why is this even configurable? Shouldn't NVIDIA know how to detect when their hardware needs to be reset a certain way? It seems like having the user defined this ENV is not a good UX.

api/v1/sriovoperatorconfig_types.go
tobiasgiese: I like the idea. The downside of using `[]corev1.EnvVar` would be that the go templating mechanism would be much more complex. We can easily range over the array, but indenting the maps isn't easy IMO. I can give it a try for sure

tobiasgiese: Implemented, see https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/733/commits/76a8f95581b0449074941997f371ac0903139837

controllers/sriovoperatorconfig_controller.go
tobiasgiese: added, see https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/733/commits/76a8f95581b0449074941997f371ac0903139837

Dockerfile.sriov-network-config-daemon
tobiasgiese: We tried several things such as `chroot` into the `/root` from the host, but without any luck.
I can add a comment to the Dockerfile and the usage in `mellanox.go` that we need this package.

cmd/sriov-network-config-daemon/start.go
tobiasgiese: Regarding your questions:

1. we discussed this in the office hours today and we have come to the conclusion that adding env vars is a very generic and non-disruptive way. We could also use a ConfigMap for this, but I'd suggest to implement this change in a follow-up PR and move all configuration to the ConfigMap.
2. This change is only necessary for certain cards, not for all cards (AFAIK). If we are rebooting the host anyway, isn't it okay to reset the FW as well? WDYT @adrianchiris @ykulazhenkov

adrianchiris: 1. @wizhaoredhat  you attended the community meeting IIRC. This was the chosen approach.

2. its not related to Nvidia Hardware its related to the server (Intel's 4th generation Xeon "Eagle Stream") and how it performs PCIe reset on boot. im not aware of a way to detect it. @SchSeba has more details.

tobiasgiese:
@zeeke @adrianchiris the GH Action ci-triggers change seems to work.
After a rebase and push no new comment has been added.
But after closing and reopening there is one. 👍🏻

cmd/sriov-network-config-daemon/start.go
adrianchiris: > Yes I did attend the meeting. But this comment was made taking some time and thinking about it more.

OK, my take,  lets not add another object (CM) into the mix which needs to be created/cleaned-up, at this point its not worth it.

> Can't we run lspci and grep for the CPU? And make decisions based on the generation of Intel processors? Maybe there is a instruction that is no longer exposed for their processor and there is a sysfs file describing that. Just thinking out loud.

Im open to concrete suggestions. im not aware of such way. i think eventually we can enable it by default. but definitely not for the first iteration.

please touch base with Sebastian he can provide you with the relevant BZ if you need more context.

tobiasgiese:
Nvidia e2e test depends on https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/742

cmd/sriov-network-config-daemon/start.go
wizhaoredhat: Sorry I apologize that I don't know all the details of the issue. I will reach out to Sebastian.

Why can't we enable it by default now? Then we can skip the feature flag entirely. Is there a side effect of the reset that the behaviour is not well defined?

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -189,6 +196,13 @@ func (p *MellanoxPlugin) Apply() error {
 		return nil
 	}
 	log.Log.Info("mellanox plugin Apply()")
+	// Only mstfwreset if the featuregate is enabled.
+	if vars.MlxPluginFwReset {
```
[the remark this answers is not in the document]
tobiasgiese: You are totally right, good catch :)

cmd/sriov-network-config-daemon/start.go
tobiasgiese: Implemented in https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/733/commits/2c4dfe01ff0ae55a2f9e0edcba3699b69d08dfd0. Thanks for the hint @SchSeba

pkg/vendors/mellanox/mellanox.go
```diff
@@ -141,6 +142,20 @@ func (m *mellanoxHelper) GetMellanoxBlueFieldMode(PciAddress string) (BlueFieldM
 	return -1, fmt.Errorf("MellanoxBlueFieldMode(): unknown device status for %s", PciAddress)
 }
 
+func (m *mellanoxHelper) MlxResetFW(pciAddresses []string) error {
+	log.Log.Info("mellanox-plugin resetFW()")
+	for _, pciAddress := range pciAddresses {
+		cmdArgs := []string{"-d", pciAddress, "-l", "3", "-y", "reset"}
```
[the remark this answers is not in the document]
adrianchiris: @noama-nv at this point the node is drained from workloads. id prefer to not send reset signal while driver is running.
if we find its problematic we can switch. either way node will reboot after.

during @tobiasgiese testing we didnt encounter issues with the current way of doing reset.

pkg/consts/constants.go
```diff
@@ -141,6 +141,9 @@ const (
 
 	// The path to the file on the host filesystem that contains the IB GUID distribution for IB VFs
 	InfinibandGUIDConfigFilePath = SriovConfBasePath + "/infiniband/guids"
+
+	// MellanoxFirmwareResetFeatureGate: enables the firmware reset via mstfwreset before a reboot
+	MellanoxFirmwareResetFeatureGate = "mellanoxFirmwareReset"
```
SchSeba, line 146: nit: please put this above wth all the other feature gates

pkg/daemon/daemon.go
```diff
@@ -299,6 +306,15 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
 		dn.disableDrain = newDisableDrain
 		log.Log.Info("Set Disable Drain", "value", dn.disableDrain)
 	}
+
+	if !reflect.DeepEqual(oldCfg.Spec.FeatureGates, newCfg.Spec.FeatureGates) {
+		dn.featureGate.Init(newCfg.Spec.FeatureGates)
+		log.Log.Info("Enabled featureGates", "featureGates", dn.featureGate.String())
```
SchSeba, line 312: nit: as we initialize this one on the daemon creation maybe here we can say something like updated featureGates?

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -189,7 +196,13 @@ func (p *MellanoxPlugin) Apply() error {
 		return nil
 	}
 	log.Log.Info("mellanox plugin Apply()")
-	return p.helpers.MlxConfigFW(attributesToChange)
+	if err := p.helpers.MlxConfigFW(attributesToChange); err != nil {
+		return err
+	}
+	if vars.MlxPluginFwReset {
+		return p.helpers.MlxResetFW(pciAddressesToReset)
```
SchSeba, line 203: return an error here can be problematic if we have multiple PCIAddress and one them them failed (for example the primary one)

SchSeba (commented):
in general looks good.

I only have an issue on what can be the case if one of the pciAddress for reset doesn't work and we return an error.

and if we reset the primary nic of the system we will not have an API connection to reconcile and we can leave the system in a bad state...

pkg/plugins/mellanox/mellanox_plugin.go
tobiasgiese: If I understand you correctly your idea would be to introduce a `errs` variable to append the errors and return it at the end of the function, right?

tobiasgiese:
> and if we reset the primary nic of the system we will not have an API connection to reconcile and we can leave the system in a bad state...

Usually the reset of the fw should work. Also the code of the `mstfwrest` binary is tested. Additionally, we have a feature gate top opt-in, which means this is not enabled for the default usage. IMHO we should rely on the robustness of the reset.
Or have you already had experience with the reset and that's why concerns?
WDYT @adrianchiris

Dockerfile.sriov-network-config-daemon
adrianchiris: +1 to add a comment

cmd/sriov-network-config-daemon/start.go
```diff
@@ -276,6 +279,22 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
 	}
 	go nodeWriter.Run(stopCh, refreshCh, syncCh)
 
+	// Init feature gates once to prevent race conditions.
+	defaultConfig := &sriovnetworkv1.SriovOperatorConfig{}
+	err = kClient.Get(context.Background(), types.NamespacedName{Namespace: vars.Namespace, Name: consts.DefaultConfigName}, defaultConfig)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Log.Info("default SriovOperatorConfig object not found. waiting for creation.")
```
adrianchiris, line 287: i think nothing is waiting for its creation the function will return and program will exit. is it not the case ?

pkg/daemon/daemon.go
```diff
@@ -299,6 +306,15 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
 		dn.disableDrain = newDisableDrain
 		log.Log.Info("Set Disable Drain", "value", dn.disableDrain)
 	}
+
+	if !reflect.DeepEqual(oldCfg.Spec.FeatureGates, newCfg.Spec.FeatureGates) {
+		dn.featureGate.Init(newCfg.Spec.FeatureGates)
+		log.Log.Info("Updated featureGates", "featureGates", dn.featureGate.String())
+	}
+
+	if dn.featureGate.IsEnabled(consts.MellanoxFirmwareResetFeatureGate) {
```
adrianchiris, line 315: you need this in start.go as well so vars.MlxPluginFwReset is initialized on start.

pkg/vendors/mellanox/mellanox.go
```diff
@@ -141,6 +142,20 @@ func (m *mellanoxHelper) GetMellanoxBlueFieldMode(PciAddress string) (BlueFieldM
 	return -1, fmt.Errorf("MellanoxBlueFieldMode(): unknown device status for %s", PciAddress)
 }
 
+func (m *mellanoxHelper) MlxResetFW(pciAddresses []string) error {
+	log.Log.Info("mellanox-plugin resetFW()")
+	for _, pciAddress := range pciAddresses {
```
adrianchiris, line 147: @tobiasgiese are you planning to try and reset all nics provided ? then return errors if they happened ?

pkg/vendors/mellanox/mellanox.go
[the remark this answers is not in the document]
adrianchiris: @tobiasgiese i hade another thought about this. since we are going to reboot right after, lets skip the driver restart (adding --skip_driver) as it will save us much time (especially on systems with multiple NICs).

cmd/sriov-network-config-daemon/start.go
tobiasgiese: Yeah, you are right. Was a copy&paste fail because the controllers are returning `reconcile.Result{}, nil`, but we have to return an error here. Good catch

pkg/daemon/daemon.go
tobiasgiese: And we also need an else clause to disable the feature again

pkg/vendors/mellanox/mellanox.go
tobiasgiese: That was the question here https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/733#discussion_r1694769683
Can implement it

SchSeba:
Hi can you please rebase this one

adrianchiris:
LGTM from my sided. just needs rebase

adrianchiris (approved):

SchSeba (approved):
LGTM

nice work!

SalDaniele:
@tobiasgiese @adrianchiris do you know which models fail without this featureGate? We expect the Nvidia DGXH100 will fail to apply mstconfig settings without this change as well.

adrianchiris:
@SalDaniele this feature originated from specific servers utilizing intel 4th generation Xeon "Eagle Stream" platform. 
where upon reboot it does not send PCIe PERST# signal to the NIC.

why do you think DGX H100 will also fail to apply mstconfig settings without this change ?

SalDaniele:
We received a customer case (OCPBUGS-42246) involving the DGX H100 platform with a cx-7 nic. Inspecting the logs, the card fails to update num vfs from the sriov config daemon on reboot, causing a boot loop. Cold boot of the system allows the vfs to be configured as expected.

SalDaniele:
After further inspection, it is possible it is due to a conflict between the SNO and Nvidia Network Operator

PR #734, opened 2024-07-11
Skip firmware reset on devices the operator doesn't control

SchSeba:
In case the user allocate vf or configure the mellanox card firmware and we don't have any policy in the operator we should not reset that configuration

### DIFF
--- a/pkg/plugins/mellanox/mellanox_plugin.go
+++ b/pkg/plugins/mellanox/mellanox_plugin.go
@@ -143,6 +143,11 @@ func (p *MellanoxPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeS
 		processedNics[pciPrefix] = true
 		pciAddress := pciPrefix + "0"
 
+		// Skip devices not configured by the operator
+		if p.nicNotConfiguredByOperator(portsMap) {
+			continue
+		}
+
 		// Skip externally managed NICs
 		if p.nicHasExternallyManagedPFs(portsMap) {
 			continue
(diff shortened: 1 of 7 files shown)


pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -206,3 +211,21 @@ func (p *MellanoxPlugin) nicHasExternallyManagedPFs(nicPortsMap map[string]sriov
...
 }
+
+// nicNotConfiguredByOperator returns true if one of the ports(interface) of the NIC is not configured by operator
+func (p *MellanoxPlugin) nicNotConfiguredByOperator(nicPortsMap map[string]sriovnetworkv1.InterfaceExt) bool {
+	existPfStatusOnStore := false
+	for _, iface := range nicPortsMap {
+		_, exist, err := p.helpers.LoadPfsStatus(iface.PciAddress)
+		if err != nil {
+			log.Log.Error(err, "failed to load PF status from disk", "address", iface.PciAddress)
+			continue
```
adrianchiris, line 222: should we fail here ?

SchSeba: I wasn't sure I copy the other function :P 
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/734/files/fa266810fd4c7bc0660cbcc750ac1df761c51038#diff-8f607793a0090ecce8926d771471ecce0f2514b77695e38bbd6bb70f8996f088R200

adrianchiris: fair enough :)

adrianchiris (approved):

pkg/plugins/mellanox/mellanox_plugin.go
adrianchiris: perhaps we shoud consider failing in both places, LMK what you think i can submit a followup PR.

LoadPfStatus will fail if file exists but it failed reading from it or its corrupt and we werent able to unmarshall

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -206,3 +211,21 @@ func (p *MellanoxPlugin) nicHasExternallyManagedPFs(nicPortsMap map[string]sriov
...
+		}
+		if exist {
+			log.Log.V(2).Info("PF configured by the operator", "interface", iface)
+			existPfStatusOnStore = true
+			break
```
zeeke, line 227:
```suggestion
			return false
```

can we get rid of the `existPfStatusOnStore` variable and early exit here?

zeeke (approved):
Only one minor comment from my side. LGTM

pkg/plugins/mellanox/mellanox_plugin.go
SchSeba: right!

pkg/plugins/mellanox/mellanox_plugin.go
SchSeba: sounds good I agree that should failed in both functions :) 

when you have time please open the PR I can review it

zeeke (approved):

PR #737, opened 2024-07-16
ci: sent ci-triggers comment only once on PR open or reopen

tobiasgiese:
If we specify a type we can stop the Github Actions bot from resending the
ci-triggers list to open PRs on every HEAD change.

> [...] if no activity types are specified, the workflow runs when a pull request
is opened or reopened or when the head branch of the pull request is updated.

source: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

### DIFF
--- a/.github/workflows/pr-ci-triggers.yml
+++ b/.github/workflows/pr-ci-triggers.yml
@@ -1,5 +1,6 @@
 on:
   pull_request_target:
+    types: [opened, reopened]
 
 jobs:
   vendors-ci-triggers-list:


zeeke (approved):
LGTM

tobiasgiese:
Pushed a new HEAD to the branch but the bot is still commenting the ci-triggers list.
Or do we need to merge this to master to take affect? AFAIK the workflows are per branch, right?

zeeke:
This might explain why it is not working:
> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

tobiasgiese:
> This might explain why it is not working:

Ah, that explains it. So we have 2 options:

- set it to `pull_request` and keep it
- test it with `pull_request` and if it works we can reset it to `pull_request_target`

WDYT?

tobiasgiese:
... but sure, simply changing it to `pull_request` does not work, that would completely undermine the security mechanism 🤦🏻

adrianchiris (approved):

adrianchiris:
merging :)

PR #721, opened 2024-06-23
add sort for the policy on the sriovOperatorConfig controller

SchSeba:
we need to be consistent with the policy order

PR #740, opened 2024-07-17
Change logic to prevent device-plugin daemonset spec change

AMacedoP:
Instead of adding a NodeSelectorRequirement per NodePolicy we add all the NodeSelectors to a single NodeSelectorRequirent so they can all be sorted. This will prevent that the daemonset's node selector changes when there are multiple NodePolicies applied.

### DIFF
--- a/controllers/helper.go
+++ b/controllers/helper.go
@@ -281,7 +281,7 @@ func setDsNodeAffinity(pl *sriovnetworkv1.SriovNetworkNodePolicyList, ds *appsv1
 }
 
 func nodeSelectorTermsForPolicyList(policies []sriovnetworkv1.SriovNetworkNodePolicy) []corev1.NodeSelectorTerm {
-	terms := []corev1.NodeSelectorTerm{}
+	expressions := []corev1.NodeSelectorRequirement{}
 	for _, p := range policies {
 		// Note(adrianc): default policy is deprecated and ignored.
 		if p.Name == constants.DefaultPolicyName {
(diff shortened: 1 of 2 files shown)


zeeke:
hi @AMacedoP , the problem you're describing has probably already been solved by:
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/721
can you check that?

That said, I'm not against the proposed changes, as they make the code clearer and more resilient.

AMacedoP:
@zeeke yes, that PR also solves my problem. How do you want to proceed with this PR? If you want I can resolve the failed test cases and merge it.

zeeke:
Please, fix the unit test and wait for the other community member's feedback.

@adrianchiris , @SchSeba WDYT?

SchSeba:
Hi @AMacedoP,

Thanks for the PR!
I am working on a different solution that I think will be better here.
My idea is to leave the device plugin with one node selector something like sriov-device-plugin-deployed
and the daemon will remove and add the label when it needs to reset the device plugin 

I think that is cleaner let me know what you think

AMacedoP:
@SchSeba that looks like a good solution. In that case what should we do with this PR?

SchSeba:
If possible I would like to close this one in favor of https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/747

@AMacedoP let me know what do you think 

Thanks!

AMacedoP:
@SchSeba will close this PR. Do you have an expected release date for a new version with the changes in #747?

Thanks!

PR #743, opened 2024-07-22
Return error on mlx plugin on firmware reset check

SchSeba:
This commit makes the functions to return error if we are not able to read from the configuration cache on the node.

### DIFF
--- a/pkg/plugins/mellanox/mellanox_plugin.go
+++ b/pkg/plugins/mellanox/mellanox_plugin.go
@@ -6,6 +6,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/helper"
 	plugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins"
 	mlx "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vendors/mellanox"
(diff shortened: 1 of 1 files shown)


SchSeba:
followup on https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/734

@adrianchiris @ykulazhenkov please take a look :)

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -194,36 +206,37 @@ func (p *MellanoxPlugin) Apply() error {
...
 	for _, iface := range nicPortsMap {
 		_, exist, err := p.helpers.LoadPfsStatus(iface.PciAddress)
 		if err != nil {
 			log.Log.Error(err, "failed to load PF status from disk", "address", iface.PciAddress)
-			continue
+			return false, err
 		}
 		if exist {
 			log.Log.V(2).Info("PF configured by the operator", "interface", iface)
-			return true
```
adrianchiris, line 225: so this was the bug ?

i see you reversed the meaning of this function

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -144,12 +144,24 @@ func (p *MellanoxPlugin) OnNodeStateChange(new *sriovnetworkv1.SriovNetworkNodeS
 		pciAddress := pciPrefix + "0"
 
 		// Skip devices not configured by the operator
-		if p.nicNotConfiguredByOperator(portsMap) {
+		isConfigured, err := p.nicConfiguredByOperator(portsMap)
+		if err != nil {
+			return false, false, err
```
adrianchiris, line 149: thats going to be a non-recoverable error right ? (also below at L160)

i guess if we get here then state file got corrupted. which "should not" happen

adrianchiris: what about first time running on the interface ? or after nodeState got re-created (we clearing the state file there) ?

this logic runs before generic plugin which eventually stores the pf status when it configures sriov interface

adrianchiris (approved):
overall LGTM, 

left some small comments

pkg/plugins/mellanox/mellanox_plugin.go
SchSeba: That is right. 
Maybe add an error message and request the user to remove the file?

adrianchiris: +1

pkg/plugins/mellanox/mellanox_plugin.go
SchSeba: done

pkg/plugins/mellanox/mellanox_plugin.go
SchSeba: done I put the info in the function itself (for both of them)

adrianchiris (approved):

SchSeba:
thanks @adrianchiris 

merging

PR #744, opened 2024-07-22
Fix MTU functional test

SchSeba:
we should not lower the MTU on a primary SDN nic.

with this change we select a PF that is not the default route of the node or not the nic connected to the ovs bridge for openshift


test/conformance/tests/test_sriov_operator.go
```diff
@@ -1724,82 +1756,152 @@ var _ = Describe("[sriov] operator", func() {
...
-						firstPod, _ = clients.Pods(namespaces.Test).Get(context.Background(), firstPod.Name, metav1.GetOptions{})
-						return firstPod.Status.Phase
-					}, 3*time.Minute, time.Second).Should(Equal(corev1.PodRunning))
+						Expect(err).ToNot(HaveOccurred())
 
-					By("Checking MTU in pod network status annotation")
-					networkStatusJSON, exist := firstPod.Annotations["k8s.v1.cni.cncf.io/network-status"]
-					Expect(exist).To(BeTrue())
-					Expect(networkStatusJSON).To(ContainSubstring("\"mtu\": 9000"))
+						Eventually(func() error {
```
zeeke, line 1780: can you use `waitForNetAttachDef()` here?

test/conformance/tests/test_sriov_operator.go
```diff
@@ -1724,82 +1756,152 @@ var _ = Describe("[sriov] operator", func() {
...
+							netAttDef := &netattdefv1.NetworkAttachmentDefinition{}
+							return clients.Get(context.Background(), runtimeclient.ObjectKey{Name: "test-mtuvolnetwork", Namespace: namespaces.Test}, netAttDef)
+						}, (10+snoTimeoutMultiplier*110)*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
 
-					var stdout, stderr string
-					Eventually(func() error {
-						stdout, stderr, err = pod.ExecCommand(clients, firstPod, "ip", "link", "show", "net1")
-						if stdout == "" {
-							return fmt.Errorf("empty response from pod exec")
+						// update the interface
```
zeeke, line 1785: please, extract a function like 



```suggestion
						intf = getInterfaceFromNodeStateByPciAddress(node, intf.PciAddr)
						
						func getInterfaceFromNodeStateByPciAddress(node, pciAddress string) {
						    nodeState := &sriovv1.SriovNetworkNodeState{}
						    err = clients.Get(context.Background(), runtimeclient.ObjectKey{Name: node, Namespace: operatorNamespace}, nodeState)
						    Expect(err).ToNot(HaveOccurred())
						    found := false
						    for _, state := range nodeState.Status.Interfaces {
							    if state.PciAddress == intf.PciAddress {
								    intf = state.DeepCopy()
								    found = true
								    break
							    }
						    }
						    Expect(found).To(BeTrue())
						}
```

WDYT?

zeeke (commented):
Left few minor comments. As far as I understood, the main change in this PR is 
```
unusedSriovDevices, err := findUnusedSriovDevices(node, sriovDeviceList)
							if err != nil {
								Skip(err.Error())
							}
							intf = unusedSriovDevices[0]
```

correct?

https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/744/files#diff-6665a7e475c21cad9c99b9463fba12769e300e6d1a6cc1323c9c3ca68e8cc120R1581

SchSeba:
Hi @zeeke when you have time please take another look :)

adrianchiris (approved):

zeeke (approved):
LGTM!

PR #746, opened 2024-07-24
Remove logic that installs rdma-core package

ykulazhenkov:
Remove logic that installs `rdma-core` package to align the code for all platforms. 
Before this commit the state is following:
* Ubuntu - the logic is completely broken and newer executed. 
   Reason: buggy [IsUbuntuCheck()](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/ee40683567d8ac5d03365e3216592e74d47538b7/pkg/host/internal/kernel/kernel.go#L448) 
  _We use grep with `--quiet` flag and `IsUbuntuCheck()` returns true only if `stdout` is not empty. Because of the `--quiet` flag, stdout will be always empty. Broken for a long time, nobody reported an issue about this._
* RHCOS - validation only logic from the begining
* RHEL - package installation (partially broken in RHEL 8 and RHEL 9) Reason: [This file](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/ee40683567d8ac5d03365e3216592e74d47538b7/pkg/host/internal/common.go#L10) Is available in RHEL 7, but not in RHEL 8 or 9

The operator currently uses quite dangerous logic to trigger loading of RDMA modules https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/ee40683567d8ac5d03365e3216592e74d47538b7/pkg/host/internal/kernel/kernel.go#L525 

I think we don't want the operator to install the package and completely unload the driver, this can break multiple use-cases (NVIDIA driver in container, Externally managed PF and potentially more)

The commit changes logic for all OS to do only a validation of RDMA modules and print warning if they are not found.

@adrianchiris  @SchSeba


adrianchiris:
we should not unload mlx5_core driver from kernel, thats dangerous as primary NIC may become unavailable.
generally in NVIDIA we instruct customers to pre install rdma-core if inbox drivers are used, hence the install step is not needed from NVIDIA POV. (plus it seems it never worked :) )

+1 on just logging if something is wrong like we do in coreos. perhaps with a meaningfull msg like `"rdma modules are not loaded, please install rdma-core package available from you package manager"`

SchSeba (commented):
I am soooo happy to see this code going away from the operator!

never like it :)

Can you please just add in the read me information that the user must install the rdma package on the system?
also you can add the in RedHat CoreOS that is not needed

adrianchiris (approved):
LGTM

ykulazhenkov:
> I am soooo happy to see this code going away from the operator!
> 
> never like it :)
> 
> Can you please just add in the read me information that the user must install the rdma package on the system? also you can add the in RedHat CoreOS that is not needed

@SchSeba  I updated the `quickstart.md` to mention that `rdma-core` packgage must be installed.

SchSeba (approved):
LGTM

PR #747, opened 2024-07-25
Redesign device plugin reset

SchSeba:
This commit introduces a new redesign on how the operator resets the device plugin

* use a general nodeSelector to avoid updating the daemonset yaml
* reduce clusterRole to role for config-daemon pod delete option ( better security)


controllers/drain_controller.go
```diff
@@ -172,6 +172,29 @@ func (dr *DrainReconcile) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 				return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
```
bn222, line 172: shorter commit name: redesign device plugin reset

controllers/drain_controller.go
```diff
@@ -436,6 +524,45 @@ func (dr *DrainReconcile) findNodePoolConfig(ctx context.Context, node *corev1.N
...
+		logger.Error(err, "failed to list device plugin pods running on node")
+		return false, err
+	}
+
+	if len(pods.Items) != 0 {
+		log.Log.Info("device plugin pod still terminating on node")
+		return false, nil
+	}
+
+	// TODO:(schseba): not sure if we should wait here for the device plugin to be running
```
bn222, line 550: yes we should, because otherwise the remaining code is 1) hard to read since the actual state is unknown, 2) if it doesn't come up, we probably want to error out

bn222: The same holds when we request the dp to restart. You probably don't want to be flipping back and forth and not blocking on state because then there will a difference between what's running and what you think is running.

controllers/drain_controller.go
```diff
@@ -172,6 +172,29 @@ func (dr *DrainReconcile) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 				return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
 			}
 
+			// check the device plugin exited and enable it again
+			// only of we have something in the node state spec
+			if len(nodeNetworkState.Spec.Interfaces) > 0 {
+				completed, err = dr.enableSriovDevicePlugin(ctx, node)
```
bn222, line 178: See my other comment, completed is a confusing name here unless we change the implementation of `enableSriovDevicePlugin`.

controllers/drain_controller.go
SchSeba: done

controllers/drain_controller.go
SchSeba: done

SchSeba:
@zeeke @adrianchiris @ykulazhenkov when you have time please take a look on this :)

controllers/drain_controller.go
```diff
@@ -243,6 +293,60 @@ func (dr *DrainReconcile) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
+			reqLogger.Error(err, "failed to annotate node with annotation", "annotation", constants.DrainComplete)
+			return ctrl.Result{}, err
+		}
+
+		reqLogger.Info("node cordoned successfully and device plugin removed")
+		dr.recorder.Event(nodeNetworkState,
+			corev1.EventTypeWarning,
+			"DrainController",
+			"node cordoned and device plugin removed completed")
+		return ctrl.Result{}, nil
```
zeeke, line 349: Any chance to code a unit test for this case? It took me a while to understand the whole controller - daemon interaction

zeeke (commented):
Left a few comments

controllers/helper.go
```diff
@@ -214,13 +190,15 @@ func syncPluginDaemonObjs(ctx context.Context,
 				return err
 			}
 			ds.Spec.Template.Spec.NodeSelector = dc.Spec.ConfigDaemonNodeSelector
+			// add the special node selector for the device plugin
+			ds.Spec.Template.Spec.NodeSelector[constants.SriovDevicePluginEnabledLabel] = constants.SriovDevicePluginEnabledLabelEnabled
```
zeeke, line 194: can't we put
```yaml
nodeSelector:
    sriovnetwork.openshift.io/device-plugin: Enabled
```
in `bindata/manifests/plugins/sriov-device-plugin.yaml` ?
should we take care of the ConfigDaemonNodeSelector?

controllers/drain_controller_helper.go
```diff
@@ -0,0 +1,468 @@
...
+	nodeStateDrainAnnotationCurrent string) (ctrl.Result, error) {
+	// nothing to do here we need to wait for the node to move back to idle
+	if nodeStateDrainAnnotationCurrent == constants.DrainComplete {
+		reqLogger.Info("node requested a drain and nodeState is on drain completed nothing todo")
+		return ctrl.Result{}, nil
+	}
+
+	// if we are on idle state we move it to drain
+	if nodeStateDrainAnnotationCurrent == constants.DrainIdle {
+		err := utils.AnnotateObject(ctx, nodeNetworkState, constants.NodeStateDrainAnnotationCurrent, constants.Draining, dr.Client)
```
zeeke, line 201: It took me a while to understand this. We're setting `nodeState = Draining` even if we are not draining. We're just cordoning. It is to avoid having too many states?

controllers/drain_controller_test.go
```diff
@@ -316,6 +316,44 @@ var _ = Describe("Drain Controller", Ordered, func() {
...
+			Expect(k8sClient.Update(ctx, nodeState)).ToNot(HaveOccurred())
+
+			simulateDaemonSetAnnotation(node, constants.DevicePluginResetRequired)
+			expectNodeStateAnnotation(nodeState, constants.DrainComplete)
+			expectNodeLabel(node, constants.SriovDevicePluginEnabledLabel, constants.SriovDevicePluginEnabledLabelDisabled)
+			simulateDaemonSetAnnotation(node, constants.DrainIdle)
+			createDevicePluginPodOnNode(ctx, node.Name)
+			expectNodeStateAnnotation(nodeState, constants.DrainIdle)
+			expectNodeLabel(node, constants.SriovDevicePluginEnabledLabel, constants.SriovDevicePluginEnabledLabelEnabled)
+		})
```
zeeke, line 355: While checking if I understood all the step, I coded this unit test:
```
It("should restart the device plugin on multiple nodes if requested", func() {
			simpleNodeStateSpec := sriovnetworkv1.SriovNetworkNodeStateSpec{Interfaces: sriovnetworkv1.Interfaces{
				{
					Name:   "test",
					NumVfs: 10,
					VfGroups: []sriovnetworkv1.VfGroup{
						{
							PolicyName:   "test",
							ResourceName: "test",
							VfRange:      "0-9",
						},
					},
				},
			}}

			node1, nodeState1 := createNode(ctx, "node-a")
			nodeState1.Spec = simpleNodeStateSpec
			node2, nodeState2 := createNode(ctx, "node-b")
			nodeState2.Spec = simpleNodeStateSpec


			simulateDaemonSetAnnotation(node1, constants.DevicePluginResetRequired)
			simulateDaemonSetAnnotation(node2, constants.DevicePluginResetRequired)
			
			expectNodeStateAnnotation(nodeState1, constants.DrainComplete)
			expectNodeStateAnnotation(nodeState2, constants.DrainComplete)
			expectNodeLabel(node1, constants.SriovDevicePluginEnabledLabel, constants.SriovDevicePluginEnabledLabelDisabled)
			expectNodeLabel(node2, constants.SriovDevicePluginEnabledLabel, constants.SriovDevicePluginEnabledLabelDisabled)

			simulateDaemonSetAnnotation(node1, constants.DrainIdle)
			simulateDaemonSetAnnotation(node2, constants.DrainIdle)

			expectNodeStateAnnotation(nodeState1, constants.DrainIdle)
			expectNodeStateAnnotation(nodeState2, constants.DrainIdle)
			expectNodeLabel(node1, constants.SriovDevicePluginEnabledLabel, constants.SriovDevicePluginEnabledLabelEnabled)
			expectNodeLabel(node2, constants.SriovDevicePluginEnabledLabel, constants.SriovDevicePluginEnabledLabelEnabled)
		})
```

But it fails and the device plugin is not enabled on any node. Is this test case wrong, or is there an underlying unmanaged condition?

controllers/drain_controller_helper.go
```diff
@@ -0,0 +1,468 @@
...
+		"DrainController",
+		"node drain completed")
+	return ctrl.Result{}, nil
+}
+
+func (dr *DrainReconcile) handleNodeDPReset(ctx context.Context,
+	reqLogger *logr.Logger,
+	node *corev1.Node,
+	nodeNetworkState *sriovnetworkv1.SriovNetworkNodeState,
+	nodeDrainAnnotation,
```
zeeke, line 191: This variable is not used. please, remove it

zeeke (commented):
Left a few comments. Overall design looks good to me but we must be super sure about all of these annotation states. I'm pretty scared of having a deadlock in some production cluster :)

controllers/helper.go
SchSeba: yep we must continue to run the device plugin only on nodes where we have the config daemon.

if I put it on the bindata it will get override no?

controllers/drain_controller_helper.go
SchSeba: right I didn't want to make this more complex than it's already :P

controllers/drain_controller_helper.go
SchSeba: right thanks!

controllers/drain_controller_test.go
SchSeba: Done please take a look.

in your unit tests you miss to update the nodeState obect with the new spec and to also create a demo device plugin pod. I fixed it and integrate the test

controllers/helper.go
zeeke: I mean we might get rid of the ConfigDaemonNodeSelector info for the device plugin. The DrainController manages the `sriovnetwork.openshift.io/device-plugin: Enabled/Disabled` label only on nodes where the config daemon is running. Having the another selector is redundant, isn't it?

pkg/consts/constants.go
```diff
@@ -67,12 +67,17 @@ const (
 	MachineConfigPoolPausedAnnotationIdle   = "Idle"
 	MachineConfigPoolPausedAnnotationPaused = "Paused"
 
+	SriovDevicePluginEnabledLabel         = "sriovnetwork.openshift.io/device-plugin"
+	SriovDevicePluginEnabledLabelEnabled  = "Enabled"
+	SriovDevicePluginEnabledLabelDisabled = "Disabled"
```
zeeke, line 72: what about ?
```suggestion
	SriovDevicePluginLabel         = "sriovnetwork.openshift.io/device-plugin"
	SriovDevicePluginLabelEnabled  = "Enabled"
	SriovDevicePluginLabelDisabled = "Disabled"
```

controllers/drain_controller_test.go
zeeke: looks good, thanks!

controllers/helper.go
SchSeba: ahhhh now I get your point! yes you are right!

adrianchiris (commented):
partial review.

left with drain controller and its helper.

its hard to review added logic since a bunch of code was moved while new code added in existing functions in the same commit :\

adrianchiris:
General comment:

to me it seems (much ?) simpler that the config daemon will be in charge of updating its own node label to evict DP pod.
in this case it just requests cordon from the drain controller.

the drain controller will then handle only what it was supposed to do originally (handle cordon/drain related operations)
which is easier to comprehend IMO.

controllers/drain_controller_helper.go
```diff
@@ -0,0 +1,467 @@
...
+		reqLogger.Error(err, "failed to label node for device plugin label",
+			"labelKey",
+			constants.SriovDevicePluginLabel,
+			"labelValue",
+			constants.SriovDevicePluginLabelDisabled)
+		return reconcile.Result{}, err
+	}
+
+	// if we manage to cordon we label the node state with drain completed and finish
+	err = utils.AnnotateObject(ctx, nodeNetworkState, constants.NodeStateDrainAnnotationCurrent, constants.DrainComplete, dr.Client)
```
adrianchiris, line 233: a bit confusing for the device plugin reset flow where we did not really drain.

pkg/daemon/daemon.go
```diff
@@ -478,6 +477,27 @@ func (dn *Daemon) nodeStateSyncHandler() error {
 		}
 	}
 
+	// if we don't need to drain, and we are on idle we need to request the device plugin reset
+	if !reqDrain && utils.ObjectHasAnnotation(dn.desiredNodeState,
+		consts.NodeStateDrainAnnotationCurrent,
+		consts.DrainIdle) {
+		log.Log.Info("nodeStateSyncHandler(): apply 'Device_Plugin_Reset_Required' annotation for node")
+		err := utils.AnnotateNode(context.Background(), vars.NodeName, consts.NodeDrainAnnotation, consts.DevicePluginResetRequired, dn.client)
```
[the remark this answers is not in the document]
adrianchiris: also why are we requesting device plugin restart in a different place than the original?

controllers/helper.go
```diff
@@ -185,42 +173,17 @@ func syncPluginDaemonObjs(ctx context.Context,
 	data.Data["ReleaseVersion"] = os.Getenv("RELEASEVERSION")
 	data.Data["ResourcePrefix"] = vars.ResourcePrefix
 	data.Data["ImagePullSecrets"] = GetImagePullSecrets()
-	data.Data["NodeSelectorField"] = GetDefaultNodeSelector()
+	data.Data["NodeSelectorField"] = GetDefaultNodeSelectorForDevicePlugin()
```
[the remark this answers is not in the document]
SchSeba: done

pkg/utils/cluster.go
```diff
@@ -161,3 +162,40 @@ func AnnotateNode(ctx context.Context, nodeName string, key, value string, c cli
...
+
+// LabelObject adds label to a kubernetes object
+func LabelObject(ctx context.Context, obj client.Object, key, value string, c client.Client) error {
+	newObj := obj.DeepCopyObject().(client.Object)
+	if newObj.GetLabels() == nil {
+		newObj.SetLabels(map[string]string{})
+	}
+
+	if newObj.GetLabels()[key] != value {
+		log.Log.V(2).Info("LabelObject(): Annotate object",
```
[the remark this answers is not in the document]
SchSeba: done

pkg/utils/cluster.go
```diff
@@ -161,3 +162,40 @@ func AnnotateNode(ctx context.Context, nodeName string, key, value string, c cli
...
+			"objectName", obj.GetName(),
+			"objectKind", obj.GetObjectKind(),
+			"labelKey", key,
+			"labelValue", value)
+		newObj.GetLabels()[key] = value
+		patch := client.MergeFrom(obj)
+		err := c.Patch(ctx,
+			newObj, patch)
+		if err != nil {
+			log.Log.Error(err, "annotateObject(): Failed to patch object")
```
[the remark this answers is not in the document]
SchSeba: done

pkg/platforms/openshift/openshift.go
```diff
@@ -228,6 +228,18 @@ func (c *openshiftContext) OpenshiftAfterCompleteDrainNode(ctx context.Context,
 		return false, err
 	}
 
+	value, exist := mcp.Annotations[consts.MachineConfigPoolPausedAnnotation]
```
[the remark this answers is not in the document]
SchSeba: I removed it from this PR

ykulazhenkov (approved):
Thanks for working on this. LGTM

Issue #749, opened 2024-07-25
Add documetnation for software-bridge management feature


ykulazhenkov:
The [Software-bridge management feature](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/doc/design/software-bridge-management.md) is currently missing documentation. 
It is required to update README.md with configuration example.

PR #750, opened 2024-07-25
[software-bridges 5/x] Enable software-bridge implementation

ykulazhenkov:
This PR enables all required code for software-bridges features. 
I will submit documentation for the feature in the next PR. I created https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/749 to track documentation update.

How to test the  feature:

* deploy the operator
* enable `manageSoftwareBridges` feature gate
```kubectl patch sriovoperatorconfigs.sriovnetwork.openshift.io -n network-operator default --patch '{ "spec": { "featureGates": { "manageSoftwareBridges": true  } } }' --type='merge'```
* create policy with the OVS config
```
kind: SriovNetworkNodePolicy
metadata:
  name: connectx6dx-switchdev
  namespace: network-operator
spec:
  eSwitchMode: switchdev
  mtu: 1500
  nicSelector:
    deviceID: 101d
    vendor: 15b3
  nodeSelector:
    node-role.kubernetes.io/worker: ""
  numVfs: 4
  resourceName: switchdev
  bridge:
    ovs: {}
```

* create OVSNetwork CR
```
kind: OVSNetwork
metadata:
  name: ovs
  namespace: network-operator
spec:
  networkNamespace: default
  ipam: |
    {
      "type": "nv-ipam",
      "poolName": "pool1"
    }
  resourceName: switchdev
  vlan: 200
```

### DIFF
--- a/pkg/plugins/generic/generic_plugin.go
+++ b/pkg/plugins/generic/generic_plugin.go
@@ -48,13 +48,14 @@ type DriverState struct {
 type DriverStateMapType map[uint]*DriverState
 
 type GenericPlugin struct {
-	PluginName          string
-	SpecVersion         string
-	DesireState         *sriovnetworkv1.SriovNetworkNodeState
-	DriverStateMap      DriverStateMapType
-	DesiredKernelArgs   map[string]bool
-	helpers             helper.HostHelpersInterface
-	skipVFConfiguration bool
+	PluginName              string
+	SpecVersion             string
+	DesireState             *sriovnetworkv1.SriovNetworkNodeState
+	DriverStateMap          DriverStateMapType
+	DesiredKernelArgs       map[string]bool
+	helpers                 helper.HostHelpersInterface
+	skipVFConfiguration     bool
+	skipBridgeConfiguration bool
 }
 
 type Option = func(c *genericPluginOptions)
@@ -68,8 +69,16 @@ func WithSkipVFConfiguration() Option {
 	}
 }
 
+// WithSkipBridgeConfiguration configures generic_plugin to skip configuration of the managed bridges
+func WithSkipBridgeConfiguration() Option {
+	return func(c *genericPluginOptions) {
+		c.skipBridgeConfiguration = true
+	}
+}
+
 type genericPluginOptions struct {
-	skipVFConfiguration bool
+	skipVFConfiguration     bool
+	skipBridgeConfiguration bool
 }
 
 const scriptsPath = "bindata/scripts/enable-kargs.sh"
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Review by adrianchiris — APPROVED** [review]
lgtm

**Comment by adrianchiris** [comment]
@SchSeba PTAL on this one.

File: pkg/plugins/generic/generic_plugin.go [hunk 1]
```diff
@@ -323,13 +346,13 @@ func (p *GenericPlugin) syncDesiredKernelArgs(kargs []string) (bool, error) {
 	return needReboot, nil
 }
 
-func (p *GenericPlugin) needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.InterfaceExts) (needDrain bool) {
+func (p *GenericPlugin) needDrainNode(desired sriovnetworkv1.SriovNetworkNodeStateSpec, current sriovnetworkv1.SriovNetworkNodeStateStatus) (needDrain bool) {
 	log.Log.V(2).Info("generic plugin needDrainNode()", "current", current, "desired", desired)
 
 	needDrain = false
-	for _, ifaceStatus := range current {
+	for _, ifaceStatus := range current.Interfaces {
```
SchSeba, line 353: now that this function check both interfaces and bridges can you move this to another function like `NeedToUpdateVfs`?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -68,8 +69,16 @@ func WithSkipVFConfiguration() Option {
 	}
 }
 
+// WithSkipBridgeConfiguration configures generic_plugin to skip configuration of the managed bridges
+func WithSkipBridgeConfiguration() Option {
```
SchSeba, line 73: I don't see this one used.

 so at the end we only validate the  `vars.ManageSoftwareBridges`. on the if state here `return vars.ManageSoftwareBridges && !p.skipBridgeConfiguration`
 
also (I know it's not part of this PR) can you please explain me when on prePhase we don't configure the vfs?
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/f8cc3364c9a16d42a666968dff90dbfbf7219260/cmd/sriov-network-config-daemon/service.go#L209-L210

ykulazhenkov: > also (I know it's not part of this PR) can you please explain me when on prePhase we don't configure the vfs?

At the `pre` stage we creatr VFs, but unbind them from a driver to let OS Network manager (e.g. NetworkManager, systemd-network) to apply custom configuration for the PF function (e.g. VF-lag). Such configuration can't be when the PF has VFs that are bound to a driver.

pkg/plugins/generic/generic_plugin.go
ykulazhenkov: done

pkg/plugins/generic/generic_plugin.go
ykulazhenkov: The WithSkipBridgeConfiguration function is used in the service command to disable bridge management during the `pre` phase

SchSeba: you are right! I miss that sorry

SchSeba (approved):
LGTM

nice work!

adrianchiris (approved):

adrianchiris:
Thx for the great work @ykulazhenkov

PR #751, opened 2024-07-29
Add a pre-delete helm hook to clean up webhook configs

almaslennikov:
Mutating and Validating webhook configs are created programmatically and are not managed by helm. To clean them up on helm delete, adding a small pre-delete hook

### DIFF
--- a/deployment/sriov-network-operator-chart/templates/predeletehook.yaml
+++ b/deployment/sriov-network-operator-chart/templates/predeletehook.yaml
@@ -1,0 +1,24 @@
+{{ if .Values.operator.admissionControllers.enabled }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ include "sriov-network-operator.fullname" . }}-post-delete-hook"
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
+spec:
+  template:
+    spec:
+      serviceAccountName: {{ include "sriov-network-operator.fullname" . }}
+      containers:
+        - name: cleanup
+          image: bitnami/kubectl:latest
+          command:
+            - /bin/sh
+            - -c
+            - |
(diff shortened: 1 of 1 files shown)


deployment/sriov-network-operator-chart/templates/predeletehook.yaml
```diff
@@ -0,0 +1,24 @@
...
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
+spec:
+  template:
+    spec:
+      serviceAccountName: {{ include "sriov-network-operator.fullname" . }}
+      containers:
+        - name: cleanup
+          image: bitnami/kubectl:latest
```
adrianchiris, line 16: overall LGTM. i wonder if we should go with that or implement small "cleanup" binary which will do these operations using client-go lib.  having this binary part of the sriov-network-operator image.

@ykulazhenkov @SchSeba @zeeke @Eoghan1232  thoughts ?

Eoghan1232: as a quick workaround, I think approach is probably fine.
but if there is a need to clean up additional resources through something like this, then a binary called from client-go lib would be a better approach probably so that way operator has control.....

deployment/sriov-network-operator-chart/templates/predeletehook.yaml
```diff
@@ -0,0 +1,24 @@
+{{ if .Values.operator.admissionControllers.enabled }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ include "sriov-network-operator.fullname" . }}-post-delete-hook"
```
ykulazhenkov, line 5: Why the name is `-post-delete-hook`? The hook is `pre-delete`

deployment/sriov-network-operator-chart/templates/predeletehook.yaml
SchSeba: two points here:
1. yes we need a binary  inside our image. use-case disconnected environments where the user needs to clone images and we have variables in helm to allow him specify the images for the operator. this one is static.
2. add a step in the run CI for k8s to test this

adrianchiris:
PR #791  addresses the same issue. closing this one.

PR #752, opened 2024-07-29
Fix virtual cluster redeploy script

SchSeba:


hack/virtual-cluster-redeploy.sh
```diff
@@ -19,10 +19,14 @@ if [ $CLUSTER_TYPE == "openshift" ]; then
   echo ${auth} > registry-login.conf
 
   internal_registry="image-registry.openshift-image-registry.svc:5000"
-  pass=$( jq .\"$internal_registry\".password registry-login.conf )
+  pass=$( jq .\"image-registry.openshift-image-registry.svc:5000\".auth registry-login.conf  )
```
adrianchiris, line 22: you can use jq to slice the string value e.g ` jq '.[1:-1]'` then pipe to base64. just suggestion current way also works.

hack/virtual-cluster-redeploy.sh
```diff
@@ -67,9 +72,38 @@ if [ $CLUSTER_TYPE == "openshift" ]; then
   export SRIOV_NETWORK_OPERATOR_IMAGE="image-registry.openshift-image-registry.svc:5000/$NAMESPACE/sriov-network-operator:latest"
   export SRIOV_NETWORK_CONFIG_DAEMON_IMAGE="image-registry.openshift-image-registry.svc:5000/$NAMESPACE/sriov-network-config-daemon:latest"
   export SRIOV_NETWORK_WEBHOOK_IMAGE="image-registry.openshift-image-registry.svc:5000/$NAMESPACE/sriov-network-operator-webhook:latest"
-fi
+  echo "## deploying SRIOV Network Operator"
+  hack/deploy-setup.sh $NAMESPACE
+else
+  source hack/env.sh
```
adrianchiris, line 78: should we have hack/deploy-operator-helm.sh with this logic ? then use it here and in run-e2e-conformance-virtual ?

SchSeba: nice idea! done

hack/virtual-cluster-redeploy.sh
SchSeba: I will leave this one for now if that is ok :)

hack/deploy-operator-helm.sh
```diff
@@ -0,0 +1,36 @@
...
+  --set images.webhook=${SRIOV_NETWORK_WEBHOOK_IMAGE} \
+  --set operator.admissionControllers.enabled=${ADMISSION_CONTROLLERS_ENABLED} \
+  --set operator.admissionControllers.certificates.certManager.enabled=${ADMISSION_CONTROLLERS_CERTIFICATES_CERT_MANAGER_ENABLED} \
+  --set sriovOperatorConfig.deploy=true"
+
+PATH=$PATH:${root}/bin
+make helm
+helm  upgrade -n ${NAMESPACE} --create-namespace \
+  $HELM_VALUES_OPTS \
+  --wait sriov-network-operator ./deployment/sriov-network-operator-chart
```
adrianchiris, line 36: nit: missing newline at the end.

adrianchiris (commented):
minor nit otherwise LGTM

hack/deploy-operator-helm.sh
SchSeba: yep :)

adrianchiris (approved):

zeeke (approved):

adrianchiris (approved):

PR #754, opened 2024-07-30
SCC-pinning for openshift workloads

kramaranya:
This PR explicitly sets the required SCC to be used to admit pods. The SCC chosen is the one that the pods are already getting admitted with, which means that this brings no change to the SCC used.

In some cases, custom SCCs can have higher priority than default SCCs, which means that they will be chosen over the default ones. This can lead to unexpected results; in order to protect openshift workloads from this, we must explicitly pin the required SCC to all our workloads in order to make sure that the expected one will be used.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -15,6 +15,8 @@ spec:
     metadata:
       labels:
         name: sriov-network-operator
+      annotations:
+        openshift.io/required-scc: restricted-v2
     spec:
       affinity:
         nodeAffinity:
--- a/bindata/manifests/daemon/daemonset.yaml
+++ b/bindata/manifests/daemon/daemonset.yaml
@@ -23,6 +23,7 @@ spec:
         openshift.io/component: network
       annotations:
         kubectl.kubernetes.io/default-container: sriov-network-config-daemon
+        openshift.io/required-scc: privileged
     spec:
       hostNetwork: true
       hostPID: true
--- a/bindata/manifests/operator-webhook/server.yaml
+++ b/bindata/manifests/operator-webhook/server.yaml
@@ -22,6 +22,8 @@ spec:
     metadata:
       labels:
         app: operator-webhook
+      annotations:
+        openshift.io/required-scc: restricted-v2
     spec:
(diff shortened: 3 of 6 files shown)


zeeke:
These changes are about setting a SecurityContextContraint for the operator's workload, in order to avoid getting assigned one by a custom priority. See [1] and [2].
Also, having the `openshift.io/required-scc` annotation in a non-Openshift cluster should be harmless, hence no need to add `{ if .IsOpenshift }` statements.

@kramaranya can you confirm my statement? Is there any other information that can be useful for reviewing this?

[1] https://docs.openshift.com/container-platform/4.16/authentication/managing-security-context-constraints.html#scc-prioritization_configuring-internal-oauth
[2] https://docs.openshift.com/container-platform/4.16/authentication/managing-security-context-constraints.html#default-sccs_configuring-internal-oauth

kramaranya:
@zeeke you are absolutely right, thanks for gathering this information.

Here is a jira [AUTH-482](https://issues.redhat.com/browse/AUTH-482), which might be useful for reviewing this. I have also updated a description of pr.

SchSeba (commented):
I think we also need this on the operator itself so please add it in the right place of the operator yamls

deploy/operator.yaml
```diff
@@ -15,6 +15,8 @@ spec:
     metadata:
       labels:
         name: sriov-network-operator
+      annotations:
```
zeeke, line 18: can you add this also in `deployment/sriov-network-operator-chart/templates/operator.yaml`?

https://github.com/kramaranya/sriov-network-operator-1/blob/6343fa158260a7f70dcf264a95e758e9393ba7ce/deployment/sriov-network-operator-chart/templates/operator.yaml#L19

It's about help deployment.

zeeke (commented):
Only one comment from my side, then LGTM.

@adrianchiris , @ykulazhenkov can you take a look at this? These annotations should not harm non-OpenShift deployments.

zeeke (approved):

ykulazhenkov (approved):
LGTM

PR #757, opened 2024-08-20
Intel switchdev

zeeke:
Ice driver supports creating VFs after the eSwitcMode is
set to switchdev. This is different than the preferred way
for `mlx5` driver.

Add specific functions to configure eSwitchMode and SriovNumVFs
for each driver.

Add a basic test to loop over all available
devices and test if setting `eSwitchMode: switchdev` works.

### DIFF
--- a/pkg/host/internal/sriov/sriov.go
+++ b/pkg/host/internal/sriov/sriov.go
@@ -898,9 +898,14 @@ func (s *sriov) SetNicSriovMode(pciAddress string, mode string) error {
 
 	dev, err := s.netlinkLib.DevLinkGetDeviceByName("pci", pciAddress)
 	if err != nil {
-		return err
+		return fmt.Errorf("can't get devlink device [%s] to set eSwitch to [%s]: %w", pciAddress, mode, err)
 	}
-	return s.netlinkLib.DevLinkSetEswitchMode(dev, mode)
+
+	err = s.netlinkLib.DevLinkSetEswitchMode(dev, mode)
+	if err != nil {
+		return fmt.Errorf("can't set eSwitch mode to [%s] on device [%s]: %w", mode, pciAddress, err)
+	}
+	return nil
 }
 
 func (s *sriov) GetLinkType(name string) string {
@@ -989,22 +994,41 @@ func (s *sriov) createVFs(iface *sriovnetworkv1.Interface) error {
 	return s.setEswitchModeAndNumVFs(iface.PciAddress, expectedEswitchMode, iface.NumVfs)
 }
 
-func (s *sriov) setEswitchMode(pciAddr, eswitchMode string) error {
-	log.Log.V(2).Info("setEswitchMode(): set eswitch mode", "device", pciAddr, "mode", eswitchMode)
-	if err := s.unbindAllVFsOnPF(pciAddr); err != nil {
-		log.Log.Error(err, "setEswitchMode(): failed to unbind VFs", "device", pciAddr, "mode", eswitchMode)
-		return err
-	}
-	if err := s.SetNicSriovMode(pciAddr, eswitchMode); err != nil {
-		err = fmt.Errorf("failed to switch NIC to SRIOV %s mode: %v", eswitchMode, err)
-		log.Log.Error(err, "setEswitchMode(): failed to set mode", "device", pciAddr, "mode", eswitchMode)
-		return err
-	}
-	return nil
-}
+type setEswitchModeAndNumVFsFn func(string, string, int) error
 
 func (s *sriov) setEswitchModeAndNumVFs(pciAddr string, desiredEswitchMode string, numVFs int) error {
+	pfDriverName, err := s.dputilsLib.GetDriverName(pciAddr)
+	if err != nil {
+		return err
+	}
+
 	log.Log.V(2).Info("setEswitchModeAndNumVFs(): configure VFs for device",
+		"device", pciAddr, "count", numVFs, "mode", desiredEswitchMode, "driver", pfDriverName)
+
+	setEswitchModeAndNumVFsByDriverName := map[string]setEswitchModeAndNumVFsFn{
+		"ice":       s.setEswitchModeAndNumVFsIce,
+		"mlx5_core": s.setEswitchModeAndNumVFsMlx,
+	}
+
+	fn, ok := setEswitchModeAndNumVFsByDriverName[pfDriverName]
+	if !ok {
+		log.Log.V(2).Info("setEswitchModeAndNumVFs(): driver not found in the support list. Using fallback implementation",
+			"device", pciAddr, "driver", pfDriverName)
+
+		// Fallback to mlx5 driver
+		fn = s.setEswitchModeAndNumVFsMlx
+	}
+
+	return fn(pciAddr, desiredEswitchMode, numVFs)
+}
+
+// setEswitchModeAndNumVFsMlx configures PF eSwitch and sriov_numvfs in the following order:
+// a. set eSwitchMode to legacy
+// b. set the desired number of Virtual Functions
+// c. unbind driver of all VFs
+// d. set eSwitchMode to `switchdev` if requested
+func (s *sriov) setEswitchModeAndNumVFsMlx(pciAddr string, desiredEswitchMode string, numVFs int) error {
+	log.Log.V(2).Info("setEswitchModeAndNumVFsMlx(): configure VFs for device",
 		"device", pciAddr, "count", numVFs, "mode", desiredEswitchMode)
 
 	// always switch NIC to the legacy mode before creating VFs. This is required because some drivers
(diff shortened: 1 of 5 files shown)


ykulazhenkov (dismissed):
LGTM

Eoghan1232:
I will do a test of this

Eoghan1232:
worked like a charm @zeeke  ;) 

let me give the PR a review now then we should be good to merge

Eoghan1232 (approved):
Changes LGTM!

once this is merged, we should also update the support HW list.
I will do this

pkg/host/internal/sriov/sriov.go
```diff
@@ -986,22 +986,30 @@ func (s *sriov) createVFs(iface *sriovnetworkv1.Interface) error {
...
+	pfDriverName, err := s.dputilsLib.GetDriverName(pciAddr)
+	if err != nil {
 		return err
 	}
-	if err := s.SetNicSriovMode(pciAddr, eswitchMode); err != nil {
-		err = fmt.Errorf("failed to switch NIC to SRIOV %s mode: %v", eswitchMode, err)
-		log.Log.Error(err, "setEswitchMode(): failed to set mode", "device", pciAddr, "mode", eswitchMode)
-		return err
+
+	type setEswitchModeAndNumVFsFn func(string, string, int) error
```
SchSeba, line 995: nit: can we move this one outside the function?

pkg/host/internal/sriov/sriov.go
```diff
@@ -986,22 +986,30 @@ func (s *sriov) createVFs(iface *sriovnetworkv1.Interface) error {
...
+
+	var setEswitchModeAndNumVFsByDriverName map[string]setEswitchModeAndNumVFsFn = map[string]setEswitchModeAndNumVFsFn{
```
SchSeba, line 997: nit: can we do something like

```
setEswitchModeAndNumVFsByDriverName map[string]setEswitchModeAndNumVFsFn := map[string]setEswitchModeAndNumVFsFn{...}
```

pkg/host/internal/sriov/sriov.go
```diff
@@ -986,22 +986,30 @@ func (s *sriov) createVFs(iface *sriovnetworkv1.Interface) error {
 	return s.setEswitchModeAndNumVFs(iface.PciAddress, expectedEswitchMode, iface.NumVfs)
 }
 
-func (s *sriov) setEswitchMode(pciAddr, eswitchMode string) error {
-	log.Log.V(2).Info("setEswitchMode(): set eswitch mode", "device", pciAddr, "mode", eswitchMode)
-	if err := s.unbindAllVFsOnPF(pciAddr); err != nil {
-		log.Log.Error(err, "setEswitchMode(): failed to unbind VFs", "device", pciAddr, "mode", eswitchMode)
+func (s *sriov) setEswitchModeAndNumVFs(pciAddr string, desiredEswitchMode string, numVFs int) error {
+	pfDriverName, err := s.dputilsLib.GetDriverName(pciAddr)
```
SchSeba, line 990: nit: can we please add a debug log level about the PCI address and the driver name?

pkg/host/internal/sriov/sriov.go
```diff
@@ -986,22 +986,30 @@ func (s *sriov) createVFs(iface *sriovnetworkv1.Interface) error {
...
+	var setEswitchModeAndNumVFsByDriverName map[string]setEswitchModeAndNumVFsFn = map[string]setEswitchModeAndNumVFsFn{
+		"ice":       s.setEswitchModeAndNumVFsIce,
+		"mlx5_core": s.setEswitchModeAndNumVFsMlx,
 	}
-	return nil
+
+	fn, ok := setEswitchModeAndNumVFsByDriverName[pfDriverName]
+	if !ok {
+		// Fallback to mlx5 driver
+		fn = s.setEswitchModeAndNumVFsMlx
```
SchSeba, line 1005: can you please add a debug message that we didn't find the driver in the supported list so we fallback

pkg/host/internal/sriov/sriov.go
```diff
@@ -986,22 +986,30 @@ func (s *sriov) createVFs(iface *sriovnetworkv1.Interface) error {
...
+	}
+
+	return fn(pciAddr, desiredEswitchMode, numVFs)
 }
 
-func (s *sriov) setEswitchModeAndNumVFs(pciAddr string, desiredEswitchMode string, numVFs int) error {
-	log.Log.V(2).Info("setEswitchModeAndNumVFs(): configure VFs for device",
+func (s *sriov) setEswitchModeAndNumVFsMlx(pciAddr string, desiredEswitchMode string, numVFs int) error {
```
SchSeba, line 1011: please add a function comment to explain the process of how we configure switchDev virtual functions for this case

pkg/host/internal/sriov/sriov.go
```diff
@@ -1011,13 +1019,46 @@ func (s *sriov) setEswitchModeAndNumVFs(pciAddr string, desiredEswitchMode strin
...
 	}
 
 	if desiredEswitchMode == sriovnetworkv1.ESwithModeSwitchDev {
 		return s.setEswitchMode(pciAddr, sriovnetworkv1.ESwithModeSwitchDev)
 	}
+
+	return nil
+}
+
+func (s *sriov) setEswitchModeAndNumVFsIce(pciAddr string, desiredEswitchMode string, numVFs int) error {
```
SchSeba, line 1034: please add a function comment to explain the process of how we configure switchDev virtual functions for this case

test/conformance/tests/test_switchdev.go
```diff
@@ -0,0 +1,95 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("[sriov] Switchdev", Ordered, func() {
+
+	BeforeAll(func() {
+		if cluster.VirtualCluster() {
+			Skip("IGB driver does not support VF statistics")
```
SchSeba, line 25: nit: please update the skip message here :)

test/conformance/tests/test_switchdev.go
```diff
@@ -0,0 +1,95 @@
...
+		sriovInfos, err := cluster.DiscoverSriov(clients, operatorNamespace)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(sriovInfos.Nodes)).ToNot(BeZero())
+
+		testNode, interfaces, err := sriovInfos.FindSriovDevicesAndNode()
+		Expect(err).ToNot(HaveOccurred())
+
+		By(fmt.Sprintf("Testing on node %s, %d devices found", testNode, len(interfaces)))
+
+		mainDevice := findMainSriovDevice(getConfigDaemonPod(testNode), interfaces)
```
SchSeba, line 47: you can use `findUnusedSriovDevices` to have the list of non main device :)

pkg/host/internal/sriov/sriov.go
zeeke: ok, moving. though that type is not supposed to be used out of this function

pkg/host/internal/sriov/sriov.go
zeeke: good point!

adrianchiris:
Please hold off with merging this one id like to go over this pr by end of week

pkg/host/internal/sriov/sriov.go
```diff
@@ -1029,6 +1047,40 @@ func (s *sriov) setEswitchModeAndNumVFs(pciAddr string, desiredEswitchMode strin
 	return nil
 }
 
+// setEswitchModeAndNumVFsIce configures PF eSwitch and sriov_numvfs in the following order:
+// a. set eSwitchMode to `switchdev` if requested
```
Eoghan1232, line 1051: we need to reset vf's to 0 before setting switchdev mode, is this handled?

pkg/host/internal/sriov/sriov.go
adrianchiris: nit: between b. and c. we also unbind all VFs on PF.
can you add this to the comment ?

note: rebind happens after part of `configSriovVFDevices`

also what about `configureHWOptionsForSwitchdev` ? is it the same for intel ice ?

adrianchiris (commented):
overall lgtm, left one small comment

pkg/host/internal/sriov/sriov.go
Eoghan1232: I believe so, otherwise driver will throw an error I believe. 
I can double check this tomorrow to confirm

Eoghan1232: I can see in some docs that we must have 0 VF's before switching to switchdev mode.

pkg/host/internal/sriov/sriov.go
Eoghan1232: I don't think we need to unbind driver, since we need to reset back to 0 vf's anyways

zeeke: Got it. Then better move the `unbindAllVFsOnPF(...)` call outside of `setEswitchMode(...)`

adrianchiris: > Regarding configureHWOptionsForSwitchdev, current implementation is about setting flow_steering_mode=smfs, which is Mellanox only (IIUC).

this also enables hw-tc-offload via ethtool.

zeeke: > > Regarding configureHWOptionsForSwitchdev, current implementation is about setting flow_steering_mode=smfs, which is Mellanox only (IIUC).
> 
> this also enables hw-tc-offload via ethtool.

right! I miss that. I checked and it should not be a problem for the ice driver:

from a local test: 
```
2024-08-29T08:21:49.27421628Z   LEVEL(-2)       sriov/sriov.go:365      EnableHwTcOffload(): enable offloading  {"device": "eno12409"}
2024-08-29T08:21:49.274526639Z  LEVEL(-2)       sriov/sriov.go:365      EnableHwTcOffload(): already enabled    {"device": "eno12409"}
```

and `flow_steering_mode` setting is guarded: if it's not supported, it doesn't hurt.

SchSeba (approved):
LGTM

nice work!

e0ne (approved):

zeeke:
Got approval from multiple sides, merging.

PR #764, opened 2024-08-27
set dev version for chart

adrianchiris:
chart in the repo should be set to an arbitrary dev version as we deploy latest components by default.

during release workflow relevant values are replaced when release chart is built.


ykulazhenkov (approved):
LGTM

zeeke (approved):

PR #768, opened 2024-08-28
Security: Reduce rbac from webhooks

SchSeba:
Always good to improve security


zeeke (approved):
we're drastically reducing the attack surface here!

LGTM

adrianchiris (approved):

PR #775, opened 2024-09-12
Preserve user defined annotations on `SriovNetworks`

zeeke:
If an SriovNetwork is created with some annotations, the operator deletes them during the creation of  the related NetworkAttachmentDefinition.

Make the reconcilers append the `last-network-namespace` annotation without deleting existing ones.

The same concept applies for `SriovIbNetworks` and `OVSNetworks`


adrianchiris (approved):
lgtm

SchSeba (approved):
LGTM

PR #776, opened 2024-09-18
Fix merge annotation function

SchSeba:
if the current obj as annotation and the updated doesn't we still want to add the ones from the current object

### DIFF
--- a/pkg/apply/merge_test.go
+++ b/pkg/apply/merge_test.go
@@ -107,6 +107,38 @@ metadata:
 	}))
 }
 
+func TestMergeOne(t *testing.T) {
+	g := NewGomegaWithT(t)
+
+	cur := UnstructuredFromYaml(t, `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: d1
+  labels:
+    label-c: cur
+  annotations:
+    annotation-c: cur`)
+
+	upd := UnstructuredFromYaml(t, `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: d1`)
+
+	// this mutates updated
+	err := MergeObjectForUpdate(cur, upd)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	g.Expect(upd.GetLabels()).To(Equal(map[string]string{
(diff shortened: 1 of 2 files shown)


pkg/apply/merge_test.go
```diff
@@ -107,6 +107,38 @@ metadata:
...
+
+	cur := UnstructuredFromYaml(t, `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: d1
+  labels:
+    c: upd
+  annotations:
+    c: upd`)
```
zeeke, line 121: nit: took me a while to get this test. can you change it in:
```suggestion
  labels:
    label-c: cur
  annotations:
    annotation-c: cur`)
```

zeeke (commented):
The gist here is: that operator should never remove a label/annotation from its managed object, right?
I guess this is because annotations and labels may come from the user.

SchSeba:
Right or from operators.

Where I was able to find this issue is in openshift we have an operator to generate the internal pull secret for every service account

```
apiVersion: v1
imagePullSecrets:
- name: network-resources-injector-sa-dockercfg-46j2c
kind: ServiceAccount
metadata:
  annotations:
    openshift.io/internal-registry-pull-secret-ref: network-resources-injector-sa-dockercfg-46j2c
creationTimestamp: "2024-09-16T12:15:10Z"
  name: network-resources-injector-sa
  namespace: openshift-sriov-network-operator
  ownerReferences:
  - apiVersion: sriovnetwork.openshift.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: SriovOperatorConfig
    name: default
    uid: 745df61d-68d7-4e97-a399-389bae71b7ba
  resourceVersion: "28652"
  uid: 6fe1b49e-943f-4c95-92aa-5cf0ab322151
secrets:
- name: network-resources-injector-sa-dockercfg-46j2c
```

without this change everytime we did a reconcile we remove this on and the openshift operator was creating a new imagepullSecret

after a few days this was running the number of secrets was huge

ykulazhenkov (approved):
LGTM

PR #777, opened 2024-09-18
[Infiniband] Fix unbinding of physical functions when configuring Infiniband virtual functions

clarkzinzow:
This PR fixes a bug in the config daemon that unbinds physical functions instead of virtual functions when configuring Infiniband virtual functions.


clarkzinzow:
Apologies, I meant to open this as a PR on a longer-lived feature branch in our fork for internal tracking! I'll close this for now.

PR #779, opened 2024-09-19
Delete webhooks when SriovOperatorConfig is deleted

zeeke:
When a user deletes the default SriovOperatorConfig resource and tries to recreate it afterward, the operator webhooks returns the error:
```
Error from server (InternalError): error when creating "/tmp/opconfig.yml": Internal error occurred: failed calling webhook "operator-webhook.sriovnetwork.openshift.io": failed to call webhook: Post "https://operator-webhook-service.openshift-sriov-network-operator.svc:443/validating-custom-resource?timeout=10s": service "operator-webhook-service" not found
```

as the webhook configuration is still present, while the Service and the DaemonSet have been deleted.

Delete all the webhook configurations when the user deletes the default SriovOperatorConfig

### DIFF
--- a/controllers/sriovoperatorconfig_controller.go
+++ b/controllers/sriovoperatorconfig_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"sort"
@@ -28,6 +29,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	kscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


SchSeba (approved):
LGTM

nice work!

ykulazhenkov (approved):
LGTM

controllers/sriovoperatorconfig_controller.go
```diff
@@ -457,3 +461,29 @@ func (r SriovOperatorConfigReconciler) setLabelInsideObject(ctx context.Context,
 
 	return nil
 }
+
+func (r SriovOperatorConfigReconciler) deleteAllWebhooks(ctx context.Context) error {
```
e0ne, line 465: please, re-use `deleteWebhookObject` method here

e0ne (changes requested):
good fix! a minor comment should be addressed to make code cleaner

controllers/sriovoperatorconfig_controller.go
zeeke: sounds good!

SchSeba (approved):
LGTM

e0ne (approved):
Thanks for addressing my comments!

controllers/sriovoperatorconfig_controller.go
```diff
@@ -81,7 +83,9 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			logger.Info("default SriovOperatorConfig object not found. waiting for creation.")
-			return reconcile.Result{}, nil
+
+			err := r.deleteAllWebhooks(ctx)
+			return reconcile.Result{}, err
```
bn222, line 88: Why don't we use owner reference? Then, all of this is automatically handled.

zeeke: Because ValidatingWebhookConfiguration and MutatingWebhookConfiguration resources are not namespaced, while the OwnerReference field works with resources in the same namespace as the owner.

PR #785, opened 2024-10-07
metrics: Fix `Metrics should have the correct labels` test

zeeke:
`sriov_kubepoddevice` metric might end up in the Prometheus database after a while, as the default scrape interval is 30s. This leads to failures in the end-to-end lane like:

```
[sriov] Metrics Exporter When Prometheus operator is available [It] Metrics should have the correct labels
/root/opr-ocp2-1/data/sriov-network-operator/sriov-network-operator/test/conformance/tests/test_exporter_metrics.go:132

  [FAILED] no value for metric sriov_kubepoddevice
```

Put the metric assertion in an `Eventually` statement


SchSeba (approved):
LGTM

ykulazhenkov (approved):
LGTM

PR #788, opened 2024-10-10
Daemon redesign - using controller-runtime

SchSeba:


cmd/sriov-network-config-daemon/service.go
```diff
@@ -163,12 +164,24 @@ func phasePre(setupLog logr.Logger, conf *systemd.SriovConfig, hostHelpers helpe
 	hostHelpers.TryEnableTun()
 	hostHelpers.TryEnableVhostNet()
 
-	return callPlugin(setupLog, PhasePre, conf, hostHelpers)
+	// Add retry here as some time the kernel is not fully loaded, and we see issues like
+	// unbindDriver(): failed to unbind driver "error": "open /sys/bus/pci/drivers/igbvf/unbind: permission denied"
+	i := 0
+	for i < 5 {
```
ykulazhenkov, line 170: can you move this change to a separate PR?
I think we should review discuss it separately.

ykulazhenkov: Was the problem that is described in the comment ever seen with this patch?https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/808 I think it can implicitly fix the issue.

SchSeba: I will try again but for sure that is not part of this PR.

I am working on another PR to add a systemd mode functional test lane, that is how I catch it.

pkg/daemon/writer.go
```diff
@@ -1,290 +0,0 @@
-package daemon
```
zeeke, line 1: It's great to see this file go away! :)

pkg/daemon/status.go
```diff
@@ -0,0 +1,142 @@
...
+		return true, nil
+	}
+
+	// check for interfaces
+	// we can't use deep equal here because if we have a vf inside a pod is name will not be available for example
+	// we use the index for both lists
+	c := current.Status.DeepCopy().Interfaces
+	d := desiredNodeState.Status.DeepCopy().Interfaces
+	for idx := range d {
+		// check if it's a new device
```
zeeke, line 82:
```suggestion
        if idx >= len(c) {
          return true, nil
        }
		// check if it's a new device
```

I don't know if it can ever happen (maybe if a new device is hot-plugged between reconcile loops). But it's safer to check for slice length

pkg/daemon/status.go
```diff
@@ -0,0 +1,142 @@
...
+	if retryErr != nil {
+		funcLog.Error(retryErr, "failed to update node state status")
+		return retryErr
+	}
+
+	dn.recordStatusChangeEvent(ctx, currentNodeState.Status.SyncStatus, status, failedMessage)
+	return nil
+}
+
+func (dn *DaemonReconcile) shouldUpdateStatus(current, desiredNodeState *sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
```
zeeke, line 60: nit: this function returns no error. can get rid of the second return parameter?

zeeke (commented):
Left a few minor comments. This PR hugely improves the config-daemon shape.

pkg/daemon/daemon.go
```diff
@@ -93,398 +55,304 @@ func New(
...
-			return fmt.Errorf("error syncing: %s, requeuing", err.Error())
-		}
-		// Finally, if no error occurs we Forget this item so it does not
-		// get queued again until another change happens.
-		dn.workqueue.Forget(obj)
-		log.Log.Info("Successfully synced")
-		return nil
-	}(obj)
 
+	// save init state
```
adrianchiris, line 135: nit: // save init state if it does not exist on host

pkg/daemon/daemon.go
```diff
@@ -93,398 +55,304 @@ func New(
...
-	var sriovResult = &systemd.SriovResult{SyncStatus: consts.SyncStatusSucceeded, LastSyncError: ""}
-	dn.desiredNodeState, err = dn.sriovClient.SriovnetworkV1().SriovNetworkNodeStates(vars.Namespace).Get(context.Background(), vars.NodeName, metav1.GetOptions{})
+	// if we are running in systemd mode we want to get the sriov result from the config-daemon that runs in systemd
+	sriovResult, exist, err := dn.checkSystemdStatus(ctx, desiredNodeState)
+	//TODO: in the case we need to think what to do if we try to apply again or not
 	if err != nil {
-		log.Log.Error(err, "nodeStateSyncHandler(): Failed to fetch node state", "name", vars.NodeName)
-		return err
+		reqLogger.Error(err, "failed to check systemd status unexpected error")
+		return ctrl.Result{}, nil
```
adrianchiris, line 181: we dont want to fail here ?

pkg/daemon/daemon.go
```diff
@@ -93,398 +55,304 @@ func New(
...
 
-				// add the error but don't requeue
-				dn.refreshCh <- Message{
-					syncStatus:    consts.SyncStatusFailed,
-					lastSyncError: sriovResult.LastSyncError,
-				}
-				<-dn.syncCh
-				return nil
-			}
+			return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
adrianchiris, line 210: nit: lets define requeueTime as a const

pkg/daemon/daemon.go
```diff
@@ -93,398 +55,304 @@ func New(
...
-	// if the operator complete the drain operator we should continue the configuration
-	if !dn.isDrainCompleted() {
-		if vars.UsingSystemdMode && dn.currentNodeState.GetGeneration() == latest {
-			serviceEnabled, err := dn.HostHelpers.IsServiceEnabled(systemd.SriovServicePath)
-			if err != nil {
-				log.Log.Error(err, "nodeStateSyncHandler(): failed to check if sriov-config service exist on host")
-				return err
-			}
-			postNetworkServiceEnabled, err := dn.HostHelpers.IsServiceEnabled(systemd.SriovPostNetworkServicePath)
+		// if there are no node drifted changes, and we are on the latest applied policy
```
adrianchiris, line 192: nit: `// if there are no host state drift changes, ...`

pkg/daemon/daemon.go
```diff
@@ -93,398 +55,304 @@ func New(
...
+		// When using systemd configuration we write the file
+		systemdConfModified, err := dn.writeSystemdConfigFile(desiredNodeState)
 		if err != nil {
-			return err
+			reqLogger.Error(err, "failed to write systemd config file")
+			return ctrl.Result{}, err
 		}
+		reqDrain = reqDrain || systemdConfModified || !exist
+		// require reboot if drain needed for systemd mode
+		reqReboot = reqReboot || systemdConfModified || reqDrain || !exist
```
adrianchiris, line 237: nit: you can drop : `|| systemdConfModified` and `|| !exist`
they are checked in `reqDrain`

pkg/daemon/daemon.go
```diff
@@ -93,398 +55,304 @@ func New(
...
-			"drain-required", reqDrain, "reboot-required", reqReboot, "disable-drain", dn.disableDrain)
+	return reqReboot, reqDrain, nil
+}
 
-		err = systemd.WriteSriovSupportedNics()
-		if err != nil {
-			log.Log.Error(err, "nodeStateSyncHandler(): failed to write supported nic ids file for systemd mode")
-			return err
-		}
+func (dn *DaemonReconcile) checkSystemdStatus(ctx context.Context, desiredNodeState *sriovnetworkv1.SriovNetworkNodeState) (*systemd.SriovResult, bool, error) {
```
adrianchiris, line 288: can you add docstring explaining the returned args ?

pkg/daemon/daemon.go
```diff
@@ -496,190 +364,169 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
+
+	err = dn.updateSyncState(ctx, desiredNodeState, syncStatus, lastSyncError)
+	if err != nil {
+		reqLogger.Error(err, "failed to update sync status")
+		return ctrl.Result{}, err
+	}
+
+	// update the lastAppliedGeneration
+	dn.lastAppliedGeneration = desiredNodeState.Generation
+	return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
adrianchiris, line 426: nit: same as previous comment. lets use a const

pkg/daemon/daemon.go
```diff
@@ -93,398 +55,304 @@ func New(
...
+	//if desiredNodeState.Status.SyncStatus != sriovResult.SyncStatus ||
+	//	desiredNodeState.Status.LastSyncError != sriovResult.LastSyncError {
+	//	err = dn.updateSyncState(ctx, desiredNodeState, sriovResult.SyncStatus, sriovResult.LastSyncError)
+	//	if err != nil {
+	//		funcLog.Error(err, "failed to update sync status")
+	//	}
+	//	return sriovResult, err
+	//}
+
+	// TODO: check if we need this
```
adrianchiris, line 333: i believe we need.

in case systemd services were enabled, and there is error in sriovResult, we want to reflect that in status and stop reconcile.

adrianchiris: @ykulazhenkov for some more feedback on this bit

adrianchiris:
once we get this in, i think we should move the node controllers under `controllers/node/` pkg

bindata/manifests/sriov-config-service/kubernetes/sriov-config-service.yaml
```diff
@@ -7,7 +7,7 @@ contents: |
 
   [Service]
   Type=oneshot
-  ExecStart=/var/lib/sriov/sriov-network-config-daemon -v 2 --zap-log-level 2 service --phase pre
+  ExecStart=/var/lib/sriov/sriov-network-config-daemon service --phase pre
```
adrianchiris, line 10: this means reboot on update ? since systemfile changes

hack/run-e2e-conformance.sh
```diff
@@ -7,4 +7,4 @@ GOPATH="${GOPATH:-$HOME/go}"
 JUNIT_OUTPUT="${JUNIT_OUTPUT:-/tmp/artifacts}"
 export PATH=$PATH:$GOPATH/bin
 
-${root}/bin/ginkgo -output-dir=$JUNIT_OUTPUT --junit-report "unit_report.xml" -v "$SUITE" -- -report=$JUNIT_OUTPUT
+${root}/bin/ginkgo --timeout 3h -output-dir=$JUNIT_OUTPUT --junit-report "unit_report.xml" -v "$SUITE" -- -report=$JUNIT_OUTPUT
```
adrianchiris, line 10: is this change related ?

pkg/daemon/status.go
SchSeba: but in theory (not sure if it's possible) you can have a new device like hot-unplug and after that hot-plug so the len will be the same but the device is not no?

pkg/daemon/status.go
SchSeba: you are right!

controllers/helper.go
```diff
@@ -108,14 +108,7 @@ type DrainStateAnnotationPredicate struct {
 }
 
 func (DrainStateAnnotationPredicate) Create(e event.CreateEvent) bool {
```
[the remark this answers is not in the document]
SchSeba: the concern is not nil. without this when a new node is added to the cluster the reconcile will not get called without this.
and if the reconcile is not created no one adds the labels to make the drain possible.

cmd/sriov-network-config-daemon/start.go
```diff
@@ -240,88 +258,108 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
...
 	setupLog.Info("Running on", "platform", vars.PlatformType.String())
 
+	// Initial supported nic IDs
 	if err := sriovnetworkv1.InitNicIDMapFromConfigMap(kubeclient, vars.Namespace); err != nil {
 		setupLog.Error(err, "failed to run init NicIdMap")
 		return err
 	}
 
-	eventRecorder.SendEvent("ConfigDaemonStart", "Config Daemon starting")
+	eventRecorder.SendEvent(stopSignalCh, "ConfigDaemonStart", "Config Daemon starting")
```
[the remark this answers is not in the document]
SchSeba: it was here before, I don't have strong feelings on it. do you want to move it below or remove it completely ?

pkg/daemon/config.go
```diff
@@ -0,0 +1,61 @@
...
+	snolog.SetLogLevel(operatorConfig.Spec.LogLevel)
+
+	newDisableDrain := operatorConfig.Spec.DisableDrain
+	if vars.DisableDrain != newDisableDrain {
+		vars.DisableDrain = newDisableDrain
+		log.Log.Info("Set Disable Drain", "value", vars.DisableDrain)
+	}
+
+	if !reflect.DeepEqual(oc.latestFeatureGates, operatorConfig.Spec.FeatureGates) {
+		vars.FeatureGate.Init(operatorConfig.Spec.FeatureGates)
```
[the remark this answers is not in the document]
SchSeba: featureGate object already implement mutex.
and disable drain is only changed here.

let me know if you think that we need for the disable drain

pkg/daemon/status.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	// check for interfaces
+	// we can't use deep equal here because if we have a vf inside a pod is name will not be available for example
+	// we use the index for both lists
+	c := current.Status.DeepCopy().Interfaces
+	d := desiredNodeState.Status.DeepCopy().Interfaces
+	for idx := range d {
+		// check if it's a new device
+		if d[idx].PciAddress != c[idx].PciAddress {
```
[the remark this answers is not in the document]
SchSeba: yes we do it base on pci address https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/5522c96101c673ad15efbc5a7acd1596283bc19c/pkg/host/internal/sriov/sriov.go#L211
so they must be the same if they are not it's a change

controllers/helper.go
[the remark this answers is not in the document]
adrianchiris: > without this when a new node is added to the cluster the reconcile will not get called without this

are you sure ? i thought by default it will call reconcile since predicate.Funcs contains implementation that will return true. which means to enqueue event for reconcile.

pkg/daemon/config.go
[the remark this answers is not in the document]
adrianchiris: ack, we didnt have b4 as well.

test/util/k8sreporter/reporter.go
```diff
@@ -6,13 +6,13 @@ import (
 	"strings"
 
 	kniK8sReporter "github.com/openshift-kni/k8sreporter"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/runtime"
```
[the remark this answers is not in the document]
SchSeba: what you mean? which imports you want in the same group?

pkg/daemon/daemon_test.go
```diff
@@ -1,337 +1,315 @@
...
-				"bindata/scripts",
-				"host/etc/sriov-operator",
-				"host/etc/sriov-operator/pci",
-				"host/etc/udev/rules.d",
-			},
-			Symlinks: map[string]string{},
-			Files: map[string][]byte{
-				"/bindata/scripts/enable-rdma.sh": []byte(""),
-				"/bindata/scripts/load-kmod.sh":   []byte(""),
+		Expect(k8sClient.DeleteAllOf(context.Background(), &sriovnetworkv1.SriovOperatorConfig{}, client.InNamespace(testNamespace))).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
SchSeba: for the operator config I don't think as we use it in a BeforeAll to prepare the env

controllers/helper.go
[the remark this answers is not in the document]
SchSeba: I give it a try and I needed to add it. without it I was not able to get the create calls :(

I can give it another try

pkg/daemon/daemon_test.go
```diff
@@ -1,337 +1,315 @@
...
-				Name:      sriovnetworkv1.SupportedNicIDConfigmap,
-				Namespace: vars.Namespace,
-			},
-			Data: map[string]string{
-				"Intel_i40e_XXV710":      "8086 158a 154c",
-				"Nvidia_mlx5_ConnectX-4": "15b3 1013 1014",
-			},
-		}
+		snolog.SetLogLevel(2)
+		vars.ClusterType = constants.ClusterTypeOpenshift
```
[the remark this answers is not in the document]
SchSeba: sure!

pkg/daemon/daemon.go
```diff
@@ -775,6 +626,57 @@ func (dn *Daemon) prepareNMUdevRule() error {
...
+		return false, err
+	}
+
+	// the node was annotated we need to wait for the operator to finish the drain
+	return true, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (dn *DaemonReconcile) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
```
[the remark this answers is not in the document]
SchSeba: It should be the default, but I like to go with explicit :)

pkg/daemon/daemon.go
```diff
@@ -775,6 +626,57 @@ func (dn *Daemon) prepareNMUdevRule() error {
...
+	}
+
+	// if we don't need to drain at all just return true so we can apply the configuration
+	return true
+}
+
+func (dn *DaemonReconcile) annotate(
+	ctx context.Context,
+	desiredNodeState *sriovnetworkv1.SriovNetworkNodeState,
+	annotationState string) (bool, error) {
```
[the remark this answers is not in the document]
SchSeba: nice catch!

pkg/daemon/daemon.go
```diff
@@ -496,190 +364,169 @@ func (dn *Daemon) nodeStateSyncHandler() error {
...
-				syncStatus:    consts.SyncStatusSucceeded,
-				lastSyncError: "",
-			}
-			// wait for writer to refresh the status
-			<-dn.syncCh
+	log.Log.V(0).Info("Interfaces not changed")
+	return false, nil
+}
+
+func (dn *DaemonReconcile) writeSystemdConfigFile(desiredNodeState *sriovnetworkv1.SriovNetworkNodeState) (bool, error) {
```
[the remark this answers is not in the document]
SchSeba: I don't think we need that in case of the daemon as we only have 1 active reconcile loop.

I use that in the drain controller to understand the context

pkg/daemon/daemon.go
SchSeba: from the code I can see that we always write the file. I am missing something?

pkg/daemon/daemon.go
SchSeba: If we return an error it will get stuck in a loop as if the result for the systemd operator return an error we don't retry to not put the node into a bootloop

pkg/daemon/daemon.go
SchSeba: it was moved out of this function so we should be ok I remove this

bindata/manifests/sriov-config-service/kubernetes/sriov-config-service.yaml
SchSeba: yes it will make a reboot one time

SchSeba:
Hi @zeeke @adrianchiris @ykulazhenkov can you please make another round of review here?

pkg/daemon/status.go
zeeke: in both case, the `if idx >= len(c) {` can save a nil dereferencing. right?

cmd/sriov-network-config-daemon/start.go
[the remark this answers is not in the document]
adrianchiris: lets remove completely imo WDYT ?

controllers/helper.go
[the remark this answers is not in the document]
adrianchiris: so this ended up as needed ?

pkg/daemon/daemon.go
adrianchiris: ack.

 should we update status with "unexpected error" ?

pkg/daemon/daemon.go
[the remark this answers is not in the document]
adrianchiris: ack

pkg/daemon/daemon.go
```diff
@@ -3,488 +3,363 @@ package daemon
...
-}
-
-type Daemon struct {
+// NodeReconciler struct holds various components necessary for reconciling an SR-IOV node.
+// It includes a Kubernetes client, SR-IOV client, and other utility interfaces.
+// The struct is designed to manage the lifecycle of an SR-IOV devices on a given node.
+type NodeReconciler struct {
 	client client.Client
 
 	sriovClient snclientset.Interface
```
adrianchiris, line 36: i think sriovClient is not used in the reconciler ? can we drop it from the NodeReconciler struct ?

pkg/daemon/daemon.go
```diff
@@ -3,488 +3,363 @@ package daemon
...
 	// kubeClient allows interaction with Kubernetes, including the node we are running on.
 	kubeClient kubernetes.Interface
```
adrianchiris, line 38: WDYT about dropping this and using controller runtime client ?
we use it to get pods 

that means we would also add pods to cache in main.

it would reduce load on api server overall this way.

adrianchiris (commented):
overall looks great !

added some small comments. once we resolve them we can merge !

controllers/helper.go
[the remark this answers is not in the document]
SchSeba: yep we need this one at least from my tests

cmd/sriov-network-config-daemon/start.go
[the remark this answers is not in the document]
SchSeba: sure done

test/util/k8sreporter/reporter.go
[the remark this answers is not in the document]
SchSeba: done

pkg/daemon/daemon.go
SchSeba: right nice catch!

pkg/daemon/daemon.go
SchSeba: done

pkg/daemon/daemon.go
SchSeba: done

e0ne (approved):
lgtm

e0ne:
LGTM. it could be merged once CI pass

pkg/daemon/daemon.go
```diff
@@ -3,488 +3,360 @@ package daemon
...
 	eventRecorder *EventRecorder
 
 	featureGate featuregate.FeatureGate
+
+	// list of disabled plugins
+	disabledPlugins []string
+
+	loadedPlugins         map[string]plugin.VendorPlugin
+	lastAppliedGeneration int64
+	disableDrain          bool
```
ykulazhenkov, line 48: it looks like we never set this value.

pkg/daemon/config.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			reqLogger.Info("OperatorConfig doesn't exist", "name", req.Name, "namespace", req.Namespace)
+			return ctrl.Result{}, nil
+		}
+		reqLogger.Error(err, "Failed to get OperatorConfig", "name", req.Name, "namespace", req.Namespace)
+		return ctrl.Result{}, err
+	}
+
+	// update log level
```
ykulazhenkov, line 42: we need to make sure that all these global variables are initialized before the NodeReconciler starts. 
Currently, FeatureGates are initialized, but not the DisableDrain flag.

adrianchiris (approved):
LGTM,

can be merged once CI pass and comments from @ykulazhenkov addressed.

ykulazhenkov (approved):

zeeke (approved):

SchSeba:
Thanks for all the help folks!

merging

PR #791, opened 2024-10-14
Adding webhook 'delete' command to remove non-namespaced objects

heyvister1:
sriov-network-operator dynamically creates `validating/mutating/clusterroles/clusterrolebinding` cluster objects, for specified `SriovOperatorConfig`. These objects are not being cleaned once sriov operator is uninstalled (via Helm/Openshift).

The proposed PR is to use a small go binary within Helm pre-delete hook. This binary will delete `default` SriovOperatorConfig, assuming added SriovOperatorConfig finalizer will make sure sriov controller is deleting all generated cluster level objects (e.g. webhooks), prior to operator deployment object is removed by Helm

### DIFF
--- a/cmd/sriov-network-operator-config-cleanup/cleanup.go
+++ b/cmd/sriov-network-operator-config-cleanup/cleanup.go
@@ -1,0 +1,83 @@
+package main
+
+import (
+	"context"
+	"time"
+
+	"github.com/spf13/cobra"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	snolog "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/log"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/client/clientset/versioned/typed/sriovnetwork/v1"
+)
+
+var (
+	namespace string
+	watchTO   int
+)
+
+func init() {
+	rootCmd.Flags().StringVarP(&namespace, "namespace", "n", "", "designated SriovOperatorConfig namespace")
+	rootCmd.Flags().IntVarP(&watchTO, "watch-timeout", "w", 10, "sriov-operator config post-delete watch timeout ")
+}
+
+func runCleanupCmd(cmd *cobra.Command, args []string) error {
+	// init logger
+	snolog.InitLog()
+	setupLog := log.Log.WithName("sriov-network-operator-config-cleanup")
+	setupLog.Info("Run sriov-network-operator-config-cleanup")
+
+	// adding context timeout although client-go Delete should be non-blocking by default
+	ctx, timeoutFunc := context.WithTimeout(context.Background(), time.Second*time.Duration(watchTO))
+	defer timeoutFunc()
+
+	restConfig := ctrl.GetConfigOrDie()
+	sriovcs, err := sriovnetworkv1.NewForConfig(restConfig)
+	if err != nil {
+		setupLog.Error(err, "failed to create 'sriovnetworkv1' clientset")
+	}
+
+	err = sriovcs.SriovOperatorConfigs(namespace).Delete(context.Background(), "default", metav1.DeleteOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		setupLog.Error(err, "failed to delete SriovOperatorConfig")
+		return err
+	}
+
+	// watching 'default' config deletion with context timeout, in case sriov-operator fails to delete 'default' config
+	watcher, err := sriovcs.SriovOperatorConfigs(namespace).Watch(ctx, metav1.ListOptions{Watch: true})
+	if err != nil {
+		setupLog.Error(err, "failed creating 'default' SriovOperatorConfig object watcher")
+		return err
+	}
+	defer watcher.Stop()
+	for {
+		select {
+		case event := <-watcher.ResultChan():
+			if event.Type == watch.Deleted {
+				setupLog.Info("'default' SriovOperatorConfig is deleted")
+				return nil
+			}
+
+		case <-ctx.Done():
+			// check whether object might has been deleted before watch event triggered
+			_, err := sriovcs.SriovOperatorConfigs(namespace).Get(context.Background(), "default", metav1.GetOptions{})
+			if err != nil {
+				if errors.IsNotFound(err) {
+					return nil
+				}
+			}
+			err = ctx.Err()
+			setupLog.Error(err, "timeout has occurred for 'default' SriovOperatorConfig deletion")
+			return err
+		}
+	}
+}
--- a/cmd/sriov-network-operator-config-cleanup/cleanup_test.go
+++ b/cmd/sriov-network-operator-config-cleanup/cleanup_test.go
@@ -1,0 +1,177 @@
+package main
+
+import (
+	"context"
+	"sync"
+
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/controllers"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/featuregate"
+	mock_platforms "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms/mock"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platforms/openshift"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/test/util"
+)
+
+type configController struct {
+	k8sManager manager.Manager
+	ctx        context.Context
+	cancel     context.CancelFunc
+	wg         *sync.WaitGroup
+}
+
+var (
+	controller               *configController
+	testNamespace            string = "sriov-network-operator"
+	defaultSriovOperatorSpec        = sriovnetworkv1.SriovOperatorConfigSpec{
+		EnableInjector:        true,
+		EnableOperatorWebhook: true,
+		LogLevel:              2,
+		FeatureGates:          nil,
+	}
+)
+
+var _ = Describe("cleanup", Ordered, func() {
+	BeforeAll(func() {
+		By("Create SriovOperatorConfig controller k8s objs")
+		config := getDefaultSriovOperatorConfig()
+		Expect(k8sClient.Create(context.Background(), config)).Should(Succeed())
+
+		somePolicy := &sriovnetworkv1.SriovNetworkNodePolicy{}
+		somePolicy.SetNamespace(testNamespace)
+		somePolicy.SetName("some-policy")
+		somePolicy.Spec = sriovnetworkv1.SriovNetworkNodePolicySpec{
+			NumVfs:       5,
+			NodeSelector: map[string]string{"foo": "bar"},
+			NicSelector:  sriovnetworkv1.SriovNetworkNicSelector{},
+			Priority:     20,
+		}
+		Expect(k8sClient.Create(context.Background(), somePolicy)).ToNot(HaveOccurred())
+		DeferCleanup(func() {
+			err := k8sClient.Delete(context.Background(), somePolicy)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		controller = newConfigController()
+
+	})
+
+	It("test webhook cleanup flow", func() {
+		controller.start()
+		defer controller.stop()
+
+		cmd := &cobra.Command{}
+		namespace = testNamespace
+		// verify that finalizer has been added, by controller, upon object creation
+		config := &sriovnetworkv1.SriovOperatorConfig{}
+		Eventually(func() []string {
+			// wait for SriovOperatorConfig flags to get updated
+			err := k8sClient.Get(context.Background(), types.NamespacedName{Name: "default", Namespace: testNamespace}, config)
+			if err != nil {
+				return nil
+			}
+			return config.Finalizers
+		}, util.APITimeout, util.RetryInterval).Should(Equal([]string{sriovnetworkv1.OPERATORCONFIGFINALIZERNAME}))
+
+		Expect(runCleanupCmd(cmd, []string{})).Should(Succeed())
+		config = &sriovnetworkv1.SriovOperatorConfig{}
+		err := util.WaitForNamespacedObjectDeleted(config, k8sClient, testNamespace, "default", util.RetryInterval, util.APITimeout)
+		Expect(err).NotTo(HaveOccurred())
+
+	})
+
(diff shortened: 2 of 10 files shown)


bindata/manifests/operator-webhook/002-rbac.yaml
```diff
@@ -9,6 +9,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: operator-webhook
+  labels:
```
rollandf, line 12: There are more objects created, do we want to label them also?

heyvister1: That is a valid point. In case others will confim that I'll add it in all generated objects. The intention was to be able to monitor non-namespaced objects created by operator

bindata/manifests/operator-webhook/002-rbac.yaml
```diff
@@ -32,6 +34,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: operator-webhook-role-binding
+  labels:
+    created-by: {{ include "sriov-network-operator.fullname" . }}
```
zeeke, line 38: I'm not an expert of Helm, but I think `sriov-network-operator.fullname` has meaning inside the `deployment/` folder only. `bindata/` folder is rendered by the operator. Am I missing something?

cmd/webhook/delete.go
```diff
@@ -0,0 +1,168 @@
+package main
```
adrianchiris, line 1: why have you decided to implement this part of the same binary as sriov-network-operator webhook ?
this is also deleting (cleaning) objects not related to this webhook. 

my preference are either:
1. new binary
2. new mode for the controller manager binary

bindata/manifests/operator-webhook/002-rbac.yaml
```diff
@@ -9,6 +9,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: operator-webhook
+  labels:
+    created-by: {{ .OperatorGeneratedResourcesLabelSelector | default "sriov-network-operator" }}
```
adrianchiris, line 13: WDYT about using : `kubernetes.io/managed-by` label ? [1]


[1] https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

deployment/sriov-network-operator-chart/templates/pre-delete-webooks.yaml
```diff
@@ -0,0 +1,24 @@
...
+          image: {{ .Values.images.webhook }}
+          command:
+            - webhook
+          args:
+            - delete
+            - --namespace
+            - {{ .Release.Namespace }}
+            - --label-selector
+            - "created-by={{ .Values.operator.admissionControllers.labelSelector | default (include "sriov-network-operator.fullname" .) }}"
+      restartPolicy: Never
```
adrianchiris, line 24: nit: missing new line

controllers/sriovoperatorconfig_controller.go
```diff
@@ -300,6 +300,7 @@ func (r *SriovOperatorConfigReconciler) syncWebhookObjs(ctx context.Context, dc
 		data.Data["OperatorWebhookCA"] = os.Getenv("ADMISSION_CONTROLLERS_CERTIFICATES_OPERATOR_CA_CRT")
 		data.Data["InjectorWebhookSecretName"] = os.Getenv("ADMISSION_CONTROLLERS_CERTIFICATES_INJECTOR_SECRET_NAME")
 		data.Data["InjectorWebhookCA"] = os.Getenv("ADMISSION_CONTROLLERS_CERTIFICATES_INJECTOR_CA_CRT")
+		data.Data["OperatorGeneratedResourcesLabelSelector"] = os.Getenv("OPERATOR_GENERATED_RESOURCES_LABEL_SELECTOR")
```
adrianchiris, line 303: i dont see this env var being set anywhere,  moreover i dont think we need this configurable do we ? 
a const is enough, it should be also used in the cleanup binary instead of providing via flag.

thoughts?

deployment/sriov-network-operator-chart/templates/pre-delete-webooks.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: batch/v1
```
adrianchiris, line 1: nit: WDYT about adding some comments about what this job does?

heyvister1: Sure, I will add more details

deployment/sriov-network-operator-chart/templates/pre-delete-webooks.yaml
adrianchiris: still missing newline

controllers/sriovoperatorconfig_controller.go
```diff
@@ -94,58 +94,89 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
...
 
-	r.FeatureGate.Init(defaultConfig.Spec.FeatureGates)
-	logger.Info("enabled featureGates", "featureGates", r.FeatureGate.String())
+	// examine DeletionTimestamp to determine if object is under deletion
+	if defaultConfig.ObjectMeta.DeletionTimestamp.IsZero() {
 
-	if !defaultConfig.Spec.EnableInjector {
-		logger.Info("SR-IOV Network Resource Injector is disabled.")
-	}
+		if !sriovnetworkv1.StringInArray(sriovnetworkv1.OPERATORCONFIGFINALIZERNAME, defaultConfig.ObjectMeta.Finalizers) {
```
adrianchiris, line 100: this seems overly complex to what we are trying to achieve IMO. plus we are now trying to `r.deleteAllWebhooks()`
twice : L87 and L166

IMO we should:

1. if object not found - do nothing
2. if object found - add finalizer
3. if deletion timestamp is set : call `r.deleteAllWebhooks()` and remove finalizer.
4. do the regular flow as it is today.

controllers/sriovoperatorconfig_controller_test.go
```diff
@@ -240,6 +246,38 @@ var _ = Describe("SriovOperatorConfig controller", Ordered, func() {
...
+			// verify that finalizer has been removed
+			var empty []string
+			config = &sriovnetworkv1.SriovOperatorConfig{}
+			Eventually(func() []string {
+				// wait for SriovOperatorConfig flags to get updated
+				err := k8sClient.Get(context.Background(), types.NamespacedName{Name: "default", Namespace: testNamespace}, config)
+				if err != nil {
+					return nil
+				}
+				return config.Finalizers
```
adrianchiris, line 277: here its enough to make sure that eventually the object no longer exists ?

cmd/sriov-network-operator-config-cleanup/cleanup.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		setupLog.Error(err, "failed initialization k8s rest config")
+	}
+
+	sriovcs, err := sriovnetworkv1.NewForConfig(config)
+	if err != nil {
+		setupLog.Error(err, "failed to create 'sriovnetworkv1' clientset")
+	}
+
+	sriovcs.SriovOperatorConfigs(namespace).Delete(context.Background(), "default", metav1.DeleteOptions{})
```
adrianchiris, line 60: this call is blocking ? or it just sets deletion timestamp ?

cmd/sriov-network-operator-config-cleanup/cleanup_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+var _ = Describe("cleanup", Ordered, func() {
+
+	defaultSriovOperatorConfig := &sriovnetworkv1.SriovOperatorConfig{
+		ObjectMeta: metav1.ObjectMeta{Name: "default", Namespace: testNamespace},
+		Spec:       defaultSriovOperatorSpec,
+	}
+
+	BeforeEach(func() {
+		Expect(k8sClient.Create(context.Background(), defaultSriovOperatorConfig)).NotTo(HaveOccurred())
```
adrianchiris, line 33: set a finalizer on it as well like in real case ?

cmd/sriov-network-operator-config-cleanup/cleanup.go
```diff
@@ -0,0 +1,68 @@
...
+	if kubeconfig != "" {
+		config, err = clientcmd.BuildConfigFromFlags("", kubeconfig)
+	} else {
+		config, err = rest.InClusterConfig()
+	}
+	if err != nil {
+		setupLog.Error(err, "failed initialization k8s rest config")
+	}
+
+	sriovcs, err := sriovnetworkv1.NewForConfig(config)
```
adrianchiris, line 55: nit: maybe use controller runtime client like we do in controllers ? for consistency.

not a must ofc, up2u

controllers/sriovoperatorconfig_controller.go
heyvister1: Ok. I'll remove initial `r.deleteAllWebhooks()` and keep one that is being called during object deletion

controllers/sriovoperatorconfig_controller_test.go
heyvister1: Above test case already verifies that mutating/validating webhooks were `deleted` upon default config deletion

cmd/sriov-network-operator-config-cleanup/cleanup.go
heyvister1: client-go Delete should be non-blocking, kubectl delete is blocking since it has a wait flag set be default
```
    --wait=true:
	If true, wait for resources to be gone before returning. This waits for finalizers.
```

cmd/sriov-network-operator-config-cleanup/cleanup_test.go
heyvister1: Ok. For this case I'll need to add mocked controller to verify that config object is deleted correctly

cmd/sriov-network-operator-config-cleanup/cleanup.go
heyvister1: done

controllers/sriovoperatorconfig_controller.go
```diff
@@ -94,58 +92,88 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			// remove our finalizer from the list and update it.
+			var found bool
+			defaultConfig.ObjectMeta.Finalizers, found = sriovnetworkv1.RemoveString(sriovnetworkv1.OPERATORCONFIGFINALIZERNAME, defaultConfig.ObjectMeta.Finalizers)
+			if found {
+				if err := r.Update(ctx, defaultConfig); err != nil {
+					return reconcile.Result{}, err
+				}
+			}
+		}
+		return reconcile.Result{}, err
```
zeeke, line 176: can you put this logic in a function 
```suggestion
		return handleSriovOperatorConfigDeletion()
```
and flip the `if defaultConfig.ObjectMeta.DeletionTimestamp.IsZero() {` condition to have an early exit?

deployment/sriov-network-operator-chart/templates/pre-delete-webooks.yaml
```diff
@@ -0,0 +1,25 @@
...
+    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
+spec:
+  template:
+    spec:
+      serviceAccountName: {{ include "sriov-network-operator.fullname" . }}
+      containers:
+        - name: cleanup
+          image: {{ .Values.images.operator }}
+          command:
+            - sriov-network-operator-config-cleanup
```
zeeke, line 21: optional: what about replacing the whole `sriov-network-operator-config-cleanup` command and files with

```
kubectl delete -n {{ .Release.Namespace }} sriovoperatorconfig/default
```

heyvister1: This was the initial intention. But we had issues with kubectl CVEs in our CI jobs, therefore we had decided to use a cmall client-go binary instead

controllers/sriovoperatorconfig_controller.go
```diff
@@ -94,58 +93,71 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
...
-	logger.Info("enabled featureGates", "featureGates", r.FeatureGate.String())
-
-	if !defaultConfig.Spec.EnableInjector {
-		logger.Info("SR-IOV Network Resource Injector is disabled.")
-	}
+	// examine DeletionTimestamp to determine if object is under deletion
+	if !defaultConfig.ObjectMeta.DeletionTimestamp.IsZero() {
+		// The object is being deleted
+		return r.handleSriovOperatorConfigDeletion(ctx, defaultConfig, logger)
+	} else {
```
zeeke, line 100: Please remove the `else` statement

cmd/sriov-network-operator-config-cleanup/cleanup.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		setupLog.Error(err, "failed to delete SriovOperatorConfig")
+		return err
+	}
+
+	// watching 'default' config deletion with context timeout, in case sriov-operator fails to delete 'default' config
+	watcher, err := sriovcs.SriovOperatorConfigs(namespace).Watch(ctx, metav1.ListOptions{Watch: true})
```
adrianchiris, line 54: there is a (probably slim) chance that obj will no longer exist when executing this line.
in this case the code below will timeout.

maybe create watcher before delete ? 
or just use good old fashioned polling ?

heyvister1: I can also check for non-exist error and exit then
```
		if errors.IsNotFound(err) {
			return nil
		}
```

controllers/sriovoperatorconfig_controller.go
```diff
@@ -434,6 +446,31 @@ func (r *SriovOperatorConfigReconciler) syncOpenShiftSystemdService(ctx context.
...
+		logger.Info("delete SriovOperatorConfig CR", "Namespace", defaultConfig.Namespace, "Name", defaultConfig.Name)
+		// make sure webhooks objects are deleted prior of removing finalizer
+		err = r.deleteAllWebhooks(ctx)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		// remove our finalizer from the list and update it.
+		var found bool
+		defaultConfig.ObjectMeta.Finalizers, found = sriovnetworkv1.RemoveString(sriovnetworkv1.OPERATORCONFIGFINALIZERNAME, defaultConfig.ObjectMeta.Finalizers)
+		if found {
```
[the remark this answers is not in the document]
adrianchiris: if statement at L453 checks for : `sriovnetworkv1.StringInArray(sriovnetworkv1.OPERATORCONFIGFINALIZERNAME, defaultConfig.ObjectMeta.Finalizers)`

what am i missing ?

cmd/sriov-network-operator-config-cleanup/cleanup.go
adrianchiris: works for me :)

controllers/sriovoperatorconfig_controller.go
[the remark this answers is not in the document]
heyvister1: you're right, my bad

cmd/sriov-network-operator-config-cleanup/cleanup.go
```diff
@@ -0,0 +1,77 @@
...
+	defer watcher.Stop()
```
adrianchiris, line 55: if err is not nil, is it safe to stop the watcher ? or should the defer be moved after the error check ?

adrianchiris (commented):
overall LGTM, need CI to be green.

added one last comment from my side , thx for the patience ! :)
before  we merge please re-work the commits so they do not reflect the review interations but rather the changes done.

e.g 

commit 1 : add finalizer to sriovoperatorconfig
commit 2: cleanup binary related changes

adrianchiris (approved):

adrianchiris (approved):

heyvister1:
@zeeke could you please add your approval for this PR?

zeeke (approved):

PR #792, opened 2024-10-14
Clean Systemd files on exit

souleb:
We have run into an issue where removing the operator leaves behind the systemd files.

If implemented all systemD related files will be cleaned up on exit on non `openshift` distributions.

### DIFF
--- a/cmd/sriov-network-config-daemon/start.go
+++ b/cmd/sriov-network-config-daemon/start.go
@@ -21,7 +21,9 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -134,7 +136,6 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
 
 	// This channel is used to ensure all spawned goroutines exit when we exit.
 	stopCh := make(chan struct{})
-	defer close(stopCh)
 
 	// This channel is used to signal Run() something failed and to jump ship.
 	// It's purely a chan<- in the Daemon struct for goroutines to write to, and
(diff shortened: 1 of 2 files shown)


cmd/sriov-network-config-daemon/start.go
```diff
@@ -294,25 +296,47 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
...
-	if err != nil {
-		setupLog.Error(err, "failed to run daemon")
+
+	// create a signal channel to catch interrupts
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, os.Interrupt)
+	signal.Notify(sigc, syscall.SIGTERM)
+
+	errChan := make(chan error)
+	defer close(errChan)
```
souleb, line 306: I used the method used here to close channels. But I believe there is a race in general, on exit (once main goes away, or if os.Exit is called) the goroutines are killed off immediately. So the goroutines listening on this channel will never get the chance to run their intended actions.

souleb:
cc @e0ne

pkg/daemon/daemon.go
```diff
@@ -207,6 +207,14 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 	for {
 		select {
 		case <-stopCh:
+			// clean files from host if we are running in systemd mode
+			if vars.UsingSystemdMode {
+				err := systemd.CleanSriovFilesFromHost(vars.ClusterType == consts.ClusterTypeOpenshift)
```
adrianchiris, line 212: so when the daemon is stopped we delete systemd files ?

if the node reboots (given pods get stop signal) wont it delete the files ? which will in turn mean the node will not have SR-IOV configured on startup  ?

souleb: What you say is correct. We should have a way to identify whether a `reboot` has been requested. In that case the cleanUp has to be a best effort because we can have an ongoing reboot and also a `spec` change in the `daemonSet`.

PR #798, opened 2024-10-28
feat: Update controller logic to handle stale SriovNetworkNodeState CRs with delay

ykulazhenkov:
Update controller logic to handle stale SriovNetworkNodeState CRs with delay

- Changed the logic in the sriov-network-operator controller to handle stale SriovNetworkNodeState CRs (those with no matching Nodes with daemon).
- Introduced a delay (30 minutes by default) before removing stale state CRs to manage scenarios where the user temporarily removes the daemon from the node but does not want to lose the state stored in the SriovNetworkNodeState.
- Added the `STALE_NODE_STATE_CLEANUP_DELAY_MINUTES` environment variable to configure the required delay in minutes (default is 30 minutes).

This functionality especially useful when the OFED container is in use. As the OFED driver loads on the host, the sriov-config-daemon is removed from this node (achieved using configDaemon nodeselector). Since loading the driver can take a considerable amount of time, we want to ensure that the SriovNetworkNodeState is not lost during this process.

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -20,8 +20,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -308,6 +307,41 @@ func (r *SriovNetworkNodePolicyReconciler) syncAllSriovNetworkNodeStates(ctx con
...
+// If the annotation is absent, the function will create one with a timestamp in future, using either the default or a configured offset.
+func (r *SriovNetworkNodePolicyReconciler) handleStaleNodeState(ctx context.Context, ns *sriovnetworkv1.SriovNetworkNodeState) error {
+	logger := log.Log.WithName("handleStaleNodeState")
+	now := time.Now().UTC()
+	keepUntilTime := ns.GetKeepUntilTime()
+	if keepUntilTime.IsZero() {
+		// keep until time annotation is not set, set a new value with default or configured offset and update the object
+		delayMinutes, err := strconv.Atoi(os.Getenv("STALE_NODE_STATE_CLEANUP_DELAY"))
+		if err != nil || delayMinutes <= 0 {
+			delayMinutes = 30 // keep objects for 30 minutes by default
```
e0ne, line 322: Let's add some logging here to indicate env variable is incorrect

e0ne: nit: we've got unused CleanupTimeout [1] variable. maybe we need to change it's value and use it here

[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/68b6c020fbba1596ddd3f2e4a70cfcdb034b7e28/test/util/util.go#L34

e0ne (changes requested):

adrianchiris (commented):
overall lgtm from my side, added small comment. 

once Ivan's comment addressed (we should use a constant IMO) im LGTM.

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -296,10 +298,7 @@ func (r *SriovNetworkNodePolicyReconciler) syncAllSriovNetworkNodeStates(ctx con
 				}
 			}
 			if !found {
-				logger.Info("Deleting SriovNetworkNodeState as node with that name doesn't exist", "nodeStateName", ns.Name)
-				err := r.Delete(ctx, &ns, &client.DeleteOptions{})
-				if err != nil {
-					logger.Error(err, "Fail to Delete", "SriovNetworkNodeState CR:", ns.GetName())
+				if err := r.handleStaleNodeState(ctx, &ns); err != nil {
```
adrianchiris, line 301: note to self: this is being called every ~5 min (resync period) worst case or when policy is updated or when policy/node changed

ykulazhenkov:
CI failure is not related to the change. The same failure occurs on the PR with dummy changes https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/800

controllers/sriovnetworknodepolicy_controller.go
ykulazhenkov: I added constant

ykulazhenkov:
@e0ne @adrianchiris  I addressed your comments. I also changed behavior a bit to completely avoid any delay in case if `STALE_NODE_STATE_CLEANUP_DELAY_MINUTES` env is explicitly set to `0`

e0ne (approved):
LGTM. Thanks for addressing my comments

adrianchiris (approved):
LGTM

adrianchiris:
@SchSeba PTAL on this one

SchSeba (approved):
LGTM

nice work!

PR #803, opened 2024-11-07
Fix MCP pause case and re-organize drain controller package

SchSeba:


controllers/helper_test.go
```diff
@@ -1,330 +0,0 @@
-/*
```
zeeke, line 1: Why remove this file?

controllers/drain_controller.go
```diff
@@ -250,93 +169,17 @@ func (dr *DrainReconcile) getObject(ctx context.Context, req ctrl.Request, objec
 	return true, nil
 }
 
-func (dr *DrainReconcile) ensureAnnotationExists(ctx context.Context, object client.Object, key string) (string, error) {
+func (dr *DrainReconcile) ensureAnnotationExists(ctx context.Context, object client.Object, key string) (string, bool, error) {
 	value, exist := object.GetAnnotations()[key]
 	if !exist {
 		err := utils.AnnotateObject(ctx, object, constants.NodeStateDrainAnnotationCurrent, constants.DrainIdle, dr.Client)
```
zeeke, line 175: this should be 
```
err := utils.AnnotateObject(ctx, object, key, constants.DrainIdle, dr.Client)
```
right?

zeeke (commented):
Just a few minor comments, otherwise LGTM

ykulazhenkov (commented):
I checked the PR and I don't have meaningful comments. I think we are good to merge it when comments from Andrea are addressed.

controllers/helper_test.go
SchSeba: this was related to the device plugin one I rebase this PR

controllers/drain_controller.go
SchSeba: right nice catch!

zeeke (approved):
LGTM!

PR #805, opened 2024-11-08
fix: Device Plugin daemonset keep restarting

### LINKED ISSUE #804: Device Plugin daemonset keep restarting
When creating multiple `SriovNetworkNodePolicy` with different `nodeSelector` we observe that the `sriov-device-plugin` daemonset is unstable. After some investigation we found that the root cause is that the operator keep changing the order of `nodeSelectorTerms` in the daemonset.

armandpicard:
Define an ordering of the NodeSelectorRequirement in the daemonset to avoid the daemonset to be unstable when policies come in different order.

Fixes #804

### DIFF
--- a/controllers/helper.go
+++ b/controllers/helper.go
@@ -326,7 +326,7 @@ func nodeSelectorTermsForPolicyList(policies []sriovnetworkv1.SriovNetworkNodePo
 				return expressionA.Key < expressionB.Key
 			}
 			if expressionA.Values[0] != expressionB.Values[0] {
-				return expressionA.Values[0] != expressionB.Values[0]
+				return expressionA.Values[0] < expressionB.Values[0]
 			}
 		}
 		return true
--- a/controllers/helper.go
+++ b/controllers/helper.go
@@ -311,6 +311,27 @@ func nodeSelectorTermsForPolicyList(policies []sriovnetworkv1.SriovNetworkNodePo
 		terms = append(terms, nodeSelector)
 	}
 
+	// sorting is needed to keep the daemon spec stable.
+	// the policies can come in a random order
+	sort.Slice(terms, func(i, j int) bool {
+		expressionsA := terms[i].MatchExpressions
+		expressionsB := terms[j].MatchExpressions
+		if len(expressionsA) != len(expressionsB) {
+			return len(expressionsA) < len(expressionsB)
+		}
+		for k := range expressionsA {
+			expressionA := expressionsA[k]
+			expressionB := expressionsB[k]
+			if expressionA.Key != expressionB.Key {
+				return expressionA.Key < expressionB.Key
+			}
+			if expressionA.Values[0] != expressionB.Values[0] {
+				return expressionA.Values[0] != expressionB.Values[0]
+			}
+		}
+		return true
+	})
+
 	return terms
 }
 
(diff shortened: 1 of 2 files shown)


controllers/helper.go
```diff
@@ -311,6 +311,27 @@ func nodeSelectorTermsForPolicyList(policies []sriovnetworkv1.SriovNetworkNodePo
...
+			return len(expressionsA) < len(expressionsB)
+		}
+		for k := range expressionsA {
+			expressionA := expressionsA[k]
+			expressionB := expressionsB[k]
+			if expressionA.Key != expressionB.Key {
+				return expressionA.Key < expressionB.Key
+			}
+			if expressionA.Values[0] != expressionB.Values[0] {
+				return expressionA.Values[0] != expressionB.Values[0]
```
heyvister1, line 329:
```suggestion
				return expressionA.Values[0] < expressionB.Values[0]
```
Function returned bool value should indicate lower value and not that they are different. You can add the following test case to check it:
```
	terms := []corev1.NodeSelectorTerm{
		{MatchExpressions: []corev1.NodeSelectorRequirement{
			{Key: "env", Values: []string{"production"}},
		}},
		{MatchExpressions: []corev1.NodeSelectorRequirement{
			{Key: "env", Values: []string{"staging"}},
		}},
		{MatchExpressions: []corev1.NodeSelectorRequirement{
			{Key: "env", Values: []string{"dev"}},
		}},
	}
```

Output would be: `dev, staging, production`

heyvister1: I also suggest to verify empty values, as it can lead to out of order index
```
            if expressionA.Key != expressionB.Key {
                return expressionA.Key < expressionB.Key
            }

            if len(expressionA.Values) == 0 {
                return true // Empty is considered less than non-empty
            }
            if len(expressionB.Values) == 0 {
                return false // Non-empty is considered greater than empty
            }
```

armandpicard: Thank for you're review ! I'll add the test case for ordering base on value.
I think we expect to always have 1 value as it's create just above, that's why I was comparing only the first values.

ykulazhenkov:
Hi, @armandpicard. Thanks for the PR.

I think these sorts [1,2] in the controllers code should keep the order stable without the changes from this PR.
[1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/controllers/sriovoperatorconfig_controller.go#L131

[2] https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/controllers/sriovnetworknodepolicy_controller.go#L127 

**Note:** ByPriority uses Priority + Name of the CR https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/92fee7bec1c14b9a018296da1119e4497a7cdda1/api/v1/helper.go#L347


The https://github.com/k8snetworkplumbingwg/sriov-network-operator/commit/91cf86cb238cde0fe9b10380f3728e54e80b5bc2 (the one that introduced [1]) was added pretty recently.

I think the issue with the unstable order of the selectors should not exist in the master branch.
Which version of the operator are you using?

armandpicard:
Hi @ykulazhenkov.
I'm not 100% sure of the exact version we use as in our case srio-network-operator is deployed by mellanox network-operator(24.7.0). By the look of the bindata in the container it is at least 8 mouth old so it doesn't have the fix.
I'll close the MR and the Issue as sriov-network-operator from master was able to get the daemonset stable with our  SrioNetworkNodePolicy's

PR #806, opened 2024-11-14
deploy: relax Operator node affinity

EmilienM:
In the context of Hypershift (Hosted Clusters with OpenShift), where a Nodepool
(terminology for a worker Node in HCP) is not a control-plane or
a master Node but a worker, we can't force the Operator to be deployed
on a master node that doesn't exist. Instead, we want to deploy it on a
worker.

The proposal here is to relax the rule and use
`preferredDuringSchedulingIgnoredDuringExecution` instead so the scheduler
will try to find a master node or fallback on other nodes if not found.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -20,14 +20,22 @@ spec:
     spec:
       affinity:
         nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-            - matchExpressions:
-              - key: node-role.kubernetes.io/master
-                operator: Exists
-            - matchExpressions:
-                - key: node-role.kubernetes.io/control-plane
-                  operator: Exists
+          # In the context of Hypershift, the SR-IOV network
+          # Operator is deployed on Nodepools which are labeled
+          # as workers. So we relax the node affinity to prefer
+          # masters/control-plane when possible otherwise we
+          # schedule where it's possible.
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - weight: 1
+              preference:
+                matchExpressions:
+                  - key: "node-role.kubernetes.io/master"
+                    operator: Exists
+            - weight: 1
+              preference:
+                matchExpressions:
+                  - key: "node-role.kubernetes.io/control-plane"
+                    operator: Exists
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master


SchSeba (approved):
LGTM

nice work!

wizhaoredhat:
LGTM!

SchSeba:
Hi @EmilienM can you take a quick look on the failed in the ocp lane please

SchSeba:
please take a look on the helm deployment

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/aecb4bb904f21c3da98e0163c461c0dc940183b9/deployment/sriov-network-operator-chart/values.yaml#L10-L24

deploy/operator.yaml
```diff
@@ -20,14 +20,26 @@ spec:
...
         nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-            - matchExpressions:
-              - key: node-role.kubernetes.io/master
-                operator: Exists
-            - matchExpressions:
-                - key: node-role.kubernetes.io/control-plane
-                  operator: Exists
+          # In the context of Hypershift, the SR-IOV network
```
EmilienM, line 23: this matches what we do in Helm charts already.

deploy/operator.yaml
```diff
@@ -20,14 +20,25 @@ spec:
...
+                  - key: "node-role.kubernetes.io/master"
+                    operator: In
+                    values: [""]
+            - weight: 1
+              preference:
+                matchExpressions:
+                  - key: "node-role.kubernetes.io/control-plane"
+                    operator: In
+                    values: [""]
+  nameOverride: ""
```
SchSeba, line 41: I don't think this line is right

deploy/operator.yaml
```diff
@@ -20,14 +20,25 @@ spec:
...
+              preference:
+                matchExpressions:
+                  - key: "node-role.kubernetes.io/master"
+                    operator: In
+                    values: [""]
+            - weight: 1
+              preference:
+                matchExpressions:
+                  - key: "node-role.kubernetes.io/control-plane"
+                    operator: In
```
SchSeba, line 39: I think here you can use Exist and remove the values

EmilienM: This was copied from https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/92fee7bec1c14b9a018296da1119e4497a7cdda1/deployment/sriov-network-operator-chart/values.yaml#L11-L24

EmilienM: I applied your suggestion though.

SchSeba (approved):

PR #807, opened 2024-11-14
Avoid using iface.Name from spec during configuration.

ykulazhenkov:
This ensures proper handling when the interface name changed for some reason.


SchSeba:
Hi @ykulazhenkov I saw you closed this one. do we still need this?

PR #808, opened 2024-11-21
Add waitForDevicesInitialization to systemd service

ykulazhenkov:
Add waitForDevicesInitialization to systemd service.

This function ensures that the network devices specified in the configuration are registered and handled by UDEV. Sometimes, the initialization of network devices might take a significant amount of time, and the sriov-config systemd service may start before the devices are fully processed, leading to failure.

### DIFF
--- a/pkg/host/internal/udev/udev.go
+++ b/pkg/host/internal/udev/udev.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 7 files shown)


e0ne (approved):
LGTM

ykulazhenkov:
@SchSeba @zeeke Please, take a look on this PR

pkg/host/internal/udev/udev.go
```diff
@@ -126,6 +127,18 @@ func (u *udev) LoadUdevRules() error {
 	return nil
 }
 
+// WaitUdevEventsProcessed calls `udevadm settle“ with provided timeout
+// The command watches the udev event queue, and exits if all current events are handled.
+func (u *udev) WaitUdevEventsProcessed(timeout int) error {
+	log.Log.V(2).Info("WaitUdevEventsProcessed()")
+	_, stderr, err := u.utilsHelper.RunCommand("udevadm", "settle", "-t", strconv.Itoa(timeout))
```
zeeke, line 134: Is this supposed to run in the host chroot? I tried running it in the config-daemon pod:
```
[root@cnfdr22 /]# udevadm settle -t 10
Running in chroot, ignoring request.
```

zeeke: Got it, I wrongly thought the systemd service was running the config-daemon via `podman run ...` .
thanks for explaining

zeeke (approved):
LGTM

PR #812, opened 2024-12-04
e2e: Avoid setting wrong routes for `host-local` IPAM

zeeke:
The IPAM configuration:

```
{
"type":"host-local",
"subnet":"10.10.10.0{"type":"host-local",
"subnet":"10.10.10.0/24",
"rangeStart":"10.10.10.171",
"rangeEnd":"10.10.10.181",
"routes":[{"dst":"0.0.0.0/0"}],
"gateway":"10.10.10.1"
}
```

Can lead to the following pod `ip route` configuration:
```
default via 10.10.10.1 dev net1
default via 10.128.0.1 dev eth0
10.10.10.0/24 dev net1 proto kernel scope link src 10.10.10.172
10.128.0.0/23 dev eth0 proto kernel scope link src 10.128.0.135
10.128.0.0/14 via 10.128.0.1 dev eth0
```

which causes connectivity issues.

Avoid setting default routes to unknown gateways.

### DIFF
--- a/test/conformance/tests/test_networkpool.go
+++ b/test/conformance/tests/test_networkpool.go
@@ -181,7 +181,7 @@ var _ = Describe("[sriov] NetworkPool", Ordered, func() {
 				},
 				Spec: sriovv1.SriovNetworkSpec{
 					ResourceName:      resourceName,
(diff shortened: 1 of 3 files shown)


ykulazhenkov (approved):
LGTM, thanks for fixing this

PR #816, opened 2024-12-13
Do not configure BlueField NICs in DPU mode

e0ne:


pkg/vendors/mellanox/mellanox.go
```diff
@@ -162,6 +163,14 @@ func (m *mellanoxHelper) MlxResetFW(pciAddresses []string) error {
 func (m *mellanoxHelper) MlxConfigFW(attributesToChange map[string]MlxNic) error {
 	log.Log.Info("mellanox-plugin configFW()")
 	for pciAddr, fwArgs := range attributesToChange {
+		bfMode, err := m.GetMellanoxBlueFieldMode(pciAddr)
+		if err != nil {
+			// NIC is not a DPU or mstconfig failed. It's safe to continue FW configuration
+			log.Log.V(2).Info("mellanox-plugin: configFW(): can't get DPU mode for NIC", "pciAddress", pciAddr)
+		}
+		if bfMode == BluefieldDpu {
+			return errors.Errorf("NIC %s is in DPU mode", pciAddr)
```
adrianchiris, line 172: can you elaborate that firmware configuration changes are not supported in this case ?
also add a comment saying why (because reboot will not re-load firmware)

SchSeba: we need to document that carefully as we have the fwreset feature we implement in the operator but that is not to switch the BF card mode

e0ne: done

adrianchiris (approved):

pkg/vendors/mellanox/mellanox.go
```diff
@@ -162,6 +163,15 @@ func (m *mellanoxHelper) MlxResetFW(pciAddresses []string) error {
 func (m *mellanoxHelper) MlxConfigFW(attributesToChange map[string]MlxNic) error {
 	log.Log.Info("mellanox-plugin configFW()")
 	for pciAddr, fwArgs := range attributesToChange {
+		bfMode, err := m.GetMellanoxBlueFieldMode(pciAddr)
+		if err != nil {
+			// NIC is not a DPU or mstconfig failed. It's safe to continue FW configuration
+			log.Log.V(2).Info("mellanox-plugin: configFW(): can't get DPU mode for NIC", "pciAddress", pciAddr)
+		}
+		if bfMode == BluefieldDpu {
+			// Host reboot won't re-load NIC firmware in DPU mode. To apply FW changes power cycle is required or mstfwreset could be used.
```
SchSeba, line 172: base on this comment should we allow the configuration if the fwreset feature gate is enabled?

e0ne: IMO, it should be done as a separate effort. We need to fix a bug now to prevent infinite boot loop

SchSeba (approved):

PR #818, opened 2024-12-17
Skip kernel parameters configuration for Ubuntu

e0ne:
It's enouph to configure ib_core module in /etc/moprobe.d/ for Ubuntu OS to change RDMA subsystem mode.

Also this commit fixes kargs.sh error because 'grubby' isn't available in official Ubuntu repositories.

### DIFF
--- a/bindata/scripts/kargs.sh
+++ b/bindata/scripts/kargs.sh
@@ -7,6 +7,14 @@ declare -a kargs=( "$@" )
 ret=0
 args=$(chroot /host/ cat /proc/cmdline)
 
+IS_OS_UBUNTU=true; [[ "$(chroot /host/ grep -i ubuntu /etc/os-release -c)" == "0" ]] && IS_OS_UBUNTU=false
+
+# Kernel args configuration isn't supported for Ubuntu now, so we shouldn't do anything here
+if ${IS_OS_UBUNTU} ; then
+    echo $ret
+    exit 0
+fi
+
 if chroot /host/ test -f /run/ostree-booted ; then
     for t in "${kargs[@]}";do
         if [[ $command == "add" ]];then
--- a/test/scripts/kargs_test.sh
+++ b/test/scripts/kargs_test.sh
@@ -6,6 +6,7 @@ SUT_SCRIPT="${SCRIPTPATH}/../../bindata/scripts/kargs.sh"
 
 
 test_RpmOstree_Add_All_Arguments() {
+    echo "ID=\"rhel\"" > ${FAKE_HOST}/etc/os-release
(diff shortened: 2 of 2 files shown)


bindata/scripts/kargs.sh
```diff
@@ -7,49 +7,53 @@ declare -a kargs=( "$@" )
 ret=0
 args=$(chroot /host/ cat /proc/cmdline)
 
-if chroot /host/ test -f /run/ostree-booted ; then
-    for t in "${kargs[@]}";do
-        if [[ $command == "add" ]];then
-          if [[ $args != *${t}* ]];then
-              if chroot /host/ rpm-ostree kargs | grep -vq ${t}; then
-                  chroot /host/ rpm-ostree kargs --append ${t} > /dev/null 2>&1
+IS_OS_UBUNTU=true; [[ "$(chroot /host/ grep -i ubuntu /etc/os-release -c)" == "0" ]] && IS_OS_UBUNTU=false
```
adrianchiris, line 10: can you add a comment that applying kargs for ubuntu is not supported and currently we do a no-op ?

adrianchiris (approved):

bindata/scripts/kargs.sh
e0ne: done

bindata/scripts/kargs.sh
```diff
@@ -7,49 +7,54 @@ declare -a kargs=( "$@" )
...
+
+# Kernerl args comfiguration isn't supported for Ubuntu now, so we shouldn't do anything here
+if ! ${IS_OS_UBUNTU} ; then
```
SchSeba, line 13: Hi let's make this one simpler, please.

can you do something like

```
if ${IS_OS_UBUNTU} ; then
  echo 0
  exit 0
fi
```

SchSeba (commented):
just a small simplification to the script and also please add a note/section in the readme that for ubuntu systems the user must configure the kernel args manually base on the list we have in the consts

```
	KernelArgPciRealloc       = "pci=realloc"
	KernelArgIntelIommu       = "intel_iommu=on"
	KernelArgIommuPt          = "iommu=pt"
	KernelArgIommuPassthrough = "iommu.passthrough=1"
	KernelArgRdmaShared       = "ib_core.netns_mode=1"
	KernelArgRdmaExclusive    = "ib_core.netns_mode=0"
```

bindata/scripts/kargs.sh
e0ne: done

SchSeba (approved):
LGTM

PR #819, opened 2024-12-19
[th/hack-env-check] hack/env.sh: move checking of environment variables outside SKIP_VAR_SET block

thom311:
When  "$SKIP_VAR_SET" is unset and the environment variables fallback to the default, the check for valid values should be done. Move the check out of the $SKIP_VAR_SET block for that.

For the current "hack/env.sh" this maybe not make an actual difference, because probably the code to assign default values will ensure that always valid value are set. Note that the openshift variant of the above code will detect the default via skopeo, which can fail. For that reason, this change makes more sense for openshift. However, also for the current code, performing the same error checking after filling out default values, ensures that the detected values are considered valid Even if that is in fact always the case, it's not entirely trivial to see.

---

A follow up will later be for openshift [here](https://github.com/openshift/sriov-network-operator/blob/518b0cc1cb3699d94a13533afef7247929c35638/hack/env.sh#L9).

Code like
```
        CNI_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-cni | jq --raw-output '.Digest')
        export SRIOV_CNI_IMAGE=${SRIOV_CNI_IMAGE:-quay.io/openshift/origin-sriov-cni@${CNI_IMAGE_DIGEST}}
```
is wrong, because `skopeo inspect docker://quay.io/openshift/origin-sriov-cni` command will fail. The first problem is that that failure will fail the entire script (`bash -e`). The second problem is that the detected value is clearly invalid. This should be dealt with.

However, I think the bash script should first also move the error checking out of the `if [ -z $SKIP_VAR_SET ]; then ... else` block.


SchSeba (approved):
LGTM and the CI is clean :)

thom311:
@SchSeba hi. Is something missing/wrong?

It's because deploy on openshift/sriov-network-operator has issues, and this patch would be the basis for addressing that.

adrianchiris (approved):

PR #821, opened 2024-12-23
add link state enable on test

SchSeba:
if we run on a system where the PF is not connected to the network we can still use it for tests but we need the link state to not be auto.


SchSeba:
@adrianchiris @zeeke please take a look quick fix for our internal CI system when the nics are not all connected to the network

adrianchiris (approved):

SchSeba:
Thanks @adrianchiris merging this one

PR #825, opened 2024-12-31
Rdma functional tests improvements

SchSeba:
This PR makes the RDMA functional tests more robust specially on single node environments where the API server will not be available during system restart


zeeke (approved):
LGTM

adrianchiris (approved):

PR #828, opened 2025-01-08
Support mtu_request for OVS

rollandf:
For using non-default MTU, OVS supports "mtu_request" field when adding a port to the bridge.

eg:
https://docs.openvswitch.org/en/latest/topics/dpdk/jumbo-frames/


rollandf:
@ykulazhenkov PTAL

ykulazhenkov (approved):
LGTM

rollandf:
@SchSeba @zeeke PTAL when you have time

zeeke (approved):
LGTM

PR #830, opened 2025-01-12
ovs: add internal interface

rollandf:
When creating a bridge with ovs-vsctl, an internal interface is added by default.
The same behavior is added in this commit

ovs-vsctl code ref:
https://github.com/openvswitch/ovs/blob/main/utilities/ovs-vsctl.c#L1597

### DIFF
--- a/pkg/host/internal/bridge/ovs/ovs_test.go
+++ b/pkg/host/internal/bridge/ovs/ovs_test.go
@@ -141,26 +141,43 @@ func createInitialDBContent(ctx context.Context, c client.Client, expectedState 
 func validateDBConfig(dbContent *testDBEntries, conf *sriovnetworkv1.OVSConfigExt) {
 	Expect(dbContent.OpenVSwitch).To(HaveLen(1))
 	Expect(dbContent.Bridge).To(HaveLen(1))
-	Expect(dbContent.Interface).To(HaveLen(1))
-	Expect(dbContent.Port).To(HaveLen(1))
+	Expect(dbContent.Interface).To(HaveLen(2))
+	Expect(dbContent.Port).To(HaveLen(2))
 	ovs := dbContent.OpenVSwitch[0]
 	br := dbContent.Bridge[0]
-	port := dbContent.Port[0]
-	iface := dbContent.Interface[0]
+	ports := make(map[string]*PortEntry, 0)
+	interfaces := make(map[string]*InterfaceEntry, 0)
+	for _, p := range dbContent.Port {
+		ports[p.Name] = p
+	}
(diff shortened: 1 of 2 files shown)


rollandf:
@ykulazhenkov PTAL

pkg/host/internal/bridge/ovs/ovs_test.go
```diff
@@ -137,12 +137,26 @@ func createInitialDBContent(ctx context.Context, c client.Client, expectedState
...
 	Expect(dbContent.Bridge).To(HaveLen(1))
-	Expect(dbContent.Interface).To(HaveLen(1))
-	Expect(dbContent.Port).To(HaveLen(1))
+	Expect(dbContent.Interface).To(HaveLen(2))
+	Expect(dbContent.Port).To(HaveLen(2))
 	ovs := dbContent.OpenVSwitch[0]
 	br := dbContent.Bridge[0]
-	port := dbContent.Port[0]
-	iface := dbContent.Interface[0]
+	var internalPort, port *PortEntry
```
adrianchiris, line 144: nit: i would create a port and interface map keyed by name.
that way if we have more later, there would be no need to do additional if conditions as below.

adrianchiris (commented):
overall LGTM.

im OK with not having api change for it and let that be the default.
later on we can change if the use-case rises (e.g need to enable/disable creation of internal port or adding some external ids to it)

@rollandf what would happen to existing bridges created that dont have internal port set ? will they be reconciled ? i think no with current logic.
do we want them to be reconciled ?

ykulazhenkov:
> do we want them to be reconciled ?

I think it is ok not to reconcile internal ports for now. If we will add such reconciliation, then all existing bridges will need to be reconfigured, meaning all nodes will be drained without any chance to opt out reconfiguration.

adrianchiris:
> I think it is ok not to reconcile internal ports for now.

agree, since adding internal port will not hinder current use-cases.

pkg/host/internal/bridge/ovs/ovs_test.go
rollandf: Done

ykulazhenkov (commented):
@rollandf 
looks good to me.

The PR currently has a merge conflict

rollandf:
> @rollandf looks good to me.
> 
> The PR currently has a merge conflict

Fixed conflict

adrianchiris (approved):
LGTM

ykulazhenkov (approved):

rollandf:
@zeeke @SchSeba PTAL

zeeke (approved):
LGTM

PR #840, opened 2025-02-09
Support draining DaemonSet pods using sriov devices

SchSeba:
with this commit we also take care of removing DaemonSet owned pods using sriov devices.

we only do it when drain is requested we don't do it for reboot requests


pkg/drain/drainer.go
```diff
@@ -131,6 +145,28 @@ func (d *Drainer) CompleteDrainNode(ctx context.Context, node *corev1.Node) (boo
 	return completed, nil
 }
 
+// removeDaemonSetsFromNode go over all the remain pods and search for DaemonSets that have SR-IOV devices to remove them
```
adrianchiris, line 148: so setting `IgnoreAllDaemonSets: false`  in drain.Helper, will delete/evict daemonset pods ?

pkg/drain/drainer.go
```diff
@@ -85,12 +86,25 @@ func (d *Drainer) DrainNode(ctx context.Context, node *corev1.Node, fullNodeDrai
...
-		if err == nil {
+		if err != nil {
+			lastErr = err
+			reqLogger.Info("drainNode(): Draining failed, retrying", "error", err)
+			return false, nil
+		}
+
+		// on full drain there is no need to try and remove pods that are owned by DaemonSets
+		// as we are going to reboot the node in any case.
+		if fullNodeDrain {
```
adrianchiris, line 97: do we care ?  (just thinking how to simplify)  why not always remove DS pods from node ?

test/conformance/tests/test_sriov_operator.go
```diff
@@ -2376,6 +2538,18 @@ func waitForPodRunning(p *corev1.Pod) *corev1.Pod {
 	return ret
 }
 
+func waitForDaemonReady(d *appsv1.DaemonSet) *appsv1.DaemonSet {
```
ykulazhenkov, line 2541: nit: waitForDaemonSetReady

pkg/drain/drainer.go
```diff
@@ -85,12 +86,19 @@ func (d *Drainer) DrainNode(ctx context.Context, node *corev1.Node, fullNodeDrai
...
+		if err != nil {
+			lastErr = err
+			reqLogger.Info("drainNode(): Draining failed, retrying", "error", err)
+			return false, nil
 		}
-		lastErr = err
-		reqLogger.Info("drainNode(): Draining failed, retrying", "error", err)
-		return false, nil
+
+		err = d.removeDaemonSetsFromNode(ctx, node.Name)
```
ykulazhenkov, line 95: I'm not entirely clear on how this works. I understand the process of selecting and removing DS pods, but I'm unsure how we prevent the DS controller from restarting pods on the node we intend to drain. Could you clarify?

test/conformance/tests/test_sriov_operator.go
SchSeba: done

pkg/drain/drainer.go
ykulazhenkov: Why the DS-managed Pod will be in Pending state?
Is it because of the cordon/taint? I think that the fact that the node is cordoned will not prevent DS pods to restart on it.
Or Is it because SRIOV resources are not exposed on the node? If so, Are we sure that SRIOV resources will not be exposed on the node at the moment when we try to kill the DS-managed pod?  How we ensure this?

I don’t have a complete picture of how the operator’s drain logic works, so I might be missing something obvious—but I think it’s worth asking :)

ykulazhenkov:
I checked the code and it looks good. I have one question about the logic. I want to make sure that we are not missing anything.

pkg/drain/drainer.go
adrianchiris: in case of reconfiguration ds pod may get back to running state if there are still sriov resources on the worker IMO.

indeed cordon does not affect daemonset pods (tested in kind cluster).
adding a custom taint with effect NoSchedule will prevent DS pod from getting rescheduled.

SchSeba: Adding a commit from the community meeting.

instead if using taint that can be harmful for the cluster (if some daemonset get deleted they will stuck until the sriov operator finish the work) 

we agree to implement in the drain controller a removal of the device plugin label from the node + the removal of the running device plugin pod

@adrianchiris @ykulazhenkov @zeeke let me know what do you think before I start coding it please :)

ykulazhenkov: +1 from me

adrianchiris: we roll more responsibilites to the drain controller.

maybe it should be the role of config daemon to ensure device plugin no longer runs on node before requesting drain ?
then put node label back after config completed.

just a different apporach to consider. im not against the current one.

adrianchiris:
@SchSeba can u rebase this one ? lets get this one merged !

SchSeba:
Hi @adrianchiris,

The PR is not ready, yet I need to implement the label remove as we spoke

PR #843, opened 2025-02-10
add support for draining single nodes

SchSeba:
With this change if a single node needs to drain to make configure changes it will remove pods using SR-IOV devices, but in case if a reboot is needed we just go and reboot.

This is an improvements from the skipDrain option we have in the sriovOperatorConfig where in case of only draining is needed to make the configuration change running pods using sriov will lost the VFs in runtime...

### DIFF
--- a/controllers/drain_controller_helper.go
+++ b/controllers/drain_controller_helper.go
@@ -74,9 +74,16 @@ func (dr *DrainReconcile) handleNodeDrainOrReboot(ctx context.Context,
 		return ctrl.Result{}, nil
 	}
 
+	// find the relevant node pool
+	nodePool, nodeList, err := dr.findNodePoolConfig(ctx, node)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


controllers/drain_controller_helper.go
```diff
@@ -172,6 +178,7 @@ func (dr *DrainReconcile) tryDrainNode(ctx context.Context, node *corev1.Node) (
 	}
 	reqLogger.Info("Max node allowed to be draining at the same time", "MaxParallelNodeConfiguration", maxUnv)
 	reqLogger.Info("Count of draining", "drainingNodes", current)
+	reqLogger.Info("Check Skip Drain on boot config", "SkipDrainOnReboot", nodePool.Spec.SkipDrainOnReboot)
```
adrianchiris, line 181: skip drain on reboot ?

SchSeba:
deprecate in favor of https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/850

PR #845, opened 2025-02-14
Start post-network-service after sriov-config

e0ne:
sriov-config-post-network-service sould be executed after sriov-config.

### DIFF
--- a/bindata/manifests/sriov-config-service/kubernetes/sriov-config-post-network-service.yaml
+++ b/bindata/manifests/sriov-config-service/kubernetes/sriov-config-post-network-service.yaml
@@ -1,7 +1,7 @@
 contents: |
   [Unit]
   Description=Configures SRIOV NIC - post network configuration
-  After=systemd-networkd-wait-online.service NetworkManager-wait-online.service openvswitch-switch.service
+  After=systemd-networkd-wait-online.service NetworkManager-wait-online.service openvswitch-switch.service sriov-config.service
   Before=kubelet.service
 
   [Service]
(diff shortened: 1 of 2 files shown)


rollandf (approved):

adrianchiris (approved):
nice catch! no idea why it didnt fail until now  😟

e0ne:
> nice catch! no idea why it didnt fail until now 😟

maybe it's related to kernel changes and/or system performance

ykulazhenkov:
I'm not against adding an explicit dependency between the services, but the logic should work without it on systems with NetworkManager or systemd-networkd if everything is working as was designed. 

How it should work on a system with NetworkManager:

- **sriov-config.service** has type=oneshot and Before=NetworkManager.service – meaning the network manager should start after the sriov-config.service process completes

- **sriov-config-post-network.service** has After=NetworkManager-wait-online.service

- **NetworkManager-wait-online.service** depends on NetworkManager.service.

So, the order should be sriov-config.service > NetworkManager > sriov-config-post-network.service.

For systems with systemd-networkd the logic is the same, but networkd services are used as dependencies.

The explicit dependency should be needed only in case if NetworkManager, system-networkd and ovs are not available on the system.

If the explicit dependency is really needed that probably means that the dependency chain doesn’t work as expected and “with bond scenarios” will not work.

Bond scenario:

1. sriov-config.service configure NIC mode and create VFs, unbind VFs from the driver at the end of execution
2. NetworkManager configures PFs to assemble a bond
3. sriov-config-post-network.service bind VFs to the driver

@e0ne  Do you have a full understanding why explicit dependency is needed?

PR #847, opened 2025-02-18
Make kargs.sh work with systemd service and for Ubuntu

e0ne:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved kernel-argument editing by selecting the correct execution environment automatically.
  * Enhanced Ubuntu GRUB handling to apply changes safely and only when necessary.
  * Improved kernel-argument updates in systemd mode by placing the required host script and applying the appropriate security labeling.
* **Tests**
  * Updated daemon controller test coverage to support Bash-based command execution.


pkg/plugins/generic/generic_plugin.go
```diff
@@ -81,7 +81,8 @@ type genericPluginOptions struct {
 	skipBridgeConfiguration bool
 }
 
-const scriptsPath = "bindata/scripts/kargs.sh"
+const daemonScriptsPath = "bindata/scripts/kargs.sh"
+const systemdScriptsPath = "//var/lib/sriov/kargs.sh"
```
ykulazhenkov, line 85: `//` at the beginning

pkg/plugins/generic/generic_plugin.go
```diff
@@ -277,7 +278,11 @@ func needDriverCheckVdpaType(state *sriovnetworkv1.SriovNetworkNodeState, driver
 // editKernelArg Tries to add the kernel args via ostree or grubby.
 func editKernelArg(helper helper.HostHelpersInterface, mode, karg string) error {
```
ykulazhenkov, line 279: Will this logic really work in the systemd mode? How it will behaves?
1. Reboot
2. systemd service exec
3. systemd service edited the grub config
4. Host booted
5. sriov-daemon pod starts
6. sriov-daemon detects that running kernel is missing required args
7.  host reboot again
8. ...
9. stable state

Is the sequence above describes the flow properly?

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -116,7 +116,7 @@ spec:
           command:
             - /bin/bash
             - -c
-            - mkdir -p /host/var/lib/sriov/ && cp /usr/bin/sriov-network-config-daemon /host/var/lib/sriov/sriov-network-config-daemon && chcon -t bin_t /host/var/lib/sriov/sriov-network-config-daemon | true # Allow systemd to run the file, use pipe true to not failed if the system doesn't have selinux or apparmor enabled
+            - mkdir -p /host/var/lib/sriov/ && cp /usr/bin/sriov-network-config-daemon /host/var/lib/sriov/sriov-network-config-daemon && chcon -t bin_t /host/var/lib/sriov/sriov-network-config-daemon | true  && cp /bindata/scripts/kargs.sh /host/var/lib/sriov/kargs.sh && chcon -t bin_t /host/var/lib/sriov/kargs.sh | true # Allow systemd to run the file, use pipe true to not failed if the system doesn't have selinux or apparmor enabled
```
ykulazhenkov, line 119: I think the original command has issue. Should it has `||` (or) instead of `|` (pipe)? The same question for the changes you added.

pkg/plugins/generic/generic_plugin.go
e0ne: in systemd mode kernel arg are configured by sriov-service so it doesn't do reboot loop

bindata/manifests/daemon/daemonset.yaml
e0ne: yes, we need to fix it and we should we need to think how to simplify this line. thanks for pointing me on this!

e0ne: just tested again: we need only one `|` to not fail if `chcon` execution failed

ykulazhenkov: Yes, this will work because `|` will mask errors from the previous command (by default, if `set -o pipefail` is not set). The `||` looks cleaner to me, but I'm okay to keep the current implementation.

ykulazhenkov (approved):

pkg/plugins/generic/generic_plugin.go
adrianchiris: @e0ne wont kernel args be applied only after reboot ?
hence two reboots will be required for this case ? 

(unless we set kargs in daemon even in systemd mode)

bindata/scripts/kargs.sh
```diff
@@ -5,54 +5,83 @@ command=$1
 shift
 declare -a kargs=( "$@" )
 ret=0
-args=$(chroot /host/ cat /proc/cmdline)
 
-IS_OS_UBUNTU=true; [[ "$(chroot /host/ grep -i ubuntu /etc/os-release -c)" == "0" ]] && IS_OS_UBUNTU=false
+if [[ -d "/bindata/scripts" ]];then
```
adrianchiris, line 9: nit :add comment, its non obvious that in this case we run in from config daemon and the other by system service.

bindata/scripts/kargs.sh
```diff
@@ -5,54 +5,83 @@ command=$1
...
+    chroot_path="/host/"
+else
+    chroot_path="/"
+fi
+
+args=$(chroot "$chroot_path" cat /proc/cmdline)
+
+IS_OS_UBUNTU=true; [[ "$(chroot "$chroot_path" grep -i ubuntu /etc/os-release -c)" == "0" ]] && IS_OS_UBUNTU=false
```
adrianchiris, line 17: you can:

```
. /etc/os-release
if [[ $ID == "ubuntu" ]]; then
..
fi
```

adrianchiris (commented):
overall LGTM. some minor nits.

bindata/scripts/kargs.sh
e0ne: done

bindata/scripts/kargs.sh
e0ne: done

e0ne: rolled back to a previous solution. this doesn't work for chroot

adrianchiris (approved):

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -125,6 +125,8 @@ spec:
               mkdir -p /host/var/lib/sriov/
               cp /usr/bin/sriov-network-config-daemon /host/var/lib/sriov/sriov-network-config-daemon
               chcon -t bin_t /host/var/lib/sriov/sriov-network-config-daemon || true # Allow systemd to run the file, use pipe true to not failed if the system doesn't have selinux or apparmor enabled
+              cp /bindata/scripts/kargs.sh /host/var/lib/sriov/kargs.sh
+              chcon -t bin_t /host/var/lib/sriov/kargs.sh | true # Allow systemd to run the file, use pipe true to not failed if the system doesn't have selinux or apparmor enabled
```
zeeke, line 129: nit: mind making congruent to L127?
```suggestion
              chcon -t bin_t /host/var/lib/sriov/kargs.sh || true # Allow systemd to run the file, use pipe true to not failed if the system doesn't have selinux or apparmor enabled
```

bindata/scripts/kargs.sh
zeeke: can you check the comment again.  it might have been dropped by a git rebase

bindata/manifests/daemon/daemonset.yaml
e0ne: done

ykulazhenkov (approved):

bindata/scripts/kargs.sh
```diff
@@ -60,4 +119,4 @@ else
     done
 fi
 
-echo $ret
+echo $ret
```
ykulazhenkov: nit: no new line at the end

bindata/scripts/kargs.sh
```diff
@@ -9,8 +9,67 @@ args=$(chroot /host/ cat /proc/cmdline)
...
+            found=false
+
+            for item in "${arr[@]}"; do
+                if [[ "$item" == "${t}" ]]; then
+                    found=true
+                    break
+                fi
+            done
+
+            if [ $found == false ];then
```
adrianchiris, line 48: what will happen if we have a karg already set with a certain value and we want to change to a different value ?

example:
current grub cmdline args have: `ib_core.netns_mode=1` then we want to set (using kargs.sh) `ib_core.netns_mode=0`

in this case we will end up with both set in grub cmdline IIUC since we compare the full key=val

i think we have the same issue with rpm-ostree as well as we use only append and delete.

imo we need to seach for the key of the karg first, if exists and its value is different then we remove and readd with new value (rpm-ostree has --replace flag)

adrianchiris: This can be addressed as a separate PR though, as we have the same issue with rpm-ostree as well.

e0ne: Let's address it in a follow-up PR both for Ubuntu and RHEL-based OSes

SchSeba: +1 from my side to handle it on a following up PR.
please if you can just open an issue so we don't forget about it

SchSeba:
Hi @e0ne,

I must say I don't follow this change.
we copy the kargs.sh into the host that is fine so when the service runs it will be able to call it.

but inside the script we still have pointers to `/host` and that will not work.
I think we need to add a check to see if we are inside the container or not.
maybe try to run a `chroot` and if that failed we know we are on the host and we need to changes stuff like `cp /host/${grub_config}`

/hold

e0ne:
> Hi guys, what about adding debian support? Looks like debian is configured similarly :)

We don't test Debian but I suppose it should work. We'll be glad to review PR with Debian support

pkg/plugins/generic/generic_plugin.go
```diff
@@ -269,7 +271,11 @@ func needDriverCheckVdpaType(state *sriovnetworkv1.SriovNetworkNodeState, driver
...
 	log.Log.Info("generic plugin editKernelArg()", "mode", mode, "karg", karg)
-	_, _, err := helper.RunCommand("/bin/sh", scriptsPath, mode, karg)
+	script := daemonScriptsPath
+	if _, err := os.Stat(consts.Host); errors.Is(err, os.ErrNotExist) {
+		script = systemdScriptsPath
+	}
+	_, _, err := helper.RunCommand("/bin/bash", script, mode, karg)
```
SchSeba, line 278: can we stay with `sh` here as before? or swtich the other calls to `bash` but please lets be consistent

pkg/daemon/daemon_test.go
```diff
@@ -119,72 +112,97 @@ var _ = Describe("Daemon Controller", Ordered, func() {
 		// general
 		hostHelper.EXPECT().Chroot(gomock.Any()).Return(func() error { return nil }, nil).AnyTimes()
 		hostHelper.EXPECT().RunCommand("/bin/sh", gomock.Any(), gomock.Any(), gomock.Any()).Return("", "", nil).AnyTimes()
+		hostHelper.EXPECT().RunCommand("/bin/bash", gomock.Any(), gomock.Any(), gomock.Any()).Return("", "", nil).AnyTimes()
```
SchSeba, line 197: please check the comment below about this

bindata/scripts/kargs.sh
```diff
@@ -9,8 +9,67 @@ args=$(chroot /host/ cat /proc/cmdline)
 
 IS_OS_UBUNTU=true; [[ "$(chroot /host/ grep -i ubuntu /etc/os-release -c)" == "0" ]] && IS_OS_UBUNTU=false
```
SchSeba: question how this one will work when we run on systemd mode?
the /host folder doesn't exist in the host itself only inside the container.

bindata/scripts/kargs.sh
```diff
@@ -9,8 +9,67 @@ args=$(chroot /host/ cat /proc/cmdline)
...
 
-# Kernel args configuration isn't supported for Ubuntu now, so we shouldn't do anything here
 if ${IS_OS_UBUNTU} ; then
+    grub_config="/etc/default/grub"
+    # Operate on the copy of the file
+    cp /host/${grub_config} /tmp/grub
```
SchSeba: question how this one will work when we run on systemd mode?
the /host folder doesn't exist in the host itself only inside the container.

e0ne: fixed

bindata/scripts/kargs.sh
e0ne: fixed now

pkg/plugins/generic/generic_plugin.go
e0ne: an updated scripts doesn't work with `sh` beacause it uses `bash`. do we want ti switch to `bash` everywhere?

bindata/scripts/kargs.sh
```diff
@@ -1,63 +1,144 @@
 #!/bin/bash
+# Copyright 2025 sriov-network-device-plugin authors
```
SchSeba: I think this copy right is wrong no?

bindata/scripts/kargs.sh
SchSeba: I don't see the comment also

bindata/scripts/kargs.sh
```diff
@@ -1,63 +1,144 @@
...
+
+args=$(chroot "$chroot_path" cat /proc/cmdline)
+
+IS_OS_UBUNTU=true; [[ "$(chroot "$chroot_path" grep -i ubuntu /etc/os-release -c)" == "0" ]] && IS_OS_UBUNTU=false
 
-# Kernel args configuration isn't supported for Ubuntu now, so we shouldn't do anything here
 if ${IS_OS_UBUNTU} ; then
+    grub_config="/etc/default/grub"
+    # Operate on the copy of the file
+    cp ${chroot_path}/${grub_config} /tmp/grub
```
SchSeba: remote the /
`${chroot_path}/${grub_config}` -> `${chroot_path}${grub_config}`

bindata/scripts/kargs.sh
```diff
@@ -1,63 +1,144 @@
...
+                        sed -i "s/\(^\s*$g\"\)\(.*\)\"/\1${new_param}\"/" /tmp/grub
+                    fi
+                done < /tmp/grub
+            fi
+        fi
+    done
+
+    if [ $ret -ne 0 ];then
+        # Update grub only if there were changes
+        cp /tmp/grub ${chroot_path}/${grub_config}
```
SchSeba, line 80: same here lets remove the /

pkg/plugins/generic/generic_plugin.go
SchSeba, line 278: can we just pass here if we are in chroot or not (boolean) and just use it in the script? this way we can remove the not clear check `if [[ -d "/bindata/scripts" ]];then`?

bindata/scripts/kargs.sh
e0ne: I'll rollback this change

bindata/scripts/kargs.sh
e0ne: done

bindata/scripts/kargs.sh
e0ne: fixed

pkg/plugins/generic/generic_plugin.go
e0ne: ok, I'll implement this in a next commit

bindata/scripts/kargs.sh
```diff
@@ -1,58 +1,124 @@
 #!/bin/bash
+
```
SchSeba, line 2: general idea here as we have full controll on the script and how we call it.

can we just add a variable when we call the script if we need chroot or not base on the `vars.chroot`
then at the beginning of the script we just make a `chroot /host` or nothing then all the code will look the same for both cases

SchSeba (commented):
I left a comment that I think will make the all implementation much easier.

let me know what do you think

bindata/scripts/kargs.sh
e0ne: done

e0ne: @SchSeba it doesn't work. for chroot we need to pass the command wich will be executed or pass the whole script as chroot param, it will be more hard to maintain.

from man:
```       
chroot - run command or interactive shell with special root directory
```

bindata/scripts/kargs.sh
```diff
@@ -1,58 +1,124 @@
...
+                        sed -i "s/\(^\s*$g\"\)\(.*\)\"/\1${new_param}\"/" /tmp/grub
+                    fi
+                done < /tmp/grub
+            fi
+        fi
+    done
+
+    if [ $ret -ne 0 ];then
+        # Update grub only if there were changes
+        cp /tmp/grub ${chroot_path}/${grub_config}
```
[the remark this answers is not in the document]
e0ne: we use chroor for eache command, so this comment isn't correct

pkg/plugins/generic/generic_plugin.go
```diff
@@ -80,7 +81,8 @@ type genericPluginOptions struct {
 	skipBridgeConfiguration bool
 }
 
-const scriptsPath = "bindata/scripts/kargs.sh"
+const daemonScriptsPath = "/bindata/scripts/kargs.sh"
```
SchSeba: nit: can you please update this one to be `containerScriptPath` and `hostScriptPath` I think it's cleaner that way

bindata/scripts/kargs.sh
```diff
@@ -1,58 +1,124 @@
...
+                if chroot "$chroot_path" rpm-ostree kargs | grep -q ${t}; then
+                    chroot "$chroot_path" rpm-ostree kargs --delete ${t} > /dev/null 2>&1
                 fi
                 let ret++
             fi
         fi
     done
 else
-    chroot /host/ which grubby > /dev/null 2>&1
+    chroot "$chroot_path" which grubby > /dev/null 2>&1
```
SchSeba, line 108: question this case will not work for UBUNTU also?
if on ubuntu we have grubby can't we just use the commands to add and remove like we do for coreOS?

bindata/scripts/kargs.sh
```diff
@@ -1,58 +1,124 @@
...
 
 command=$1
 shift
 declare -a kargs=( "$@" )
 ret=0
-args=$(chroot /host/ cat /proc/cmdline)
 
-IS_OS_UBUNTU=true; [[ "$(chroot /host/ grep -i ubuntu /etc/os-release -c)" == "0" ]] && IS_OS_UBUNTU=false
+if [[ -d "/bindata/scripts" ]];then
+    chroot_path="/host/"
```
SchSeba, line 11: in general I think it's better to have `command='chroot /host'` and `command=''` then just use it in all the script

bindata/scripts/kargs.sh
```diff
@@ -1,58 +1,124 @@
...
+else
+    chroot_path="/"
+fi
+
+args=$(chroot "$chroot_path" cat /proc/cmdline)
+
+IS_OS_UBUNTU=true; [[ "$(chroot "$chroot_path" grep -i ubuntu /etc/os-release -c)" == "0" ]] && IS_OS_UBUNTU=false
 
-# Kernel args configuration isn't supported for Ubuntu now, so we shouldn't do anything here
 if ${IS_OS_UBUNTU} ; then
```
SchSeba, line 20: please check the comment below and let me know if there is a reason that ubuntu is different from a grubby based system. so we can just use the command to add and remove kernel args

bindata/scripts/kargs.sh
e0ne: unfortunately, there's no grubby in official Canonical repos :(

bindata/scripts/kargs.sh
e0ne: there's no cli in oficial Ubuntu repos to modify kernel args

pkg/plugins/generic/generic_plugin.go
e0ne: done

bindata/scripts/kargs.sh
e0ne: IMO, `chroot_path` helps to understand logic better rather then `comand`. also it simplifies commands execution in the script

bindata/scripts/kargs.sh
```diff
@@ -1,58 +1,124 @@
...
+    grub_config="/etc/default/grub"
```
ykulazhenkov: Editing /etc/default/grub is unreliable because files in /etc/default/grub.d/*.cfg are loaded afterward and can override its values. Ubuntu cloud images are one common example, as 50-cloudimg-settings.cfg often redefines GRUB_CMDLINE_LINUX_DEFAULT. Could we use a late operator-owned drop-in, such as /etc/default/grub.d/99-sriov-network-operator.cfg, instead?

SchSeba: nice catch I was not aware of that!
+1 on this change

e0ne: done

pkg/plugins/generic/generic_plugin.go
```diff
@@ -269,7 +271,11 @@ func needDriverCheckVdpaType(state *sriovnetworkv1.SriovNetworkNodeState, driver
 // editKernelArg Tries to add the kernel args via ostree or grubby.
 func editKernelArg(helper helper.HostHelpersInterface, mode, karg string) error {
 	log.Log.Info("generic plugin editKernelArg()", "mode", mode, "karg", karg)
-	_, _, err := helper.RunCommand("/bin/sh", scriptsPath, mode, karg)
+	script := containerScriptPath
+	if _, err := os.Stat(consts.Host); errors.Is(err, os.ErrNotExist) {
+		script = hostScriptPath
+	}
```
zeeke, line 277: can you use `vars.UsingSystemdMode` :


```suggestion
    if vars.UsingSystemdMode {
        script = hostScriptPath
    }
```

Issue #846, opened 2025-02-18
sriov-network-operator consumes a large amount of memory due to caching machineconfig


crliu3227:
When there are numerous nodes in the cluster, or the cluster has undergone multiple upgrades, a substantial number of machineconfig resources will accumulate within the cluster. 
When we use mgr.client to get machineconfig, it will stay in cache and cost a large amount of memory.
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/b2c487c1a2421e684d0618ca14b438b28e500872/controllers/sriovoperatorconfig_controller.go#L416

PR #848, opened 2025-02-19
use kubeclient to get and delete machineconfig

crliu3227:
Use kubeclient to get and delete machineconfig, to fix https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/846

### DIFF
--- a/controllers/sriovoperatorconfig_controller.go
+++ b/controllers/sriovoperatorconfig_controller.go
@@ -56,10 +56,10 @@ import (
 // SriovOperatorConfigReconciler reconciles a SriovOperatorConfig object
 type SriovOperatorConfigReconciler struct {
 	client.Client
-	Scheme         *runtime.Scheme
-	PlatformHelper platforms.Interface
-	FeatureGate    featuregate.FeatureGate
-	KubeClient     client.Client
+	Scheme            *runtime.Scheme
+	PlatformHelper    platforms.Interface
+	FeatureGate       featuregate.FeatureGate
+	UncachedAPIReader client.Reader
 }
 
 //+kubebuilder:rbac:groups=sriovnetwork.openshift.io,resources=sriovoperatorconfigs,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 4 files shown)


controllers/sriovoperatorconfig_controller.go
```diff
@@ -59,6 +59,7 @@ type SriovOperatorConfigReconciler struct {
 	Scheme         *runtime.Scheme
 	PlatformHelper platforms.Interface
 	FeatureGate    featuregate.FeatureGate
+	KubeClient     client.Client
```
zeeke, line 62: Can we be more specific and use `mgr.GetAPIReader()`. It would be clearer if anyone else needs to do operations avoiding the cache,

```suggestion
	uncachedAPIReader     client.Reader
```

zeeke (commented):
@crliu3227 thanks for dealing with this issue. 

It would be great if we could disable the cache for all MachineConfig when we create the Manager in [main.go#L134](https://github.com/crliu3227/sriov-network-operator/blob/efb10630b79c4e2e3f26aad27d9e5c0450f26235/main.go#L134)

According to this article, it should be possible:
https://medium.com/@timebertt/kubernetes-controllers-at-scale-clients-caches-conflicts-patches-explained-aa0f7a8b4332#8181

WDYT?

controllers/sriovoperatorconfig_controller.go
crliu3227: OK, I will modify this part and resubmit the PR again

zeeke (approved):
LGTM

controllers/sriovoperatorconfig_controller.go
```diff
@@ -413,7 +414,8 @@ func (r *SriovOperatorConfigReconciler) syncOpenShiftSystemdService(ctx context.
 
 	if cr.Spec.ConfigurationMode != sriovnetworkv1.SystemdConfigurationMode {
 		obj := &machinev1.MachineConfig{}
-		err := r.Get(context.TODO(), types.NamespacedName{Name: consts.SystemdServiceOcpMachineConfigName}, obj)
+		// use uncached api reader to get machineconfig to reduce memory footprint
+		err := r.UncachedAPIReader.Get(context.TODO(), types.NamespacedName{Name: consts.SystemdServiceOcpMachineConfigName}, obj)
```
ykulazhenkov, line 418: Why not to use `ctx` that is passed to `syncOpenShiftSystemdService` as  arg?

crliu3227: Not sure why `ctx` wasn't used as a arg before, so no modifications have been made here.

SchSeba: Hi @crliu3227 please lets use the context from the function probably it was missed in the review process :)

adrianchiris: also dont we need to disable cache for MachineConfig in `mgr` (under main) ?

crliu3227: > Hi @crliu3227 please lets use the context from the function probably it was missed in the review process :)

Ok

crliu3227: > also dont we need to disable cache for MachineConfig in `mgr` (under main) ?

SriovNetworkPoolConfigReconciler will reconiler hardware offload MachineConfig, disable cache for MachineConfig might not be very appropriate, so add a label to MachineConfigs interested by sriov-network-operator ?

e0ne (approved):

SchSeba (approved):

PR #849, opened 2025-02-19
Remove leader election for the operator

SchSeba:
for the operator we don't need a leader election.
to avoid multiple instances running we change the upgrade policy to Recreate.

We always deploy the sriov operator with replica 1 so no need really to have leader election


main.go
```diff
@@ -48,9 +48,6 @@ import (
 	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/controllers"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/featuregate"
-	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/leaderelection"
```
zeeke, line 51: I think we can delete this package completely

And this will increase our code coverage :rocket:

SchSeba: nice catch thanks!

PR #850, opened 2025-02-19
Allow to drain on single node

SchSeba:
With this change we can stop using the skipDrain for single node clusters.

when a daemon request for drain we remove all the pods using sriov devices from the node, but in case the daemon requests a reboot and we are on a single node cluster we just mark the node as ready for reboot.

NOTE: we check the number of nodes in general not with the daemonSelector or the pool nodeSelector as in that case pods not using sriov devices can in theory move to other nodes.


adrianchiris (approved):

zeeke (approved):
LGTM

PR #852, opened 2025-02-24
Remove vf configuration file from host

SchSeba:
In case a policy gets removed when a node is rebooting after the reboot the number of virtual functions will be 0 and we will not remove the related configuration files from the host.

This commit fixes that issue

### DIFF
--- a/test/conformance/tests/test_sriov_operator.go
+++ b/test/conformance/tests/test_sriov_operator.go
@@ -1711,6 +1711,54 @@ var _ = Describe("[sriov] operator", Ordered, func() {
 			})
 		})
 
+		Context("Daemon reset with shutdown", Ordered, func() {
+			var testNode string
+			var policy *sriovv1.SriovNetworkNodePolicy
+			resourceName := "resetresource"
+
+			BeforeAll(func() {
+				isSingleNode, err := cluster.IsSingleNode(clients)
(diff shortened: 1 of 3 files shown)


adrianchiris (approved):

test/conformance/tests/test_sriov_operator.go
```diff
@@ -1722,6 +1722,72 @@ var _ = Describe("[sriov] operator", func() {
...
+				Eventually(func(g Gomega) bool {
+					node, err := clients.CoreV1Interface.Nodes().Get(context.Background(), testNode, metav1.GetOptions{})
+					g.Expect(err).ToNot(HaveOccurred())
+					for _, con := range node.Status.Conditions {
+						if con.Type == corev1.NodeReady && con.Status != corev1.ConditionTrue {
+							return true
+						}
+					}
+					return false
+				}, waitingTime, 1*time.Second).Should(BeTrue())
```
zeeke, line 1767: Can you refactor this to something like:
```
waitForNodeToBeReady()
waitForNodeToBeNotReady()
```
or
```
waitForNode(corev1.NodeReady, corev1.ConditionTrue)
waitForNode(corev1.NodeReady, corev1.ConditionFalse)
```

It took me a while to spot the differences between the calls

e0ne (approved):

zeeke (approved):

PR #856, opened 2025-03-04
Reach MetaData service over IPv6

dkokkino:


pkg/platforms/openstack/openstack.go
```diff
@@ -113,17 +117,73 @@ func New(hostManager host.HostManagerInterface) OpenstackInterface {
...
+
 // GetOpenstackData gets the metadata and network_data
 func getOpenstackData(mountConfigDrive bool) (metaData *OSPMetaData, networkData *OSPNetworkData, err error) {
 	metaData, networkData, err = getOpenstackDataFromConfigDrive(mountConfigDrive)
 	if err != nil {
 		log.Log.Error(err, "GetOpenStackData(): non-fatal error getting OpenStack data from config drive")
-		metaData, networkData, err = getOpenstackDataFromMetadataService()
-		if err != nil {
-			return metaData, networkData, fmt.Errorf("GetOpenStackData(): error getting OpenStack data: %w", err)
+		// Attempt to reach MetaData over IPv6 then over IPv4 for both HTTPS and HTTP
```
MaysaMacedo, line 153: I believe most of the deployments will be IPv4, so maybe we should try IPv4 first and if it fails, we try IPv6.

pkg/platforms/openstack/openstack.go
```diff
@@ -113,17 +117,73 @@ func New(hostManager host.HostManagerInterface) OpenstackInterface {
...
+		reachedMetaData := false
+		for i, baseURL := range ospBaseURLS {
+			isIPv6 := i == 0
```
MaysaMacedo, line 156: One idea: maybe we could remove the usage of a loop given the only possible addresses are either IPv4 or IPv6 and that list won't grow.

pkg/platforms/openstack/openstack.go
```diff
@@ -113,17 +117,73 @@ func New(hostManager host.HostManagerInterface) OpenstackInterface {
...
 }
 
+func getActiveInterfaceName() (string, error) {
+	interfaces, err := network.Interfaces()
+	if err != nil {
+		return "", err
+	}
+
+	for _, intf := range interfaces {
+		if intf.Flags&network.FlagUp != 0 && intf.Flags&network.FlagLoopback == 0 {
```
MaysaMacedo, line 127: We should check if there is at least one IPv6 address defined in the interface to be used

pkg/platforms/openstack/openstack.go
dkokkino: Okay i changed the logic to firstly check IPv4 then IPv6

pkg/platforms/openstack/openstack.go
dkokkino: Okay i added this check thanks for the feedback

pkg/platforms/openstack/openstack.go
dkokkino: I agree a loop is not necessary but I prefer using one because it allows us to exit as soon as we find the first address that can reach the metadata gracefully.

pkg/platforms/openstack/openstack.go
```diff
@@ -113,17 +117,85 @@ func New(hostManager host.HostManagerInterface) OpenstackInterface {
...
+
+	return "", fmt.Errorf("no active non-loopback interface found with an IPv6 address")
+}
+
+func constructMetaDataURLs(baseURL, activeInterface string, isIPv6 bool) []string {
+	var urls []string
+	encodedSign := "%25"
+	port := "80"
+	if isIPv6 {
+		urls = append(urls, ospHTTPS+"["+baseURL+encodedSign+activeInterface+"]:"+port)
```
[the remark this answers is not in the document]
dkokkino: Hi @SchSeba I am following this documentation.

https://docs.openstack.org/nova/latest/user/metadata.html#the-metadata-service:~:text=As%20with%20all,http%3A//169.254.169.254/....

PR #858, opened 2025-03-05
remove 5 second timeout

SchSeba:
We don't want to have this low timeout better to use the default for the client.

This put the k8s api server under a huge load a big clusters


ykulazhenkov (approved):

zeeke (approved):
LGTM

PR #860, opened 2025-03-10
 Sync node policies when node labels change only

zeeke:
Following from [1], fix the condition for calling the
 SriovNetworkNodePolicy reconciler on node updates, which
 is "only when a label changes".

 [1] https://github.com/k8snetworkplumbingwg/sriov-network-operator/commit/baa41c97adeb9249f30c5707f4bc8deee5e30c31#diff-d443e0d8dad7ac4f492a0601edbf9bfeec42ce8a496c020c0949e22b26442336R138-R183


ykulazhenkov (approved):

zeeke:
thanks @ykulazhenkov , merging this

PR #864, opened 2025-03-24
kernel: Avoid setting `iommu.passthrough` on ARM system

zeeke:
The kernel argument `iommu.passthgouh` can break some GPU scenarios when set to `1` [1]:

 >  We recommend that you add the iommu.passthrough=0 kernel parameter until this
issue is resolved

Avoid set the argument, as it would be hard for the user to undone the effect of this operation, as the config-daemon would reconcile the parameter list.

[1] https://docs.nvidia.com/grace-linux-install-guide.pdf

### DIFF
--- a/pkg/consts/constants.go
+++ b/pkg/consts/constants.go
@@ -139,12 +139,11 @@ const (
 		`IMPORT{program}="/etc/udev/switchdev-vf-link-name.sh $attr{phys_port_name}", ` +
 		`NAME="%s_$env{NUMBER}"`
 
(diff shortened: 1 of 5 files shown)


adrianchiris:
lets go with this approach for now. if we end up needing to set this in some scenarios i think we should look at adding a proper API (part of sriovnetwokrnodepool) for general host kernel args.

SchSeba (approved):
LGTM

SchSeba:
Thanks merging!

PR #866, opened 2025-03-25
Fix status update

SchSeba:
we do a patch to update the status but we need to copy the object meta from the latest get object if not the patch command will failed if the object did change


SchSeba:
Hi @zeeke @adrianchiris please take a look on this one a small fix that was found out by our QE

zeeke (approved):
LGTM

e0ne (approved):

PR #872, opened 2025-03-31
Skip device plugin configuration for non existing nodes

SchSeba:
In case where a node was removed but a nodeState still exist we return error instead of allowing the other nodes to be configured as expected

this commit fix this issue


SchSeba:
@ykulazhenkov @zeeke can you please take a look?

zeeke (approved):
LGTM

adrianchiris (approved):

PR #879, opened 2025-04-07
fix: copy pci.ids file to host in systemd mode

almaslennikov:
Config daemon relies on /usr/share/hwdata/pci.ids file for PCI devices discovery. It always exists in the container, but may be missing on the host. In this case, when we try to run the config daemon in the systemd mode, it will fail with an error. Let's copy this file to host if it doesn't exist there when we set up the systemd mode.


rollandf (approved):

e0ne (approved):

zeeke (approved):
LGTM

PR #881, opened 2025-04-10
fix: switch to semantic DeepEquals check from apimachinery

almaslennikov:
The switch to controller-runtime in thhe https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/788/commits/683101d9497db308d7840e2068f5731ed5952f0c commit introduced an interesting regression where there would be `{}` empty obj vs. `nil` obj discrepancies between semantically the same objects (e.g. OVS Bridged in spec and status). 
In the switchdev ovs use case it leads to SriovNetworkNodeState constantly shifting between the InProgress and Succeeded statuses and to sriov-network-device-plugin being recreated every few minutes.
Let's switch to using the `DeepEquals` function from k8s api-machinery, which treats empty and nil objects as equal.
I've changed all usages of `reflect.DeepEquals` to `equality.Semantic.DeepEquals`. Please, let me know if some of them should keep using the former one.

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"reflect"
 	"regexp"
 	"slices"
 	"sort"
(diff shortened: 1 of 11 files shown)


SchSeba (commented):
nice!

can I please ask you to implement some unit-tests to cover this one

almaslennikov:
Hi @SchSeba, added a unit test to check the semantical equality for OVS bridges

ykulazhenkov (approved):

PR #885, opened 2025-04-17
e2e: Improve `GetNodeSecureBootState`

zeeke:
If the file `/host/sys/kernel/security/lockdown` does not exists, the `cat` command populates `stderr`.


SchSeba (approved):
LGTM

PR #891, opened 2025-05-05
Use context to annotate the node object

SchSeba:
Same as we do for the NodeState to annotate the node object we need to use the context and not a background one


pkg/daemon/daemon.go
```diff
@@ -666,16 +666,19 @@ func (dn *NodeReconciler) annotate(
...
 		log.Log.Error(err, "Failed to annotate node")
 		return err
 	}
 
 	funcLog.Info(fmt.Sprintf("apply '%s' annotation for nodeState", annotationState))
-	if err := utils.AnnotateObject(context.Background(), desiredNodeState,
+	if err := utils.AnnotateObject(ctx, desiredNodeState,
 		consts.NodeStateDrainAnnotation,
 		annotationState, dn.client); err != nil {
+		log.Log.Error(err, "Failed to annotate nodeState")
```
ykulazhenkov, line 681: do we need to use funcLog here?

adrianchiris (approved):

ykulazhenkov (approved):

SchSeba:
Thanks!
merging this one

PR #893, opened 2025-05-12
add re-queue when we wait for drain to complete

SchSeba:
in case we have waiting for a drain and the user change the label in the node annotation manually it can make the configuration stuck as the daemon will not re-apply the right label to correct the user changes.

This will also help cover issue if exist in the draining logic in a way that the daemon will be able to retry again and not stuck waiting for a reconcile

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -260,9 +260,11 @@ func (dn *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 			reqLogger.Error(err, "failed to handle drain")
 			return ctrl.Result{}, err
 		}
-		// drain is still in progress we don't need to re-queue the request as the operator will update the annotation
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -260,9 +260,11 @@ func (dn *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 			reqLogger.Error(err, "failed to handle drain")
 			return ctrl.Result{}, err
 		}
-		// drain is still in progress we don't need to re-queue the request as the operator will update the annotation
+		// drain is still in progress we will still requeue the request in case there is an un-expect state in the draining
```
adrianchiris, line 263: please add a TODO to remove this (hopefully by OCP 4.20).

adrianchiris (commented):
a few nits otherwise LGTM

adrianchiris (approved):

SchSeba:
Thanks! merging

PR #894, opened 2025-05-12
Support namespaced {Sriov,SriovIB,OVS}Networks 

zeeke:
SriovNetwork,IbSriovNetwork,OVSNetwork object (XNetwork resources from now on) must be created in the operator's namespace and the `.Spec.NetworkNamespace` field defines where the
controller should create the NetworkAttachmentDefinition resource. This constraint can be a problem
in clusters where applications are managed by non cluster administrators.

These changes makes the `genericNetworkReconciler` to accept XNetwork resources in namespaces different than
the operator's one. In such cases, the field `.Spec.NetworkNamespace` must be empty, and a validating webhook
ensure this constraint.

Webhook validation for networks .Spec.NetworkNamespace field 

Move `[sriov] operator No SriovNetworkNodePolicy ...` test cases to its
own test file.

Implement test case to verify controller and webhook logic


controllers/generic_network_controller.go
```diff
@@ -252,3 +252,72 @@ func (r *genericNetworkReconciler) deleteNetAttDef(ctx context.Context, cr Netwo
...
+	}
+
+	if !errors.IsNotFound(err) {
+		return nil, err
+	}
+
+	operatorNamespacedName := req.NamespacedName
+	operatorNamespacedName.Namespace = vars.Namespace
+
+	err = r.Get(ctx, operatorNamespacedName, instance)
```
SchSeba: why we need to also check in the operator namespace?
can you please add a comment to explain why so I will remember next time I look at this code :P

controllers/sriovnetwork_controller_test.go
```diff
@@ -342,6 +342,45 @@ var _ = Describe("SriovNetwork Controller", Ordered, func() {
...
+				err = k8sClient.Delete(ctx, &cr)
+				Expect(err).NotTo(HaveOccurred())
+
+				expectedOwnerReference := metav1.OwnerReference{
+					Kind:       "SriovNetwork",
+					APIVersion: sriovnetworkv1.GroupVersion.String(),
+					UID:        cr.UID,
+					Name:       cr.Name,
+				}
+				Expect(netAttDef.ObjectMeta.OwnerReferences).To(ContainElement(expectedOwnerReference))
```
SchSeba, line 383: can you add a check here that the net-attach-def gets deleted after we remove the sriovNetwork please?

pkg/webhook/validate_networks.go
```diff
@@ -0,0 +1,43 @@
...
+func validateOVSNetwork(cr *sriovnetworkv1.OVSNetwork, operation v1.Operation) (bool, []string, error) {
+	err := validateNetworkNamespace(cr)
+	if err != nil {
+		return false, nil, err
+	}
+	return true, nil, nil
+}
+
+func validateNetworkNamespace(cr controllers.NetworkCRInstance) error {
+	if cr.GetNamespace() != vars.Namespace && cr.NetworkNamespace() != "" {
```
SchSeba, line 38: please also add a check in case we are doing it on the sriov-network-operator it must contain NetworkNamespace no?

pkg/webhook/validate_networks.go
```diff
@@ -0,0 +1,43 @@
...
+		return fmt.Errorf(".Spec.NetworkNamespace field can't be specified if the resource is not in the %s namespace", vars.Namespace)
```
SchSeba, line 39: also do we want to block the creation there is already a network from the operator namespace with the same name?

or if there already a net-attach-def on the requested namespace?

test/conformance/tests/test_no_policy.go
```diff
@@ -0,0 +1,299 @@
+package tests
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
SchSeba, line 12: nit: can you re-order this to


```
	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
	
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	corev1 "k8s.io/api/core/v1"
	k8serrors "k8s.io/apimachinery/pkg/api/errors"
	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
	netattdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"

	sriovv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
	"github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/cluster"
	"github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/discovery"
	"github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/namespaces"
	"github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/nodes"
```

test/conformance/tests/test_no_policy.go
```diff
@@ -0,0 +1,299 @@
...
+					g.Expect(afterLogs).ToNot(
+						ContainElement(
+							ContainSubstring("Start to sync webhook objects"),
+						),
+					)
+				}, 3*time.Minute, 5*time.Second).Should(Succeed())
+			})
+		})
+
+		DescribeTable("should gracefully restart quickly", func(webookEnabled bool) {
```
SchSeba: This is the one that I removed so can you leave it in the original place?

test/conformance/tests/test_no_policy.go
```diff
@@ -0,0 +1,299 @@
...
+				setFeatureFlag("metricsExporter", false)
+				Eventually(func(g Gomega) {
+					_, err := clients.ServiceMonitors(operatorNamespace).Get(context.Background(), "sriov-network-metrics-exporter", metav1.GetOptions{})
+					g.Expect(k8serrors.IsNotFound(err)).To(BeTrue())
+				}).WithTimeout(time.Minute).WithPolling(time.Second).Should(Succeed())
+			})
+		})
+
+		Context("Namespaced network objects", func() {
+			BeforeAll(func() {
```
SchSeba: nit: can you remove this empty one?

test/conformance/tests/test_no_policy.go
```diff
@@ -0,0 +1,299 @@
...
+				Expect(string(k8serrors.ReasonForError(err))).
+					To(ContainSubstring(".Spec.NetworkNamespace field can't be specified if the resource is not in the "))
+			},
+				Entry("SriovNetwork", &sriovv1.SriovNetwork{ObjectMeta: metav1.ObjectMeta{Name: "sriovnet1", Namespace: namespaces.Test}, Spec: sriovv1.SriovNetworkSpec{NetworkNamespace: "default"}}),
+				Entry("SriovIBNetwork", &sriovv1.SriovIBNetwork{ObjectMeta: metav1.ObjectMeta{Name: "sriovibnet1", Namespace: namespaces.Test}, Spec: sriovv1.SriovIBNetworkSpec{NetworkNamespace: "default"}}),
+				Entry("OVSNetwork", &sriovv1.OVSNetwork{ObjectMeta: metav1.ObjectMeta{Name: "ovsnet1", Namespace: namespaces.Test}, Spec: sriovv1.OVSNetworkSpec{NetworkNamespace: "default"}}),
+			)
+		})
+	})
+})
```
SchSeba, line 259: can you add a check here that if network in the operator namespace exist it should win always ones created by the user on it's own namespace

controllers/generic_network_controller.go
zeeke: this is because we reconcile NetworkAttachmentDefinitions updates, and here we have to fetch the source object.

the cases are

```
|| Reconciled object   || source object ||
| netAttachDef(nsX/net1) | SriovNetwork(operatorNs/net1) with networkNamespace: nsX |
| netAttachDef(nsX/net1) | SriovNetwork(nsX/net1) |
| SriovNetwork(operatorNs/net1) | SriovNetwork(operatorNs/net1)  |
| SriovNetwork(nsX/net1)  | SriovNetwork(nsX/net1) |
 ```

adding a comment

File: pkg/webhook/validate_networks.go (hunk 3 above)
Comment (zeeke) [reply pkg/webhook/validate_networks.go:38]: It's not mandatory, and if the SriovNetwork.Spec.NetworkNamespace is empty, the NetworkattachmentDefinition is created in the sriov-network-operator namespace.

This might not sound very clean, but in openshift the `openshift-sriov-network-operator` namespace is listed as global:

```
$ oc get cm -n openshift-multus multus-daemon-config -o yaml | grep global
        "globalNamespaces": "default,openshift-multus,openshift-sriov-network-operator,openshift-cnv",
```

It's usefult to create one NetAttachDefinition and use it in multiple namespace

File: controllers/sriovnetwork_controller_test.go (hunk 2 above)
Comment (zeeke) [reply controllers/sriovnetwork_controller_test.go:383]: unfortunately no, OwnerReferences are cleaned by kubernetes controllers, which are not available in test env. :confused: 

We have to trust them to work correctly

File: pkg/webhook/validate_networks.go (hunk 4 above)
Comment (zeeke) [reply pkg/webhook/validate_networks.go:39]: That's a little bit debatable: having two or more objects contending the same NetAttachDef is handled in 
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/897

It's the same problem, viewed from the perspective of the webhook, and we already have the same with an SriovNetwork + SriovIBNetwork pointing to the same object.

I prefer to address it in #897 or in another subsequent PR, after discussing in a community meeting. WDYT?

File: test/conformance/tests/test_no_policy.go (hunk 6 above)
Comment (zeeke) [reply test/conformance/tests/test_no_policy.go]: To leave it where it was, I have to put back all the `No SriovNetworkPolicy` test tree. We'll get a conflict in any case and I'll solve it. Let's merge 
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/901

before this

File: test/conformance/tests/test_no_policy.go (hunk 8 above)
Comment (zeeke) [reply test/conformance/tests/test_no_policy.go:259]: This case is handled  in 
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/897

but no such priority has been implemented yet. That PR handles these collisions by making the first object win.

for example, if a namespaced network is created, then the created netattachdef has the owner annotation. When the operatornamespaced network is created, it finds the target netattachdefs already owned by another object and the controller does nothing.

File: pkg/webhook/validate_networks_test.go [hunk 9]
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/controllers"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+func TestValidate_NetworkNamespace(t *testing.T) {
+	g := NewGomegaWithT(t)
+	defer func(previous string) { vars.Namespace = previous }(vars.Namespace)
+	vars.Namespace = "operator-namespace"
+
+	validNetworkNamespaces := []controllers.NetworkCRInstance{
```
adrianchiris: any chance to have a "[]tcase" (list of test case struct) which contains the test name, the NetworkCRInstance and a `shouldFail` boolean ? then the test would iterate over entries , run validateNetworkNamespace and compare to expected result ?

controllers/generic_network_controller.go
```diff
@@ -72,56 +74,47 @@ type genericNetworkReconciler struct {
...
+		// Request object not found, could have been deleted after reconcile request.
+		// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+		// Return and don't requeue
+		return reconcile.Result{}, nil
+	}
+
+	if instance.NetworkNamespace() != "" && instance.GetNamespace() != vars.Namespace {
+		reqLogger.Error(
+			fmt.Errorf("bad value for NetworkNamespace"),
+			".Spec.NetworkNamespace can't be specified if the resource belongs to a namespace other than the operator's",
```
adrianchiris: nit do we want to specify "jsonpath" style field refs ?

e.g `.spec.networkNamespace`, `.metadata.namespace` ?

controllers/generic_network_controller.go
adrianchiris: can we handle this case in an eventHandler registered in SetupWithManager ?

so instead of using `handler.EnqueuRequestForObject{}` we would implement our own with `handler.EnqueueRequestFromMapFn` there we can try to get the "source" object from current net-attach-def namespace or from the operator ns.

then Reconcile would get the correct object it needs to reconcile, WDYT ?

controllers/generic_network_controller.go
zeeke: yep, that would be better for the user point of view

pkg/webhook/validate_networks.go
SchSeba: works for me

test/conformance/tests/test_no_policy.go
SchSeba: now that 901 is merged I think we can remove this one

test/conformance/tests/test_no_policy.go
SchSeba: sounds good

SchSeba (approved):
from my side LGTM.

to me it's not an hard request to have the object retreval in the handler, but it's nice to have.

zeeke:
I added the handleNetAttachDef function and it actually made the logic a little straightforward.

@adrianchiris @SchSeba please take a look

SchSeba (approved):
I link this new approach!

I learned something new :) 

LGTM

controllers/generic_network_controller.go
```diff
@@ -202,11 +210,36 @@ func (r *genericNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		Watches(&netattdefv1.NetworkAttachmentDefinition{}, &handler.EnqueueRequestForObject{}).
+		Watches(&netattdefv1.NetworkAttachmentDefinition{}, handler.EnqueueRequestsFromMapFunc(r.handleNetAttDef)).
 		Watches(&corev1.Namespace{}, &namespaceHandler).
 		Complete(r.controller)
 }
 
+func (r *genericNetworkReconciler) handleNetAttDef(ctx context.Context, obj client.Object) []reconcile.Request {
+	instance := r.controller.GetObject()
+	nadNamespacedName := types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}
+	err := r.Get(ctx, nadNamespacedName, instance)
```
adrianchiris, line 225: potentially we might have both existing and the other (from the sriov network operator namespace) is the actual owner.

maybe we should add both if they exist ?

controllers/generic_network_controller.go
```diff
@@ -163,6 +164,13 @@ func (r *genericNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 				return reconcile.Result{}, nil
 			}
 
+			if instance.GetNamespace() == netAttDef.Namespace {
```
adrianchiris: at L190 if the nad already exists, it will get overriden by the incoming network object either from the operator namespace or from the target namespace.

if we have two network objects with conflicting configs the last one getting reconciled will be the winner.
this is a bit unpredictable.

i suggest to check if the object has owner ref and the network object is in the target namespace (without the spec.networkNamespace field set) then update else skip.

and ofc the other way around. if the object does not have owner ref, only update it if the network object is in the operator namespace with spec.networkNamespace field set

controllers/generic_network_controller.go
```diff
@@ -91,37 +92,37 @@ func (r *genericNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	instanceFinalizers := instance.GetFinalizers()
+
+	if instance == nil {
+		// Request object not found, could have been deleted after reconcile request.
+		// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+		// Return and don't requeue
+		return reconcile.Result{}, nil
+	}
+
+	if instance.NetworkNamespace() != "" && instance.GetNamespace() != vars.Namespace {
```
e0ne, line 103: do we want to add error field into `*NetworkStatus` objects?

controllers/generic_network_controller.go
zeeke: makes sense! fixing

controllers/generic_network_controller.go
zeeke: This is the problem of multiple network objects referring the same NAD, though these changes makes the situation a little worse. 

I'm handling it here:
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/897

Is it ok to handle it in that subsequent PR?

controllers/generic_network_controller.go
zeeke: It's a good idea, but I would discuss it after 
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/918

Maybe the conditions on CRDs are enough. What do you think?

zeeke:
> Hi @zeeke,
> 
> our QE did the following test on this PR
> 
> * Create sriovNetwork/net1 in ns1
> * Operator creates NAD/net1 in ns1
> * Update sriovNetwork/net1 with new resourceName
> * NAD/net1 is also updated (but no changes in spec as expected)
> 
> And on the last step when he check the NAD/net1 the owner reference was removed. So we probably have an issue with the update.

fixed.

@adrianchiris @e0ne @SchSeba please, take another look

controllers/generic_network_controller.go
adrianchiris: ack

controllers/generic_network_controller.go
```diff
@@ -86,42 +87,45 @@ func (r *genericNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
+			reqLogger.Error(err, "XXX2")
```
adrianchiris: please remove debug prints ::)

controllers/generic_network_controller.go
```diff
@@ -86,42 +87,45 @@ func (r *genericNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
-	instanceFinalizers := instance.GetFinalizers()
+
+	if instance == nil {
+		// Request object not found, could have been deleted after reconcile request.
+		// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+		// Return and don't requeue
+		reqLogger.Info("XXX3")
```
adrianchiris: please remove debug prints ::)

controllers/sriovnetwork_controller_test.go
```diff
@@ -342,6 +342,68 @@ var _ = Describe("SriovNetwork Controller", Ordered, func() {
...
 				Should(Succeed())
 		})
+
+		Context("When the SriovNetwork namespace is not equal to the operator one", func() {
+			BeforeAll(func() {
+				nsBlue := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "ns-blue"}}
+				Expect(k8sClient.Create(context.Background(), nsBlue)).ToNot(HaveOccurred())
+				DeferCleanup(func() {
+					By("deleting ns-blue")
+					err := k8sClient.Delete(ctx, nsBlue)
```
adrianchiris: with envtest you cant really delete namespace, it will just stay dangling.

https://book.kubebuilder.io/reference/envtest#namespace-usage-limitation

controllers/sriovnetwork_controller_test.go
```diff
@@ -342,6 +342,68 @@ var _ = Describe("SriovNetwork Controller", Ordered, func() {
...
+				cr.Spec.ResourceName = "resource_y"
+				err = k8sClient.Patch(ctx, &cr, dynclient.MergeFrom(original))
+				Expect(err).NotTo(HaveOccurred())
+
+				// Check that the OwnerReference persists
+				netAttDef = &netattdefv1.NetworkAttachmentDefinition{}
+
+				Eventually(func(g Gomega) {
+					netAttDef = &netattdefv1.NetworkAttachmentDefinition{}
+					err = util.WaitForNamespacedObject(netAttDef, k8sClient, "ns-blue", cr.GetName(), util.RetryInterval, util.Timeout)
```
adrianchiris: nit: no need to call this wait func, object already exists, you can just call g.Expect(dynclient.Get(....)).To(Succeed())
its all wrapped in Eventually block anyway

controllers/sriovnetwork_controller_test.go
```diff
@@ -342,6 +342,68 @@ var _ = Describe("SriovNetwork Controller", Ordered, func() {
...
+					g.Expect(err).NotTo(HaveOccurred())
+					g.Expect(netAttDef.GetAnnotations()["k8s.v1.cni.cncf.io/resourceName"]).To(Equal("openshift.io/resource_y"))
+					g.Expect(netAttDef.ObjectMeta.OwnerReferences).To(ContainElement(expectedOwnerReference))
+				}).WithPolling(100 * time.Millisecond).WithTimeout(5 * time.Second).Should(Succeed())
+
+				// Delete the SriovNetwork
+				err = k8sClient.Delete(ctx, &cr)
+				Expect(err).NotTo(HaveOccurred())
```
adrianchiris, line 403: lets wait for the underlying net-attach-def to be deleted ?

controllers/sriovnetwork_controller_test.go
```diff
@@ -342,6 +342,68 @@ var _ = Describe("SriovNetwork Controller", Ordered, func() {
 				MustPassRepeatedly(10).
 				Should(Succeed())
 		})
+
+		Context("When the SriovNetwork namespace is not equal to the operator one", func() {
```
adrianchiris, line 348: can you add a case where the user requests to create nad in a different ns than the current one ?
then check nad doesnt get created with a `Consistently` block ?

controllers/generic_network_controller.go
```diff
@@ -202,11 +215,41 @@ func (r *genericNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		Watches(&corev1.Namespace{}, &namespaceHandler).
 		Complete(r.controller)
 }
 
+func (r *genericNetworkReconciler) handleNetAttDef(ctx context.Context, obj client.Object) []reconcile.Request {
+	ret := []reconcile.Request{}
+	instance := r.controller.GetObject()
+	nadNamespacedName := types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}
+
+	log.Log.WithName("XXX").Info("handleNetAttDef", "nadNamespacedName", nadNamespacedName, "vars.Namespace", vars.Namespace)
```
adrianchiris: nit: remove debug logs (XXX/XXX11 etc)

adrianchiris (commented):
final nits otherwise LGTM

controllers/generic_network_controller.go
zeeke: sure! sorry for that leftover

controllers/sriovnetwork_controller_test.go
zeeke: I didn't know that, thanks!

I added a `cleanNetworksInNamespace` function to remove all the resources after the test

controllers/sriovnetwork_controller_test.go
zeeke: Added!

controllers/sriovnetwork_controller_test.go
```diff
@@ -390,3 +475,12 @@ func generateExpectedNetConfig(cr *sriovnetworkv1.SriovNetwork) string {
...
 	return configStr
 }
+
+func cleanNetworksInNamespace(namespace string) {
+	ctx := context.Background()
+	err := k8sClient.DeleteAllOf(ctx, &sriovnetworkv1.SriovNetwork{}, client.InNamespace(namespace))
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	k8sClient.DeleteAllOf(ctx, &netattdefv1.NetworkAttachmentDefinition{}, client.InNamespace(namespace))
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
```
adrianchiris: can you also wait for both objects to be gone from the kubernetes api ?

zeeke: sure, I put the delete and list calls in an eventually block

controllers/sriovnetwork_controller_test.go
```diff
@@ -390,3 +475,24 @@ func generateExpectedNetConfig(cr *sriovnetworkv1.SriovNetwork) string {
 	}
 	return configStr
 }
+
+func cleanNetworksInNamespace(namespace string) {
+	ctx := context.Background()
+	EventuallyWithOffset(1, func(g Gomega) {
+		err := k8sClient.DeleteAllOf(ctx, &sriovnetworkv1.SriovNetwork{}, client.InNamespace(namespace))
+		g.Expect(1, err).NotTo(HaveOccurred())
```
adrianchiris: g.Expect(err) ? here and below

adrianchiris:
lgtm, lets have UT pass

adrianchiris (approved):

PR #896, opened 2025-05-30
e2e: Filter devices in `FindOneMellanoxSriovDevice`

zeeke:
Honor environment variable `SRIOV_NODE_AND_DEVICE_NAME_FILTER` in `FindOneMellanoxSriovDevice(...)` function.

cc @evgenLevin

### DIFF
--- a/test/util/cluster/cluster.go
+++ b/test/util/cluster/cluster.go
@@ -247,24 +247,24 @@ func (n *EnabledNodes) FindOneVfioSriovDevice() (string, sriovv1.InterfaceExt) {
 
 // FindOneMellanoxSriovDevice retrieves a valid sriov device for the given node.
 func (n *EnabledNodes) FindOneMellanoxSriovDevice(node string) (*sriovv1.InterfaceExt, error) {
-	s, ok := n.States[node]
-	if !ok {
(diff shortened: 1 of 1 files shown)


adrianchiris (approved):
lgtm

test/util/cluster/cluster.go
```diff
@@ -247,24 +247,27 @@ func (n *EnabledNodes) FindOneVfioSriovDevice() (string, sriovv1.InterfaceExt) {
...
 	// TODO: remove this when mlx support secure boot/lockdown mode
 	if n.IsSecureBootEnabled[node] {
 		return nil, fmt.Errorf("secure boot is enabled on the node mellanox cards are not supported")
 	}
 
-	for _, itf := range s.Status.Interfaces {
-		if itf.Vendor == mlxVendorID && sriovv1.IsSupportedModel(itf.Vendor, itf.DeviceID) {
-			return &itf, nil
+	var errs []error
+	for _, node := range n.Nodes {
```
SchSeba: I think there is a problem here we don't use the node name anymore??

zeeke: right! fixing

PR #897, opened 2025-06-06
networks: Add Owner annotation to NetAttachDefs

zeeke:
depends on:
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/894

NetworkattachmentDefinitions are usually created in namespaces that are different
than the {Sriov,SriovIb,OVS}Networks, so the OwnerReference object field can't be
used to endorse the ownership reference.

Add a `sriovnetwork.openshift.io/owner` annotation to put on the NetAttachDef
to avoid multiple {Sriov,SriovIb,OVS}Network objects does not override changes.

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -19,6 +19,7 @@ import (
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	intstrutil "k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
@@ -651,6 +652,7 @@ func (cr *SriovIBNetwork) RenderNetAttDef() (*uns.Unstructured, error) {
 	} else {
 		data.Data["SriovNetworkNamespace"] = cr.Spec.NetworkNamespace
 	}
+	data.Data["Owner"] = OwnerRefToString(cr)
 	data.Data["SriovCniResourceName"] = os.Getenv("RESOURCE_PREFIX") + "/" + cr.Spec.ResourceName
 
 	data.Data["StateConfigured"] = true
@@ -719,6 +721,7 @@ func (cr *SriovNetwork) RenderNetAttDef() (*uns.Unstructured, error) {
 	} else {
 		data.Data["SriovNetworkNamespace"] = cr.Spec.NetworkNamespace
 	}
+	data.Data["Owner"] = OwnerRefToString(cr)
 	data.Data["SriovCniResourceName"] = os.Getenv("RESOURCE_PREFIX") + "/" + cr.Spec.ResourceName
 	data.Data["SriovCniVlan"] = cr.Spec.Vlan
 
@@ -837,6 +840,7 @@ func (cr *OVSNetwork) RenderNetAttDef() (*uns.Unstructured, error) {
 	} else {
 		data.Data["NetworkNamespace"] = cr.Spec.NetworkNamespace
 	}
+	data.Data["Owner"] = OwnerRefToString(cr)
 	data.Data["CniResourceName"] = os.Getenv("RESOURCE_PREFIX") + "/" + cr.Spec.ResourceName
 
 	if cr.Spec.Capabilities == "" {
@@ -993,4 +997,15 @@ func (s *SriovNetworkNodeState) ResetKeepUntilTime() bool {
 	delete(annotations, consts.NodeStateKeepUntilAnnotation)
 	s.SetAnnotations(annotations)
 	return true
+}
+
+func OwnerRefToString(cr client.Object) string {
+	if cr == nil {
+		return "owner-object-not-found"
+	}
+	if cr.GetObjectKind().GroupVersionKind().Empty() {
+		return "unknown/" + cr.GetNamespace() + "/" + cr.GetName()
+	}
+
+	return cr.GetObjectKind().GroupVersionKind().GroupKind().String() + "/" + cr.GetNamespace() + "/" + cr.GetName()
 }
--- a/controllers/generic_network_controller.go
+++ b/controllers/generic_network_controller.go
@@ -38,6 +38,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
 )
@@ -189,6 +190,20 @@ func (r *genericNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 	} else {
 		reqLogger.Info("NetworkAttachmentDefinition CR already exist")
+
+		foundOwner := found.GetAnnotations()[consts.OwnerRefAnnotation]
+		expectedOwner := netAttDef.GetAnnotations()[consts.OwnerRefAnnotation]
+
+		// Note for the future: the `foundOwner != ""` condition can be removed to make the operator not touching the NetworkAttachmentDefinition created
+		// by the user.
+		if foundOwner != "" && foundOwner != expectedOwner {
+			reqLogger.Info("A NetworkAttachmentDefinition with the same name already exists and it does not belong to this resource",
+				"Namespace", netAttDef.Namespace, "Name", netAttDef.Name,
+				"CurrentOwner", foundOwner, "ExpectedOwner", expectedOwner,
+			)
+			return reconcile.Result{}, nil
+		}
+
 		if !equality.Semantic.DeepEqual(found.Spec, netAttDef.Spec) || !equality.Semantic.DeepEqual(found.GetAnnotations(), netAttDef.GetAnnotations()) {
 			reqLogger.Info("Update NetworkAttachmentDefinition CR", "Namespace", netAttDef.Namespace, "Name", netAttDef.Name)
 			netAttDef.SetResourceVersion(found.GetResourceVersion())
--- a/controllers/generic_network_controller_test.go
+++ b/controllers/generic_network_controller_test.go
@@ -1,0 +1,222 @@
+package controllers
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	netattdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/test/util"
+)
+
+var _ = Describe("All Network Controllers", Ordered, func() {
+	var cancel context.CancelFunc
+	var ctx context.Context
+
+	BeforeAll(func() {
+		By("Setup controller manager")
+		k8sManager, err := setupK8sManagerForTest()
+		Expect(err).ToNot(HaveOccurred())
+
+		err = (&SriovNetworkReconciler{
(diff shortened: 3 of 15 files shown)


api/v1/helper.go
```diff
@@ -651,6 +651,7 @@ func (cr *SriovIBNetwork) RenderNetAttDef() (*uns.Unstructured, error) {
 	} else {
 		data.Data["SriovNetworkNamespace"] = cr.Spec.NetworkNamespace
 	}
+	data.Data["Owner"] = cr.GroupVersionKind().GroupKind().String() + "/" + cr.GetNamespace() + "/" + cr.Name
```
adrianchiris: nit: add helper func for this bit ? 

not a must, its fine if u prefer this way.

adrianchiris: also keep in mind that getting the object using a direct client (i.e not backed by cache) will not populate the GVK.  i dont think we should have any issues with this though.

pkg/consts/constants.go
```diff
@@ -98,6 +98,8 @@ const (
 	MCPPauseAnnotationState = "sriovnetwork.openshift.io/state"
 	MCPPauseAnnotationTime  = "sriovnetwork.openshift.io/time"
 
+	OwnerAnnotation = "sriovnetwork.openshift.io/owner"
```
adrianchiris: nit: `owner-ref` ?

fine to keep as is if u prefer.

controllers/generic_network_controller.go
```diff
@@ -180,6 +181,20 @@ func (r *genericNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		expectedOwner := netAttDef.GetAnnotations()[consts.OwnerAnnotation]
+
+		// Note for the future: the `foundOwner != ""` condition can be removed to make the oeprator not touching the NetworkAttachmentDefinition created
+		// by the user.
+		if foundOwner != "" && foundOwner != expectedOwner {
+			reqLogger.Info("A NetworkAttachmentDefinition with the same name already exists and it does not belong to this resource",
+				"Namespace", netAttDef.Namespace, "Name", netAttDef.Name,
+				"CurrentOwner", foundOwner, "ExpectedOwner", expectedOwner,
+			)
+			return reconcile.Result{}, err
```
adrianchiris: nit: return reconcile.Result{}, nil

controllers/generic_network_controller.go
```diff
@@ -180,6 +181,20 @@ func (r *genericNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+
```
adrianchiris, line 206: in case foundOwner is "" and there is a CR in the operator namespace then it should be the one eventually owning this object.

with current impl, the first network CR that gets reconciled will "take ownership" of the net-attach-def potentially overriding it.

maybe we should check if net-attach-def has no owner AND a network CR exists in the operator namespace pointing to it, we should also do nothing like above.

controllers/generic_network_controller_test.go
```diff
@@ -0,0 +1,187 @@
...
+			wg.Wait()
+		})
+	})
+
+	Context("owner-reference annotations", func() {
+		AfterEach(func() {
+			k8sClient.DeleteAllOf(ctx, &sriovnetworkv1.SriovNetwork{}, client.InNamespace(testNamespace))
+			k8sClient.DeleteAllOf(ctx, &sriovnetworkv1.SriovIBNetwork{}, client.InNamespace(testNamespace))
+			k8sClient.DeleteAllOf(ctx, &sriovnetworkv1.OVSNetwork{}, client.InNamespace(testNamespace))
+			k8sClient.DeleteAllOf(ctx, &netattdefv1.NetworkAttachmentDefinition{}, client.InNamespace("default"))
```
adrianchiris: should we wait for those resources to be gone before returning ?

api/v1/helper.go
```diff
@@ -994,3 +998,27 @@ func (s *SriovNetworkNodeState) ResetKeepUntilTime() bool {
...
+		return &SriovNetwork{}, namespace, name, nil
+
+	case "SriovIBNetwork.sriovnetwork.openshift.io":
+		return &SriovIBNetwork{}, namespace, name, nil
+
+	case "OVSNetwork.sriovnetwork.openshift.io":
+		return &OVSNetwork{}, namespace, name, nil
+	}
+
+	return nil, "", "", fmt.Errorf("unknown ownerRef groupKind %s", groupKind)
```
SchSeba: just nit we can use default here instead of the return right?

api/v1/helper.go
zeeke: I created the function `OwnerRefToString` in the latest changes.

For GVK, I thought the problem was only when creating an object locally. In any case, I think our test cases on the controllers keep us pretty safe.

pkg/consts/constants.go
zeeke: `owner-ref` sounds good

controllers/generic_network_controller.go
zeeke: I'm sure to follow this. Maybe it's getting too complicated:

With the old implementation (before these changes):
- if a NetAttachDef net1 exists and no networkCR is pointing to it, the controller do nothing. This PR maintains this behavior.
- if a NetAttachDef net1 exists and multiple networkCRs are pointing to it, the controller starts contending it in a loop (see attachments). This PR assign the ownership to the first controller it runs (we can call it "randomly"), which is fine, since the NetAttachDef was unstable and not usable.
- If a NetAttachDef net1 exists and one networkCR was pointing to it, then the controller was already overriding it, as it was its owner. This PR explicitly marks the NetAttachDef as owned by the networkCR. 

Did I get the scenario you have in mind?


[sriov-network-operator_multiple_networks.log](https://github.com/user-attachments/files/21142440/sriov-network-operator_multiple_networks.log)
[sriov-networks.yaml.txt](https://github.com/user-attachments/files/21142446/sriov-networks.yaml.txt)

controllers/generic_network_controller_test.go
zeeke: we could, though I didn't get any flake during during tests

api/v1/helper.go
```diff
@@ -994,3 +998,27 @@ func (s *SriovNetworkNodeState) ResetKeepUntilTime() bool {
 	s.SetAnnotations(annotations)
 	return true
 }
+
+func OwnerRefToString(cr client.Object) string {
+	return cr.GetObjectKind().GroupVersionKind().GroupKind().String() + "/" + cr.GetNamespace() + "/" + cr.GetName()
```
adrianchiris, line 1010: maybe we should panic in case the gvk is unset ? just to be on the safe side to catch dev/test errors

controllers/generic_network_controller.go
```diff
@@ -283,61 +274,66 @@ func (r *genericNetworkReconciler) namespaceHandlerCreate(ctx context.Context, e
...
+		if !ok {
+			continue
+		}
 
-	instanceFinalizers := instance.GetFinalizers()
-	if !sriovnetworkv1.StringInArray(sriovnetworkv1.NETATTDEFFINALIZERNAME, instanceFinalizers) {
-		instance.SetFinalizers(append(instanceFinalizers, sriovnetworkv1.NETATTDEFFINALIZERNAME))
-		if err := r.Update(ctx, instance); err != nil {
-			return err
+		obj, namespace, name, err := sriovnetworkv1.StringToOwnerRef(owner)
```
adrianchiris: should we limit the cleanup to specific object kind ? according to the object returned in the `controller.GetObject()`  ?

controllers/generic_network_controller.go
adrianchiris: i see you point. makes sense.

i was referring to the following case where we have:
```
kind: SriovNetwork
metadata:
  name: sriov-net1
  namespace: openshift-sriov-network-operator
spec:
  resourceName: sriovnic
  networkNamespace: default
  ipam: |
    {
      "type": "host-local",
      "subnet": "10.56.217.0/24",
      "rangeStart": "10.56.217.100",
      "rangeEnd": "10.56.217.200",
      "routes": [{
        "dst": "0.0.0.0/0"
      }],
      "gateway": "10.56.217.1"
    }
  vlan: 100
  linkState: enable
  maxTxRate: 1000
  minTxRate: 100
  spoofChk: "on"
  trust: "on"
  capabilities: |
    {
      "mac": true,
      "ips": true
    }
---
kind: SriovNetwork
metadata:
  name: sriov-net1
  namespace: default
spec:
  resourceName: sriovnic
  networkNamespace: default
  ipam: |
    {
      <some other ipam config>
    }
  vlan: 100
  linkState: enable
  maxTxRate: 1000
  minTxRate: 100
  spoofChk: "on"
  trust: "on"
  capabilities: |
    {
      "mac": true,
      "ips": true
    }
---
```

i guess its the same thing and its ok to align to one of the sriovNetworks.

api/v1/helper.go
zeeke: panicking looks too much to me.
I'd rather return a string like `unknown/xxx/yyy`. 

I also tried to implement a smarter way using the CRDs scheme (inspired to [printers/typesetter.go](https://github.com/kubernetes/cli-runtime/blob/f536649dab677f59b7d75aca093d3312c2019cad/pkg/printers/typesetter.go#L41-L58) ), but the code became too big for the problem we're trying to solve here, IMO.

controllers/generic_network_controller.go
zeeke: good idea. The problem is that I didn't find a nice way to check that two pointers have the same underlying type. 
The only one I found is 

```
func isType(a, b interface{}) bool {
    return fmt.Sprintf("%T", a) == fmt.Sprintf("%T", b)
}
```

https://stackoverflow.com/questions/34820469/can-i-compare-variable-types-with-type-in-golang

which is not really nice :(

controllers/generic_network_controller.go
```diff
@@ -88,17 +88,15 @@ func (r *genericNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			return reconcile.Result{}, nil
+			return reconcile.Result{}, r.garbageCollect(ctx)
```
adrianchiris: for network CR instances that are in the operator namespace and point to a different namespace, are we relying on garbageCollect to clean them ? 

i think the approach is problematic, since we are removing finalizers at L97, once the object is deleted, it will be removed from k8s API. if for some reason garbage collect fails and controller is restarted we will never retry.

then we rely on the next delete event to trigger garbage collect (which might take a while or never come ?)

i wonder if we should just keep finalizers on network CR instances or at least for those in the operator namespace then we explicitly delete what we need.

zeeke: sounds good. The first commit is about adding the owner-ref annotation and ensure 2 NetworkObject are not clashing on the same NetAttachDef. The second commit is about using the `owner-ref` annotation on the netattachdef instead of networkobject's finalizers, which I was a sort of bonus change.

I'm dropping the second commit to focus this PR on having the `owner-ref` annotation. We can discuss about dropping the finalizers later, if needed

controllers/generic_network_controller.go
```diff
@@ -189,6 +190,20 @@ func (r *genericNetworkReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 	} else {
 		reqLogger.Info("NetworkAttachmentDefinition CR already exist")
+
+		foundOwner := found.GetAnnotations()[consts.OwnerRefAnnotation]
+		expectedOwner := netAttDef.GetAnnotations()[consts.OwnerRefAnnotation]
+
+		// Note for the future: the `foundOwner != ""` condition can be removed to make the operator not touching the NetworkAttachmentDefinition created
+		// by the user.
+		if foundOwner != "" && foundOwner != expectedOwner {
```
SchSeba, line 199: just a quick question if we have `!equality.Semantic.DeepEqual(found.GetAnnotations(), netAttDef.GetAnnotations()` below do we still need this validation here?

SchSeba (approved):
LGTM

just a small nit, not blocking

api/v1/helper.go
zeeke: this commit has been dropped:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/897#discussion_r2318890962

it was about replacing finalizers delete flow with another based on garbage collecting. Now this PR only avoids multiple network objects overrides the same netAttachDef

controllers/generic_network_controller.go
zeeke: yes, we do:

in case we have:
```
kind: NetworkAttachmentDefinition
metadata:
  name: foo
  namespace: foo-ns
  annotations:
     owner-ref: SriovNetwork/sriov-operator-ns/foo
spec:
  <sriov cni config>
  ```
which comes from an SriovNetwork in the sriov-operator-ns namespace


when reconciling another network object:
```
kind: SriovNetwork
metadata:
  name: foo
  namespace: foo-ns
spec: ...
```

the generated net-attach-def has `owner-ref: SriovNetwork/foo-ns/foo`, which would make the condition  `!equality.Semantic.DeepEqual(found.GetAnnotations(), netAttDef.GetAnnotations()` true and the netAttachDef woudl be overridden.

The policy, as of now, is the first reconciler networkobject is the owner

adrianchiris (approved):

PR #898, opened 2025-06-09
security: Add `NetworkPolicies` to operands

zeeke:
`NetworkPolicy.networking.k8s.io` objects [1] allow defining which network traffic can reach a pod or can go leave a pod, in a kubernated cluster.

In order to reduce the surface attack all the operands that are not hostNetworked, should be covered by NetworkPolicies, explicitly setting the connections they require to work.

[1] https://kubernetes.io/docs/concepts/services-networking/network-policies/

### DIFF
--- a/test/conformance/tests/test_sriov_operator.go
+++ b/test/conformance/tests/test_sriov_operator.go
@@ -17,6 +17,7 @@ import (
 	admission "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkv1 "k8s.io/api/networking/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


test/conformance/tests/test_sriov_operator.go
```diff
@@ -1168,7 +1169,8 @@ var _ = Describe("[sriov] operator", Ordered, func() {
 				assertObjectIsNotFound("network-resources-injector", &rbacv1.ClusterRole{})
 				assertObjectIsNotFound("network-resources-injector-role-binding", &rbacv1.ClusterRoleBinding{})
 				assertObjectIsNotFound("network-resources-injector-config", &admission.MutatingWebhookConfiguration{})
-				assertObjectIsNotFound("nri-control-switches", &corev1.ConfigMap{})
+				assertObjectIsNotFound("network-resources-injector-allow-traffic-api-server", &corev1.Service{})
```
SchSeba: I think this one is not right

test/conformance/tests/test_sriov_operator.go
```diff
@@ -1168,7 +1169,8 @@ var _ = Describe("[sriov] operator", Ordered, func() {
...
+				assertObjectIsNotFound("nri-control-switches", &networkv1.NetworkPolicy{})
```
SchSeba: this one should be ConfigMap no?

test/conformance/tests/test_sriov_operator.go
zeeke: great catch! fixing

SchSeba (approved):
LGTM!

just a small nic in the commit message there is a typo for `kubernated` -> `kubernetes`

SchSeba:
Hi @adrianchiris when you have time please take a look on this one and let me know if we can merge it

PR #899, opened 2025-06-10
Support aws

SchSeba:
## Summary by CodeRabbit

* **New Features**
  * AWS platform support: SR-IOV discovery and management on EC2, plus AWS-specific conformance tests.

* **Improvements**
  * Case-insensitive automatic platform detection from provider IDs.
  * Expanded virtual SR-IOV device discovery and batch configuration flows with improved logging and driver handling.
  * NetFilter docs and outputs standardized to include OpenStack and AWS formats.

* **Tests**
  * Broader unit/integration coverage and platform-gated conformance suites for AWS.

### DIFF
--- a/pkg/host/internal/sriov/sriov.go
+++ b/pkg/host/internal/sriov/sriov.go
@@ -208,6 +208,15 @@ func (s *sriov) SetVfAdminMac(vfAddr string, pfLink, vfLink netlink.Link) error 
 	return nil
 }
 
+// getDeviceClass parses the device class from the device
+func getDeviceClass(device *ghw.PCIDevice) (int64, error) {
+	devClass, err := strconv.ParseInt(device.Class.ID, 16, 64)
+	if err != nil {
+		return 0, fmt.Errorf("unable to parse device class: %v", err)
+	}
+	return devClass, nil
+}
+
 func (s *sriov) DiscoverSriovDevices(storeManager store.ManagerInterface) ([]sriovnetworkv1.InterfaceExt, error) {
 	log.Log.V(2).Info("DiscoverSriovDevices")
 	pfList := []sriovnetworkv1.InterfaceExt{}
@@ -223,7 +232,7 @@ func (s *sriov) DiscoverSriovDevices(storeManager store.ManagerInterface) ([]sri
 	}
 
 	for _, device := range devices {
-		devClass, err := strconv.ParseInt(device.Class.ID, 16, 64)
+		devClass, err := getDeviceClass(device)
 		if err != nil {
 			log.Log.Error(err, "DiscoverSriovDevices(): unable to parse device class, skipping",
 				"device", device)
@@ -326,7 +335,7 @@ func (s *sriov) DiscoverSriovVirtualDevices() ([]sriovnetworkv1.InterfaceExt, er
 	}
 
 	for _, device := range devices {
-		devClass, err := strconv.ParseInt(device.Class.ID, 16, 64)
+		devClass, err := getDeviceClass(device)
 		if err != nil {
 			log.Log.Error(err, "DiscoverSriovVirtualDevices(): unable to parse device class for device, skipping",
 				"device", device)
@@ -915,25 +924,30 @@ func (s *sriov) ConfigSriovDevicesVirtual(storeManager store.ManagerInterface, i
 					"numVfs", ifaceToConfigure.Iface.NumVfs)
 				return errors.New("NumVfs > 1")
 			}
+			// Validate the VFGroups we need to have only one VFGroup
 			if len(ifaceToConfigure.Iface.VfGroups) != 1 {
-				log.Log.Error(nil, "ConfigSriovDeviceVirtual(): missing VFGroup")
-				return errors.New("NumVfs != 1")
+				log.Log.Error(nil, "ConfigSriovDeviceVirtual(): unexpected number of VFGroups", "vfGroups", ifaceToConfigure.Iface.VfGroups)
+				return errors.New("unexpected number of VFGroups")
 			}
+
 			addr := ifaceToConfigure.Iface.PciAddress
-			log.Log.V(2).Info("ConfigSriovDeviceVirtual()", "address", addr)
+			log.Log.V(2).Info("ConfigSriovDeviceVirtual()", "pciAddress", addr)
 			driver := ""
 			vfID := 0
-			for _, group := range ifaceToConfigure.Iface.VfGroups {
-				log.Log.V(2).Info("ConfigSriovDeviceVirtual()", "group", group)
-				if sriovnetworkv1.IndexInRange(vfID, group.VfRange) {
-					log.Log.V(2).Info("ConfigSriovDeviceVirtual()", "indexInRange", vfID)
-					if sriovnetworkv1.StringInArray(group.DeviceType, vars.DpdkDrivers) {
-						log.Log.V(2).Info("ConfigSriovDeviceVirtual()", "driver", group.DeviceType)
-						driver = group.DeviceType
-					}
-					break
-				}
+
+			group := ifaceToConfigure.Iface.VfGroups[0]
+			// Validate the VF ID is in the range for the VFGroup
+			if !sriovnetworkv1.IndexInRange(vfID, group.VfRange) {
+				log.Log.Error(nil, "ConfigSriovDeviceVirtual(): VF ID is not in the range", "vfID", vfID, "vfRange", group.VfRange)
+				return errors.New("VF ID is not in the range")
 			}
+
+			// check if we need to bind to a DPDK driver
+			if sriovnetworkv1.StringInArray(group.DeviceType, vars.DpdkDrivers) {
+				log.Log.V(2).Info("ConfigSriovDeviceVirtual()", "driver", group.DeviceType)
+				driver = group.DeviceType
+			}
+
 			if driver == "" {
 				log.Log.V(2).Info("ConfigSriovDeviceVirtual(): bind default")
 				if err := s.kernelHelper.BindDefaultDriver(addr); err != nil {
--- a/pkg/host/internal/sriov/sriov_test.go
+++ b/pkg/host/internal/sriov/sriov_test.go
@@ -1171,7 +1171,7 @@ var _ = Describe("SRIOV", func() {
 					LinkAdminState: "down",
 				}})
 			Expect(err).To(HaveOccurred())
-			Expect(err.Error()).To(ContainSubstring("NumVfs != 1"))
+			Expect(err.Error()).To(ContainSubstring("unexpected number of VFGroups"))
 		})
 
 		It("should fail when BindDefaultDriver fails", func() {
@@ -1423,10 +1423,7 @@ var _ = Describe("SRIOV", func() {
 				}})).NotTo(HaveOccurred())
 		})
 
-		It("should configure with default driver when VfRange doesn't include vfID 0", func() {
-			hostMock.EXPECT().BindDefaultDriver("0000:d8:00.0").Return(nil)
-			storeManagerMode.EXPECT().SaveLastPfAppliedStatus(gomock.Any()).Return(nil)
-
+		It("should return error when VfRange doesn't include vfID 0", func() {
 			Expect(s.ConfigSriovDevicesVirtual(storeManagerMode,
 				[]sriovnetworkv1.Interface{{
 					Name:       "enp216s0f0np0",
@@ -1443,7 +1440,7 @@ var _ = Describe("SRIOV", func() {
 					PciAddress:     "0000:d8:00.0",
 					NumVfs:         1,
 					LinkAdminState: "down",
-				}})).NotTo(HaveOccurred())
+				}})).To(HaveOccurred())
 		})
 
 		It("should configure with default driver when DeviceType is not a DPDK driver", func() {
@@ -1489,30 +1486,6 @@ var _ = Describe("SRIOV", func() {
 					PciAddress: "0000:d8:00.0",
 					NumVfs:     1,
 					Mtu:        1500,
-				}})).NotTo(HaveOccurred())
-		})
-
-		It("should configure when EswitchMode needs update", func() {
-			hostMock.EXPECT().BindDpdkDriver("0000:d8:00.0", "vfio-pci").Return(nil)
-			storeManagerMode.EXPECT().SaveLastPfAppliedStatus(gomock.Any()).Return(nil)
-
-			Expect(s.ConfigSriovDevicesVirtual(storeManagerMode,
-				[]sriovnetworkv1.Interface{{
-					Name:        "enp216s0f0np0",
-					PciAddress:  "0000:d8:00.0",
-					NumVfs:      1,
-					EswitchMode: "switchdev",
-					VfGroups: []sriovnetworkv1.VfGroup{{
-						VfRange:      "0-0",
-						ResourceName: "test-resource0",
-						PolicyName:   "test-policy0",
-						DeviceType:   "vfio-pci",
-					}},
-				}},
-				[]sriovnetworkv1.InterfaceExt{{
-					PciAddress:  "0000:d8:00.0",
-					NumVfs:      1,
-					EswitchMode: "legacy",
 				}})).NotTo(HaveOccurred())
 		})
 
--- a/pkg/platform/aws/aws.go
+++ b/pkg/platform/aws/aws.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
 
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -17,24 +18,34 @@ import (
 const (
 	// metadataBaseURL is the base URL for the EC2 instance metadata service.
 	metadataBaseURL = "http://169.254.169.254/latest/meta-data/"
-	// macsPath is the path to list MAC addresses. Note the trailing slash,
-	// which is standard for directory-like listings in the metadata service.
+	// macsPath is the path to list MAC addresses.
 	macsPath = "network/interfaces/macs/"
-	// subnetIDSuffix is appended to a specific MAC address path to get its subnet ID.
-	subnetIDSuffix = "/subnet-id"
+	// subnetIDPath is the path suffix for fetching subnet ID from metadata service.
+	subnetIDPath = "subnet-id"
+	// awsNetworkIDPrefix is the prefix used for AWS network ID in NetFilter field.
+	// Format: "aws/NetworkID:<subnet-id>"
+	awsNetworkIDPrefix = "aws/NetworkID:"
 )
 
+// Aws implements the platform.Interface for AWS EC2 instances.
+// It handles SR-IOV device discovery for virtual functions in AWS environments,
+// using metadata from the EC2 instance metadata service.
 type Aws struct {
-	hostHelpers       helper.HostHelpersInterface
-	loadedDevicesInfo sriovnetworkv1.InterfaceExts
+	hostHelpers        helper.HostHelpersInterface
+	InitialDevicesInfo sriovnetworkv1.InterfaceExts
 }
 
+// New creates a new Aws platform instance.
+// Returns a configured Aws platform or an error if initialization fails.
 func New(hostHelpers helper.HostHelpersInterface) (*Aws, error) {
 	return &Aws{
 		hostHelpers: hostHelpers,
 	}, nil
 }
 
+// Init initializes the AWS platform by loading device information.
+// If a checkpoint exists, it loads device info from the saved node state.
+// Otherwise, it queries the EC2 instance metadata service.
 func (a *Aws) Init() error {
 	ns, err := a.hostHelpers.GetCheckPointNodeState()
 	if err != nil {
@@ -50,10 +61,13 @@ func (a *Aws) Init() error {
 	return nil
 }
 
+// Name returns the name of the AWS platform.
 func (a *Aws) Name() string {
 	return string(consts.AWS)
 }
 
+// GetVendorPlugins returns the virtual plugin as the main plugin for AWS.
+// AWS platforms only use the virtual plugin, with no additional plugins.
 func (a *Aws) GetVendorPlugins(_ *sriovnetworkv1.SriovNetworkNodeState) (plugin.VendorPlugin, []plugin.VendorPlugin, error) {
 	virtual, err := virtualplugin.NewVirtualPlugin(a.hostHelpers)
 	return virtual, []plugin.VendorPlugin{}, err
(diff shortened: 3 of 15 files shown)


SchSeba:
depend on https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/902
/hold

pkg/consts/platforms.go
```diff
@@ -0,0 +1,43 @@
...
+	VirtualAWS
+)
+
+func (e PlatformTypes) String() string {
+	switch e {
+	case Baremetal:
+		return "Baremetal"
+	case VirtualOpenStack:
+		return "Virtual/Openstack"
+	case VirtualAWS:
```
e0ne: nit: we can leave just AWS

pkg/consts/platforms.go
```diff
@@ -0,0 +1,43 @@
...
+	VirtualOpenStack
+	// VirtualAWS platform
+	VirtualAWS
+)
+
+func (e PlatformTypes) String() string {
+	switch e {
+	case Baremetal:
+		return "Baremetal"
+	case VirtualOpenStack:
```
e0ne: nit: we can omit 'Virtual' prefix

pkg/orchestrator/openshift/openshift.go
```diff
@@ -84,33 +58,24 @@ func New() (OpenshiftContextInterface, error) {
...
-	// if it's not an openshift cluster we just return true that the operator manage to drain the node
-	if !c.IsOpenshiftCluster() {
-		return true, nil
-	}
-
+func (c *OpenshiftContext) BeforeDrainNode(ctx context.Context, node *corev1.Node) (bool, error) {
 	// if the operator is running on hypershift variation of openshift there is no machine config operator
 	// just return true here
-	if c.IsHypershift() {
+	if c.Flavor() == consts.ClusterFlavorHypershift {
```
e0ne, line 86: I prefer to leave `IsHypershift` method to make if statements less complex

pkg/vars/vars.go
```diff
@@ -84,7 +85,7 @@ var (
 func init() {
 	Namespace = os.Getenv("NAMESPACE")
 
-	ClusterType = os.Getenv("CLUSTER_TYPE")
+	ClusterType = consts.ClusterType(os.Getenv("CLUSTER_TYPE"))
```
e0ne, line 93: Why do we have a function in `consts' package?

pkg/daemon/daemon.go
```diff
@@ -32,86 +32,69 @@ import (
...
+	platformInterface platform.Interface
 
 	eventRecorder *EventRecorder
 
 	featureGate featuregate.FeatureGate
 
-	// list of disabled plugins
-	disabledPlugins []string
+	additionalPlugins []plugin.VendorPlugin
+	mainPlugin        plugin.VendorPlugin
```
e0ne, line 43: What is main plugin? Let's use 'defalutPlugin` as an environment name

pkg/daemon/plugin.go
```diff
@@ -1,108 +1,40 @@
...
+	mainPlugin, additionalPlugins, err := dn.platformInterface.GetPlugins(ns)
+	if err != nil {
+		funcLog.Error(err, "Failed to load plugins", "platform", vars.PlatformType, "orchestrator", vars.ClusterType)
 	}
 
-	pluginList := make([]string, 0, len(loadedPlugins))
-	for pluginName := range loadedPlugins {
-		pluginList = append(pluginList, pluginName)
+	// check for the main plugin disabled
+	// TODO: we really want to allow the disable of the main plugin?
```
e0ne: It's not clear what is main plugin? Can we rename it to defaultPlugin or genericPlugin?

pkg/daemon/plugin.go
```diff
@@ -1,108 +1,40 @@
...
-		genericPlugin, err := GenericPlugin(helpers)
-		if err != nil {
-			log.Log.Error(err, "loadPlugins(): failed to load the generic plugin")
-			return nil, err
-		}
-		pluginName := genericPlugin.Name()
-		if !isPluginDisabled(pluginName, disabledPlugins) {
-			loadedPlugins[pluginName] = genericPlugin
-		}
+	mainPlugin, additionalPlugins, err := dn.platformInterface.GetPlugins(ns)
```
e0ne: Why did we rename vendorPlugins to additionalPlugins?

pkg/utils/utils.go
```diff
@@ -3,21 +3,30 @@ package utils
...
 
+const (
+	requestTimeout = 5 * time.Second
+)
+
 //go:generate ../../bin/mockgen -destination mock/mock_utils.go -source utils.go
 type CmdInterface interface {
 	Chroot(string) (func() error, error)
 	RunCommand(string, ...string) (string, string, error)
+	HTTPGetFetchData(string) (string, error)
```
e0ne, line 29: It shouldn't be a part of `CmdInterface`

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,188 @@
...
+			Mac:        iface.Mac,
+		}
+		a.loadedDevicesInfo[idx] = iface
+	}
+	return a.loadedDevicesInfo, nil
+}
+
+func (a *Aws) DiscoverBridges() (sriovnetworkv1.Bridges, error) {
+	if vars.ManageSoftwareBridges {
+		return sriovnetworkv1.Bridges{}, fmt.Errorf("bridge configuration not supported on openstack platform")
```
e0ne: openstack -> AWS

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	return sriovnetworkv1.Bridges{}, nil
+}
+
+// CreateOpenstackDevicesInfo create the openstack device info map
```
e0ne: Openstack -> AWS

pkg/platform/aws/aws_test.go
```diff
@@ -0,0 +1,223 @@
...
+			Expect(err).To(MatchError("aws platform not supported in systemd"))
+		})
+	})
+
+	Context("Bridge Discovery", func() {
+		It("should return an error if ManageSoftwareBridges is true", func() {
+			vars.ManageSoftwareBridges = true
+			defer func() { vars.ManageSoftwareBridges = false }() // Cleanup
+			_, err := awsPlatform.DiscoverBridges()
+			Expect(err).To(MatchError("bridge configuration not supported on openstack platform"))
```
e0ne: openstack -> AWS

e0ne (changes requested):

pkg/platform/aws/aws.go
SchSeba: done

pkg/platform/aws/aws.go
SchSeba: done

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return err
+	}
+
+	if ns == nil {
+		err = a.createDevicesInfo()
+		return err
+	}
+
+	a.createDevicesInfoFromNodeStatus(ns)
```
zeeke, line 60: if `ns == nil`, this call produces a nil exception.

isn't it supposed to be in the `else` branch?

pkg/host/internal/sriov/sriov.go
```diff
@@ -311,6 +311,63 @@ func (s *sriov) DiscoverSriovDevices(storeManager store.ManagerInterface) ([]sri
 	return pfList, nil
 }
 
+func (s *sriov) DiscoverSriovVirtualDevices() ([]sriovnetworkv1.InterfaceExt, error) {
```
zeeke, line 323: Can we rename this to `DiscoverSriovVirtualPF`?

pkg/platform/aws/aws.go
SchSeba: if it's nil we create and return

pkg/host/internal/sriov/sriov.go
SchSeba: I change it to `DiscoverSriovVirtualDevices` let me know if that is fine

zeeke: ok, but it's the same name as before. :smiley:

SchSeba: yep I wanted it to continue and be similiar to `DiscoverSriovDevices` just for virtual.

if you still prefer to change let me know and I will  no problem :)

zeeke: ok, looks good

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,194 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/helper"
+	plugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins"
+	virtualplugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/virtual"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
```
adrianchiris, line 17: nit: add docstrings throughout

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,194 @@
...
+		a.loadedDevicesInfo[idx] = iface
+	}
+	return a.loadedDevicesInfo, nil
+}
+
+// DiscoverBridges is not supported on AWS platform.
+// Returns ErrOperationNotSupportedByPlatform as AWS does not support software bridge management.
+func (a *Aws) DiscoverBridges() (sriovnetworkv1.Bridges, error) {
+	if vars.ManageSoftwareBridges {
+		return sriovnetworkv1.Bridges{}, vars.ErrOperationNotSupportedByPlatform
```
adrianchiris: shouldnt we always return this ? like we do in openstack platform

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return sriovnetworkv1.Bridges{}, nil
+}
+
+// createDevicesInfo creates the AWS device info map
+// This function is used to create the AWS device info map from the metadata server.
+func (a *Aws) createDevicesInfo() error {
+	funcLog := log.Log.WithName("getDataFromMetadataService")
```
adrianchiris: nit: createDevicesInfo ?

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,194 @@
...
+	a.loadedDevicesInfo = make(sriovnetworkv1.InterfaceExts, 0)
+	funcLog.Info("getting aws network info from metadata server")
+	metaData, err := a.hostHelpers.HTTPGetFetchData(metadataBaseURL + macsPath)
```
adrianchiris: nit: you can use url lib : https://pkg.go.dev/net/url#JoinPath

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return fmt.Errorf("error getting aws meta_data from %s: %v", metadataBaseURL+macsPath, err)
+	}
+
+	// If the endpoint returns an empty body (e.g., no MACs available or an issue),
+	// treat it as no MACs found rather than an error. The caller can then decide how to handle an empty list.
```
adrianchiris, line 153: maybe add logging here as there would be no devices loaded

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,194 @@
...
+	// MAC addresses are returned separated by newlines, each ending with a '/'.
+	// Example raw response: "0e:1a:95:aa:12:a1/\n0e:92:d3:ee:52:1b/"
+	macEntries := strings.Split(metaData, "\n")
+	if len(macEntries) == 0 {
+		return nil
+	}
+
+	macAddressToSubNetID := map[string]string{}
+	for _, macEntry := range macEntries {
+		subnetIDURL := metadataBaseURL + macsPath + macEntry + subnetIDSuffix
```
adrianchiris: same as above (use url.JoinPath) then you can strip any "/" from mac entries at L141 insead of at L157 if u like

adrianchiris: also can you add in comment an example raw response

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	for _, iface := range pfList {
+		subnetID, exist := macAddressToSubNetID[iface.Mac]
+		if !exist {
+			continue
+		}
+
+		subnetID = strings.TrimPrefix(subnetID, "subnet-")
+		iface.NetFilter = fmt.Sprintf("aws/NetworkID:%s", subnetID)
```
adrianchiris: not related to this but i saw we have `OpenstackNetworkID` defined in `api/v1/helper.go` perhaps its worth moving it to openstack platform impl.

also perhaps we should have aws/NetworkID as const in the top of the file ?

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,194 @@
...
+	return nil, vars.ErrOperationNotSupportedByPlatform
+}
+
+// DiscoverSriovDevices discovers VFs on a virtual platform
+func (a *Aws) DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
+	funcLog := log.Log.WithName("DiscoverSriovDevices")
+	log.Log.V(2).Info("discovering sriov devices")
+
+	// TODO: check if we want to support hot plug here in the future
+	for idx, iface := range a.loadedDevicesInfo {
```
adrianchiris: for cleanliness should we generate a (deep) copy of the loadedDevicesInfo and perform updates on the copy ?

else caller can (accidently) alter fields in the returned slice as its only shallow copied

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,194 @@
...
+		hasDriver, driver := a.hostHelpers.HasDriver(iface.PciAddress)
+		if !hasDriver {
+			funcLog.Error(nil, "device doesn't have a driver, skipping",
```
adrianchiris: since we skip, this means that for this iface entry we return whatever was stored in `a.loadedDevicesInfo` is that what we want ?

pkg/platform/aws/aws.go
adrianchiris, line 60: why do we want two flows here ? 

eventually entires are updated when calling DiscoverSriovDevices

pkg/platform/aws/aws.go
SchSeba: good catch!
I remove this check, we don't really need this one we can allow the drvier to be empty if that is the case.

pkg/platform/aws/aws.go
SchSeba: This is the same as in case of openstack.
in case the sriov-config-daemon gets restated and the VF is allocated to a pod we will not create it in the nodeState. that is why the check file was implemented.

pkg/platform/aws/aws.go
SchSeba: done

pkg/platform/aws/aws.go
SchSeba: done

SchSeba:
Hi @adrianchiris when you have time please give another look :)

pkg/platform/aws/aws.go
adrianchiris: the network devices there are VF only without PF right ? this is why we advertise a device with as single VF ?

adrianchiris: right! 10x for the refresher.

pkg/host/internal/sriov/sriov.go
```diff
@@ -311,6 +311,63 @@ func (s *sriov) DiscoverSriovDevices(storeManager store.ManagerInterface) ([]sri
...
+		return nil, fmt.Errorf("DiscoverSriovVirtualDevices(): error getting PCI info: %v", err)
+	}
+
+	devices := pci.Devices
+	if len(devices) == 0 {
+		return nil, fmt.Errorf("DiscoverSriovVirtualDevices(): could not retrieve PCI devices")
+	}
+
+	for _, device := range devices {
+		devClass, err := strconv.ParseInt(device.Class.ID, 16, 64)
```
adrianchiris: should we move this check to a separate func and call it here and in DiscoverSriovDevcies ?

e.g `GetDeviceClass(device)`

not a must by any means. fine to keep as is.

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,222 @@
...
+
+	pfList, err := a.hostHelpers.DiscoverSriovVirtualDevices()
+	if err != nil {
+		return err
+	}
+
+	for _, iface := range pfList {
+		subnetID, exist := macAddressToSubNetID[iface.Mac]
+		if !exist {
+			continue
```
adrianchiris, line 197: do we want to log anything here ?

pkg/host/internal/sriov/sriov.go
```diff
@@ -843,49 +900,98 @@ func (s *sriov) checkForConfigAndReset(ifaceStatus sriovnetworkv1.InterfaceExt,
...
+			if ifaceToConfigure.Iface.NumVfs > 1 {
+				log.Log.Error(nil, "ConfigSriovDeviceVirtual(): in a virtual environment, only one VF per interface",
+					"numVfs", ifaceToConfigure.Iface.NumVfs)
+				return errors.New("NumVfs > 1")
+			}
+			if len(ifaceToConfigure.Iface.VfGroups) != 1 {
+				log.Log.Error(nil, "ConfigSriovDeviceVirtual(): missing VFGroup")
+				return errors.New("NumVfs != 1")
+			}
+			addr := ifaceToConfigure.Iface.PciAddress
```
adrianchiris, line 933: i would add a check here that vf index 0  is part of the first (and only) vf group
then below proceed with the configuration of the single VF we have and save some indentation below WDYT ?

pkg/host/internal/sriov/sriov_test.go
```diff
@@ -1033,6 +1033,633 @@ var _ = Describe("SRIOV", func() {
...
+					}},
+				}},
+				[]sriovnetworkv1.InterfaceExt{{
+					PciAddress: "0000:d8:00.0",
+					NumVfs:     1,
+					Mtu:        1500,
+				}})).NotTo(HaveOccurred())
+		})
+
+		It("should configure when EswitchMode needs update", func() {
```
adrianchiris: should it configure ? if eswitchmode was provided in spec, do we want to fail ?

test/conformance/tests/test_aws_platform.go
```diff
@@ -0,0 +1,135 @@
...
+	})
+
+	AfterAll(func() {
+		err := namespaces.Clean(operatorNamespace, namespaces.Test, clients, discovery.Enabled())
+		Expect(err).ToNot(HaveOccurred())
+		WaitForSRIOVStable()
+	})
+
+	Describe("Generic SriovNetworkNodePolicy", func() {
+		It("should configure vf in vfio-pci mode", func() {
```
adrianchiris, line 116: do we want a test for non vfio-pci ?

test/util/cluster/cluster.go
```diff
@@ -371,29 +372,47 @@ func SetDisableNodeDrainState(clients *testclient.ClientSet, operatorNamespace s
...
+	}
+	if err != nil {
+		return false, fmt.Errorf("secureboot check error stdout:[%s] stderr:[%s]: %w", stdout, stderr, err)
+	}
+
+	return strings.Contains(stdout, "[integrity]") || strings.Contains(stdout, "[confidentiality]"), nil
+}
+
+// createPodOnNodeWithHostPath creates a pod on the given node with the host path mounted
+func createPodOnNode(clients *testclient.ClientSet, nodeName, namespace string) (*corev1.Pod, error) {
```
adrianchiris, line 399: do we want to pass in host path ? or the comment is wrong ?

test/util/cluster/cluster.go
```diff
@@ -405,24 +424,112 @@ func GetNodeSecureBootState(clients *testclient.ClientSet, nodeName, namespace s
...
+		return consts.Baremetal, fmt.Errorf("no nodes found in the selector")
+	}
+
+	nodeObj := &corev1.Node{}
+	err = clients.Get(context.Background(), runtimeclient.ObjectKey{Name: ss[0].Name}, nodeObj)
+	if err != nil {
+		return consts.Baremetal, fmt.Errorf("failed to get node %s %v", ss[0].Name, err)
+	}
+
+	for key, pType := range vars.PlatformsMap {
```
adrianchiris: nit: consider moving this logic to a common func and use it here, and in config daemon `start.go`

adrianchiris (commented):
should we have any docs around this feature ?

pkg/platform/aws/aws.go
SchSeba: yep this are only VFs in this case.

pkg/host/internal/sriov/sriov.go
SchSeba: sure no problem.

done

pkg/host/internal/sriov/sriov.go
SchSeba: I like the idea!

pkg/platform/aws/aws.go
SchSeba: done

test/conformance/tests/test_aws_platform.go
SchSeba: I will add a test BUT we need to remember this is going to use the host-device CNI as the sriov-cni doesn't support moving VFs without access to the PF.

I check both the OCP and the k8s CI lanes contains the containers-cni package so we should be ok

test/util/cluster/cluster.go
SchSeba: I update the string.
this is a pod with host Network hostpath and privilege

test/util/cluster/cluster.go
SchSeba: nice idea!
done

pkg/host/internal/sriov/sriov_test.go
SchSeba: I removed this test. we don't do anything related to eswitch in case of virtual devices.

In a future PR I am going to update the validation

SchSeba:
Hi @adrianchiris done with the comments :)

I hope it's ready now to get merge.

pkg/platform/aws/aws.go
```diff
@@ -0,0 +1,223 @@
...
+// Returns a copy of the device info to prevent callers from modifying internal state.
+func (a *Aws) DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
+	log.Log.V(2).Info("discovering sriov devices")
+
+	result := make([]sriovnetworkv1.InterfaceExt, 0, len(a.InitialDevicesInfo))
+
+	// TODO: check if we want to support hot plug here in the future
+	for _, iface := range a.InitialDevicesInfo {
+		// Create a copy to avoid modifying the original
+		ifaceCopy := iface
```
adrianchiris: we should deep copy it imo just to be on the safe side. i.e call `iface.DeepCopy()`

only VFs slice will be shallow copied and we override it below so current impl is ok but using deep copy would be safer imo (if in the future this struct is extended).

pkg/platform/openstack/openstack.go
```diff
@@ -30,6 +30,9 @@ const (
 	// with the "config-2" label.
 	//https://docs.openstack.org/nova/latest/user/config-drive.html
 	configDriveLabel = "config-2"
+	// openstackNetworkIDPrefix is the prefix used for OpenStack network ID in NetFilter field.
+	// Format: "openstack/NetworkID:<network-uuid>"
+	openstackNetworkIDPrefix = "openstack/NetworkID:"
```
adrianchiris, line 35: do we need now in `api/v1/helper.go` the following ?:

```
// NetFilterType Represents the NetFilter tags to be used
type NetFilterType int
const (
	// OpenstackNetworkID network UUID
	OpenstackNetworkID NetFilterType = iota
)

...


func (e NetFilterType) String() string {
	switch e {
	case OpenstackNetworkID:
		return "openstack/NetworkID"
	default:
		return fmt.Sprintf("%d", int(e))
	}
}
```

since netfilter has a specific format perhaps it makes sense to have a generic NetFilterType struct that would be used to construct the the netfilter identifier for the interface in openstack and aws platform

adrianchiris: also we should update the docstring in sriovnetworknodepolicy_types.go:

```
	// Infrastructure Networking selection filter. Allowed value "openstack/NetworkID:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
	NetFilter string `json:"netFilter,omitempty"`
```

test/conformance/tests/test_aws_platform.go
SchSeba: done I added a netdevice test also using the host-device cni

pkg/platform/aws/aws.go
SchSeba: done

pkg/platform/openstack/openstack.go
SchSeba: I removed the netfilter and add the requested comment.

if that is fine lets leave out the `NetFilterType` and we can implement that in the future if needed.

I am going to open a following up PR to add validation for both virtual platform and I can add it there

adrianchiris: ack

zeeke (approved):

PR #902, opened 2025-06-11
Create platform and orchestrator packages

SchSeba:


cmd/sriov-network-config-daemon/service.go
```diff
@@ -234,36 +246,7 @@ func (s *ServiceConfig) callPlugin(phase string) error {
...
-	case consts.VirtualOpenStack:
-		switch phase {
-		case PhasePre:
-			configPlugin, err = newVirtualPluginFunc(s.hostHelper)
-			if err != nil {
-				return nil, fmt.Errorf("failed to create virtual plugin %v", err)
-			}
-		case PhasePost:
-			s.log.Info("skip post configuration phase for virtual cluster")
-			return nil, nil
```
zeeke, line 263: The virtualOpenStack case has been dropped. The 
OpenStack.SystemdGetPlugins function is a stub
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/902/commits/838fc5e764f030c51bc31707d0ae3d1e485c7962#diff-a4d9507a11b9b1d03919ccbe8546be858a1f01459e7f4a9c65d5e4da28e3a053R76

pkg/daemon/plugin.go
```diff
@@ -1,108 +1,40 @@
...
-
-		if vars.ClusterType != consts.ClusterTypeOpenshift {
-			k8sPlugin, err := K8sPlugin(helpers)
-			if err != nil {
-				log.Log.Error(err, "loadPlugins(): failed to load the k8s plugin")
-				return nil, err
-			}
+func (dn *NodeReconciler) loadPlugins(ns *sriovnetworkv1.SriovNetworkNodeState, disabledPlugins []string) error {
+	funcLog := log.Log.WithName("loadPlugins").WithValues("platform", vars.PlatformType, "orchestrator", vars.ClusterType)
+	funcLog.Info("loadPlugins(): loading plugins")
```
zeeke: nit: reduce nois

```suggestion
	funcLog.Info("loading plugins", "disabled", disabledPlugins)
```

pkg/daemon/plugin.go
```diff
@@ -1,108 +1,40 @@
...
 		}
 	}
 
-	return vendorPlugins, nil
+	additionalPluginsName := make([]string, len(dn.additionalPlugins))
+	for idx, plugin := range dn.additionalPlugins {
+		additionalPluginsName[idx] = plugin.Name()
+	}
+
+	log.Log.Info("loadPlugins(): loaded plugins", "mainPlugin", dn.mainPlugin.Name(), "additionalPlugins", additionalPluginsName)
```
zeeke:
```suggestion
	log.Log.Info("loaded plugins", "mainPlugin", dn.mainPlugin.Name(), "additionalPlugins", additionalPluginsName)
```

pkg/host/internal/sriov/sriov.go
```diff
@@ -929,6 +929,63 @@ func (s *sriov) GetLinkType(name string) string {
 	return s.encapTypeToLinkType(link.Attrs().EncapType)
 }
 
+func (s *sriov) GetInterfaces() ([]sriovnetworkv1.InterfaceExt, error) {
```
zeeke: This function is never called. is it a leftover?

SchSeba:
Hi @adrianchiris @zeeke @e0ne as requested here is the design proposal for this feature https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/917
please review it when you have the time thanks!

pkg/host/internal/sriov/sriov.go
SchSeba: right nice catch!

cmd/sriov-network-config-daemon/service.go
SchSeba: you are right!

cmd/sriov-network-config-daemon/service.go
```diff
@@ -59,36 +54,52 @@ var (
 	}
 	phaseArg string
 
-	newGenericPluginFunc  = generic.NewGenericPlugin
-	newVirtualPluginFunc  = virtual.NewVirtualPlugin
-	newHostHelpersFunc    = helper.NewDefaultHostHelpers
-	newPlatformHelperFunc = platforms.NewDefaultPlatformHelper
+	newPlatformFunc = func(hostHelpers helper.HostHelpersInterface) (platform.Interface, error) {
```
zeeke: we can avoid this inline function
```suggestion
	newPlatformFunc    = platform.New
```

SchSeba: nice catch! done

pkg/consts/platforms.go
```diff
@@ -0,0 +1,39 @@
...
+type ClusterFlavor string
+
+const (
+	ClusterFlavorVanillaK8s ClusterFlavor = "vanilla"
+	ClusterFlavorOpenshift  ClusterFlavor = "default"
+	ClusterFlavorHypershift ClusterFlavor = "hypershift"
+)
+
+// PlatformTypes
+type PlatformTypes int
```
adrianchiris: should we be consistent with the above consts and just have this as a string ? then we can drop the String() method ?

pkg/host/manager.go
```diff
@@ -50,6 +50,11 @@ type hostManager struct {
 	types.SystemdInterface
 }
 
+func NewDefaultHostManager() (HostManagerInterface, error) {
```
adrianchiris, line 53: who is going to use this one ? i see its not used in this PR

pkg/utils/utils.go
```diff
@@ -3,21 +3,30 @@ package utils
...
+	"time"
 
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
 )
 
+const (
+	requestTimeout = 5 * time.Second
```
adrianchiris: nit: httpRequestTimeout ?

pkg/utils/utils.go
```diff
@@ -49,6 +59,36 @@ func (u *utilsHelper) Chroot(path string) (func() error, error) {
 	}, nil
 }
 
+func (u *utilsHelper) HTTPGetFetchData(url string) (string, error) {
```
adrianchiris, line 62: who is using this one ?

pkg/daemon/status.go
```diff
@@ -98,32 +96,25 @@ func (dn *NodeReconciler) shouldUpdateStatus(current, desiredNodeState *sriovnet
...
-			}
-		}
+	funcLog := log.Log.WithName("updateStatusFromHost")
+	funcLog.Info("Getting host network status")
+	ifaces, err := dn.platformInterface.DiscoverSriovDevices()
+	if err != nil {
+		funcLog.Error(err, "failed to discover sriov devices")
+		return err
+	}
+	bridges, err := dn.platformInterface.DiscoverBridges()
```
adrianchiris: i wonder if we should should check `vars.ManagedSoftwareBridges` here and conditionally call discoverBridges() method
then the platform implementation can just fail if not supported.

pkg/daemon/status.go
```diff
@@ -98,32 +96,25 @@ func (dn *NodeReconciler) shouldUpdateStatus(current, desiredNodeState *sriovnet
...
+		funcLog.Error(err, "failed to discover bridges")
+		return err
 	}
 
 	nodeState.Status.Interfaces = ifaces
 	nodeState.Status.Bridges = bridges
-	nodeState.Status.System.RdmaMode, err = dn.HostHelpers.DiscoverRDMASubsystem()
+	nodeState.Status.System.RdmaMode, err = dn.hostHelpers.DiscoverRDMASubsystem()
+	if err != nil {
+		funcLog.Error(err, "failed to discover rdma subsystem")
```
adrianchiris, line 121: wdyt about returning err here and then nil below for consistency with other func calls above.

not critical by any means

pkg/daemon/plugin.go
```diff
@@ -1,108 +1,41 @@
...
-		loadedPlugins = loadedVendorPlugins
-
-		if vars.ClusterType != consts.ClusterTypeOpenshift {
-			k8sPlugin, err := K8sPlugin(helpers)
-			if err != nil {
-				log.Log.Error(err, "loadPlugins(): failed to load the k8s plugin")
-				return nil, err
-			}
+func (dn *NodeReconciler) loadPlugins(ns *sriovnetworkv1.SriovNetworkNodeState, disabledPlugins []string) error {
+	funcLog := log.Log.WithName("loadPlugins").WithValues("platform", vars.PlatformType, "orchestrator", vars.ClusterType)
```
adrianchiris, line 13: WDYT about having a `Name()` method for the platform and orchestrator interfaces ? then it can be used here ?

pkg/platform/platform.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platform/baremetal"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platform/openstack"
+	plugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+//go:generate ../../bin/mockgen -destination mock/mock_platform.go -source platform.go
+type Interface interface {
+	Init() error
+	GetHostHelpers() helper.HostHelpersInterface
```
adrianchiris: why do we need this method? i see all use the same host helper

pkg/platform/platform.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/helper"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platform/baremetal"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platform/openstack"
+	plugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+//go:generate ../../bin/mockgen -destination mock/mock_platform.go -source platform.go
+type Interface interface {
```
adrianchiris, line 19: please add detailed docstring for the interface and all of its methods

pkg/daemon/plugin.go
```diff
@@ -1,108 +1,41 @@
...
+	if err != nil {
+		funcLog.Error(err, "Failed to load plugins", "platform", vars.PlatformType, "orchestrator", vars.ClusterType)
+		return err
 	}
 
-	pluginList := make([]string, 0, len(loadedPlugins))
-	for pluginName := range loadedPlugins {
-		pluginList = append(pluginList, pluginName)
+	// check for the main plugin disabled
+	// TODO: we really want to allow the disable of the main plugin?
```
adrianchiris: we dont, we should fail if disabledPlugins contain the main plugin IMO. WDYT?

adrianchiris: erroring out would be dead code same as L23 as we check if plugin can be disabled in daemon cmd start.go.
but i think its fine to add this extra protection at this level

pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/orchestrator/openshift"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+//go:generate ../../bin/mockgen -destination mock/mock_orchestrator.go -source orchestrator.go
+type Interface interface {
+	// ClusterType represent to a cluster type
+	ClusterType() consts.ClusterType
+	// Flavor represents the internal flavor of the cluster
+	Flavor() consts.ClusterFlavor
```
adrianchiris, line 32: add: if the implementation does not have a flavor then `consts.ClusterFlavorDefault` flavor should be returned

pkg/consts/platforms.go
```diff
@@ -0,0 +1,39 @@
...
+
+const (
+	ClusterTypeOpenshift  ClusterType = "openshift"
+	ClusterTypeKubernetes ClusterType = "kubernetes"
+)
+
+type ClusterFlavor string
+
+const (
+	ClusterFlavorVanillaK8s ClusterFlavor = "vanilla"
```
adrianchiris: do we really need the vanilla flavor ? of cluster type kubernetes there is a single flavor

IMO we can:
1. drop `ClusterFlavorVanillaK8s`
2. rename `ClusterFlavorOpenshift` to `ClusterFlavorDefault`
3. use `ClusterFlavorDefault` in both openshift (non hypershift) and kubernetes orchestrator implementations

pkg/consts/platforms.go
```diff
@@ -0,0 +1,39 @@
+package consts
+
+import "fmt"
+
+type ClusterType string
```
adrianchiris, line 4: nit: add docstrings throughout this file

pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/orchestrator/kubernetes"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/orchestrator/openshift"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+//go:generate ../../bin/mockgen -destination mock/mock_orchestrator.go -source orchestrator.go
+type Interface interface {
+	// ClusterType represent to a cluster type
```
adrianchiris: the sentence is not coherent.

pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,36 @@
...
+	ClusterType() consts.ClusterType
+	// Flavor represents the internal flavor of the cluster
+	Flavor() consts.ClusterFlavor
+	// BeforeDrainNode run a platform-specific logic before draining the node
```
adrianchiris: can you add more info please when this and the below is called ?

cmd/sriov-network-config-daemon/service.go
```diff
@@ -59,36 +54,50 @@ var (
...
+	err = sc.readConf()
+	if err != nil {
+		return nil, sc.updateSriovResultErr(phaseArg, err)
+	}
+
+	// init globals
+	vars.PlatformType = sc.sriovConfig.PlatformType
+	platformInterface, err := newPlatformFunc(hostHelpers)
+	if err != nil {
+		return nil, fmt.Errorf("failed to creeate serviceConfig: %w", err)
```
adrianchiris: nit: create

cmd/sriov-network-config-daemon/service.go
```diff
@@ -59,36 +54,50 @@ var (
...
+	}
+
+	err = sc.readConf()
+	if err != nil {
+		return nil, sc.updateSriovResultErr(phaseArg, err)
+	}
+
+	// init globals
+	vars.PlatformType = sc.sriovConfig.PlatformType
+	platformInterface, err := newPlatformFunc(hostHelpers)
```
adrianchiris: nit: move above ServiceConfig creation so you can assign platformInterface directly in struct initialization

pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,36 @@
...
+	BeforeDrainNode(context.Context, *corev1.Node) (bool, error)
+	// AfterCompleteDrainNode run a platform-specific logic after node draining was completed
+	AfterCompleteDrainNode(context.Context, *corev1.Node) (bool, error)
+}
+
+func New() (Interface, error) {
```
adrianchiris: lets make this more like a factory and pass in the cluster type as arg WDYT ?

pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,36 @@
...
+	switch vars.ClusterType {
+	case consts.ClusterTypeOpenshift:
+		return openshift.New()
+	case consts.ClusterTypeKubernetes:
+		return kubernetes.New()
+	default:
+		return nil, fmt.Errorf("unknown orchestration type: %s", vars.ClusterType)
```
adrianchiris: unknown cluster type ?

pkg/orchestrator/orchestrator_suite_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	logf.SetLogger(zap.New(
+		zap.WriteTo(GinkgoWriter),
+		zap.UseDevMode(true),
+		func(o *zap.Options) {
+			o.TimeEncoder = zapcore.RFC3339NanoTimeEncoder
+		}))
+
+	// Go to project root directory
+	err = os.Chdir("../..")
```
adrianchiris, line 55: nit: why not provide the full path in testenv below ?

pkg/orchestrator/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+)
+
+type Kubernetes struct{}
```
adrianchiris, line 12: add docstrings to the file please

pkg/orchestrator/orchestrator.go
adrianchiris: also some info on the expected return values and their meaning

pkg/orchestrator/openshift/openshift.go
```diff
@@ -21,56 +21,30 @@ import (
...
-	OpenshiftBeforeDrainNode(context.Context, *corev1.Node) (bool, error)
-	OpenshiftAfterCompleteDrainNode(context.Context, *corev1.Node) (bool, error)
-
-	GetNodeMachinePoolName(context.Context, *corev1.Node) (string, error)
-	ChangeMachineConfigPoolPause(context.Context, *mcv1.MachineConfigPool, bool) error
-}
-
 // OpenshiftContext contains metadata and structs utilized to interact with Openshift clusters
-type openshiftContext struct {
+type OpenshiftContext struct {
```
adrianchiris: OpenshiftOrchestrator ? also update docstring to say it implements the Orchestrator interface

pkg/orchestrator/openshift/openshift.go
```diff
@@ -21,56 +21,30 @@ import (
...
-
+func New() (*OpenshiftContext, error) {
 	kubeClient, err := client.New(vars.Config, client.Options{Scheme: vars.Scheme})
 	if err != nil {
 		return nil, err
 	}
 
-	openshiftFlavor := OpenshiftFlavorDefault
+	openshiftFlavor := consts.ClusterFlavorOpenshift
 	infraClient, err := client.New(vars.Config, client.Options{
```
adrianchiris: do we really need infraCLient ? i think we can just use kubeClient from above

pkg/orchestrator/openshift/openshift.go
```diff
@@ -289,14 +249,9 @@ func (c *openshiftContext) OpenshiftAfterCompleteDrainNode(ctx context.Context,
...
 
-func (c *openshiftContext) GetNodeMachinePoolName(ctx context.Context, node *corev1.Node) (string, error) {
-	// if it's not an openshift cluster we return error
-	if !c.IsOpenshiftCluster() {
-		return "", fmt.Errorf("not an openshift cluster")
-	}
-
-	// hyperShift cluster don't have machine config
-	if c.IsHypershift() {
+func (c *OpenshiftContext) GetNodeMachinePoolName(ctx context.Context, node *corev1.Node) (string, error) {
```
adrianchiris: can you change all methods that are not related to the interface implementation to be private ?

pkg/orchestrator/openshift/openshift.go
```diff
@@ -84,33 +58,24 @@ func New() (OpenshiftContextInterface, error) {
...
+	return c.openshiftFlavor
 }
 
-func (c *openshiftContext) OpenshiftBeforeDrainNode(ctx context.Context, node *corev1.Node) (bool, error) {
-	// if it's not an openshift cluster we just return true that the operator manage to drain the node
-	if !c.IsOpenshiftCluster() {
-		return true, nil
-	}
-
+func (c *OpenshiftContext) BeforeDrainNode(ctx context.Context, node *corev1.Node) (bool, error) {
```
adrianchiris: add docstrings to interface methods

pkg/platform/platform.go
```diff
@@ -0,0 +1,36 @@
...
+	Init() error
+	GetHostHelpers() helper.HostHelpersInterface
+
+	DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error)
```
adrianchiris, line 32: please add info on the expected return values of these methods as well as some information when they are called and what they are expected to do.

pkg/platform/platform.go
```diff
@@ -0,0 +1,36 @@
...
+	DiscoverBridges() (sriovnetworkv1.Bridges, error)
+
```
adrianchiris, line 39: nit: drop the newlines here and above L19

pkg/platform/platform.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+//go:generate ../../bin/mockgen -destination mock/mock_platform.go -source platform.go
+type Interface interface {
+	Init() error
+	GetHostHelpers() helper.HostHelpersInterface
+
+	DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error)
+	DiscoverBridges() (sriovnetworkv1.Bridges, error)
```
adrianchiris, line 38: we should have a `NotSupportedError` defined in this file which is returned in case DiscoverBridges functionality is not supported by the plugin

pkg/platform/platform.go
adrianchiris, line 19: should we have a Name() method to identify the implementation of the platform ?

pkg/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,36 @@
...
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/orchestrator/kubernetes"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/orchestrator/openshift"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+//go:generate ../../bin/mockgen -destination mock/mock_orchestrator.go -source orchestrator.go
+type Interface interface {
```
adrianchiris, line 19: should we have a Name() method to identify the implementation of the orchestrator ?

also add docstring to the interface

pkg/platform/baremetal/baremetal_suite_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	logf.SetLogger(zap.New(
+		zap.WriteTo(GinkgoWriter),
+		zap.UseDevMode(true),
+		func(o *zap.Options) {
+			o.TimeEncoder = zapcore.RFC3339NanoTimeEncoder
+		}))
+
+	// Go to project root directory
+	err = os.Chdir("../../..")
```
adrianchiris, line 52: why not provide the full path below ?

pkg/platform/baremetal/baremetal.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/orchestrator"
+	plugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins"
+	genericplugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/generic"
+	intelplugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/intel"
+	k8splugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/k8s"
+	mellanoxplugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/mellanox"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+var VendorPluginMap = map[string]func(helpers helper.HostHelpersInterface) (plugin.VendorPlugin, error){
```
adrianchiris, line 24: docstrings throughout please :)

pkg/platform/baremetal/baremetal.go
```diff
@@ -0,0 +1,118 @@
...
+	"8086": intelplugin.NewIntelPlugin,
+	"15b3": mellanoxplugin.NewMellanoxPlugin,
+}
+
+type Baremetal struct {
```
adrianchiris, line 31: please mention in docstring that this implements platform interface

pkg/platform/platform.go
```diff
@@ -0,0 +1,36 @@
...
+
+	GetPlugins(ns *sriovnetworkv1.SriovNetworkNodeState) (plugin.VendorPlugin, []plugin.VendorPlugin, error)
+	SystemdGetPlugin(phase string) (plugin.VendorPlugin, error)
```
adrianchiris: nit: SystemdGetVendorPlugin ? so its consistent with return value ?

pkg/platform/platform.go
```diff
@@ -0,0 +1,36 @@
...
+
+//go:generate ../../bin/mockgen -destination mock/mock_platform.go -source platform.go
+type Interface interface {
+	Init() error
+	GetHostHelpers() helper.HostHelpersInterface
+
+	DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error)
+	DiscoverBridges() (sriovnetworkv1.Bridges, error)
+
+	GetPlugins(ns *sriovnetworkv1.SriovNetworkNodeState) (plugin.VendorPlugin, []plugin.VendorPlugin, error)
```
adrianchiris: nit: GetVendorPlugins ? so its consistent with what is getting returned ?

pkg/platform/baremetal/baremetal.go
```diff
@@ -0,0 +1,118 @@
...
+			genericplugin.WithSkipBridgeConfiguration())
+	case consts.PhasePost:
+		return genericplugin.NewGenericPlugin(bm.hostHelpers)
+	default:
+		return nil, fmt.Errorf("invalid phase %s", phase)
+	}
+}
+
+func (bm *Baremetal) loadVendorPlugins(ns *sriovnetworkv1.SriovNetworkNodeState) ([]plugin.VendorPlugin, error) {
+	vendorPluginsMap := map[string]plugin.VendorPlugin{}
```
adrianchiris: nit: loadedVendorPluginsMap ? or just loadedPluginsMap ?

vendorPluginsMap is too similar to VendorPluginMap so its a bit confusing to read through

pkg/platforms/openstack/openstack.go
```diff
@@ -325,177 +227,3 @@ func getPCIAddressFromMACAddress(macAddress string, nics []*net.NIC) (string, er
...
-	}
-	return pfList, nil
-}
-
-func (o *openstackContext) CreateOpenstackDevicesInfoFromNodeStatus(networkState *sriovnetworkv1.SriovNetworkNodeState) {
-	devicesInfo := make(OSPDevicesInfo)
-	for _, iface := range networkState.Status.Interfaces {
-		devicesInfo[iface.PciAddress] = &OSPDeviceInfo{MacAddress: iface.Mac, NetworkID: iface.NetFilter}
-	}
-
```
adrianchiris, line 499: while at it can you fix getBodyFromURL func to defer `resp.Body.Close()` after successfull get request ?

```
func getBodyFromURL(url string) ([]byte, error) {
	log.Log.V(2).Info("Getting body from", "url", url)
	resp, err := retryablehttp.Get(url)
	if err != nil {
		return nil, err
	}
    defer resp.Body.Close()
	rawBytes, err := io.ReadAll(resp.Body)
	if err != nil {
		return nil, err
	}
	return rawBytes, nil
}
```

pkg/platform/openstack/openstack.go
```diff
@@ -0,0 +1,265 @@
...
+	dputils "github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/utils"
+
+	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/helper"
+	plugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins"
+	virtualplugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/virtual"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
```
adrianchiris, line 19: please add docstring throughout

pkg/platform/openstack/openstack.go
```diff
@@ -0,0 +1,265 @@
...
+	ospNetworkDataURL  = ospMetaDataBaseURL + "/" + ospNetworkDataJSON
+	ospMetaDataURL     = ospMetaDataBaseURL + "/" + ospMetaDataJSON
+	// Config drive is defined as an iso9660 or vfat (deprecated) drive
+	// with the "config-2" label.
+	//https://docs.openstack.org/nova/latest/user/config-drive.html
+	configDriveLabel = "config-2"
+)
+
+var (
+	ospNetworkDataFile = ospMetaDataDir + "/" + ospNetworkDataJSON
```
adrianchiris, line 36: something wierd is happening with this one. we initialize it here but we override it in:

`getOpenstackDataFromConfigDrive()` and only use it within that file

pkg/platform/openstack/openstack.go
```diff
@@ -0,0 +1,265 @@
...
+			log.Log.Error(err, "DiscoverSriovDevicesVirtual(): unable to parse device class for device, skipping",
+				"device", device)
+			continue
+		}
+		if devClass != consts.NetClass {
+			// Not network device
+			continue
+		}
+
+		deviceInfo, exist := o.openStackDevicesInfo[device.Address]
```
adrianchiris, line 129: AFAIS this contains only network devices, so parsing and checking device class above is not really needed. but i guess its not related to this PR

adrianchiris:
once merged, lets sync the changes back to the design doc so it can serve as reference for implementors

pkg/host/manager.go
SchSeba: This with the http in utils is going to be used by the AWS, Oracle and the follow PR I am going to propose to refactor the openstack one.

This allows us to create unit-tests for the packages

pkg/utils/utils.go
SchSeba: going to be used by the AWS, Oracle and the follow PR I am going to propose to refactor the openstack one.

This allows us to create unit-tests for the packages

pkg/platform/platform.go
SchSeba: going to be used by the AWS, Oracle and the follow PR I am going to propose to refactor the openstack one.

This allows us to create unit-tests for the packages

I wanted this to be a singleton so we are sure everyone is using the same instance

adrianchiris: in the implementation you can pass in the same hostHelper.
i dont think  this method is needed in the platform interface.

pkg/orchestrator/orchestrator_suite_test.go
SchSeba: I just copy this from all the other tests we have :P
I prefer to leave it like that if that is ok.

pkg/orchestrator/openshift/openshift.go
SchSeba: I prefer to leave them public so we can unit-test them using an `openshift_test` package.

pkg/platform/openstack/openstack.go
```diff
@@ -0,0 +1,286 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/helper"
+	plugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins"
+	virtualplugin "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/plugins/virtual"
+)
+
+var (
+	// errNotSupported is returned when an operation is not supported by the OpenStack platform.
+	// This matches the platform.ErrNotSupported sentinel error.
+	errNotSupported = errors.New("operation not supported by platform")
```
adrianchiris: nit: maybe move this one to platform.go and make it public (ErrNotSupported) ?

pkg/platform/platform.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// DiscoverSriovDevices discovers all SR-IOV capable devices on the host.
+	// This is called during status updates to get the current state of SR-IOV devices.
+	// Returns a list of discovered interfaces with their SR-IOV capabilities, or an error if discovery fails.
+	DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error)
+
+	// DiscoverBridges discovers software bridges managed by the operator.
+	// This is called during status updates when vars.ManageSoftwareBridges is enabled.
+	// Returns a list of discovered bridges, or an error if discovery fails.
+	// If the platform does not support bridge discovery, it should return an empty list and nil error.
```
adrianchiris: > an empty list and nil error

an empty list and ErrNotSupported error ?

pkg/platform/platform.go
adrianchiris: as agreed this is removed from the interface definition

pkg/orchestrator/orchestrator_suite_test.go
adrianchiris: ack

pkg/orchestrator/openshift/openshift.go
adrianchiris: was not addressed.

pkg/orchestrator/openshift/openshift.go
adrianchiris: ack

pkg/platform/openstack/openstack.go
adrianchiris: i see its already there, just need to use it in `DiscoverBridges()` impl

pkg/platform/baremetal/baremetal_suite_test.go
adrianchiris: discussed above

pkg/platform/openstack/openstack.go
adrianchiris: not related to this MR this logic was there before. feel free to close this one you read it.

pkg/platform/openstack/openstack.go
adrianchiris: feel free to close this after you read it.

SchSeba: in general after this PR goes in I am going to rework the openstack so it's easier to test it via unit-tests so I am closing this one for now

pkg/platform/openstack/openstack.go
SchSeba: I move the error to be under vars. We can't have it under platform as it will make a circular dependency.

pkg/platform/platform.go
SchSeba: nice catch! done

pkg/orchestrator/openshift/openshift.go
SchSeba: done

adrianchiris (approved):

SchSeba:
Thanks! Merging this one

PR #903, opened 2025-06-18
add validation in e2e for multiple resource injection

SchSeba:
Small add to existing e2e test just to validate that pod spec change with the right number of sriov vf resources injected

### DIFF
--- a/test/conformance/tests/test_sriov_operator.go
+++ b/test/conformance/tests/test_sriov_operator.go
@@ -954,13 +954,31 @@ var _ = Describe("[sriov] operator", Ordered, func() {
 					pod.DefineWithNetworks([]string{sriovNetworkName, sriovNetworkName, sriovNetworkName, sriovNetworkName, sriovNetworkName}),
 					node,
 				)
+				By("Checking the sriov resource does not exist in resource request and limit spec")
+				_, exist := testPodA.Spec.Containers[0].Resources.Requests[corev1.ResourceName(fmt.Sprintf("openshift.io/%s", resourceName))]
+				Expect(exist).To(BeFalse())
+				_, exist = testPodA.Spec.Containers[0].Resources.Limits[corev1.ResourceName(fmt.Sprintf("openshift.io/%s", resourceName))]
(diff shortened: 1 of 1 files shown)


test/conformance/tests/test_sriov_operator.go
```diff
@@ -954,13 +954,31 @@ var _ = Describe("[sriov] operator", Ordered, func() {
 					pod.DefineWithNetworks([]string{sriovNetworkName, sriovNetworkName, sriovNetworkName, sriovNetworkName, sriovNetworkName}),
 					node,
 				)
+				By("Checking the sriov resource does not exist in resource request and limit spec")
+				_, exist := testPodA.Spec.Containers[0].Resources.Requests[corev1.ResourceName(fmt.Sprintf("openshift.io/%s", resourceName))]
+				Expect(exist).To(BeFalse())
+				_, exist = testPodA.Spec.Containers[0].Resources.Limits[corev1.ResourceName(fmt.Sprintf("openshift.io/%s", resourceName))]
+				Expect(exist).To(BeFalse())
+
```
zeeke: this is testing the `DefineWithNetworks` function not adding resources to the container spec, which will be added after creation by the network-resource-injector. Isn't it redundant?

SchSeba: right that is just to cover that no one will add the resource in the function and this tests will pass even if the resource injector is not doing is job

zeeke: ok, it makes sense

zeeke (approved):
LGTM

test/conformance/tests/test_sriov_operator.go
adrianchiris: TBH its abit confusing that we protect the test code from other test code.

if this is not a real concern, lets remove this bit IMO. if it is a real concern, can you please add a comment why this check was added ?

other than that, lgtm

SchSeba: removed :)

PR #904, opened 2025-06-24
fix: set SR-IOV NUM_VFS to 0 for mellanox NICs before fw reset

almaslennikov:
we are using --skip-driver option with mlxfwreset to speed up the process. However, if there are still VFs created for the PF, the command will fail. Let's reset the VFs to mitigate the error.

### DIFF
--- a/pkg/plugins/mellanox/mellanox_plugin.go
+++ b/pkg/plugins/mellanox/mellanox_plugin.go
@@ -206,7 +206,7 @@ func (p *MellanoxPlugin) Apply() error {
 		return err
 	}
 	if vars.FeatureGate.IsEnabled(consts.MellanoxFirmwareResetFeatureGate) {
-		return p.helpers.MlxResetFW(pciAddressesToReset)
+		return p.helpers.MlxResetFW(pciAddressesToReset, mellanoxNicsStatus)
 	}
 	return nil
 }
--- a/pkg/vendors/mellanox/mellanox.go
+++ b/pkg/vendors/mellanox/mellanox.go
@@ -12,6 +12,7 @@ import (
 
 	sriovnetworkv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/host"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
 )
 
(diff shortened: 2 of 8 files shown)


zeeke (approved):

SchSeba (approved):
LGTM

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -206,6 +206,29 @@ func (p *MellanoxPlugin) Apply() error {
 		return err
 	}
 	if vars.FeatureGate.IsEnabled(consts.MellanoxFirmwareResetFeatureGate) {
+		for _, pciAddress := range pciAddressesToReset {
```
adrianchiris: should we move this logic to MlxResetFW ? i think its better that the logic is contained there.
(method signature might need to change to get the mellanoxNicsStatus map

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -206,6 +206,29 @@ func (p *MellanoxPlugin) Apply() error {
...
+			err := p.helpers.SetSriovNumVfs(pciAddress, 0)
+			if err != nil {
+				log.Log.Error(err, "failed to set SR-IOV number of VFs to 0 before firmware reset", "pciAddress", pciAddress)
+				return err
+			}
+
+			if mlx.IsDualPort(pciAddress, mellanoxNicsStatus) {
+				otherIfaceSpec := mlx.GetOtherPortSpec(pciAddress, mellanoxNicsSpec)
```
adrianchiris: we would want to reset the VFs on the second PF regardless of it being specified in spec, else the reset command fails no ?

pkg/vendors/mellanox/mellanox.go
```diff
@@ -412,7 +412,7 @@ func isLinkTypeRequireChange(iface sriovnetworkv1.Interface, ifaceStatus sriovne
 	return false, nil
 }
 
-func getOtherPortSpec(pciAddress string, mellanoxNicsSpec map[string]sriovnetworkv1.Interface) *sriovnetworkv1.Interface {
+func GetOtherPortSpec(pciAddress string, mellanoxNicsSpec map[string]sriovnetworkv1.Interface) *sriovnetworkv1.Interface {
```
adrianchiris: if we roll the new logic in the resetFW cmd this can remain local.

pkg/plugins/mellanox/mellanox_plugin.go
almaslennikov: Moved the new logic to MlxResetFW

pkg/plugins/mellanox/mellanox_plugin.go
almaslennikov: Right, changed to reset the second PF if it exists, regardless of whether it's in the spec

pkg/vendors/mellanox/mellanox.go
almaslennikov: Moved the new logic to MlxResetFW

pkg/vendors/mellanox/mellanox.go
```diff
@@ -144,10 +145,25 @@ func (m *mellanoxHelper) GetMellanoxBlueFieldMode(PciAddress string) (BlueFieldM
 	return -1, fmt.Errorf("MellanoxBlueFieldMode(): unknown device status for %s", PciAddress)
 }
 
-func (m *mellanoxHelper) MlxResetFW(pciAddresses []string) error {
+func (m *mellanoxHelper) MlxResetFW(pciAddresses []string, mellanoxNicsStatus map[string]map[string]sriovnetworkv1.InterfaceExt, hostHelper host.HostManagerInterface) error {
```
SchSeba: can we just change the order here https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/1921bed36fa8d97e5c7989b2c02a67fcdaa587e2/pkg/helper/host.go#L29 and add the hostInterface to here?
```
func New(utilsHelper utils.CmdInterface) MellanoxInterface {
	return &mellanoxHelper{
		utils: utilsHelper,
	}
}
```

adrianchiris (approved):
LGTM, Thx @almaslennikov for working on this !
please address the comment from @SchSeba  and we are good to go IMO

adrianchiris:
@SchSeba PTAL

PR #911, opened 2025-07-09
Add dependabot config

sebrandon1:
Creates a dependabot config to automatically open PRs with updates for this repo.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,7 +20,7 @@ updates:
       - "dependencies"
       - "docker"
 
-  # Go modules
+  # Go modules - k8s dependencies
   - package-ecosystem: "gomod"
     directory: "/"
     schedule:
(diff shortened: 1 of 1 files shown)


SchSeba (approved):
Thanks!

LGTM

.github/dependabot.yml
```diff
@@ -0,0 +1,31 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+      - "docker"
+
+  # Go modules
+  - package-ecosystem: "gomod"
```
adrianchiris, line 24: should we exclude controller runtime ? this is sometimes more complex to handle e.g for major ver updates
also its better to group k8s updates:

this config should work well:

```
version: 2
updates:
  # Docker images
  - package-ecosystem: "docker"
    directory: "/"
    schedule:
      interval: "daily"
    commit-message:
      prefix: "chore:"

  # GitHub Actions
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
      day: "monday"
    commit-message:
      prefix: "chore:"

  - package-ecosystem: "gomod"
    directory: "/"
    schedule:
      interval: "weekly"
      day: "tuesday"
    groups:
      kubernetes:
        patterns: [ "k8s.io/*" ]
    ignore:
      # Ignore k8s-operator-libs, controller-runtime, and Kubernetes major and minor updates. These should be done manually.
      - dependency-name: "sigs.k8s.io/controller-runtime"
        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
      - dependency-name: "github.com/NVIDIA/k8s-operator-libs"
        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
      - dependency-name: "k8s.io/*"
        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
    commit-message:
      prefix: **"chore:"**
```

sebrandon1: 😄 These are the only ones that I care about that keep getting updated.

adrianchiris: OK, so suggest to split into 3 groups:

1. k8s. (pattern k8s.io/*)
2. ctrl runtime (pattern: sigs.k8s.io/controller-runtime/*)
3. general-dependencies (using exclude key with the above patterns)

wdyt ?

then if we have issues with either group its possible to solve manually and not block other updates.

SchSeba: +1 from my side.

sebrandon1: Yup that sounds good I can split into 3 groups.

PR #920, opened 2025-08-12
Add GOPROXY env support for Dockerfiles

e0ne:
Usage: IMAGE_BUILD_OPTS="GOPROXY=direct" make image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,8 @@
 FROM golang:1.23 AS builder
+
+ARG GOPROXY
+ENV GOPROXY=$GOPROXY
+
 WORKDIR /go/src/github.com/k8snetworkplumbingwg/sriov-network-operator
 COPY . .
 RUN make _build-manager BIN_PATH=build/_output/cmd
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,4 +1,8 @@
 FROM golang:1.23 AS builder
+
+ARG GOPROXY=direct
```
adrianchiris: should we keep this unset so go uses the default ?

see:
```
GOPROXY defaults to https://proxy.golang.org,direct. Under that configuration, the go command first contacts the Go module mirror run by Google, then falls back to a direct connection if the mirror does not have the module. See https://proxy.golang.org/privacy for the mirror's privacy policy. The GOPRIVATE and GONOPROXY environment variables may be set to prevent specific modules from being downloaded using proxies. See [Privacy](https://go.dev/ref/mod#private-module-privacy) for information on private proxy configuration.

See [Module proxies](https://go.dev/ref/mod#module-proxy) and [Resolving a package to a module](https://go.dev/ref/mod#resolve-pkg-mod) for more information on how proxies are used.
```
https://go.dev/ref/mod#environment-variables

using direct as default is not a good way to go IMO

e0ne: makes sense. done

PR #922, opened 2025-08-26
chore: bump libovsdb lib

adrianchiris:
current lib path is deprecated, use
github.com/ovn-kubernetes/libovsdb instead


swtaylor:
Looks like this is how they are handling it upstream: 
https://github.com/ovn-kubernetes/libovsdb/blob/18e5bb119abe1e29ca07b02083ebdef39242e6ff/client/client_test.go#L958-L963

PR #923, opened 2025-08-27
Avoid setting Drivers selector for `vfio-pci` devices

zeeke:
Devices configured by SriovNetworkNodePolicy are already selected by the `PfNames` and `RootDevices` fields.

There is no need to add a Drivers selector, and it makes the following scenario not work: a. One node has a mellanox device
b. One node has intel devices
c. The user wants to configure both node to run DPDK applications
  c1. policy1 selects all nodes, mellanox devices, setting `isRdma: true, resourceName: x`
  c2. policy2 selects all nodes, intel devices, setting `deviceType: vfio-pci, resourceName: x`
d. A pod is configured to consume a resource `x`, so it can be scheduled in both nodes.

On the mellanox node, if the nic selector has `Drivers: ["vfio-pci"]`, no VFs are put in the resource pool.

The above scenario can be worked around by changing policy1 and policy2 `nodeSelector` fields, to target their a more specific node. Still, it should be fixed.

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -584,10 +584,7 @@ func createDevicePluginResource(
 	if len(p.Spec.NicSelector.RootDevices) > 0 {
 		netDeviceSelectors.RootDevices = append(netDeviceSelectors.RootDevices, p.Spec.NicSelector.RootDevices...)
 	}
-	// Removed driver constraint for "netdevice" DeviceType
-	if p.Spec.DeviceType == constants.DeviceTypeVfioPci {
-		netDeviceSelectors.Drivers = append(netDeviceSelectors.Drivers, p.Spec.DeviceType)
-	}
+
 	// Enable the selection of devices using NetFilter
 	if p.Spec.NicSelector.NetFilter != "" {
 		// Loop through interfaces status to find a match for NetworkID or NetworkTag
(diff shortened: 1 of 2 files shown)


adrianchiris:
can you provide the final sriov-network-device plugin config that gets generated ?

I wonder how can a pool with `Drivers: ["vfio-pci"]` and `isRDMA: true` can expose resources since if a driver is bound to vfio-pci it should not have rdma resources.

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -584,10 +584,7 @@ func createDevicePluginResource(
 	if len(p.Spec.NicSelector.RootDevices) > 0 {
 		netDeviceSelectors.RootDevices = append(netDeviceSelectors.RootDevices, p.Spec.NicSelector.RootDevices...)
 	}
-	// Removed driver constraint for "netdevice" DeviceType
```
SchSeba, line 587: Hi I think this change is not right.

can you add a unit test to test partitional device?
I think this change will break the case where we have one PF that we split in 2 
10 vfio-pci (resource name: vfio)
10 netdevice (resource name: netdev)

I think if we don't add the driver the device plugin will select all of the VFs to the vfio pool

SchSeba (approved):

controllers/sriovnetworknodepolicy_controller.go
zeeke: I added [2cbd5b5](https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/923/commits/2cbd5b5d8e4e81d8bacde16425bb4f299e686495).

The gist is that the scenario with vfio-pci and netdevice must go with pf partitioning, using the `<pfname>#<vfindexes>` syntax. The vf indexes are enough to select the right VFs. 

The Drivers selector, on the other hand, is not accurate enough (what happens if I have two vfio-pci resources?), so it can be dropped, as it is redundant.

SchSeba (approved):
LGTM

nice work!

PR #927, opened 2025-09-15
Fix `make install` target

e0ne:
We use `make install` as a part of e2e tests setup. No need to use kustomize to apply generaged CRDs by kubectl.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,11 +86,11 @@ run: skopeo install
 
 # Install CRDs into a cluster
 install: manifests kustomize
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -86,7 +86,7 @@ run: skopeo install
 
 # Install CRDs into a cluster
 install: manifests kustomize
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+	kubectl apply -f $(CRD_BASES)
```
SchSeba, line 89: Please do the same of uninstall

e0ne: done

SchSeba (approved):
LGTM

adrianchiris (approved):

PR #948, opened 2025-09-30
feat: support OVS bridge secure mode

rollandf:
Support fail-mode field for OVS Bridges.

"fail-mode=secure"  means that the OpenFlow flow table starts out empty.

See https://docs.openvswitch.org/en/latest/tutorials/ovs-advanced/#setup


nvidia-ci-cd:

PR #917, opened 2025-07-21
Design proposal for the platform-orchestrator-abstraction

SchSeba:

PR #952, opened 2025-10-09
Setting option for node draining by external controllers 

heyvister1:
SRIOV OP internal drain controller can be disabled, through `USE_EXTERNAL_DRAINER`. For example draining can be performed by external [NVIDIA maintenance operator](https://github.com/Mellanox/maintenance-operator)

## Summary by CodeRabbit

* **New Features**
  * Opt-in external drainer mode via env/Helm flag (USE_EXTERNAL_DRAINER) to delegate node drain operations to external maintenance tooling.

* **Behavior**
  * Operator skips internal drain reconciliation for nodes marked for external draining; pool config changes won’t apply during those external drains.

* **Documentation**
  * Design docs updated with notes on external drain-controller delegation and effects.

* **Tests**
  * Added tests for external drainer flows and annotation handling.


controllers/helper.go
```diff
@@ -138,6 +138,12 @@ func GetImagePullSecrets() []string {
 	}
 }
 
+// UseMaintenanceOperatorDrainer indicates if internal drain controller is disabled
+// and draining will be done by external NVIDIA maintenance operator
+func UseMaintenanceOperatorDrainer() bool {
```
SchSeba: can we please have this in the `vars` and `consts` folder and not here.

SchSeba (commented):
In general I am fine with having this here. but can we please have a more generic name for the variable?

controllers/helper.go
heyvister1: done

pkg/daemon/daemon.go
```diff
@@ -532,6 +543,17 @@ func (dn *NodeReconciler) handleDrain(ctx context.Context, desiredNodeState *sri
 		return false, nil
 	}
 
+	// add external drainer annotation if enabled
+	if vars.UseExternalDrainer {
+		if err := utils.AnnotateNode(ctx,
```
adrianchiris: can we annotate nodeState instead ?

as we store related drain state in nodeState obj

main.go
```diff
@@ -295,6 +271,46 @@ func main() {
 	}
 }
 
+func setupDrainController(mgr ctrl.Manager, restConfig *rest.Config,
```
adrianchiris, line 274: Perhaps we should take a bit different approach.

we could always start the drain controller and then skip drain requests if use-external-drainer annotation is set.

that way, if there are any "in-flight" drains they will complete even if we switched on the external-drainer functionality.

WDYT ?

pkg/daemon/daemon.go
heyvister1: Done

e0ne (approved):
LGTM

doc/design/parallel-node-config.md
```diff
@@ -60,6 +60,8 @@ Node annotation, `sriovnetwork.openshift.io/state` and SriovNetworkNodeState ann
 
 *NOTE:* In the future we are going to drop the node annotation and only use the SriovNetworkNodeState
 
+*NOTE:* Internal drain controller can be disabled by exposing the following `USE_EXTERNAL_DRAINER` env variable. This means that drain operations will be done externally, for example by utilizing [NVIDIA maintenance OP](https://github.com/Mellanox/maintenance-operator). In addition, `SriovNetworkPoolConfig` will not take any effect during drain procedure, since the maintenance operator will be in charge of parallel node operations.
```
adrianchiris: i dont think we ever talked about changes to existing design docs if new addition changes that feature's behaviour.
i guess its ok but it might pose unmanageable in the future.

@SchSeba @zeeke thoughts ? should we have this addition here or just in main README ?

main.go
```diff
@@ -295,6 +271,45 @@ func main() {
...
+	platformsHelper platforms.Interface, scheme *runtime.Scheme) error {
+	if vars.UseExternalDrainer {
+		setupLog.Info("'UseExternalDrainer' is set, draining will be done externally")
```
adrianchiris, line 280: please add a note on why we still setup the drain controller here even if UseExternalDrainer is set.

adrianchiris (approved):
LGTM added minor nit to clarify why we need to start drain controller when external drainer is used.

main.go
heyvister1: Done

controllers/drain_controller_test.go
```diff
@@ -408,7 +418,7 @@ func simulateDaemonSetAnnotation(node *corev1.Node, drainAnnotationValue string)
 		ToNot(HaveOccurred())
 }
 
-func createNode(ctx context.Context, nodeName string) (*corev1.Node, *sriovnetworkv1.SriovNetworkNodeState) {
+func createNode(ctx context.Context, nodeName string, useExternalDrainer bool) (*corev1.Node, *sriovnetworkv1.SriovNetworkNodeState) {
```
zeeke: please, avoid the boolean parameter as it reduce readability. 
as alternatives:
a. change the parameter to `additionalAnnotation map[string]string`
b. add a tweak parameter like

`tweak func(*corev1.Node, *sriovnetworkv1.SriovNetworkNodeState)` so that a function can customize the creation of the node objects

c. remove the parameter and update the k8s objects after the creation

I prefer `b`, but the others would work too

controllers/drain_controller_test.go
```diff
@@ -426,12 +436,16 @@ func createNode(ctx context.Context, nodeName string) (*corev1.Node, *sriovnetwo
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      nodeName,
 			Namespace: vars.Namespace,
-			Labels: map[string]string{
+			Annotations: map[string]string{
```
zeeke, line 437: Any idea about how these tests were working before this line change?

main.go
zeeke: what about using the same annotations as the regular drain flow? 
- the daemon would set the sriovnetworknodestate.DesiredState == "Drain_Required" 
- the in-operator drainer would do nothing
- the external drainer would drain the node using its own logic, then set the node.CurrentState=DrainComplete 

would it be a cleaner implementation?

doc/design/parallel-node-config.md
SchSeba: I am fine adding that change also here not only in the README and update the updatedDate in the doc.
+ add the link to this PR (Example https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/917/files#diff-dc6ff1ecffa429d5635307cc009fe47ad0d51f41d549dbf88f67752c7cfbebc6R9)

controllers/drain_controller_test.go
SchSeba: if I remember right the controller adds them if they don't exist.

pkg/daemon/daemon.go
```diff
@@ -408,6 +408,20 @@ func (dn *NodeReconciler) apply(ctx context.Context, desiredNodeState *sriovnetw
 		return ctrl.Result{}, err
 	}
 
+	// remove external drainer nodestate annotation if exists
+	annotations := desiredNodeState.GetAnnotations()
+	if _, ok := annotations[consts.NodeStateExternalDrainerAnnotation]; ok {
```
SchSeba: can you explain why we need this here?

main.go
SchSeba: It's more complicated.

if we are in the middle, and we update the config-daemon yaml it will start new pods that will add the label.
in this case we will be in the middle of configuration with the 2 labels in parallel

SchSeba (commented):
I must say I don't follow this sorry.

the daemon will still request drain and this check https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/952/files#diff-a53b7b593d3d778e62eaeeafa40088656f9212bfa2c2b7991df15fa78e60b0f0R256
will never pass.

also you can have a race as we do the annotation update twice in

```
	// add external drainer nodestate annotation if flag is enabled
	if vars.UseExternalDrainer {
		err := utils.AnnotateObject(ctx, desiredNodeState,
			consts.NodeStateExternalDrainerAnnotation, "true", dn.client)
		if err != nil {
			funcLog.Error(err, "failed to add nodestate external drainer annotation")
			return false, err
		}
	}

	// annotate both node and node state with drain or reboot
	annotation := consts.DrainRequired
	if reqReboot {
		annotation = consts.RebootRequired
	}
	return true, dn.annotate(ctx, desiredNodeState, annotation)
}
```

and I don't understand how the daemon will know that he is able to continue with the configuration because the drain was done.

controllers/drain_controller_test.go
heyvister1: I have used option "a"

controllers/drain_controller_test.go
heyvister1: drain_controller.go nodeState predicate was filtering events according to labels and not annotations
[https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/controllers/helper.go#L122-L123](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/controllers/helper.go#L122-L123)

main.go
adrianchiris: maybe we should not over complicate and just document that enabling/disabling external drainer should be done when no sriov configuration is required.

else, what im thinking is:
1. we skip reconcile nodestate with external drainer label set in the drain controller
2. config daemon would need to check if its in the middle of configuration (i.e requested drain or drain in progress) for such case it would not add the external drainer annotation to nodestate

then for 2:

if previously it was with external drainer, its expected that the external drainer to handle the drain and update the node state annotation accordingly regardless of the current `UseExternalDrainer` config

if previously it was without external drainer then its expected that the sriov-operator drain controller to do the drain regardless of the current `UseExternalDrainer` config

main.go
adrianchiris: @zeeke thats the idea. the external drainer would use the same set of annotations.

in fact the current implementation of the external drainer is using the same set of annotations as the sriov operator drain controller but delegates the drain operation to nvidia mainternance operator.

[1] https://github.com/Mellanox/network-operator/blob/a528454e52d35243d8e47e00a8aeafbd182447f6/controllers/drain_controller.go#L94

pkg/daemon/daemon.go
adrianchiris: we remove the external drainer annotation from node state as its no longer needed.

we could keep it here and remove it when drain is requested if `vars.UseExternalDrainer is disabled` and there is no in progress drains but i think its better to remove it when we are done with sriov configuration. WDYT ?

heyvister1:
> I must say I don't follow this sorry.
> 
> the daemon will still request drain and this check https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/952/files#diff-a53b7b593d3d778e62eaeeafa40088656f9212bfa2c2b7991df15fa78e60b0f0R256 will never pass.
> 
> also you can have a race as we do the annotation update twice in
> 
> ```
> 	// add external drainer nodestate annotation if flag is enabled
> 	if vars.UseExternalDrainer {
> 		err := utils.AnnotateObject(ctx, desiredNodeState,
> 			consts.NodeStateExternalDrainerAnnotation, "true", dn.client)
> 		if err != nil {
> 			funcLog.Error(err, "failed to add nodestate external drainer annotation")
> 			return false, err
> 		}
> 	}
> 
> 	// annotate both node and node state with drain or reboot
> 	annotation := consts.DrainRequired
> 	if reqReboot {
> 		annotation = consts.RebootRequired
> 	}
> 	return true, dn.annotate(ctx, desiredNodeState, annotation)
> }
> ```
> 
> and I don't understand how the daemon will know that he is able to continue with the configuration because the drain was done.

Hey @SchSeba Me and @adrianchiris went through the code again. Can you please explain why you're suspecting race-condition may occur, while patching nodeState annotations sequentially (twice)?
Using MergeFrom Patch, for annotations map will modify annotations map according to the diff computed between the in-memory object and the modified object, with new annotation key/value. 
So in our case, both annotation key/values will get updated without causing object update conflicts

controllers/helper.go
```diff
@@ -119,8 +119,8 @@ func (DrainStateAnnotationPredicate) Update(e event.UpdateEvent) bool {
 		return false
 	}
 
-	oldAnno, hasOldAnno := e.ObjectOld.GetLabels()[constants.NodeStateDrainAnnotationCurrent]
-	newAnno, hasNewAnno := e.ObjectNew.GetLabels()[constants.NodeStateDrainAnnotationCurrent]
+	oldAnno, hasOldAnno := e.ObjectOld.GetAnnotations()[constants.NodeStateDrainAnnotationCurrent]
```
adrianchiris, line 122: 10x for the fix :)

pkg/daemon/daemon.go
```diff
@@ -517,6 +531,16 @@ func (dn *NodeReconciler) handleDrain(ctx context.Context, desiredNodeState *sri
 		return false, nil
 	}
 
+	// add external drainer nodestate annotation if flag is enabled
+	if vars.UseExternalDrainer {
+		err := utils.AnnotateObject(ctx, desiredNodeState,
```
adrianchiris: in this case i believe we should add this annotation only if the current drain state is `idle` and the currently set desired drain state is also idle

i.e we are at a point where:
1. no drain is in progress 
2. no drain was previously requested

that is in case we want to support switching to/from use external drainer.

for the case where we switch from external drainer to built-in drainer, its expected that the external drainer to still be active to handle any in-progress drain requests ofc.

adrianchiris: lets discuss this in community meeting id like to hear @SchSeba thoughts on it.

main.go
zeeke: got it, but I don't get why we need a new annotation `NodeStateExternalDrainerAnnotation = "sriovnetwork.openshift.io/use-external-drainer"`.

https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/952/files#diff-0e683dcc8c5eeb8433c366296d8cb81d12ff06fea8a08637949e040a3aa334e3R75

since the draincontroller is turnedoff (either not started or skips every reconcile)

adrianchiris: so, its not turned off :)
it just skips reconcile of sriovnetworknodestate with this annotation.

`sriovnetwork.openshift.io/use-external-drainer` annotation is used to mark an sriovnetworknodestate to be handled by an external drainer and the built-in drain controller would skip reconcile of those objects.

the reason we dont turn off the built-in drainer completely is because we want to handle cases of "in-flight" sriov configuration where the drain operation already started by the built-in drainer.

the sriov config daemon would need to be smart about adding this annotation (only when the drain state is idle)
so we dont have two drain contorllers (the external one and the sriov operator one)

SchSeba:
Hi @heyvister1 can you please check the coderabbit comments and see if they are relevant

pkg/consts/constants.go
```diff
@@ -77,15 +77,16 @@ const (
...
-	DrainIdle                       = "Idle"
-	DrainRequired                   = "Drain_Required"
-	RebootRequired                  = "Reboot_Required"
-	Draining                        = "Draining"
-	DrainComplete                   = "DrainComplete"
+	NodeDrainAnnotation                = "sriovnetwork.openshift.io/state"
+	NodeStateDrainAnnotation           = "sriovnetwork.openshift.io/desired-state"
+	NodeStateDrainAnnotationCurrent    = "sriovnetwork.openshift.io/current-state"
+	DesiredMachineConfigAnnotation     = "machineconfiguration.openshift.io/desiredConfig"
+	NodeStateExternalDrainerAnnotation = "sriovnetwork.openshift.io/use-external-drainer"
```
SchSeba, line 83: nit: please move this one above the MachineConfig one

pkg/daemon/daemon.go
```diff
@@ -517,6 +531,20 @@ func (dn *NodeReconciler) handleDrain(ctx context.Context, desiredNodeState *sri
...
+			utils.ObjectHasAnnotation(desiredNodeState, consts.NodeStateDrainAnnotation, consts.DrainIdle) {
+			err := utils.AnnotateObject(ctx, desiredNodeState,
+				consts.NodeStateExternalDrainerAnnotation, "true", dn.client)
+			if err != nil {
+				funcLog.Error(err, "failed to add node external drainer annotation")
+				return false, err
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
SchSeba: this was the potential deadlock I was thinking about (nice to see it manage to catch it)

pkg/daemon/daemon.go
SchSeba: let me add here was I understand and please correct me if I am wrong here.

adding the `NodeStateExternalDrainerAnnotation` will just tell the external provider to handle the drain, but the external provider will watch the same existing annotations we have today 

```
	NodeStateDrainAnnotation           = "sriovnetwork.openshift.io/desired-state"
	NodeStateDrainAnnotationCurrent    = "sriovnetwork.openshift.io/current-state"
```

and will also change the `NodeStateDrainAnnotationCurrent` to act as the drain_controller.


if all of the above is right, I propose

in the daemon reconcile loop we check for the environment variable and we add/remove it ONLY when both annotations are on `idle`.

that is all so we have this only in one please of the code and not split between to different places.

I think that is cleaner.
please let me know what do you think

pkg/daemon/daemon.go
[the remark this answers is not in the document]
heyvister1: In this scenario deadlock won't happen since internal drainer controller is still working and not disabled, so it should manage such case, while external controller flow is skipped in this case.

pkg/consts/constants.go
heyvister1: Done

heyvister1:
@SchSeba , @adrianchiris PTAL upon last commit changes, handling Sebastian's proposal
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/952#discussion_r2655474396

pkg/consts/constants.go
heyvister1: Done

doc/design/parallel-node-config.md
```diff
@@ -60,6 +60,8 @@ Node annotation, `sriovnetwork.openshift.io/state` and SriovNetworkNodeState ann
...
+
```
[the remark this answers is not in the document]
adrianchiris: +1 on addressing this comment.

pkg/daemon/daemon.go
```diff
@@ -661,6 +668,45 @@ func (dn *NodeReconciler) annotate(
 	return nil
 }
 
+// manages addition/removal of external drainer annotation upon node state objects
+func (dn *NodeReconciler) addRemoveExternalDrainerAnnotation(ctx context.Context, desiredNodeState *sriovnetworkv1.SriovNetworkNodeState) error {
+	funcLog := log.Log.WithName("addRemoveExternalDrainerAnnotation")
```
adrianchiris: lets use the log from context:

`reqLogger := log.FromContext(ctx).WithName("addRemoveExternalDrainerAnnotation")`

pkg/daemon/daemon.go
```diff
@@ -661,6 +668,45 @@ func (dn *NodeReconciler) annotate(
...
+
+	// external drainer annotation will be added/removed only when both desired/current node state are in 'Idle' state
```
adrianchiris, line 681: i believe we need to take into consideration also when the annotation does not exist.

e.g first time node state is created, the annotation is only set later in the reconcile process (and moved to idle after successfull first configuration)

so i would proceed with setting this annotation when either:

1. both annotations are unset
2. both annotations are idle

pkg/daemon/daemon.go
```diff
@@ -661,6 +668,45 @@ func (dn *NodeReconciler) annotate(
...
+	}
+
+	if !vars.UseExternalDrainer {
+		// remove external drainer nodestate annotation if exists
+		annotations := desiredNodeState.GetAnnotations()
+		if _, ok := annotations[consts.NodeStateExternalDrainerAnnotation]; ok {
+			funcLog.Info("remove external drainer nodestate annotation", "annotation", consts.NodeStateExternalDrainerAnnotation)
+			original := desiredNodeState.DeepCopy()
+			delete(annotations, consts.NodeStateExternalDrainerAnnotation)
+			desiredNodeState.SetAnnotations(annotations)
```
adrianchiris: is this needed ? i believe you get a reference for the map at L683
which you then update.

no copy takes place for this map

pkg/daemon/daemon.go
```diff
@@ -661,6 +668,45 @@ func (dn *NodeReconciler) annotate(
...
+			// Patch only the annotations
+			if err := dn.client.Patch(ctx, desiredNodeState, client.MergeFrom(original)); err != nil {
+				funcLog.Error(err, "failed to patch nodestate after removing external drainer annotation")
+				return err
+			}
+		}
+		return nil
+	}
+
+	// check if external drainer environment variable is enabled, before adding external drainer annotation
```
adrianchiris: i dont understand the comment. maybe just remove it ? it doesnt add much here IMO

pkg/daemon/daemon.go
```diff
@@ -149,6 +149,17 @@ func (dn *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
 	}
 
+	// add/remove external drainer annotation in case external drainer
+	changed, err := dn.addRemoveExternalDrainerAnnotation(ctx, desiredNodeState)
+	if err != nil {
+		reqLogger.Error(err, "failed to add/remove external drainer annotation")
+		return ctrl.Result{}, err
+	}
+	if changed {
+		reqLogger.Info("external drainer annotation changed, return to allow drain controller to react")
```
adrianchiris: the last part of the log is missleading 

we may have the idle annot already there. we will requeue since we patch the node state annotation which will trigger a reconcile.

main.go
zeeke: ok, understood. thanks

PR #962, opened 2025-11-06
e2e: Refactor `PF shutdown` test case

zeeke:
Shutting down a PF to test if the NIC can be used for pod-to-pod connectivity causes a race condition, as the config-daemon would try to set it up again during reconciliation. Also, it would trigger a node drain, making the test case very flaky.

Instead, convert the test case to verify that a NIC with
 no carrier provides intra-node connectivity.

### DIFF
--- a/test/conformance/tests/test_policy_configuration.go
+++ b/test/conformance/tests/test_policy_configuration.go
@@ -456,7 +456,7 @@ var _ = Describe("[sriov] operator", Ordered, func() {
 			})
 			Context("PF shutdown", func() {
 				// 29398
-				It("Should be able to create pods successfully if PF is down.Pods are able to communicate with each other on the same node", func() {
+				It("Should be able to create pods successfully on a NIC with NO_CARRIER. Pods are able to communicate with each other on the same node", func() {
 					if cluster.VirtualCluster() {
 						// https://bugzilla.redhat.com/show_bug.cgi?id=2214976
 						Skip("Bug in IGB driver")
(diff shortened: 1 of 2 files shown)


test/conformance/tests/test_policy_configuration.go
```diff
@@ -473,17 +473,15 @@ var _ = Describe("[sriov] operator", Ordered, func() {
 					testNode = sriovInfos.Nodes[0]
 					sriovDeviceList, err := sriovInfos.FindSriovDevices(testNode)
 					Expect(err).ToNot(HaveOccurred())
-					unusedSriovDevices, err := findUnusedSriovDevices(testNode, sriovDeviceList)
+					noCarrierSriovDevices, err := findNoCarrierSriovDevices(testNode, sriovDeviceList)
```
adrianchiris, line 476: do we also require it to also be unsused?

adrianchiris (approved):
so we require to have a nic with cable unplugged for this to actually run.

test/conformance/tests/test_policy_configuration.go
zeeke: An "unused device", in the context of `findUnusedSriovDevices`, means it is not the default route, and it is not assegned to an ovs bridge as master (what happen in OVNk).

https://github.com/zeeke/sriov-network-operator-1/blob/90b11c8214118ba6177e322db0a041fd553b48cf/test/conformance/tests/test_sriov_operator.go#L1649

So, if the device has no-carrier, it is also "unused".
(Maybe it can be the uplink of a test ovs bridge? 🤔 )

PR #967, opened 2025-11-23
Add support for network interface alternative names

cgoncalves:
### Overview

This PR adds support for network interface alternative names (altnames) throughout the SR-IOV Network Operator. Interface alternative names provide additional flexibility when selecting network interfaces in policies, particularly useful in environments where interfaces may have predictable alternative naming schemes.

This change is backward compatible. Existing policies using primary interface names will continue to work without modification.

### How to Use

Alternative names are automatically discovered and stored in the `SriovNetworkNodeState` CR. When defining a `SriovNetworkNodePolicy`, you can now reference interfaces using either their primary name or any of their alternative names in the `nicSelector.pfNames` field.

**Example:**
```yaml
kind: SriovNetworkNodePolicy
metadata:
  name: policy-using-altname
  namespace: sriov-network-operator
spec:
  deviceType: netdevice
  nicSelector:
    pfNames: ["sriov1"]  # Can use altname instead of primary interface name
    vendor: "8086"
  nodeSelector:
    feature.node.kubernetes.io/network-sriov.capable: "true"
```

## Release Notes

* **New Features**
  * Policy-level support for matching network interfaces by primary or alternative names
  * Automatic discovery of alternative interface names for physical functions

* **Documentation**
  * Added guide for discovering and viewing alternative interface names in SriovNetworkNodeState
  * Added documentation for using primary and alternative names in SR-IOV network policies

### DIFF
--- a/pkg/host/internal/sriov/sriov.go
+++ b/pkg/host/internal/sriov/sriov.go
@@ -272,8 +272,9 @@ func (s *sriov) DiscoverSriovDevices(storeManager store.ManagerInterface) ([]sri
 
 		altNames, err := s.netlinkLib.GetAltNames(pfNetName)
 		if err != nil {
-			log.Log.Error(err, "DiscoverSriovDevices(): unable to get alternative names for device, skipping", "device", device)
-			continue
+			log.Log.V(2).Info("DiscoverSriovDevices(): unable to get alternative names for device, continuing with empty altnames",
+				"device", device.Address, "reason", err.Error())
+			altNames = []string{}
 		}
 
 		link, err := s.netlinkLib.LinkByName(pfNetName)


pkg/webhook/validate.go
```diff
@@ -465,8 +473,19 @@ func validateNicModel(selector *sriovnetworkv1.SriovNetworkNicSelector, iface *s
 				pfNames = append(pfNames, p)
 			}
 		}
-		if !sriovnetworkv1.StringInArray(iface.Name, pfNames) {
-			return fmt.Errorf("interface name: %s not found in physical function names", iface.PciAddress)
+		// Check if the interface name or any of its alternative names match pfNames
+		nameMatches := sriovnetworkv1.StringInArray(iface.Name, pfNames)
```
SchSeba: we have the same also in https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/967/files#diff-371d95e79c61d8cc1e6791d27e009289552efcaada38f9b2fc75ebb487c90ea7R614

so maybe we can create a function and put it under the api helpers so it can be reused?

cgoncalves: Good idea, done.

README.md
```diff
@@ -288,6 +288,30 @@ status:
 
 From this example, in status field, the user can find out there are 2 SRIOV capable NICs on node 'work-node-1'; in spec field, user can learn what the expected configure is generated from the combination of SriovNetworkNodePolicy CRs.  In the virtual deployment case, a single VF will be associated with each device.
 
+### Discovering Alternative Names
```
adrianchiris: nit: lets be explicit here

DIscovering Alternative Names for Network interfaces ?

api/v1/helper.go
```diff
@@ -594,7 +623,8 @@ func (selector *SriovNetworkNicSelector) Selected(iface *InterfaceExt) bool {
 				pfNames = append(pfNames, p)
 			}
 		}
-		if !StringInArray(iface.Name, pfNames) {
+
+		if !NameOrAltNameMatchesPfNames(iface.Name, iface.AltNames, pfNames) {
```
adrianchiris, line 612: have we validated that sriov-network-device-plugin works properly with altnames ?

api/v1/helper.go
```diff
@@ -306,6 +321,20 @@ func StringInArray(val string, array []string) bool {
 	return false
 }
 
+// NameOrAltNameMatchesPfNames checks if the given name or any of the alternative names
+// match any entry in the pfNames list.
+func NameOrAltNameMatchesPfNames(name string, altNames []string, pfNames []string) bool {
```
adrianchiris, line 311: nit:
```
AnyNameMatchesPfNames(names []string, pfNames []string) {
  for _, n := range names {
    // check and return
  }
}
```

api/v1/helper.go
```diff
@@ -994,3 +1024,27 @@ func OwnerRefToString(cr client.Object) string {
...
+		return name
+	}
+	// Check all interfaces in the nodeState
+	for _, iface := range nodeState.Status.Interfaces {
+		// Check if the name matches the actual interface name
+		if iface.Name == name {
+			return name
+		}
+		// Check if the name is in the alternative names list
+		for _, altName := range iface.AltNames {
```
adrianchiris: can use StringInArray() like func above

pkg/host/internal/sriov/sriov.go
```diff
@@ -270,6 +260,12 @@ func (s *sriov) DiscoverSriovDevices(storeManager store.ManagerInterface) ([]sri
 			continue
 		}
 
+		altNames, err := s.netlinkLib.GetAltNames(pfNetName)
+		if err != nil {
+			log.Log.Error(err, "DiscoverSriovDevices(): unable to get alternative names for device, skipping", "device", device)
+			continue
```
adrianchiris: is it an overkill to skip the device in such case ? 
should this be best effort like LinkSpeed or linkAdminState

tbh im of either way, just want to hear others thoughts.

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -569,7 +594,8 @@ func createDevicePluginResource(
 		}
 	}
 	if len(p.Spec.NicSelector.PfNames) > 0 {
-		netDeviceSelectors.PfNames = append(netDeviceSelectors.PfNames, p.Spec.NicSelector.PfNames...)
+		resolvedPfNames := resolvePfNames(p.Spec.NicSelector.PfNames, nodeState)
+		netDeviceSelectors.PfNames = sriovnetworkv1.UniqueAppend(netDeviceSelectors.PfNames, resolvedPfNames...)
```
adrianchiris, line 598: so we will just append the actual interface name to PfNames.

we might end up with PfNames = [alt1#1-10, real1#1-10]

where real1 is the netdev name of a device with alt1 set as altname

will device plugin handle this as expected ?

api/v1/helper.go
cgoncalves: I'd rather keep it as is. Your suggested function would require the creation of a new slice with altnames+name to be passed in as argument.

pkg/host/internal/sriov/sriov.go
cgoncalves: Agreed. In the worst case when altnames retrieval fails (timing, transient netlink error), it can cause a boot loop.

controllers/sriovnetworknodepolicy_controller.go
cgoncalves: The code uses `sriovnetworkv1.UniqueAppend` to add these resolved names to `netDeviceSelectors.PfNames`.

If you have two policies, one using `alt1` and another using `real1` (pointing to the same interface), both will resolve to `real1`. `UniqueAppend` will then ensure `real1` appears only once in the final list.

api/v1/helper.go
cgoncalves: Per our internal testing, yes. Any specific concern or scenario you would like to see tested?

adrianchiris: nothing specific

api/v1/helper.go
adrianchiris: ack. makes sense

zeeke (approved):

PR #982, opened 2025-12-15
fix(cpu): Early return for s390x in GetCPUVendor to avoid ghw errors

ashokpariya0:
On s390x, ghw.CPU() returns empty Processors slice, causing errors and warnings. Add early return based on runtime.GOARCH to skip ghw call and return correct vendor.

This keeps existing workflow intact for x86/arm64 while fixing s390x detection.

### Why we need this change??
We use  ghw library(https://github.com/jaypipes/ghw) for cpu vendor detection which does not support s390x arch.
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/2f98be352bdd485cb646337e72e39a1be146bf34/pkg/host/internal/cpu/cpu.go#L21

and we see below warning/error in generic plugin log
```
WARNING: expected to find 'processor' key in /proc/cpuinfo attributes
WARNING: expected to find 'processor' key in /proc/cpuinfo attributes

WARNING: failed to find attributes for logical processor 0
WARNING: failed to find attributes for logical processor 1
WARNING: failed to find attributes for logical processor 10

wrong CPU information retrieved: cpu (0 physical packages, 0 cores, 0 hardware threads)
```

This is because it expect vendor_id in /proc/cpuinfo which is not available in s390x case like below
```
cpu number      : 0
physical id     : 2
core id         : 0
book id         : 2
drawer id       : 2
dedicated       : 0
address         : 0
siblings        : 12
cpu cores       : 6
version         : 00
identification  : 271F08
machine         : 9175
cpu MHz dynamic : 5508
cpu MHz static  : 5508
```
On s390x, the format is different (global vendor_id, # processors, then per-processor blocks with cpu number, book id, drawer id)
```
# cat /proc/cpuinfo 
vendor_id       : IBM/S390
# processors    : 38
bogomips per cpu: 28901.00
max thread id   : 1

cpu number      : 0
physical id     : 2
core id         : 0
book id         : 2
drawer id       : 2
dedicated       : 0
address         : 0
siblings        : 12
cpu cores       : 6
version         : 00
identification  : 271F08
machine         : 9175
cpu MHz dynamic : 5508
cpu MHz static  : 5508

cpu number      : 1
```

ghw fails to parse processors  and returns empty Processors slice

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CPU vendor detection for s390x architecture systems.

### DIFF
--- a/pkg/host/internal/cpu/cpu.go
+++ b/pkg/host/internal/cpu/cpu.go
@@ -2,6 +2,7 @@ package cpu
 
 import (
 	"fmt"
+	"runtime"
 
 	ghwPkg "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/host/internal/lib/ghw"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/host/types"
@@ -18,6 +19,13 @@ func New(ghwLib ghwPkg.GHWLib) *cpuInfoProvider {
 }
 
 func (c *cpuInfoProvider) GetCPUVendor() (types.CPUVendor, error) {
+	// s390x: ghw library does not support s390x arch.
+	// ghw.CPU() returns empty Processors slice, causing errors and warnings.
+	// Early return correct vendor to skip ghw call entirely on s390x.
+	if runtime.GOARCH == "s390x" {
+		return types.CPUVendorS390X, nil
+	}
+
 	cpuInfo, err := c.ghwLib.CPU()
 	if err != nil {
 		return -1, fmt.Errorf("can't retrieve the CPU vendor: %w", err)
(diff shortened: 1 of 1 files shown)


SchSeba (commented):
Hi @ashokpariya0 thanks for the contribution, can I ask if you can add support to the GHW lib?

I think that will be a much better solution.

If that will not be possible (rejected by the package owner) I am fine merging this one

ashokpariya0:
> Hi @ashokpariya0 thanks for the contribution, can I ask if you can add support to the GHW lib?
> 
> I think that will be a much better solution.
> 
> If that will not be possible (rejected by the package owner) I am fine merging this one

>
I analyzed the changes needed in the GHW library to add proper CPU info detection(not just vendor_id), and it looks like this will take some time, and we would prefer to handle it in a follow-up PR.

For the current use case, we only need the vendor_id, which is fixed for s390 (IBM/S390). 
We are not using any other fields from /proc/cpuinfo.

We can add something similar to what NFD does (for example:
https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/source/cpu/cpu.go#L293-L328
), but that would introduce additional and unnecessary parsing complexity for this specific case.

so, i think the existing changes are sufficient for the current case and provide a minimal fix. 
We can work on adding full entire CPU detection support in the GHW library in a follow-up PR on- https://github.com/jaypipes/ghw

ashokpariya0:
@SchSeba We have added changes in ghw, we can merge pr-https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1029 and close this pr.

PR #981, opened 2025-12-12
Add 'Block Device Plugin Until Configured' design doc

ykulazhenkov:
This design document proposes a mechanism to prevent the SR-IOV Device Plugin from announcing resources to the kubelet until the sriov-config-daemon has finished applying VF configuration on the node.

The feature introduces a new feature gate 'blockDevicePluginUntilConfigured' that adds an init container to the sriov-device-plugin DaemonSet. This init container blocks the device plugin main container from starting until the sriov-config-daemon signals that configuration is complete.

This solves a timing issue that can occur when VFs are created by an external script or application (using 'externallyManaged: true' policy), where pods could receive partially configured VFs or lose VF connectivity when configuration is applied.

Implementation of the proposal https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/994 

## Summary by CodeRabbit

* **New Features**
  * Feature gate to delay device plugin resource announcement until node network configuration completes.

* **Documentation**
  * Added design docs describing the initialization workflow, deployment behavior when enabled, annotation-based coordination, periodic unblocking, upgrade/downgrade considerations, and a test plan.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #994, opened 2025-12-19
feat: block device plugin until SR-IOV config applied

ykulazhenkov:
Add blockDevicePluginUntilConfigured feature gate that prevents the SR-IOV device plugin from starting until the sriov-config-daemon has applied the configuration for the node.

When enabled, the device plugin daemonset runs an init container that sets a wait-for-config annotation on its pod. The init container then waits until the sriov-config-daemon removes this annotation, which happens after the daemon has applied the SR-IOV configuration for the node.

This feature addresses the race condition where the device plugin starts and reports available resources before the configuration is actually applied, which can lead to pods being scheduled prematurely.

Key changes:
- Add wait-for-config subcommand to sriov-network-config-daemon
- Add init container to device plugin daemonset (when feature enabled)
- Add logic in daemon to remove annotation after config is applied
- Add Role/RoleBinding for device plugin pod access

Design doc https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/981

## Summary by CodeRabbit

* **New Features**
  - Added blockDevicePluginUntilConfigured feature gate to control device-plugin startup until configuration is applied
  - Device plugin can be blocked/unblocked until config is ready; optional init flow and a wait-for-config command enable this behavior
* **Permissions**
  - Expanded pod patch permissions and added role/rolebinding to allow plugin pod access for the unblock flow
* **Tests**
  - New unit and integration tests covering wait-for-config, daemon, and utility behaviors

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,7 +421,7 @@ Feature gates are used to enable or disable specific features in the operator.
 
 6. **Block Device Plugin Until Configured** (`blockDevicePluginUntilConfigured`)
   - **Description:** Prevents the SR-IOV device plugin from starting until the sriov-config-daemon has applied the SR-IOV configuration for the node. When enabled, the device plugin daemonset runs an init container that sets a wait-for-config annotation on its pod and waits until the sriov-config-daemon removes this annotation after applying the configuration. This addresses the race condition where the device plugin starts and reports available resources before the configuration is actually applied, which can lead to pods being scheduled prematurely.
-  - **Default:** Disabled
+  - **Default:** Enabled
 
 ### Enabling Feature Gates
 
--- a/bindata/manifests/plugins/sriov-device-plugin.yaml
+++ b/bindata/manifests/plugins/sriov-device-plugin.yaml
@@ -23,6 +23,14 @@ spec:
         component: network
         type: infra
         openshift.io/component: network
+        # NOTE: The controller uses equality.Semantic.DeepDerivative(in.Spec, ds.Spec)
+        # to detect changes in the DaemonSet's spec and decide if an update is needed.
+        # To ensure the init container is properly managed when toggling the
+        # BlockDevicePluginUntilConfigured feature gate, we define an explicit field
+        # (init-container-enabled) that is always present in the DaemonSet labels.
+        # Its value reflects the state of the feature gate and guarantees spec changes
+        # are propagated, ensuring the init container is added or removed as required.
+        init-container-enabled: "{{ .BlockDevicePluginUntilConfigured }}"
     spec:
       hostNetwork: true
       nodeSelector:
@@ -59,11 +67,7 @@ spec:
               fieldPath: metadata.namespace
       {{- end }}
       containers:
-      {{- if .BlockDevicePluginUntilConfigured }}
-      - name: sriov-device-plugin-main
-      {{- else }}
       - name: sriov-device-plugin
-      {{- end }}
         image: {{.SRIOVDevicePluginImage}}
         args:
         - --log-level=10
--- a/cmd/sriov-network-config-daemon/waitforconfig.go
+++ b/cmd/sriov-network-config-daemon/waitforconfig.go
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
 	snolog "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/log"
@@ -141,7 +142,7 @@ func startWaitForConfigManager(setupLog logr.Logger, config *rest.Config, podNam
 
 	// Set annotation on pod to signal that we are waiting for config
 	setupLog.Info("Setting annotation on pod", "annotation", consts.DevicePluginWaitConfigAnnotation)
-	err = setAnnotationOnPod(context.Background(), setupLog, tempClient, podName)
+	err = setAnnotationOnPod(ctx, setupLog, tempClient, podName)
 	if err != nil {
 		setupLog.Error(err, "failed to set annotation on pod")
 		return err
@@ -152,6 +153,7 @@ func startWaitForConfigManager(setupLog logr.Logger, config *rest.Config, podNam
 	// Watch only specific pod object
 	selector := fields.SelectorFromSet(fields.Set{"metadata.name": podName.Name})
 	mgr, err := ctrl.NewManager(config, ctrl.Options{
+		Metrics: metricsserver.Options{BindAddress: "0"},
 		Cache: cache.Options{
 			DefaultNamespaces: map[string]cache.Config{podName.Namespace: {}},
 			ByObject: map[client.Object]cache.ByObject{
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -370,6 +370,20 @@ func (dn *NodeReconciler) checkSystemdStatus() (*hosttypes.SriovResult, bool, er
 // 7. Updating the lastAppliedGeneration to the current generation.
 func (dn *NodeReconciler) apply(ctx context.Context, desiredNodeState *sriovnetworkv1.SriovNetworkNodeState, reqReboot bool, sriovResult *hosttypes.SriovResult) (ctrl.Result, error) {
 	reqLogger := log.FromContext(ctx).WithName("Apply")
+
+	// Restart the device plugin *before* applying configuration if the
+	// BlockDevicePluginUntilConfiguredFeatureGate feature is enabled.
+	// With this gate enabled, the device plugin will remain blocked until it is
+	// explicitly unblocked after configuration (see waitForDevicePluginPodAndTryUnblock).
+	// If the feature gate is not enabled, preserve legacy behavior by
+	// restarting the device plugin *after* configuration is applied.
+	if vars.FeatureGate.IsEnabled(consts.BlockDevicePluginUntilConfiguredFeatureGate) {
+		if err := dn.restartDevicePluginPod(ctx); err != nil {
+			reqLogger.Error(err, "failed to restart device plugin on the node")
+			return ctrl.Result{}, err
+		}
+	}
+
 	// apply the additional plugins after we are done with drain if needed
 	for _, p := range dn.additionalPlugins {
 		err := p.Apply()
@@ -395,13 +409,16 @@ func (dn *NodeReconciler) apply(ctx context.Context, desiredNodeState *sriovnetw
 		return ctrl.Result{}, dn.rebootNode()
 	}
 
-	if err := dn.restartDevicePluginPod(ctx); err != nil {
-		reqLogger.Error(err, "failed to restart device plugin on the node")
-		return ctrl.Result{}, err
-	}
 	if vars.FeatureGate.IsEnabled(consts.BlockDevicePluginUntilConfiguredFeatureGate) {
 		if err := dn.waitForDevicePluginPodAndTryUnblock(ctx, desiredNodeState); err != nil {
 			reqLogger.Error(err, "failed to wait for device plugin pod to start and try to unblock it")
+			return ctrl.Result{}, err
+		}
+	} else {
+		// if the feature gate is not enabled we preserver the old behavior
+		// and restart device plugin after configuration is applied
+		if err := dn.restartDevicePluginPod(ctx); err != nil {
+			reqLogger.Error(err, "failed to restart device plugin on the node")
 			return ctrl.Result{}, err
 		}
 	}
@@ -452,7 +469,7 @@ func (dn *NodeReconciler) tryUnblockDevicePlugin(ctx context.Context,
 	for _, pod := range devicePluginPods {
 		if err := utils.RemoveAnnotationFromObject(ctx, &pod,
 			consts.DevicePluginWaitConfigAnnotation, dn.client); err != nil {
-			return fmt.Errorf("failed to remove wait-for-config annotation from pod: %w", err)
+			return fmt.Errorf("failed to remove %s annotation from pod: %w", consts.DevicePluginWaitConfigAnnotation, err)
 		}
 	}
 	return nil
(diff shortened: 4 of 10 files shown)


cmd/sriov-network-config-daemon/main.go
```diff
@@ -44,5 +45,6 @@ func init() {
 func main() {
 	if err := rootCmd.Execute(); err != nil {
 		log.Log.Error(err, "error executing sriov-network-config-daemon")
+		os.Exit(1)
```
ykulazhenkov, line 48: This change submitted as a separate PR https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/993

cmd/sriov-network-config-daemon/waitforconfig.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	// Set annotation on pod to signal that we are waiting for config
+	setupLog.Info("Setting annotation on pod", "annotation", consts.DevicePluginWaitConfigAnnotation)
+	err = setAnnotationOnPod(context.Background(), setupLog, tempClient, podName)
+	if err != nil {
+		setupLog.Error(err, "failed to set annotation on pod")
+		return err
+	}
+	setupLog.Info("Annotation set successfully, waiting for removal")
```
zeeke: If the device-plugin is restarted for any reason (not due to a config-daemon work), there is a risk it get stuck waiting for someone to remove the annotation. right?

ykulazhenkov: The sriov-config-daemon regularly checks the node's current configuration. With this PR, we've expanded the check to also verify that no device plugins are in a locked state. So, even if device plugin is restarted it will be eventually unlocked by the config daemon

cmd/sriov-network-config-daemon/waitforconfig.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// Configure Manager
+	// Watch only specific pod object
+	selector := fields.SelectorFromSet(fields.Set{"metadata.name": podName.Name})
+	mgr, err := ctrl.NewManager(config, ctrl.Options{
+		Cache: cache.Options{
+			DefaultNamespaces: map[string]cache.Config{podName.Namespace: {}},
```
adrianchiris, line 158: note on k8s API access:
1. we watch for specific pod in a specific namespace.
2. is active until the annotation is removed

its lifespan is expected to be relatively short and in a stable state (WRT sriov configuration) will not even be active.

bindata/manifests/plugins/sriov-device-plugin.yaml
```diff
@@ -39,8 +39,31 @@ spec:
...
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+      {{- end }}
       containers:
+      {{- if .BlockDevicePluginUntilConfigured }}
```
ykulazhenkov: This is a hack to remove the init container when the FeatureGate is disabled.

We need this because, before updating the DaemonSet (DS), we use the following check:
```
		// DeepDerivative checks for changes only comparing non-zero fields in the source struct.
		// This skips default values added by the api server.
		// References in https://github.com/kubernetes-sigs/kubebuilder/issues/592#issuecomment-625738183

		// Note(Adrianc): we check Equality of OwnerReference as we changed sriov-device-plugin owner ref
		// from SriovNetworkNodePolicy to SriovOperatorConfig, hence even if there is no change in spec,
		// we need to update the obj's owner reference.

		if equality.Semantic.DeepEqual(in.OwnerReferences, ds.OwnerReferences) &&
			equality.Semantic.DeepDerivative(in.Spec, ds.Spec) {
			logger.V(1).Info("Daemonset spec did not change, not updating")
			return nil
		}
```

We cannot solve this using a special label/annotation in this case:

1. If we add a label/annotation only when the FeatureGate is enabled, we won’t be able to roll back from “enabled” to “disabled” because `DeepDerivative` only compares fields that are present (non-zero) in the source object.
2. If we always add a label/annotation reflecting the FeatureGate state (for example, `init-container-enabled: "true"/"false"`), then installing the first operator version that includes this change would trigger a full restart of the device plugin across the cluster-even if the device plugin image and configuration are unchanged.
		
So, we need to rely on some field that is already present in the DaemonSet spec.

pkg/daemon/daemon.go
```diff
@@ -536,64 +575,123 @@ func (dn *NodeReconciler) handleDrain(ctx context.Context, desiredNodeState *sri
...
+		}
+		// If annotation is not found within the timeout, log a warning and proceed.
+		// The device plugin pod will be unblocked by the periodic check logic in tryUnblockDevicePlugin.
+		log.Log.Info("waitForDevicePluginPodAndUnblock(): WARNING: device plugin pod with " +
+			"wait-for-config annotation not found within timeout")
+	}
+	if len(devicePluginPods) > 0 {
+		// try to unblock all device plugin pods we retrieved with the latest loop iteration
+		if err := dn.tryUnblockDevicePlugin(ctx, desiredNodeState, devicePluginPods); err != nil {
+			return err
```
zeeke: in case of timeout (`stdErrors.Is(err, context.DeadlineExceeded) == true`), would it be better to exit the function  instead of calling `dn.tryUnblockDevicePlugin(...)` with the devicepluginpods from the last iteration?

README.md
```diff
@@ -419,6 +419,10 @@ Feature gates are used to enable or disable specific features in the operator.
   - **Description:** Enables the firmware reset via `mstfwreset` before a system reboot. This feature is specific to Mellanox network devices and is used to ensure that the firmware is properly reset during system maintenance.
   - **Default:** Disabled
 
+6. **Block Device Plugin Until Configured** (`blockDevicePluginUntilConfigured`)
+  - **Description:** Prevents the SR-IOV device plugin from starting until the sriov-config-daemon has applied the SR-IOV configuration for the node. When enabled, the device plugin daemonset runs an init container that sets a wait-for-config annotation on its pod and waits until the sriov-config-daemon removes this annotation after applying the configuration. This addresses the race condition where the device plugin starts and reports available resources before the configuration is actually applied, which can lead to pods being scheduled prematurely.
+  - **Default:** Disabled
```
adrianchiris: lets go with default Enabled

bindata/manifests/plugins/sriov-device-plugin.yaml
adrianchiris: when upgrading the operator, the device plugin image is usually updated as well so device plugin would get restarted regardless IMO.

so fine to go with the labels approach.

regardless, please add some comment on why we do this.

cmd/sriov-network-config-daemon/waitforconfig.go
```diff
@@ -0,0 +1,204 @@
...
+	// Create a temporary client to set the annotation immediately
+	tempClient, err := client.New(config, client.Options{})
+	if err != nil {
+		setupLog.Error(err, "failed to create kubernetes client")
+		return err
+	}
+
+	// Set annotation on pod to signal that we are waiting for config
+	setupLog.Info("Setting annotation on pod", "annotation", consts.DevicePluginWaitConfigAnnotation)
+	err = setAnnotationOnPod(context.Background(), setupLog, tempClient, podName)
```
adrianchiris: nit: use ctx ?

cmd/sriov-network-config-daemon/waitforconfig.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	if !utils.ObjectHasAnnotationKey(pod, consts.DevicePluginWaitConfigAnnotation) {
+		logger.Info("Annotation removed, device plugin can proceed")
+		r.Cancel()
+		return ctrl.Result{}, nil
+	}
+
+	logger.Info("Annotation still present, waiting...")
+	return ctrl.Result{RequeueAfter: consts.DaemonRequeueTime}, nil
```
adrianchiris, line 93: we watch for pods so we will get called  once the pod is updated. but i guess its better to have that in sync with the daemon.

pkg/daemon/daemon.go
```diff
@@ -415,14 +435,33 @@ func (dn *NodeReconciler) apply(ctx context.Context, desiredNodeState *sriovnetw
...
 	return ctrl.Result{RequeueAfter: consts.DaemonRequeueTime}, nil
 }
 
+// tryUnblockDevicePlugin checks if the device plugin can be unblocked
+func (dn *NodeReconciler) tryUnblockDevicePlugin(ctx context.Context,
+	desiredNodeState *sriovnetworkv1.SriovNetworkNodeState, devicePluginPods []corev1.Pod) error {
+	funcLog := log.Log.WithName("tryUnblockDevicePlugin")
+	funcLog.Info("check if we need to remove the wait-for-config annotation")
+	// we want to unblock the device plugin only if the desired state contains configuration for the interfaces
+	if len(desiredNodeState.Spec.Interfaces) == 0 {
```
adrianchiris, line 465: why do we care ?

if spec.Interfaces == 0 then device plugin config is empty -> no resources will be exposed.

in this case device plugin will not be ready if no policy applies for the node ?

pkg/daemon/daemon.go
```diff
@@ -415,14 +435,33 @@ func (dn *NodeReconciler) apply(ctx context.Context, desiredNodeState *sriovnetw
...
+		funcLog.Info("desired node state has no interfaces, keep the wait-for-config annotation")
+		return nil
+	}
+	for _, pod := range devicePluginPods {
+		if err := utils.RemoveAnnotationFromObject(ctx, &pod,
+			consts.DevicePluginWaitConfigAnnotation, dn.client); err != nil {
+			return fmt.Errorf("failed to remove wait-for-config annotation from pod: %w", err)
```
adrianchiris: nit: `mt.Errorf("failed to remove %s annotation from pod: %w", consts.DevicePluginWaitConfigAnnotation, err)`

pkg/daemon/daemon.go
```diff
@@ -536,64 +575,123 @@ func (dn *NodeReconciler) handleDrain(ctx context.Context, desiredNodeState *sri
...
 	}
+	return nil
+}
 
+// waitForDevicePluginPodAndTryUnblock waits for the new device plugin pod to start and set the wait-for-config annotation. This allows us to unblock
+// the new device plugin instance within the same reconciliation loop without needing to wait
+// for the periodic check. We expect to have at least one device plugin pod for the node.
+func (dn *NodeReconciler) waitForDevicePluginPodAndTryUnblock(ctx context.Context, desiredNodeState *sriovnetworkv1.SriovNetworkNodeState) error {
+	funcLog := log.Log.WithName("waitForDevicePluginPodAndTryUnblock")
+	funcLog.Info("waiting for device plugin to set wait-for-config annotation")
```
adrianchiris: nit: log the annotation from consts pkg

pkg/daemon/daemon.go
```diff
@@ -536,64 +575,123 @@ func (dn *NodeReconciler) handleDrain(ctx context.Context, desiredNodeState *sri
...
+	var devicePluginPods []corev1.Pod
+	err := wait.PollUntilContextTimeout(ctx, time.Second, 2*time.Minute, true,
+		func(ctx context.Context) (bool, error) {
+			var err error
+			devicePluginPods, err = dn.getDevicePluginPodsForNode(ctx)
+			if err != nil {
+				log.Log.Error(err, "waitForDevicePluginPodAndUnblock(): failed to get device plugin pod while waiting for a new pod to start")
```
adrianchiris: nit: use funcLog ?

pkg/daemon/daemon.go
```diff
@@ -536,64 +575,123 @@ func (dn *NodeReconciler) handleDrain(ctx context.Context, desiredNodeState *sri
...
+	funcLog.V(2).Info("try to restart device plugin pod")
+	devicePluginPods, err := dn.getDevicePluginPodsForNode(ctx)
+	if err != nil {
+		return err
+	}
+	for _, pod := range devicePluginPods {
+		podUID := pod.UID
+		funcLog.V(2).Info("Found device plugin pod, deleting it",
+			"pod-name", pod.Name, "pod-uid", podUID)
 		err = dn.client.Delete(ctx, &pod)
```
adrianchiris, line 629: we might delete other instances of device plugin if they exist in the operator namespace.

not a new issue. just an observation

bindata/manifests/plugins/sriov-device-plugin.yaml
ykulazhenkov: yes, make sense. I changed the logic to rely on a label

pkg/daemon/daemon.go
ykulazhenkov: done

README.md
ykulazhenkov: enabled by default

cmd/sriov-network-config-daemon/waitforconfig.go
ykulazhenkov: done

cmd/sriov-network-config-daemon/waitforconfig.go
ykulazhenkov: Yes, we are performing a watch. This explicit requeue is used to periodically print log messages so users know the controller is still running.

pkg/daemon/daemon.go
ykulazhenkov: done

pkg/daemon/daemon.go
ykulazhenkov: done

pkg/daemon/daemon.go
ykulazhenkov: done

pkg/daemon/daemon.go
ykulazhenkov: Yes, if someone happens to place the device plugin instance in the sriov-network-operator namespace, it will be removed.

pkg/daemon/daemon.go
ykulazhenkov: This serves as an additional safeguard against the "early announcement" of devices that have not yet been configured by the operator. In theory, there can be a situation where the ConfigMap for the device plugin is already updated, but the config daemon has not yet received the spec (spec is not set yet in SriovNetworkNodeState).

The controller is expected to remove the device plugin from the node if there is no configuration (which is the current behavior), so device plugin pods should not remain stuck in the "init" state indefinitely.

pkg/featuregate/featuregate.go
```diff
@@ -2,33 +2,57 @@ package featuregate
...
+	// The provided map is merged with the default features state.
 	Init(features map[string]bool)
 	// String returns string representation of the feature state
 	String() string
 }
 
-// New returns default implementation of the FeatureGate interface
+// New returns default implementation of the FeatureGate interface with the default features state
 func New() FeatureGate {
 	return &featureGate{
```
adrianchiris, line 36: nit: `return NewWithDefaultFeaturers(DefaultFeatureStates)` ?

minor, not a must by any means.

ykulazhenkov: This is valid, but I don't think it will significantly enhance code quality, so let's leave it as it is.

ykulazhenkov:
Thanks for the review @zeeke @e0ne @adrianchiris! There are 3 approvals on the PR. As far as I recall, we have a "no self-merge" rule for this repo. Someone needs to hit the merge button.

PR #999, opened 2025-12-25
fix: check for sriov result file in systemd mode

rollandf:
If the result file doesn't exist but the systemd service is enabled, the sriovResultExists == true.
Fix this logic by checking if the result actually exists

## Summary by CodeRabbit

* **New Features**
  * Exposed systemd status checking capability as a public API.

* **Bug Fixes**
  * Improved systemd status detection to accurately reflect actual system state rather than relying on default assumptions.

* **Tests**
  * Added comprehensive test suite covering systemd status verification across multiple scenarios and configurations.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -169,7 +169,7 @@ func (dn *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	}
 
 	// if we are running in systemd mode we want to get the sriov result from the config-daemon that runs in systemd
-	sriovResult, sriovResultExists, err := dn.checkSystemdStatus()
+	sriovResult, sriovResultExists, err := dn.CheckSystemdStatus()
 	//TODO: in the case we need to think what to do if we try to apply again or not
 	if err != nil {
 		reqLogger.Error(err, "failed to check systemd status unexpected error")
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon_internal_test.go
```diff
@@ -0,0 +1,56 @@
...
+	. "github.com/onsi/gomega"
+	"go.uber.org/mock/gomock"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+	mock_helper "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/helper/mock"
+	hosttypes "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/host/types"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
+)
+
+var _ = Describe("Daemon checkSystemdStatus", func() {
```
SchSeba: can we please add this into the daemon_test that already exist, not sure if that is only me be it feel weird to have another test file :P

pkg/daemon/daemon_internal_test.go
```diff
@@ -0,0 +1,56 @@
...
+		mockCtrl = gomock.NewController(GinkgoT())
+		hostHelper = mock_helper.NewMockHostHelpersInterface(mockCtrl)
+		reconciler = &NodeReconciler{
+			hostHelpers: hostHelper,
+		}
+		vars.UsingSystemdMode = true
+	})
+
+	AfterEach(func() {
+		mockCtrl.Finish()
```
SchSeba: I think we should also revert the `vars.UsingSystemdMode` if this tests are running before the general daemon it can impact the other tests

pkg/daemon/daemon_internal_test.go
```diff
@@ -0,0 +1,56 @@
...
+	})
+
+	Context("when systemd services are enabled", func() {
```
SchSeba: Consider adding test cases for additional scenarios:

When systemd services are not enabled
When ReadSriovResult() returns an error
When not in systemd mode (vars.UsingSystemdMode = false)

rollandf: Done

pkg/daemon/daemon_internal_test.go
rollandf: Done

pkg/daemon/daemon_internal_test.go
rollandf: Done, I had to make the func public.
Tell me what you think

SchSeba (approved):

PR #1001, opened 2025-12-30
feat: Make SR-IOV Network Operator working in STIG-Enabled Kubernetes

rollandf:
… Cluster

## Summary by CodeRabbit

* **Security**
  * Hardened container runtime for operator, cleanup hook, and webhook: dropped capabilities, enforced non-root execution, enabled seccomp runtime profile, and prevented privilege escalation.
  * Docker image now runs processes as a non-privileged (non-root) user at runtime.

### DIFF
--- a/bindata/manifests/operator-webhook/server.yaml
+++ b/bindata/manifests/operator-webhook/server.yaml
@@ -82,6 +82,12 @@ spec:
         securityContext:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - ALL
+          runAsNonRoot: true
+          seccompProfile:
+            type: RuntimeDefault
         resources:
           requests:
             cpu: 10m
--- a/bindata/manifests/webhook/server.yaml
+++ b/bindata/manifests/webhook/server.yaml
@@ -85,6 +85,12 @@ spec:
         securityContext:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - ALL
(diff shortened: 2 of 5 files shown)


bindata/manifests/operator-webhook/server.yaml
```diff
@@ -82,6 +82,12 @@ spec:
         securityContext:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - ALL
+          runAsNonRoot: true
```
SchSeba, line 88: do we need to change the dockerfile for it also?

bindata/manifests/webhook/server.yaml
```diff
@@ -85,6 +85,12 @@ spec:
         securityContext:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - ALL
+          runAsNonRoot: true
```
SchSeba, line 91: do we need to open a PR on the resource injector to not use root as the user? I am not sure if that is the case already :P

rollandf: We already are using non-root:
https://github.com/k8snetworkplumbingwg/network-resources-injector/blob/master/Dockerfile#L22

bindata/manifests/operator-webhook/server.yaml
rollandf: We already are using non-root:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/Dockerfile.webhook#L9

PR #1003, opened 2026-01-01
Design: implement control grpc to the device plugin

SchSeba:
## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive design documentation for a new gRPC-based control interface for the SR-IOV Network Device Plugin, including architecture specifications, workflow descriptions, API definitions, implementation notes, upgrade pathways, and test strategies.

### DIFF
--- a/doc/design/device-plugin-grpc-control.md
+++ b/doc/design/device-plugin-grpc-control.md
@@ -1,0 +1,362 @@
+---
+title: Device Plugin gRPC Control Interface
+authors:
+  - TBD
+reviewers:
+  - TBD
+creation-date: 01-01-2026
+last-updated: 01-01-2026
+---
+
+# Device Plugin gRPC Control Interface
+
+## Summary
+
+This design proposes implementing a gRPC server in the SR-IOV Network Device Plugin
+that enables the sriov-config-daemon to establish a long-lived connection and control
+when the device plugin publishes SR-IOV VF resources to kubelet. The config-daemon
+passes the `SriovNetworkNodeState` generation to the device plugin, which stores it
+and only starts exposing devices when connected. On disconnect, the device plugin
+stops reporting devices and cleans them from kubelet.
+
+## Motivation
+
+### Problem Description
+
+When using policies with `externallyManaged: true`, VFs are created by an external
+script or application instead of the sriov-config-daemon. The SR-IOV Network Operator
+creates the sriov-device-plugin configuration and enables its DaemonSet pod at the
+same time it starts applying SR-IOV state to the node.
+
+This creates a race condition where the sriov-device-plugin may start and announce
+resources to kubelet before VFs are fully configured by the sriov-config-daemon.
+
+#### Race Condition Sequence
+
+After node provisioning or host reboot:
+
+1. External script creates VFs (partially configured, e.g., no RDMA GUID assigned)
+2. sriov-device-plugin pod starts, discovers VFs and announces them to kubelet
+3. Kubernetes scheduler schedules pods to this node
+4. Pods start with partially configured VFs
+5. sriov-config-daemon applies VF configuration (unbind/bind VF driver)
+6. Pods that were not evicted lose VF connectivity
+
+#### Existing Proposal: Init Container Approach (PR #981)
+
+PR #981 proposes adding an init container to the device plugin DaemonSet that blocks
+until the config-daemon signals configuration completion via pod annotations. While
+functional, this approach has several limitations:
+
+**Complexity:**
+- Requires adding an init container to the device plugin DaemonSet
+- Uses Kubernetes API for coordination (pod annotation watching)
+- Requires periodic polling by config-daemon to detect device plugin restarts
+- Adds latency due to annotation-based signaling through the API server
+
+**Limitations:**
+- Cannot dynamically pause device reporting without pod restart
+- Device plugin has no awareness of which configuration generation it should serve
+- Disconnect/reconnect scenarios require pod restarts
+- No direct synchronization between daemon and device plugin processes
+
+### Use Cases
+
+* As a user, I want pods to only receive SR-IOV VFs after they are fully configured
+* As a user, I want the config-daemon to have direct control over when the device
+  plugin starts and stops advertising resources
+* As a user, I want the device plugin to automatically stop advertising resources
+  if the config-daemon is not connected or crashes
+* As a user, I want the device plugin to be aware of the configuration generation
+  it is serving
+
+### Goals
+
+* Implement a gRPC server in the device plugin for config-daemon communication
+* Enable the config-daemon to control device plugin resource announcement lifecycle
+* Provide generation-based synchronization between operator and device plugin
+* Automatically clean up resources when the connection is lost
+* Maintain backward compatibility when the feature is not used
+
+### Non-Goals
+
+* Changing the drain behavior for DaemonSet pods
+* Modifying how externally managed VFs are created
+* Changing the VF configuration process itself
+* Replacing the existing kubelet device plugin gRPC interface
+
+## Proposal
+
+Add a new gRPC server in the sriov-network-device-plugin that listens on a Unix
+domain socket. The sriov-config-daemon establishes a long-lived streaming connection
+to this server and signals when the device plugin should start exposing devices.
+
+### Workflow Description
+
+#### Device Plugin Startup (Controlled Mode)
+
+When the device plugin starts with the control interface enabled:
+
+1. Device plugin starts and initializes the gRPC control server
+2. Device plugin waits for a connection from the config-daemon
+3. When the config-daemon connects and sends an `EnableDevices` RPC with the
+   `SriovNetworkNodeState` generation, the device plugin:
+   - Stores the generation for reference
+   - Starts the resource servers and registers with kubelet
+   - Begins the ListAndWatch loop to expose devices
+4. The connection remains open as a health check mechanism
+5. If the connection is lost (daemon restart, crash, etc.), the device plugin:
+   - Stops all resource servers
+   - Cleans up kubelet registrations
+   - Waits for a new connection
+
+#### Config Daemon Behavior
+
+When the config-daemon completes VF configuration:
+
+1. After successfully applying SR-IOV configuration, connect to the device plugin
+   gRPC server
+2. Call `EnableDevices` RPC with the current `SriovNetworkNodeState` generation
+3. Maintain the connection for the lifetime of the configuration
+4. On reconfiguration that requires device plugin restart:
+   - Close the existing connection (triggers device cleanup)
+   - Wait for configuration to complete
+   - Establish a new connection with the new generation
+
+#### Sequence Diagram
+
+```mermaid
+sequenceDiagram
+    participant ES as External Script
+    participant DP as Device Plugin
+    participant GS as gRPC Server (in DP)
+    participant CD as Config Daemon
+    participant K as Kubelet
+
+    ES->>ES: Create VFs
+    DP->>GS: Start gRPC Control Server
+    DP->>DP: Discover Devices (but don't register)
+    Note over DP: Waiting for control connection
+    
+    CD->>CD: Apply VF Configuration
+    CD->>GS: Connect + EnableDevices(generation=5)
+    GS->>DP: Signal to start
+    DP->>DP: Store generation=5
+    DP->>K: Register & ListAndWatch
+    K->>K: Update Node Capacity
+    
+    Note over CD,GS: Long-lived connection maintained
+    
+    CD->>CD: New configuration needed
+    CD->>GS: Disconnect
+    GS->>DP: Connection lost signal
+    DP->>K: Stop ListAndWatch, cleanup
+    K->>K: Remove SR-IOV resources
+    
+    CD->>CD: Apply new VF Configuration
+    CD->>GS: Connect + EnableDevices(generation=6)
+    GS->>DP: Signal to start
+    DP->>DP: Store generation=6
+    DP->>K: Register & ListAndWatch
+```
+
+### API Extensions
+
+#### gRPC Service Definition
+
+```protobuf
+syntax = "proto3";
+
+package sriovdp.control.v1;
+
+option go_package = "github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/control/v1";
+
+// ControlService provides an interface for the config-daemon to control
+// the device plugin's resource announcement lifecycle.
+service ControlService {
+  // EnableDevices establishes a long-lived connection that signals the device
+  // plugin to start exposing devices to kubelet. The connection acts as a
+  // health check - when it closes, the device plugin stops advertising resources.
+  // The client sends the initial request with the generation, and the server
+  // streams status updates back.
+  rpc EnableDevices(EnableDevicesRequest) returns (stream DevicePluginStatus);
+}
+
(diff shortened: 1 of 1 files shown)


doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+package sriovdp.control.v1;
+
+option go_package = "github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/control/v1";
+
+// ControlService provides an interface for the config-daemon to control
+// the device plugin's resource announcement lifecycle.
+service ControlService {
+  // EnableDevices establishes a long-lived connection that signals the device
+  // plugin to start exposing devices to kubelet. The connection acts as a
```
adrianchiris, line 178: i think its optionally loading the new configuration then start exposing devices to kubelet

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+  // health check - when it closes, the device plugin stops advertising resources.
```
adrianchiris, line 179: you mean the device plugin will re-advertize zero devices.

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+  // The client sends the initial request with the generation, and the server
+  // streams status updates back.
+  rpc EnableDevices(EnableDevicesRequest) returns (stream DevicePluginStatus);
```
adrianchiris, line 182: what will ensure that wen EnableDevicesRequest is sent, the config map file is actually updated in the device plugin pod

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+}
+
+// DevicePluginStatus is streamed back to the config-daemon to provide
+// visibility into the device plugin state.
+message DevicePluginStatus {
+  // Current state of the device plugin
+  State state = 1;
+  
+  // Number of resource pools currently being served
+  int32 resource_pool_count = 2;
```
adrianchiris, line 202: what will we do with this information ?

same for the below

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+  
+  // Optional error message if state is ERROR
+  string error_message = 5;
+  
+  enum State {
+    UNKNOWN = 0;
+    INITIALIZING = 1;
+    SERVING = 2;
+    ERROR = 3;
+    STOPPING = 4;
```
adrianchiris, line 218: no STOP ?

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+  
+  // The generation currently being served
+  int64 serving_generation = 4;
+  
+  // Optional error message if state is ERROR
+  string error_message = 5;
+  
+  enum State {
+    UNKNOWN = 0;
+    INITIALIZING = 1;
```
adrianchiris, line 215: when will we be in this state

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+| **Additional containers** | Yes (init container) | No |
+| **Communication method** | Pod annotations via K8s API | Direct gRPC over Unix socket |
+| **Latency** | Higher (API server roundtrip) | Lower (direct IPC) |
+| **Generation awareness** | No | Yes |
+| **Dynamic pause/resume** | Requires pod restart | Connection close/reconnect |
+| **Failure detection** | Polling required | Immediate (connection closed) |
+| **Resource overhead** | Additional container image | Minimal (in-process server) |
+| **Kubernetes API load** | Annotation updates | None |
+| **Implementation scope** | Operator only | Both operator and device plugin |
+| **Complexity** | Moderate | Higher initial, simpler operation |
```
adrianchiris, line 290: the complexity for init container is low id say. its like ~300 lines of actual code and alot of it is boiler plate (logs, create client etc)

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+# Device Plugin gRPC Control Interface
+
+## Summary
+
+This design proposes implementing a gRPC server in the SR-IOV Network Device Plugin
+that enables the sriov-config-daemon to establish a long-lived connection and control
+when the device plugin publishes SR-IOV VF resources to kubelet. The config-daemon
+passes the `SriovNetworkNodeState` generation to the device plugin, which stores it
+and only starts exposing devices when connected. On disconnect, the device plugin
```
adrianchiris, line 19: a note: the latter part was never done by device plugin. (probe function has a TODO to implement)

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+#### Race Condition Sequence
+
+After node provisioning or host reboot:
+
+1. External script creates VFs (partially configured, e.g., no RDMA GUID assigned)
+2. sriov-device-plugin pod starts, discovers VFs and announces them to kubelet
+3. Kubernetes scheduler schedules pods to this node
+4. Pods start with partially configured VFs
+5. sriov-config-daemon applies VF configuration (unbind/bind VF driver)
+6. Pods that were not evicted lose VF connectivity
```
adrianchiris, line 43: do you mean daemonset pods that consume VFs ?

even today we dont handle it upon sriov reconfiguration regardless of externally managed right ?
having a switch to delay device plugin started indeed solves the intial config but not reconfiguration.
this however is a separate problem which i believe we can solve by draining ds pods as well (only those with sriov prefixed resources imo)

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+#### Existing Proposal: Init Container Approach (PR #981)
+
+PR #981 proposes adding an init container to the device plugin DaemonSet that blocks
+until the config-daemon signals configuration completion via pod annotations. While
+functional, this approach has several limitations:
+
+**Complexity:**
+- Requires adding an init container to the device plugin DaemonSet
```
adrianchiris, line 52: this is not a limitation its an implementation detail

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+- Uses Kubernetes API for coordination (pod annotation watching)
```
adrianchiris, line 53: this is not a limitation, its an implementation detail IMO.

if using k8s API for coordination is a limitation then we sure took a few wrong turns with the operator as well :)

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+- Requires periodic polling by config-daemon to detect device plugin restarts
```
adrianchiris, line 54: i believe current approach would be more heavy on the config-daemon no ?

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+- Adds latency due to annotation-based signaling through the API server
```
adrianchiris, line 55: what do you mean by this ?

i think the main limitations of this apprach is:

1. another entity creating a watch on each node to API server (although limited to a specific NS and pod and relatively short lived)
2. detection of device plugin reset may be delayed up until polling interval

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+**Limitations:**
+- Cannot dynamically pause device reporting without pod restart
+- Device plugin has no awareness of which configuration generation it should serve
+- Disconnect/reconnect scenarios require pod restarts
```
adrianchiris, line 60: this is a limitation of the current implementation and not the init container approach.

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+#### Device Plugin Startup (Controlled Mode)
+
+When the device plugin starts with the control interface enabled:
+
+1. Device plugin starts and initializes the gRPC control server
+2. Device plugin waits for a connection from the config-daemon
+3. When the config-daemon connects and sends an `EnableDevices` RPC with the
+   `SriovNetworkNodeState` generation, the device plugin:
+   - Stores the generation for reference
+   - Starts the resource servers and registers with kubelet
```
adrianchiris, line 105: what ensures the generation matches the config file?

maybe push the desired config through this API

adrianchiris: also IMO the proposed service can be a non streaming API, then use the standard grpc healh server for watching when the service goes down


see https://github.com/grpc/grpc/blob/master/doc/health-checking.md

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+1. After successfully applying SR-IOV configuration, connect to the device plugin
+   gRPC server
+2. Call `EnableDevices` RPC with the current `SriovNetworkNodeState` generation
+3. Maintain the connection for the lifetime of the configuration
+4. On reconfiguration that requires device plugin restart:
+   - Close the existing connection (triggers device cleanup)
+   - Wait for configuration to complete
+   - Establish a new connection with the new generation
+
```
adrianchiris, line 125: missing: on connection lost flow.

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+}
+
+// EnableDevicesRequest is sent by the config-daemon to enable device advertising.
+message EnableDevicesRequest {
+  // The generation of the SriovNetworkNodeState that triggered this enable.
+  // This allows the device plugin to track which configuration it is serving.
+  int64 node_state_generation = 1;
+  
+  // Optional: node name for logging/debugging purposes
+  string node_name = 2;
```
adrianchiris, line 192: why is this needed ? even if optional

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+}
+
+// DevicePluginStatus is streamed back to the config-daemon to provide
```
adrianchiris, line 195: any rules on when to stream this ?

i.e for each EnableDevicesRequest its expected to get exactly one DevicePluginStatus msg ?

doc/design/device-plugin-grpc-control.md
adrianchiris: apart of SERVING and ERROR i dont understand when the others be transmitted.

even for error, if we are going to close the streaming connection (and probably exit) we could use the regular gRPC error  IMO.

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+When `--controlled-mode=false` (default), the device plugin behaves exactly as
+before - starting resource servers immediately on startup. This ensures existing
+deployments continue to work without changes.
+
+#### Error Handling
+
+- **Connection timeout:** Config-daemon retries connection with exponential backoff
+- **Unexpected disconnect:** Device plugin stops servers and waits for reconnection
+- **Config-daemon crash:** Device plugin detects closed connection and cleans up
```
adrianchiris, line 274: same as expected disconnect

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+- **Device plugin restart:** Config-daemon detects disconnect and reconnects
+
+### Comparison with Init Container Approach (PR #981)
+
+| Aspect | Init Container (PR #981) | gRPC Control Interface |
+|--------|--------------------------|------------------------|
+| **Additional containers** | Yes (init container) | No |
+| **Communication method** | Pod annotations via K8s API | Direct gRPC over Unix socket |
+| **Latency** | Higher (API server roundtrip) | Lower (direct IPC) |
+| **Generation awareness** | No | Yes |
```
adrianchiris, line 284: what is the benefit of generation awareness ? 

AFAIU its to handle cases where u send the same EnableDevicesRequest in the same connection to prevent device plugin from reloading config.

this limitation is non existent in the init container approach.

also TBH im not sure with the current flow config daemon can send EnableDevicesRequest with the same generation twice in a row in the same connection

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+6. **Cleaner Lifecycle Management:** Connection-based lifecycle is more intuitive
+   than annotation-based coordination.
+
+7. **Future Extensibility:** The gRPC interface can be extended with additional
+   RPCs for features like:
+   - Device-level enable/disable
+   - Configuration hot-reload
+   - Detailed device status queries
+
```
adrianchiris, line 316: we should list some downsides relative to the init container appraoch as well IMO.

1. relatively complex logic introduced in both config daemon and device plugin.
3. risk of API incompatibility which we need to take care of (e.g config daemon at versionX of gRPC API and device plugin at version X+1) or via versa
4. device plugin in its current form is pretty stable, init container approach introduces 0 changes to device plugin


generally we are aiming to switch to DRA so need to ask ourselves if the above is really worth it.
(also a note, in DRA we also access k8s API :) )

doc/design/device-plugin-grpc-control.md
```diff
@@ -0,0 +1,362 @@
...
+* Device plugin waits for connection in controlled mode
+* Device plugin starts servers after EnableDevices RPC
+* Device plugin stops servers on connection close
+* Generation is correctly stored and reported
+* Backward compatibility when controlled mode is disabled
+
+#### Integration Tests
+
+* End-to-end flow with config-daemon and device plugin
+* Verify devices appear in kubelet only after EnableDevices
```
adrianchiris, line 351: this is integration test or e2e test ?

doc/design/device-plugin-grpc-control.md
zeeke: the point here is that a any config-daemon needs talk to the device-plugin instance that is running on the same node. Using the api server is suboptimal in this case (limits the scalability, can introduce bugs as a config-daemon might target the wrong device-plugin), and a direct wire, like gRPC over a socketfile, might be a better solution. 

Some decision we made with the operator are not that good, after all

doc/design/device-plugin-grpc-control.md
ykulazhenkov: > maybe push the desired config through this API

I agree with Adrian. 

If we are considering adding a control socket or "externally managed mode" 😄 to the SR-IOV device plugin, it might be beneficial to design it so that the device plugin's configuration can be dynamically updated through this interface.

doc/design/device-plugin-grpc-control.md
ykulazhenkov: note: the proposed implementation of the "Init Container Approach" uses sriov-config-daemon binary (and image), so no extra image pull is needed.

doc/design/device-plugin-grpc-control.md
ykulazhenkov: note: the proposed implementation of the "Init Container Approach" uses periodic checks, but this checks rely on client's cache, meaning no extra calls to the k8s API are made.

SchSeba:
closing this one for now.
I will create a new one after we have the DRA implementation to replace the device plugin

PR #1009, opened 2026-01-11
update github action to use the go verion step for manifest

SchSeba:


rollandf (approved):

zeeke (approved):

zeeke:
next time we bump the go version in actions file, we might use the `go-version-file` parameter:
https://github.com/actions/setup-go?tab=readme-ov-file#go-version-file

PR #1012, opened 2026-01-11
Conditions/nodestate

SchSeba:
## Summary by CodeRabbit

* **New Features**
  * Added Ready, Progressing, and Draining status conditions for node-state visibility.
  * Added drain lifecycle reporting for pending, draining, completed, and error states.
  * Exposed condition statuses as CRD printer columns.
  * Added status updates for configuration progress, drain waits, errors, and observed generations.

* **Bug Fixes**
  * Improved drain error reporting and recovery.
  * Preserved condition transition times and independently managed condition types.
  * Improved stability detection with a continuous stable period.

* **Documentation**
  * Documented node-state conditions, reasons, ownership, and stability behavior.


test/util/cluster/cluster.go
```diff
@@ -297,15 +298,22 @@ func SriovStable(operatorNamespace string, clients *testclient.ClientSet) (bool,
 }
 
 func stateStable(state sriovv1.SriovNetworkNodeState) (bool, error) {
-	switch state.Status.SyncStatus {
-	case "Succeeded":
-		return true, nil
-	// When the config daemon is restarted the status will be empty
-	// This doesn't mean the config was applied
-	case "":
+	readyCond := meta.FindStatusCondition(state.Status.Conditions, sriovv1.ConditionReady)
```
adrianchiris, line 301: shouldnt ready condition be set to true only if Progressing condition is set and is false and draining condition is set and is false ?

if so maybe we should just check Ready condition here ? WDYT?

doc/api/node-state-api.md
```diff
@@ -145,6 +145,47 @@ status:
...
+|--------|--------|-------------|
+| `DrainNotNeeded` | False | No drain operation is in progress or required |
+| `DrainPending` | True | Drain is waiting for an available slot before it can start |
+| `DrainingNode` | True | Node drain is actively in progress |
+| `DrainFailed` | True | Drain is in progress but encountering errors (e.g., PDB violations) |
+| `DrainCompleted` | False | Drain operation completed successfully |
+
+The `Draining` condition is owned by the drain controller via Server-Side Apply, while `Ready` and `Progressing` are owned by the config daemon. This ownership split prevents conflicts when both components update status concurrently.
+
+Stability is defined as: `Ready=True`, `Progressing!=True`, `Draining!=True`.
```
adrianchiris, line 188: nit: since there is only True and false maybe say:
Ready=true, Progressing=False, Draining=False ?

in any case for stable state we expect all 3 conditions to be present right ?

doc/design/kubernetes-conditions-integration.md
```diff
@@ -397,8 +375,8 @@ worker-pool   5         5             True    False         False      15m
 
 # Check node state with all conditions
 kubectl get sriovnetworknodestate -o wide
-NAME      SYNC STATUS   READY   PROGRESSING   DEGRADED   DRAINPROGRESS   DRAINDEGRADED   DRAINCOMPLETE   AGE
-worker1   Succeeded     True    False         False      False           False           True            1h
+NAME      SYNC STATUS   READY   PROGRESSING   DRAINING   AGE
```
adrianchiris, line 378: arent we missing some columns ? e.g the current and desired sync state ?

pkg/status/patcher.go
```diff
@@ -109,6 +119,109 @@ func (p *Patcher) ApplyCondition(ctx context.Context, obj client.Object, conditi
...
 }
 
+// ApplyStatus applies non-condition status fields and owned conditions in a single
+// Server-Side Apply call. The field manager claims ownership of exactly the keys
+// present in statusFields plus the supplied condition entries. Conditions owned by
+// other field managers are not included in the payload, so their ownership is
+// preserved.
+func (p *Patcher) ApplyStatus(ctx context.Context, obj client.Object, statusFields map[string]interface{}, ownedConditions []metav1.Condition) error {
+	oldConditions := getConditionsFromObject(obj)
+
```
adrianchiris, line 129: should we ensure that statusFields["conditions"] is not set ?

pkg/drain/drainer.go
```diff
@@ -169,9 +167,21 @@ func createDrainHelper(kubeClient kubernetes.Interface, ctx context.Context, ful
...
-		ErrOut: writer{func(msg string, kv ...interface{}) { logger.Error(nil, strings.ReplaceAll(msg, "\n", ""), kv...) }},
+		Ctx: ctx,
+		Out: writer{func(msg string, kv ...interface{}) { logger.Info(strings.ReplaceAll(msg, "\n", "")) }},
+		// ErrOut captures errors from the drain library, including retry errors.
+		// We call the onError callback here to report eviction failures immediately,
+		// before the drain timeout is reached. Informational warnings (e.g. about
+		// ignored DaemonSet pods) are logged but not reported as errors.
+		ErrOut: writer{func(msg string, kv ...interface{}) {
+			cleanMsg := strings.ReplaceAll(msg, "\n", "")
+			if strings.HasPrefix(cleanMsg, "WARNING") {
```
adrianchiris, line 178: does drain helper pass msg with WARNING ? i looked through https://github.com/kubernetes/kubectl/blob/master/pkg/drain/drain.go

didnt find any strings written to the writer with WARNING prefix

pkg/drain/drainer.go
```diff
@@ -77,38 +89,23 @@ func (d *Drainer) DrainNode(ctx context.Context, node *corev1.Node, fullNodeDrai
...
-		}
-		lastErr = err
-		reqLogger.Info("drainNode(): Draining failed, retrying", "error", err)
-		return false, nil
-	}); err != nil {
-		if wait.Interrupted(err) {
-			reqLogger.Info("drainNode(): failed to drain node", "steps", backoff.Steps, "error", lastErr)
-		}
-		reqLogger.Info("drainNode(): failed to drain node", "error", err)
+
```
adrianchiris, line 95: good call to simplify here as the controller will retry.

pkg/drain/drainer.go
adrianchiris, line 131: we dont want to have some callback be called here on failure like in DrainNodes (BeforeDrainNode) ?

controllers/drain_controller_helper.go
```diff
@@ -303,3 +365,35 @@ func (dr *DrainReconcile) findNodePoolConfig(ctx context.Context, node *corev1.N
 		return defaultPoolConfig, defaultNodeLists, nil
 	}
 }
+
+// updateDrainConditions updates the drain-related conditions on the SriovNetworkNodeState
+func (dr *DrainReconcile) updateDrainConditions(ctx context.Context, nodeNetworkState *sriovnetworkv1.SriovNetworkNodeState, state sriovnetworkv1.DrainState, errorMessage string) error {
+	reqLogger := ctx.Value(constants.LoggerContextKey).(logr.Logger).WithName("updateDrainConditions")
+
+	// Get the latest version of the nodeNetworkState
+	latestState := &sriovnetworkv1.SriovNetworkNodeState{}
```
adrianchiris, line 374: why do we need to do a get call here instead of using the passed object ?

controllers/drain_controller_helper.go
```diff
@@ -53,6 +55,12 @@ func (dr *DrainReconcile) handleNodeIdleNodeStateDrainingOrCompleted(ctx context
 		return ctrl.Result{}, err
 	}
 
+	// Update drain conditions to idle state
+	if err := dr.updateDrainConditions(ctx, nodeNetworkState, sriovnetworkv1.DrainStateIdle, ""); err != nil {
```
adrianchiris, line 59: just to make sure i understand. if this update fails then we will requeue, and if the drain state annotation is already idle, we will update conditions part of the drain controller main reconcile loop right ?

adrianchiris (commented):
wasnt able to go through all of it unfortunately. pretty big PR. so this one is a partial review.

test/util/cluster/cluster.go
SchSeba: Ready and Draining are independently owned, so all three checks are required here we can miss active drains after successful sync

controllers/drain_controller_helper.go
SchSeba: yep

controllers/drain_controller_helper.go
SchSeba: yes we want to get the latest object before we do the SSA, just as a best practice.

pkg/drain/drainer.go
SchSeba: I add a comment `full drain is not important here, onError callback not needed for uncordon` we don't really need it here as we don't have a running routine

SchSeba (commented):
Hi @zeeke @rollandf when you have time please take a look on this one

controllers/drain_controller_helper.go
```diff
@@ -89,6 +99,19 @@ func (dr *DrainReconcile) handleNodeDrainOrReboot(ctx context.Context,
...
 
+	drainingCondition := meta.FindStatusCondition(nodeNetworkState.Status.Conditions, sriovnetworkv1.ConditionDraining)
+	if nodeStateDrainAnnotationCurrent == constants.Draining &&
+		drainingCondition != nil &&
+		drainingCondition.Status == metav1.ConditionFalse &&
+		drainingCondition.Reason == sriovnetworkv1.ReasonDrainCompleted {
+		reqLogger.Info("drain already completed, reconciling stale node state annotation")
+		if err := utils.AnnotateObject(ctx, nodeNetworkState, constants.NodeStateDrainAnnotationCurrent, constants.DrainComplete, dr.Client); err != nil {
+			reqLogger.Error(err, "failed to annotate node with annotation", "annotation", constants.DrainComplete)
+			return ctrl.Result{}, err
```
zeeke, line 110: I expected the status conditions to only report the status of the cluster. Here, based on a condition's value, we update an annotation used for the draining logic.

Am I missing soemthing about this case?

controllers/drain_controller_helper.go
```diff
@@ -105,15 +128,43 @@ func (dr *DrainReconcile) handleNodeDrainOrReboot(ctx context.Context,
 		}
 	}
 
+	// Keep the latest unique drain errors (e.g. PDB violations) so the condition
+	// message names the pods blocking the drain without flooding the API server.
+	seenDrainErrors := make(map[string]struct{})
+	var drainErrors []string
+	onDrainError := func(drainErr error) {
```
zeeke, line 135: can we put this callback in a mutex lock? 
```suggestion
	var drainErrorsMu sync.Mutex
	seenDrainErrors := make(map[string]struct{})
	var drainErrors []string
	onDrainError := func(drainErr error) {
		drainErrorsMu.Lock()
		defer drainErrorsMu.Unlock()
```

the ErrOut field draing.Helper is called in a loop of goroutines
https://github.com/kubernetes/kubectl/blob/master/pkg/drain/drain.go#L328C18-L328C24

pkg/daemon/status.go
```diff
@@ -5,59 +5,92 @@ import (
...
+	}
 
-		funcLog.V(2).Info("update nodeState status",
-			"CurrentSyncStatus", currentNodeState.Status.SyncStatus,
-			"CurrentLastSyncError", currentNodeState.Status.LastSyncError,
-			"NewSyncStatus", desiredNodeState.Status.SyncStatus,
-			"NewFailedMessage", desiredNodeState.Status.LastSyncError)
+	statusFieldsChanged := !currentNodeState.Status.StatusFieldsEqual(&desiredNodeState.Status)
 
-		err := dn.client.Status().Patch(ctx, desiredNodeState, client.MergeFrom(currentNodeState))
```
zeeke, line 42: The `Patch` command was updating the desiredNodeState object. new code does not update it and I think it's a regression.

This AI generated test proves it
https://github.com/zeeke/sriov-network-operator-1/pull/1/changes/6ff560d5569f9eb417afc012252f95f1393a0577

controllers/drain_controller_helper.go
SchSeba: You're not missing something — this is crash recovery, not using the condition as the drain-slot source of truth.

We persist `DrainCompleted` on the condition *before* annotating `current-state=DrainComplete`. If the annotate fails, the next reconcile still sees the annotation as `Draining`. Without this check we would call `DrainNode` again on a node that already finished draining.

Drain-slot counting still uses the annotations. The condition is only used here to finish that incomplete transition.

SchSeba:
Thanks @zeeke!
can you make another round of review please

pkg/daemon/status.go
```diff
@@ -5,59 +5,114 @@ import (
...
 
+// syncDesiredObjectMeta copies live ObjectMeta onto desiredNodeState.
+// Server-Side Apply does not mutate desiredNodeState the way Status().Patch did,
+// so callers rely on this refresh to observe generation and annotation changes.
+func syncDesiredObjectMeta(desiredNodeState, live *sriovnetworkv1.SriovNetworkNodeState) {
+	desiredNodeState.ObjectMeta = live.ObjectMeta
+}
+
+// configurationOwnedConditions returns the condition entries owned by the config daemon.
+func configurationOwnedConditions(conditions []metav1.Condition) []metav1.Condition {
```
zeeke, line 102: can you rename to `getConfigDaemonConditions` or `getConditionsOwnedByConfigDaemon` ?

zeeke (approved):
only one minor comment from my side

pkg/daemon/status.go
SchSeba: is `configurationOwnedConditions` good as we are in the daemon package already?

zeeke: ok, let's infer from the context

zeeke (approved):

Issue #1015, opened 2026-01-13
ConnectX-6 Lx fails to switch to switchdev mode when NumVfs > 0 (bad resource state 0x9, syndrome 0x98afbb)


ashokpariya0:
```
# lspci -nn
0101:00:00.0 Ethernet controller [0200]: Mellanox Technologies MT2894 Family [ConnectX-6 Lx] [15b3:101f]
0102:00:00.1 Ethernet controller [0200]: Mellanox Technologies MT2894 Family [ConnectX-6 Lx] [15b3:101f]
```

below error is seen in sriov-config-daemon
```
2026-01-06T10:20:44.462525754Z	LEVEL(-2)	sriov/sriov.go:1190	SetNicSriovMode()	{"device": "0101:00:00.0", "mode": "switchdev"}
2026-01-06T10:20:47.174119372Z	ERROR	sriov/sriov.go:632	configSriovPFDevice(): fail to set NumVfs for device	{"device": "0101:00:00.0", "error": "can't set eSwitch mode to [switchdev] on device [0101:00:00.0]: invalid argument"}
2026-01-06T10:20:47.174441773Z	ERROR	sriov/sriov.go:681	configSriovInterfaces(): fail to configure sriov interface. resetting interface.	{"address": "0101:00:00.0", "error": "can't set eSwitch mode to [switchdev] on device [0101:00:00.0]: invalid argument"}
```

Details error is below:

```
2026-01-06T10:20:39.630110212Z	INFO	sriovnetwork	sriov/sriov.go:848	NeedToUpdateSriov(): EswitchMode needs update	{"desired": "switchdev", "current": "legacy"}
2026-01-06T10:20:39.630126944Z	LEVEL(-2)	sriov/sriov.go:681	configSriovInterfaces(): start sriov configuration
2026-01-06T10:20:39.630133312Z	LEVEL(-2)	sriov/sriov.go:812	configSriovDevice(): configure sriov device	{"device": "0101:00:00.0", "config": {"pciAddress":"0101:00:00.0","numVfs":8,"name":"ens2048f0np0","eSwitchMode":"switchdev","vfGroups":[{"resourceName":"swtichdevens2048f0np0","deviceType":"netdevice","vfRange":"0-7","policyName":"test-switchdev-policy-h65rl"}]}, "skipVFConfiguration": false}
[...]

PR #1017, opened 2026-01-14
mlx5: switch to switchdev before setting sriov_numvfs

ashokpariya0:
mlx5-based Network Express Card firmware(MT2894 Family [ConnectX-6 Lx]) requires eSwitch mode to be switched to switchdev before SR-IOV VFs are created. This change reorders the mlx5-specific SR-IOV configuration sequence to switch to switchdev prior to setting sriov_numvfs, enabling successful VF provisioning on affected hardware.

This PR fixes: https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/1015

**Problem**:
The current mlx5-specific SR-IOV configuration sequence performs:
legacy → unbind VFs → set sriov_numvfs → switchdev
On some mlx5 devices and firmware combinations, SR-IOV VFs can only be
successfully created after the eSwitch is already in switchdev mode

**Solution**
Reorder the mlx5 configuration flow to
legacy → unbind VFs → switchdev → set sriov_numvfs
This ensures the eSwitch mode is correctly set before VF creation,
allowing successful provisioning on affected hardware

Fix verified with MT2894 Family [ConnectX-6 Lx] :
```
lspci -nn
0101:00:00.0 Ethernet controller [0200]: Mellanox Technologies MT2894 Family [ConnectX-6 Lx] [15b3:101f]
0102:00:00.1 Ethernet controller [0200]: Mellanox Technologies MT2894 Family [ConnectX-6 Lx] [15b3:101f]
```

PR #1019, opened 2026-01-20
Fix VF creation race by waiting for VF symlinks before switchdev

ashokpariya0:
Attempting to unbind VFs or switch the PF to switchdev mode before all VFs are ready causes race conditions, leading to mode switch failures with "invalid argument" errors.

**Why we need this change?**

This PR fixed issue: https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/1015

for func- setEswitchModeAndNumVFsMlx(https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/11b55058a43fdbfa08e02c0df5604c64fd27e1c3/pkg/host/internal/sriov/sriov.go#L1162), currently we do following,
// setEswitchModeAndNumVFsMlx configures PF eSwitch and sriov_numvfs in the following order:
// a. set eSwitchMode to legacy
// b. set the desired number of Virtual Functions
// c. unbind driver of all VFs
// d. set eSwitchMode to `switchdev` if requested

on s390 arch, with following mellanox device
```
# lspci -nn
0101:00:00.0 Ethernet controller [0200]: Mellanox Technologies MT2894 Family [ConnectX-6 Lx] [15b3:101f]
0102:00:00.1 Ethernet controller [0200]: Mellanox Technologies MT2894 Family [ConnectX-6 Lx] [15b3:101f]
```

VF creation is asynchronous, and SetSriovNumVfs() does not block until all VF sysfs entries are fully created. As a result, unbinding VFs or switching the PF to switchdev mode too early can lead to race conditions.

In PR https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1017
, an alternative sequence was proposed, changing the order from:
```
legacy → unbind VFs → set sriov_numvfs → switchdev
```

to:
```
legacy → unbind VFs → switchdev → set sriov_numvfs
```

In this PR, instead of reordering the operations, we explicitly wait until all expected VF sysfs symlinks are present after setting sriov_numvfs. Only once VF creation is fully completed do we proceed with unbinding VFs and switching the PF to switchdev mode, eliminating the race condition on s390x.

### DIFF
--- a/pkg/host/internal/sriov/sriov.go
+++ b/pkg/host/internal/sriov/sriov.go
@@ -1166,44 +1166,57 @@ func (s *sriov) waitForVFLinks(pciAddr string, expectedNum int, maxTimeout time.
 		"expected", expectedNum,
 		"maxTimeout", maxTimeout)
 
-	start := time.Now()
-	sleepDuration := 500 * time.Millisecond
+	err := wait.PollUntilContextTimeout(
+		context.Background(),
+		time.Second, // poll interval
+		maxTimeout,  // overall timeout
+		true,        // run immediately
+		func(ctx context.Context) (bool, error) {
+			vfAddrs, err := s.dputilsLib.GetVFList(pciAddr)
+			if err != nil {
+				log.Log.V(2).Info("waitForVFLinks(): GetVFList failed, retrying", "err", err)
+				return false, nil
+			}
 
-	for time.Since(start) < maxTimeout {
-		vfAddrs, err := s.dputilsLib.GetVFList(pciAddr)
-		if err != nil {
-			log.Log.V(2).Info("waitForVFLinks(): GetVFList failed, retrying", "err", err)
-			time.Sleep(sleepDuration)
-			continue
-		}
+			if len(vfAddrs) < expectedNum {
+				return false, nil
+			}
 
-		current := len(vfAddrs)
-		if current >= expectedNum {
 			// Check all have physfn symlink
-			allReady := true
 			for _, vfAddr := range vfAddrs {
-				linkPath := filepath.Join(vars.FilesystemRoot, consts.SysBusPciDevices, vfAddr, "physfn")
+				linkPath := filepath.Join(
+					vars.FilesystemRoot,
+					consts.SysBusPciDevices,
+					vfAddr,
+					"physfn",
+				)
+
 				if _, err := os.Lstat(linkPath); err != nil {
-					log.Log.V(2).Info("waitForVFLinks(): physfn symlink missing", "vf", vfAddr, "err", err)
-					allReady = false
-					break
+					log.Log.V(2).Info(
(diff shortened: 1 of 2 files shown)


pkg/host/internal/sriov/sriov.go
```diff
@@ -1154,6 +1154,58 @@ func (s *sriov) setEswitchModeAndNumVFs(pciAddr string, desiredEswitchMode strin
...
+
+	start := time.Now()
+	sleepDuration := 500 * time.Millisecond
+
+	for time.Since(start) < maxTimeout {
+		vfAddrs, err := s.dputilsLib.GetVFList(pciAddr)
+		if err != nil {
+			log.Log.V(2).Info("waitForVFLinks(): GetVFList failed, retrying", "err", err)
+			time.Sleep(sleepDuration)
+			continue
```
zeeke: what about returning the error here?
do we know any transient error from reading the symlink from the directory?

ashokpariya0: looking at code- https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/master/pkg/utils/utils.go#L168-L197

it mainly interacts with sysfs (/sys/bus/pci/.../virtfn*) using Lstat, Glob, and EvalSymlinks.
so we don't want to return error immediately bacause temporary sysfs inconsistencies
in current context, as GetVFList is retryable and  we relies on the overall timeout to detect real failures

pkg/host/internal/sriov/sriov.go
```diff
@@ -1154,6 +1154,58 @@ func (s *sriov) setEswitchModeAndNumVFs(pciAddr string, desiredEswitchMode strin
 	return fn(pciAddr, desiredEswitchMode, numVFs)
 }
 
+func (s *sriov) waitForVFLinks(pciAddr string, expectedNum int, maxTimeout time.Duration) error {
```
ykulazhenkov, line 1157: Can we use "wait helpers" from `k8s.io/apimachinery/pkg/util/wait` as in https://github.com/ashokpariya0/sriov-network-operator/blob/d3a17f34f30e4c026e278b36900a6a9363d4cd71/pkg/host/internal/sriov/sriov.go#L173? 

I'm not sure if we need a backoff here, so it is up to you to use `wait.PollUntilContextTimeout` (fixed interval retries + timeout) or `wait.ExponentialBackoffWithContext`.

ykulazhenkov (approved):
LGTM

pkg/host/internal/sriov/sriov.go
ashokpariya0: Done, [change](https://github.com/k8snetworkplumbingwg/sriov-network-operator/compare/d3a17f34f30e4c026e278b36900a6a9363d4cd71..e9aeeb1c3bb47be0e8e2e1b19be148b1ac2f44bf)

PR #1020, opened 2026-01-20
Decrease log level if device link speed can't be evaluated

e0ne:


pkg/host/internal/network/network.go
```diff
@@ -241,11 +241,12 @@ func (n *network) GetNetDevNodeGUID(pciAddr string) string {
 }
 
 func (n *network) GetNetDevLinkSpeed(ifaceName string) string {
-	log.Log.V(2).Info("GetNetDevLinkSpeed(): get LinkSpeed", "device", ifaceName)
+	funcLog := log.Log.WithValues("device", ifaceName)
+	funcLog.V(2).Info("GetNetDevLinkSpeed(): get LinkSpeed", "device", ifaceName)
```
zeeke: `"device", ifaceName` is redundant, right? we can remove it

Also, I'd completely remove the logging activity here and leave the WARNING below. 
WDYT?

e0ne: makes sense. done

PR #1021, opened 2026-01-31
Fix "message too long" error for InfiniBand devices with many VFs

SchSeba:
When configuring InfiniBand devices with many VFs, LinkByName() can fail with "message too long" because the kernel's netlink response exceeds the ~16KB message limit. This happens because LinkByName requests all VF info (including GUIDs) via RTEXT_FILTER_VF, which generates very large messages for IB devices.

This commit adds a fallback mechanism:
- Add LinkByNameForSetVf() method that reads minimal link info (name and index) directly from sysfs instead of using netlink
- Update configSriovVFDevices() to catch "message too long" errors and fall back to LinkByNameForSetVf()
- Add unit test covering the fallback scenario

The fallback only triggers when the specific "message too long" error occurs, preserving normal behavior for Ethernet devices and IB devices with fewer VFs.

### DIFF
--- a/pkg/host/internal/lib/netlink/mock/mock_netlink.go
+++ b/pkg/host/internal/lib/netlink/mock/mock_netlink.go
@@ -196,6 +196,21 @@ func (mr *MockNetlinkLibMockRecorder) LinkByName(name any) *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "LinkByName", reflect.TypeOf((*MockNetlinkLib)(nil).LinkByName), name)
 }
 
+// LinkByNameForSetVf mocks base method.
+func (m *MockNetlinkLib) LinkByNameForSetVf(name string) (netlink.Link, error) {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "LinkByNameForSetVf", name)
+	ret0, _ := ret[0].(netlink.Link)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
+// LinkByNameForSetVf indicates an expected call of LinkByNameForSetVf.
+func (mr *MockNetlinkLibMockRecorder) LinkByNameForSetVf(name any) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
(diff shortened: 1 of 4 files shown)


SchSeba:
closing this as the issue is in the kernel.

more info:

DESCRIPTION:
When querying InfiniBand devices with SR-IOV VFs enabled, the kernel triggers
a WARNING in rtnl_getlink() and returns EMSGSIZE (Message too long).

AFFECTED KERNEL: 5.14.0-570.76.1.el9_6.x86_64
WORKING KERNEL: 5.14.0-570.73.1.el9_6.x86_64

ROOT CAUSE:
Commit 30aad41721e0 ("net/core: Add support for getting VF GUIDs") added
VF GUID retrieval to netlink messages but did not update rtnl_vfinfo_size()
to allocate buffer space for this data.

UPSTREAM FIX:
Commit bb7bdf636cef ("rtnetlink: Allocate vfinfo size for VF GUIDs when supported")
by Mark Bloch (NVIDIA), merged April 2025.

STACK TRACE FROM AFFECTED SYSTEM:
WARNING: CPU: 90 PID: 92077 at net/core/rtnetlink.c:3903 rtnl_getlink+0x3df/0x3f0
Kernel: 5.14.0-570.76.1.el9_6.x86_64
Hardware: Dell Inc. PowerEdge R760xa
Call Trace:
rtnl_getlink+0x3df/0x3f0
rtnetlink_rcv_msg+0x159/0x410
netlink_rcv_skb+0x54/0x100
netlink_unicast+0x245/0x370
netlink_sendmsg+0x206/0x440
__sys_sendto+0x1dc/0x1f0

REPRODUCTION:

Enable SR-IOV VFs on InfiniBand device:
echo 16 > /sys/bus/pci/devices/0000:0d:00.0/sriov_numvfs
Query interface:
ip link show dev ibp13s0
Observe: "RTNETLINK answers: Message too long"
IMPACT:

SR-IOV Network Operator cannot discover InfiniBand PF devices
ib-sriov-cni cannot configure VF GUIDs for pods
Kubernetes pod networking fails for InfiniBand workloads

PR #1025, opened 2026-02-05
Fix netlink buffer size for InfiniBand devices with many VFs

bbenshab:
# Fix netlink buffer size for InfiniBand devices with many VFs

## Problem

When configuring SR-IOV VFs on InfiniBand devices, the operator encounters "message too long" netlink errors when the device has 16 or more VFs. This causes VF configuration to fail and triggers automatic rollback, preventing successful SR-IOV setup.

### Error Logs (Before Fix)

```
ERROR	baremetal/baremetal.go:56	DiscoverSriovDevices(): unable to get Link for device, skipping	{"device": "0000:29:00.0", "error": "message too long"}
ERROR	baremetal/baremetal.go:56	DiscoverSriovDevices(): unable to get Link for device, skipping	{"device": "0000:40:00.0", "error": "message too long"}
ERROR	baremetal/baremetal.go:56	DiscoverSriovDevices(): unable to get Link for device, skipping	{"device": "0000:4f:00.0", "error": "message too long"}
```

### Root Cause

The vishvananda/netlink library uses a default netlink receive buffer size of approximately 32KB. InfiniBand devices with many VFs return link information that exceeds this buffer size, resulting in ENOMSG ("message too long") errors from the kernel.

This issue is tracked in: https://issues.redhat.com/browse/OCPBUGS-74637

## Solution

This PR adds a new method `LinkByNameWithLargeBuffer()` to the netlink wrapper interface that creates a custom netlink handle with a 32MB receive buffer. This buffer size is sufficient to handle InfiniBand devices with 16+ VFs.

The fix is applied automatically for Mellanox devices (vendor ID 15b3) during the discovery phase in `DiscoverSriovDevices()`, as Mellanox devices are commonly InfiniBand which need larger buffers with many VFs.

### Changes

1. **pkg/host/internal/lib/netlink/netlink.go**
   - Added `LinkByNameWithLargeBuffer(name string) (Link, error)` to `NetlinkLib` interface
   - Implemented method using `handle.SetSocketReceiveBufferSize(32MB, true)` to force override kernel limits
   - Added reference to OCPBUGS-74637 in code comments

2. **pkg/host/internal/sriov/sriov.go**
   - Updated `DiscoverSriovDevices()` to use large buffer for Mellanox devices (vendor ID 15b3)
   - Automatically applied during device discovery phase for all InfiniBand hardware

## Testing

Successfully tested on production OpenShift 4.17 cluster with 11 InfiniBand devices across multiple hardware types.

### Test Environment
[...]

PR #1026, opened 2026-02-07
Use ip commands to bypass netlink PAGE_SIZE limit for InfiniBand devices

bbenshab:
### Problem

When configuring InfiniBand devices with 16 VFs, the SR-IOV Network Operator fails with "message too long" errors:

```
ERROR configSriovVFDevices(): unable to get PF link for device: message too long
```

### Root Cause

This is caused by a **kernel regression** (see [OCPBUGS-74889](https://issues.redhat.com/browse/OCPBUGS-74889) and [RHEL-145522](https://issues.redhat.com/browse/RHEL-145522)) where the Linux kernel limits individual netlink messages to PAGE_SIZE (~4KB on x86_64).

InfiniBand devices with 16 VFs return metadata (MAC, VLAN, Node GUID, Port GUID, RDMA stats per VF) that exceeds this 4KB limit. The kernel enforces this limit at the message level, making it impossible to query link information via netlink for IB devices with many VFs.

**Note**: While this is a kernel regression that should ultimately be fixed in the kernel, we cannot wait for a kernel fix as **we need to support InfiniBand devices now**. This workaround provides a working solution that is isolated to IB device handling.

### Solution

Skip netlink queries entirely for InfiniBand devices and use direct `ip link set` commands to configure VF GUIDs:

1. **configSriovVFDevices()**: Skip `LinkByName()` for IB devices, set `pfLink=nil`
2. **VF GUID configuration**: Call `configureVfGUIDWithIPCommand()` when `pfLink=nil`
3. **New function**: `configureVfGUIDWithIPCommand()` generates random GUIDs and uses `ip link set` commands directly
4. **Set PF link up**: Use sysfs + `ip link set` for IB devices, netlink for non-IB devices

**Why this works:**
- `ip link set` commands use netlink SETLINK operations that don't query VF data
- Only configuration operations are performed, no large responses returned
- Reading link state from sysfs avoids netlink queries
- Kernel accepts configuration commands without hitting PAGE_SIZE limit

### Changes

```diff
+ Added imports: os/exec, crypto/rand
+ Modified configSriovVFDevices(): Skip netlink for IB, set pfLink=nil
+ Modified VF GUID config: Use ip commands when pfLink=nil  
+ New function configureVfGUIDWithIPCommand(): Direct ip command execution
+ Modified "Set PF link up": Use sysfs/ip for IB, netlink for others
```

### Testing

**Environment:**
- Hardware: NVIDIA ConnectX-7 InfiniBand adapters
- Configuration: 12 NICs per node, 16 VFs per NIC
- Platform: OpenShift 4.20, RHCOS 9.6
- Kernel: 5.14.0-570.73.1.el9_6.x86_64

### References

- **Kernel Regression**: [OCPBUGS-74889](https://issues.redhat.com/browse/OCPBUGS-74889)
- **RHEL Bug**: [RHEL-145522](https://issues.redhat.com/browse/RHEL-145522)
- **Previous PR** (buffer approach): [#1025](https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1025) - closed, wrong approach
- **Detailed explanation**: [PR #1021 comment](https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1021#issuecomment-3829309941)
- **Implementation**: [bbenshab/sriov-network-operator@86140528](https://github.com/bbenshab/sriov-network-operator/commit/86140528)

### Note on Kernel Regression

While this issue originates from a kernel regression, we cannot wait for a kernel fix because:
1. InfiniBand support is required now for production deployments
2. Kernel fixes take months to propagate through releases
3. This workaround is clean, isolated, and maintainable
4. It will continue to work even after kernel is fixed

This PR provides a working solution that unblocks InfiniBand deployments while the kernel team addresses the underlying regression.

### DIFF
--- a/pkg/host/internal/infiniband/infiniband.go
+++ b/pkg/host/internal/infiniband/infiniband.go
@@ -1,6 +1,7 @@
 package infiniband
 
 import (
+	"crypto/rand"
 	"errors"
 	"fmt"
 	"io/fs"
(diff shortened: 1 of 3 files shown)


SchSeba:
closing this one it was a kernel regression there is no need for changes in the operator

PR #1033, opened 2026-02-16
fix: add timeout to sysfs writes to prevent daemon hang

zeeke:
Kernel drivers (e.g. i40e) can block indefinitely when writing to sriov_numvfs if the device is in a bad state. For example, the following error has been hit on a `Intel XXV710` NIC:

```
Feb 16 13:53:01 worker0 kernel: 06c73374b594186: left promiscuous mode
Feb 16 13:53:01 worker0 kernel: i40e 0000:3b:00.0: Setting MAC 5e:28:32:f0:80:20 on VF 1
Feb 16 13:53:02 worker0 kernel: i40e 0000:3b:00.0: Bring down and up the VF interface to make this change effective.
Feb 16 13:53:02 worker0 kernel: i40e 0000:3b:00.0: Unable to configure VFs, other operation is pending.
Feb 16 13:53:02 worker0 kernel: i40e 0000:3b:00.0: Unable to configure VFs, other operation is pending.
Feb 16 13:53:02 worker0 kernel: 152a5b6a3b44739: left promiscuous mode
```

Replace direct `os.WriteFile` calls in SetSriovNumVfs with a new `WriteFileWithTimeout` utility that runs the write in a goroutine and returns a timeout error after 2 minutes.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -128,3 +128,23 @@ func GetChrootExtension() string {
 	}
 	return fmt.Sprintf("chroot %s%s", vars.FilesystemRoot, consts.Host)
 }
+
+// WriteFileWithTimeout writes data to a file with a timeout.
+// This is useful for writing to sysfs files where the kernel driver may block
+// indefinitely if it is in a bad state.
+// Note: if the timeout expires, the write goroutine will remain blocked in the
+// kernel; it cannot be canceled but will be cleaned up when the process exits.
+func WriteFileWithTimeout(path string, data []byte, perm os.FileMode, timeout time.Duration) error {
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+	ch := make(chan error, 1)
+	go func() {
+		ch <- os.WriteFile(path, data, perm)
+	}()
+	select {
+	case err := <-ch:
+		return err
+	case <-timer.C:
+		return fmt.Errorf("timeout writing to file %s after %v", path, timeout)
+	}
+}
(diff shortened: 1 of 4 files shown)


SchSeba:
Hi @zeeke can you check the tests please

zeeke:
Unit test flake 
```
[FAILED] Timed out after 1.001s.
  The function passed to Eventually failed at /home/runner/work/sriov-network-operator/sriov-network-operator/controllers/sriovoperatorconfig_controller_test.go:132 with:
  Expected
      <[]string | len:0, cap:0>: nil
  not to be empty
  In [BeforeEach] at: /home/runner/work/sriov-network-operator/sriov-network-operator/controllers/sriovoperatorconfig_controller_test.go:142 @ 02/25/26 17:23:03.536
```

is fixed in
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1044

pkg/utils/utils.go
```diff
@@ -128,3 +128,23 @@ func GetChrootExtension() string {
 	}
 	return fmt.Sprintf("chroot %s%s", vars.FilesystemRoot, consts.Host)
 }
+
+// WriteFileWithTimeout writes data to a file with a timeout.
+// This is useful for writing to sysfs files where the kernel driver may block
+// indefinitely if it is in a bad state.
+// Note: if the timeout expires, the write goroutine will remain blocked in the
+// kernel; it cannot be canceled but will be cleaned up when the process exits.
+func WriteFileWithTimeout(path string, data []byte, perm os.FileMode, timeout time.Duration) error {
```
SchSeba, line 137: can we put this one under the utils interface this way we can also mock it for testing in the sriov package?

zeeke: If we don't have a real need for it, I'd avoid putting it behind an interface. 
It can be called (and hence tested) in the unit test, as it was the `os.WriteFile(...)` function.

SchSeba: 👍

PR #1040, opened 2026-02-19
e2e: Retry cluster check

zeeke:
Wrap the cluster stability check with an Eventually block, so the test can pass in case of temporary errors (e.g. image pull timeouts).

### DIFF
--- a/test/validation/tests/test_validation.go
+++ b/test/validation/tests/test_validation.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


test/validation/tests/test_validation.go
```diff
@@ -115,17 +115,15 @@ var _ = Describe("validation", func() {
...
-			CheckStable()
+			Eventually(func(g Gomega) {
+				res, err := cluster.SriovStable(operatorNamespace, clients)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(res).To(BeTrue(), "SR-IOV status is not stable")
+
+				isClusterReady, err := cluster.IsClusterStable(clients)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(isClusterReady).To(BeTrue(), "Cluster is not stable")
+			}).Should(Succeed())
```
SchSeba: so we want to have timeout here or we just use the default?

zeeke: definitely a good timeout. fixing
thanks!

e0ne (approved):
LGTM

PR #1044, opened 2026-02-24
add kmod for mstfwreset

SchSeba:
mstfwreset needs kmod to be installed on the system to run.

Why your command was immune before v4.32.0
Tracing your specific flow (--skip_driver, -l 3 / PCI_RESET, NIC device, x86 Linux):

MlnxDriverFactory().getDriverObj() -- with --skip_driver, the skip=True flag bypasses all driver detection. The MlnxDriverLinux class uses sysfs (/sys/bus/pci/drivers), not lsmod, so this was never a problem.
mstRestart() -- uses lsmod internally, but is gated by not IS_MSTFLINT, which is False when running as mstfwreset.py. So this path was never reached from the mstflint project.
resetPciAddrPPC() -- uses isModuleLoaded("mst_ppc_pci_reset"), but only runs on PPC64 architecture. Not hit on x86.
reset_flow_switch() -- uses isModuleLoaded("sx_core"), but only for switch devices, not NIC devices.

GPU driver safety check -- uses lsmod | grep -E, but only triggered when reset_sync=FW AND method=HOT_RESET, which are not the defaults for your command.
None of those paths were hit by your command. The only lsmod call you hit is the one added in v4.32.0 in reset_flow_host(), which runs unconditionally for all Linux reset/query operations.
Summary

Version
lsmod required for your command?
<= v4.31.x No
v4.32.0 (April 2025) Yes -- checks for mst_pciconf
v4.35.0 (Jan 2026) Yes -- checks for mstflint_access

The purpose of this check is to pass mst_driver_is_loaded to CmdRegMroq, which uses it to determine reset sync/method capabilities. The check itself is informational (it doesn't block execution if the module isn't loaded), but the lsmod command being unavailable is what causes the hard failure.

### DIFF
--- a/controllers/sriovoperatorconfig_controller_test.go
+++ b/controllers/sriovoperatorconfig_controller_test.go
@@ -139,7 +139,7 @@ var _ = Describe("SriovOperatorConfig controller", Ordered, func() {
 				}
 				err = k8sClient.Update(ctx, config)
 				Expect(err).NotTo(HaveOccurred())
-			}).Should(Succeed())
+			}, util.APITimeout, util.RetryInterval).Should(Succeed())
 
 		})
 


pkg/vendors/mellanox/mellanox.go
```diff
@@ -151,18 +151,20 @@ func (m *mellanoxHelper) MlxResetFW(pciAddresses []string, mellanoxNicsStatus ma
 	log.Log.Info("mellanox-plugin resetFW()")
 	var errs []error
 	for _, pciAddress := range pciAddresses {
+		// we try to set the number of VFs to 0 before firmware reset to avoid the case where the number of VFs is not 0 after the firmware reset
+		// but in case the SRIOV_EN is false, we can't set the number of VFs to 0. not sriov_numvf file in sysfs.
```
adrianchiris: nit: not -> no

pkg/vendors/mellanox/mellanox.go
```diff
@@ -151,18 +151,20 @@ func (m *mellanoxHelper) MlxResetFW(pciAddresses []string, mellanoxNicsStatus ma
...
 		err := m.hostHelper.SetSriovNumVfs(pciAddress, 0)
 		if err != nil {
 			log.Log.Error(err, "failed to set SR-IOV number of VFs to 0 before firmware reset", "pciAddress", pciAddress)
-			return err
 		}
 
 		if IsDualPort(pciAddress, mellanoxNicsStatus) {
 			otherPortPCIAddress := getOtherPortPCIAddress(pciAddress)
+			// we try to set the number of VFs to 0 before firmware reset to avoid the case where the number of VFs is not 0 after the firmware reset
+			// but in case the SRIOV_EN is false, we can't set the number of VFs to 0. not sriov_numvf file in sysfs.
```
adrianchiris: nit: no

pkg/vendors/mellanox/mellanox.go
```diff
@@ -151,18 +151,20 @@ func (m *mellanoxHelper) MlxResetFW(pciAddresses []string, mellanoxNicsStatus ma
 	log.Log.Info("mellanox-plugin resetFW()")
 	var errs []error
 	for _, pciAddress := range pciAddresses {
+		// we try to set the number of VFs to 0 before firmware reset to avoid the case where the number of VFs is not 0 after the firmware reset
+		// but in case the SRIOV_EN is false, we can't set the number of VFs to 0. not sriov_numvf file in sysfs.
 		err := m.hostHelper.SetSriovNumVfs(pciAddress, 0)
```
adrianchiris, line 154: should we check sriov_numvfs file exist in SetSriovNumVfs ? if we set to 0 and file does not exist -> succeed ? then we should keep original logic

SchSeba: I like the idea!
done

PR #1048, opened 2026-03-05
Clean stale VF representor ports from OVS bridges during switchdev reconciliation

almaslennikov:
After a host reboot, VF representor ports can remain attached to OVS bridges when pods were not deleted before the reboot. When the config daemon recreates the VFs, the new representors reuse those names and can fail during pod creation.

This change cleans the managed bridge before switchdev VF recreation:

- Finds the bridge and PF uplink name from the managed bridge store using the PF PCI address.
- Enumerates the interfaces currently attached to that bridge and removes every VF representor matching the operator udev convention `{pfName}_{vfIndex}`, including sparse stale indices from an older VF count.
- Detaches the PF uplink as part of the same operation.
- Runs cleanup before VF recreation even when the NIC is already in legacy mode after reboot.
- Checks every configured uplink when locating a PF in a multi-uplink bridge.

The bridge cleanup APIs were renamed to describe the combined uplink and VF representor operation and now require only the PF PCI address.

Validation:

- `go test ./pkg/host/... ./pkg/helper/... -count=1` in Linux with Go 1.26.5
- `golangci-lint v2.7.2 run ./pkg/host/... ./pkg/helper/...`

## Summary by CodeRabbit

* **Enhancements**
  * Managed bridge cleanup now detaches PF uplinks and associated VF representor interfaces.
  * SR-IOV switchdev configuration cleans up interfaces before recreating VFs, including after reboot.
  * Improved handling for multiple uplinks, missing representors, and non-contiguous VF indices.

* **Bug Fixes**
  * Prevents stale interfaces from remaining attached during SR-IOV reconfiguration.
  * Preserves underlying configuration errors for clearer troubleshooting.

* **Tests**
  * Added coverage for detachment success, error propagation, multi-uplink bridges, and switchdev configuration.

### DIFF
--- a/pkg/host/internal/bridge/bridge.go
+++ b/pkg/host/internal/bridge/bridge.go
@@ -2,6 +2,7 @@ package bridge
 
 import (
 	"context"
+	"fmt"
 
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -68,14 +69,13 @@ func (b *bridge) ConfigureBridges(bridgesSpec sriovnetworkv1.Bridges, bridgesSta
 	return nil
 }
 
-// DetachInterfaceFromManagedBridge detach interface from a managed bridge,
-// this step is required before applying some configurations to PF, e.g. changing of eSwitch mode.
-// The function detach interface from managed bridges only.
-func (b *bridge) DetachInterfaceFromManagedBridge(pciAddr string) error {
-	log.Log.V(1).Info("DetachInterfaceFromManagedBridge(): detach interface", "pciAddr", pciAddr)
-	if err := b.ovs.RemoveInterfaceFromOVSBridge(context.Background(), pciAddr); err != nil {
-		log.Log.Error(err, "DetachInterfaceFromManagedBridge(): failed to detach interface from OVS bridge", "pciAddr", pciAddr)
-		return err
+// DetachUplinkAndVFRepresentorsFromManagedBridge detaches a PF uplink and all of
+// its VF representors from a managed bridge.
+func (b *bridge) DetachUplinkAndVFRepresentorsFromManagedBridge(pciAddr string) error {
+	log.Log.V(1).Info("DetachUplinkAndVFRepresentorsFromManagedBridge(): detach uplink and VF representors", "pciAddr", pciAddr)
+	if err := b.ovs.DetachUplinkAndVFRepresentorsFromOVSBridge(context.Background(), pciAddr); err != nil {
+		log.Log.Error(err, "DetachUplinkAndVFRepresentorsFromManagedBridge(): failed to detach uplink and VF representors from OVS bridge", "pciAddr", pciAddr)
+		return fmt.Errorf("failed to detach uplink and VF representors for device %s: %w", pciAddr, err)
 	}
 	return nil
 }
--- a/pkg/host/internal/bridge/ovs/ovs.go
+++ b/pkg/host/internal/bridge/ovs/ovs.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -47,8 +48,9 @@ type Interface interface {
 	GetOVSBridges(ctx context.Context) ([]sriovnetworkv1.OVSConfigExt, error)
 	// RemoveOVSBridge removes managed OVS bridge by name
 	RemoveOVSBridge(ctx context.Context, bridgeName string) error
-	// RemoveInterfaceFromOVSBridge interface from the managed OVS bridge
-	RemoveInterfaceFromOVSBridge(ctx context.Context, ifaceAddr string) error
+	// DetachUplinkAndVFRepresentorsFromOVSBridge detaches a PF uplink and all of its
+	// VF representors from the managed OVS bridge identified by the PF PCI address.
+	DetachUplinkAndVFRepresentorsFromOVSBridge(ctx context.Context, pciAddress string) error
 }
 
 // New creates new instance of the OVS interface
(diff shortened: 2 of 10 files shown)


pkg/host/internal/bridge/ovs/ovs.go
```diff
@@ -298,31 +300,34 @@ func (o *ovs) RemoveOVSBridge(ctx context.Context, bridgeName string) error {
...
 	defer cancel()
-	funcLog := log.Log.WithValues("pciAddress", pciAddress)
+	funcLog := log.Log.WithValues("pciAddress", pciAddress, "pfName", pfName, "numVfs", numVfs)
 	funcLog.V(1).Info("RemoveInterfaceFromOVSBridge(): remove interface from managed bridge")
 	knownConfigs, err := o.store.GetManagedOVSBridges()
 	if err != nil {
 		funcLog.Error(err, "RemoveInterfaceFromOVSBridge(): failed to read data from store")
 		return fmt.Errorf("failed to read data from store: %v", err)
 	}
-	var relatedBridges []*sriovnetworkv1.OVSConfigExt
```
adrianchiris, line 312: why is this change needed ? i dont think its related to what you are trying to fix is it ?

pkg/host/internal/bridge/ovs/ovs.go
```diff
@@ -346,8 +351,18 @@ func (o *ovs) RemoveInterfaceFromOVSBridge(ctx context.Context, pciAddress strin
 		return nil
 	}
 
-	funcLog.V(2).Info("RemoveInterfaceFromOVSBridge(): remove interface from the bridge")
-	if err := o.deleteInterfaceByName(ctx, dbClient, brConf.Uplinks[0].Name); err != nil {
+	// Remove stale VF representor interfaces
+	for i := 0; i < numVfs; i++ {
+		repName := fmt.Sprintf("%s_%d", pfName, i)
```
adrianchiris: note: thats how we name representors via udev in the operator.

adrianchiris: i wonder if its always safe to use the passed in numVFs  e.g node state spec changed during reboot
we could list the ports and just delete ports that have this type of name format. not sure if its a good idea just a thought

rollandf (approved):

pkg/host/internal/bridge/ovs/ovs.go
almaslennikov: The old code only checked kc.Uplinks[0], so for multi-uplink bridges (like multiplane setups with more than one PF on each bridge), if the PF being detached was at index 1, the lookup would miss it entirely and skip the cleanup. The refactor iterates all uplinks to find the matching PCI address. Also simplified from a relatedBridges slice to a single brConf pointer with early break — a PCI address can only belong to one bridge, so collecting multiple matches and warning was unnecessary.

pkg/host/internal/bridge/ovs/ovs.go
almaslennikov: Added a code comment referencing the udev naming convention.

Re iterating over numVFs, it is safe because deleteInterfaceByName is a no-op when the interface doesn't exist in OVSDB — so if numVfs is larger than what's actually in the bridge, the extra iterations are harmless. On the other hand, if there were previously more VFs than the currently configured numVfs, the leftover representors won't conflict with the current configuration since they have different names. If the user changes numVfs back, they will get cleaned up. Added a comment in the code explaining this.

almaslennikov:
CI failures don't seem related

pkg/host/internal/bridge/ovs/ovs.go
```diff
@@ -346,8 +354,22 @@ func (o *ovs) RemoveInterfaceFromOVSBridge(ctx context.Context, pciAddress strin
...
 	}
 
-	funcLog.V(2).Info("RemoveInterfaceFromOVSBridge(): remove interface from the bridge")
-	if err := o.deleteInterfaceByName(ctx, dbClient, brConf.Uplinks[0].Name); err != nil {
+	// Remove stale VF representor interfaces for indices 0..numVfs-1.
+	// Representor names follow the udev naming convention used by the operator: {pfName}_{vfIndex}.
+	// We only clean representors up to the currently configured numVfs. If previously more VFs
+	// were configured, leftover representors at higher indices don't conflict with the current
+	// configuration and will be cleaned when numVfs is changed to cover those indices again.
+	for i := 0; i < numVfs; i++ {
```
SchSeba: can we not use the numVfs here and just use a running number until the first representor that doesn't exist?
this way we will always clean everything.

pkg/host/internal/bridge/ovs/ovs.go
```diff
@@ -298,31 +300,37 @@ func (o *ovs) RemoveOVSBridge(ctx context.Context, bridgeName string) error {
 	return nil
 }
 
-// RemoveInterfaceFromOVSBridge removes interface from the managed OVS bridge
-func (o *ovs) RemoveInterfaceFromOVSBridge(ctx context.Context, pciAddress string) error {
+// RemoveInterfaceFromOVSBridge removes PF uplink and stale VF representor interfaces
+// from the managed OVS bridge identified by the PF's PCI address.
+func (o *ovs) RemoveInterfaceFromOVSBridge(ctx context.Context, pciAddress string, pfName string, numVfs int) error {
```
SchSeba: if we have the pciAddress we can find the interface name right? or the other way around also possible.
I don't need we should send both to this function.

another nit is maybe we need to change the function name because it's a function to remove a PF from a bridge + all the VF representors from the same PR

pkg/host/internal/bridge/bridge.go
```diff
@@ -68,12 +68,11 @@ func (b *bridge) ConfigureBridges(bridgesSpec sriovnetworkv1.Bridges, bridgesSta
...
 
-// DetachInterfaceFromManagedBridge detach interface from a managed bridge,
-// this step is required before applying some configurations to PF, e.g. changing of eSwitch mode.
-// The function detach interface from managed bridges only.
-func (b *bridge) DetachInterfaceFromManagedBridge(pciAddr string) error {
-	log.Log.V(1).Info("DetachInterfaceFromManagedBridge(): detach interface", "pciAddr", pciAddr)
-	if err := b.ovs.RemoveInterfaceFromOVSBridge(context.Background(), pciAddr); err != nil {
+// DetachInterfaceFromManagedBridge detach PF uplink and stale VF representor interfaces
+// from a managed bridge. The function detach interface from managed bridges only.
+func (b *bridge) DetachInterfaceFromManagedBridge(pciAddr string, pfName string, numVfs int) error {
```
SchSeba: same here maybe detachUplinkAndVFRepresentorsFromManagedBridge

pkg/host/internal/sriov/sriov.go
```diff
@@ -109,7 +109,7 @@ func (s *sriov) ResetSriovDevice(ifaceStatus sriovnetworkv1.InterfaceExt) error
 			return err
 		}
 		log.Log.V(2).Info("ResetSriovDevice(): reset eswitch mode and number of VFs", "mode", eswitchMode)
-		if err := s.setEswitchModeAndNumVFs(ifaceStatus.PciAddress, eswitchMode, 0); err != nil {
+		if err := s.setEswitchModeAndNumVFs(ifaceStatus.PciAddress, ifaceStatus.Name, eswitchMode, 0); err != nil {
```
SchSeba: I think we can just pass the `ifaceStatus` all the info is there if we really need the name check my other comment

pkg/host/internal/sriov/sriov.go
```diff
@@ -1123,12 +1123,22 @@ func (s *sriov) createVFs(iface *sriovnetworkv1.Interface) error {
...
 		}
 	}
-	return s.setEswitchModeAndNumVFs(iface.PciAddress, expectedEswitchMode, iface.NumVfs)
+	// Clean stale VF representor interfaces and detach PF uplink from the managed
+	// bridge before VF re-creation. After host reboot, representors from the previous
+	// lifecycle remain in the bridge; the PF uplink also needs to be detached because
+	// setEswitchModeAndNumVFsMlx() skips detachPFFromBridge() when NIC is already
+	// in legacy mode (which is always the case after reboot).
+	if expectedEswitchMode == sriovnetworkv1.ESwithModeSwitchDev {
+		if err := s.detachPFFromBridge(iface.PciAddress, iface.Name, iface.NumVfs); err != nil {
```
SchSeba: same here we pass 3 times information from the same struct we can just send the struct right?

SchSeba (commented):
nice work I left some comments on this one. 

I think we can do some cleanup to the PR and send less/different information between the functions

SchSeba:
Hi @almaslennikov please check the comments when you have time :)

pkg/host/internal/bridge/ovs/ovs.go
almaslennikov: Updated in 6e959cf77. I switched this to enumerate the interfaces actually attached to the managed bridge and remove every exact {pfName}_{numeric VF index} match. This handles sparse stale ports too, so cleanup no longer depends on the current numVfs or stops at a gap. I added coverage with indices 0 and 7 while index 1 is absent.

pkg/host/internal/bridge/bridge.go
almaslennikov: Updated in 6e959cf77. The bridge API is now DetachUplinkAndVFRepresentorsFromManagedBridge(pciAddr), with the corresponding OVS API renamed as well.

pkg/host/internal/sriov/sriov.go
almaslennikov: Addressed in 6e959cf77. createVFs now passes only iface.PciAddress to the cleanup operation; pfName and numVfs are no longer propagated because the bridge layer derives the name and enumerates the actual attached representors.

pkg/host/internal/bridge/ovs/ovs.go
almaslennikov: Updated in 6e959cf77. The OVS API is now DetachUplinkAndVFRepresentorsFromOVSBridge(ctx, pciAddress). It resolves both the managed bridge and PF name from the matching PCI address in the local bridge store, so callers no longer pass both values. The new name also describes the combined cleanup.

pkg/host/internal/bridge/ovs/ovs.go
almaslennikov: Followed up on this in 6e959cf77. Cleanup now lists the interfaces actually attached to the managed bridge and filters them using the PF name derived from the local store, so every stale representor is removed regardless of the current numVfs.

pkg/host/internal/sriov/sriov.go
almaslennikov: Addressed in 6e959cf77 by removing the added PF name argument from the eSwitch call chain entirely. The bridge layer derives the name from its managed store using the PCI address, so setEswitchModeAndNumVFs keeps its original focused signature.

almaslennikov:
Addressed the latest CodeRabbit follow-up in be7118530:

- Preserved underlying error identity with `%w` through the OVS, bridge, and SR-IOV detachment path.
- Added OVS error-path tests for managed-store lookup and stale-store cleanup failures.
- Added a `ConfigSriovInterfaces` failure-path test proving bridge-detachment errors stop VF creation and remain discoverable with `errors.Is`.

Validation rerun in Linux with Go 1.26.5:

- `go test ./pkg/host/... ./pkg/helper/... -count=1`
- `golangci-lint v2.7.2 run ./pkg/host/... ./pkg/helper/...` — 0 issues

pkg/host/internal/sriov/sriov.go
```diff
@@ -1206,6 +1206,16 @@ func (s *sriov) createVFs(iface *sriovnetworkv1.Interface) error {
 			return nil
 		}
 	}
+	// Clean stale VF representor interfaces and detach PF uplink from the managed
+	// bridge before VF re-creation. After host reboot, representors from the previous
+	// lifecycle remain in the bridge; the PF uplink also needs to be detached because
+	// setEswitchModeAndNumVFsMlx() skips bridge cleanup when NIC is already
+	// in legacy mode (which is always the case after reboot).
+	if expectedEswitchMode == sriovnetworkv1.ESwithModeSwitchDev {
```
SchSeba, line 1214: please add 
Missing test coverage:

There are two gaps:

Switchdev spec + current switchdev mode + ManageSoftwareBridges=true: No test verifies that detach is called when the device is already in switchdev mode. This is the "current status is switchdev" half of your question. Though in practice this likely returns early at line 1203 (numVfs and mode already match), so detachUplink wouldn't be reached anyway.

Legacy spec + ManageSoftwareBridges=true: No test explicitly verifies that DetachUplinkAndVFRepresentorsFromManagedBridge is NOT called when the spec is legacy. This is implicitly covered because gomock would fail if an unexpected call were made, but there's no explicit test with ManageSoftwareBridges=true and a legacy-mode spec. All the existing legacy-mode tests leave ManageSoftwareBridges at its default (false).

pkg/host/internal/bridge/ovs/ovs.go
```diff
@@ -298,63 +299,130 @@ func (o *ovs) RemoveOVSBridge(ctx context.Context, bridgeName string) error {
...
+	for _, iface := range interfaces {
+		if _, found := interfaceIDs[iface.UUID]; found {
+			interfaceNames = append(interfaceNames, iface.Name)
+		}
+	}
+	sort.Strings(interfaceNames)
+	return interfaceNames, nil
+}
+
+func isVFRepresentorName(interfaceName, pfName string) bool {
```
SchSeba, line 419: this is the best option to find if this is a vf representor for a specific PF?
if that is the case please add a doc string to this function with an example how the vf representors looks like

pkg/host/internal/bridge/ovs/ovs.go
```diff
@@ -298,63 +299,130 @@ func (o *ovs) RemoveOVSBridge(ctx context.Context, bridgeName string) error {
...
+	ports := []*PortEntry{}
+	if err := dbClient.List(ctx, &ports); err != nil {
+		return nil, fmt.Errorf("failed to list ports: %w", err)
+	}
+
+	interfaceIDs := map[string]struct{}{}
+	for _, port := range ports {
+		if bridge.HasPort(port.UUID) {
+			for _, interfaceID := range port.Interfaces {
+				interfaceIDs[interfaceID] = struct{}{}
```
SchSeba, line 396: is it possible to have more then one interface on a port?

pkg/host/internal/bridge/ovs/ovs.go
```diff
@@ -298,63 +299,130 @@ func (o *ovs) RemoveOVSBridge(ctx context.Context, bridgeName string) error {
...
+	sort.Strings(interfaceNames)
+	return interfaceNames, nil
+}
+
+func isVFRepresentorName(interfaceName, pfName string) bool {
+	vfIndex, found := strings.CutPrefix(interfaceName, pfName+"_")
+	if !found || vfIndex == "" {
+		return false
+	}
+	for _, char := range vfIndex {
```
SchSeba: can we just try to convert this to int and return if we manage to do it?

SchSeba (commented):
great work!

I left some small comments after that we can merge this PR

pkg/host/internal/sriov/sriov.go
almaslennikov: Added explicit coverage in 7e04cd5e3. The switchdev case verifies that matching VF count and current switchdev mode returns early without detaching, because no VF recreation occurs. The legacy case sets ManageSoftwareBridges=true, requires VF recreation, and verifies that the path reaches driver selection without calling bridge detachment.

pkg/host/internal/bridge/ovs/ovs.go
almaslennikov: Added the doc comment and table coverage in 7e04cd5e3. This helper checks the exact naming convention produced by the operator-created udev rule; for example, PF enp216s0f0np0 has representor enp216s0f0np0_7, while enp216s0f0np0_backup is rejected.

pkg/host/internal/bridge/ovs/ovs.go
almaslennikov: Yes. OVSDB Port.interfaces is a set, and a bonded port can reference multiple Interface rows. The loop intentionally collects every interface UUID attached to a bridge port; I added a comment in 7e04cd5e3 to make that case explicit.

pkg/host/internal/bridge/ovs/ovs.go
almaslennikov: Updated in 7e04cd5e3. The suffix is now parsed with strconv.Atoi and accepted only when conversion succeeds and the VF index is non-negative; the guard also preserves rejection of an explicitly signed +index.

SchSeba (approved):
LGTM nice work!

PR #1057, opened 2026-03-15
SRIOV persistent logging enhancement

gavrielg1:
SRIOV persistent logging enhancement - 
host-based log storage with rotation for the sriov-network-config-daemon

### DIFF
--- a/doc/design/sriov-persistent-logging-enhancement.md
+++ b/doc/design/sriov-persistent-logging-enhancement.md
@@ -1,0 +1,263 @@
+# Persistent Log Storage for Config Daemon
+
+## Summary
+
+Add persistent, host-based log storage with rotation for the sriov-network-config-daemon so that logs survive container restarts and node reboots. Today, when the config daemon triggers a node reboot (e.g., after SR-IOV firmware changes or VF reconfiguration), all container logs are lost because they only exist in the container's stdout/stderr stream. This makes it extremely difficult to debug why a reboot was initiated.
+
+## Motivation
+
+The config daemon is responsible for applying SR-IOV configuration on each node. Part of that process may require a node reboot (e.g., Mellanox firmware changes, enabling/disabling SR-IOV, totalVfs changes). The reboot is performed via systemd-run reboot after chrooting into `/host`. Once the node reboots, the kubelet restarts, the DaemonSet pod is recreated, and all previous container logs are gone.
+
+This creates a significant debugging gap: operators cannot determine what the config daemon was doing or why it decided to reboot the node.
+
+## Use Cases
+
+1. **Post-reboot debugging**: After a config daemon-initiated reboot, the operator or cluster admin needs to inspect pre-reboot logs to understand what configuration was being applied and why the reboot was required.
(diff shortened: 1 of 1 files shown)


rollandf (commented):
Thanks for the design doc. This feature will be helpful for debugging.

I am missing any reference to `systemd` mode where the config daemon is running directly on the host, which means that writing to `/host` path will probably fail.

Currently in this mode, the logs are available in `journactl` but can be flushed.

The systemd service doesn't talk to the Kubernetes API. It reads config from /etc/sriov-operator/sriov-interface-config.yaml (the SriovConfig struct).
The LogConfig settings would need to be included in the SriovConfig/hosttypes.SriovConfig struct and written to disk by WriteConfFile(), or the systemd service needs hardcoded/flag-based log config.

SchSeba:
+1 we need to also handle the systemd mode

gavrielg1:
> Thanks for the design doc. This feature will be helpful for debugging.
> 
> I am missing any reference to `systemd` mode where the config daemon is running directly on the host, which means that writing to `/host` path will probably fail.
> 
> Currently in this mode, the logs are available in `journactl` but can be flushed.
> 
> The systemd service doesn't talk to the Kubernetes API. It reads config from /etc/sriov-operator/sriov-interface-config.yaml (the SriovConfig struct). The LogConfig settings would need to be included in the SriovConfig/hosttypes.SriovConfig struct and written to disk by WriteConfFile(), or the systemd service needs hardcoded/flag-based log config.

in systemd mode, journactl allows to display logs from previous boots, so this case is covered already by the system logs.

doc/design/sriov-persistent-logging-enhancement.md
```diff
@@ -0,0 +1,301 @@
...
+- **Option B**: Write the log file to a path that resolves correctly in both contexts. Since `/host` maps to `/` on the host, the log path inside chroot would be `/var/log/sriov-network-config-daemon/config-daemon.log`, which is valid. We can open the file using this host-relative path before chroot, and lumberjack rotation will work because chroot sections are short-lived and the file is re-opened at the original path when chroot exits. However, this relies on timing assumptions and could fail if rotation occurs during chroot.
+- **Option C**: Write to a tmpfs or emptyDir mount that is also bind-mounted to the host. However, this does not survive pod restarts.
+
+**Recommended approach**: Option A — Use a dedicated logging mechanism that is isolated from chroot operations. This provides a robust solution that avoids race conditions entirely. The zap logger will buffer writes in memory, and the underlying file writer (lumberjack) will operate in the main namespace, never entering the chrooted environment. This eliminates the risk of path resolution failures during log rotation, regardless of when rotation occurs relative to chroot operations.
```
SchSeba, line 72: please check is this is working as I remember that for cni's in general we had to configured https://github.com/k8snetworkplumbingwg/sriov-cni/blob/master/cmd/sriov/main.go#L16

I am not sure if that is still the case with newer golang versions and the additional go routine will be isolated from the chroot command

SchSeba (commented):
LGTM just a critical comment we must be sure

PR #1061, opened 2026-03-23
feat: integrate dra-driver-sriov as device-plugin alternative

rollandf:
Add optional Dynamic Resource Allocation mode behind the dynamicResourceAllocation feature gate: deploy the SR-IOV DRA driver instead of the network device plugin, and keep SR-IOV configuration aligned with Multus/NAD-oriented networking.

Deployment & RBAC
- Add bindata manifests under bindata/manifests/dra-driver/: basic DeviceClass, ServiceAccount, ClusterRole/Role (incl. pod access), ClusterRoleBinding, DaemonSet with init container (image from SRIOV_DRA_DRIVER_IMAGE / SRIOV_NETWORK_CONFIG_DAEMON_IMAGE).
- Extend operator and Helm chart ClusterRoles for DRA driver APIs and resource.k8s.io DeviceClass management.
- Chart: sriovDraDriver image value and operator env wiring.

CRDs
- Vendor SriovResourcePolicy and DeviceAttributes CRDs (chart + test util) aligned with dra-driver-sriov.

Operator controllers
- SriovOperatorConfig: when DRA is enabled, sync DRA driver objects and skip device-plugin DaemonSet path; when disabled, tear down DRA DaemonSet and related sync uses device plugin as today.
- SriovNetworkNodePolicy: when DRA is enabled, generate/sync DeviceAttributes (per resource pool), SriovResourcePolicy (per node) from SriovNetworkNodePolicy, extended-resource DeviceClass objects (resource.k8s.io), with cleanup when policies or DRA mode change; keep existing device-plugin ConfigMap flow when DRA is off.
- Helper: syncDRADriverObjs, cleanupDRADriverObjs, GetNodeSelectorForDRADriver.

Config daemon
- Discover and coordinate with sriov-dra-driver pods (labels/behavior) alongside existing device-plugin logic.

Dependencies & API
- Add github.com/k8snetworkplumbingwg/dra-driver-sriov module and register SriovResourcePolicy / DeviceAttributes types in the manager scheme.

Documentation
- Add doc/design/DRA-integration.md and extend doc/advanced-features.md for the DRA feature gate and behavior.

Tests
- Extend controller tests for DRA sync/cleanup paths and helper behavior.

## Summary by CodeRabbit

* **New Features**
  * Added feature-gated Dynamic Resource Allocation (DRA) support with a dedicated driver DaemonSet and mode switching from the existing device plugin.
  * Introduced new DRA Kubernetes resources and CRDs: `DeviceAttributes`, `SriovResourcePolicy`, and `DeviceClass`, with supporting RBAC.
  * Extended Helm configuration with DRA driver image/config (CDI root and default interface prefix).

* **Documentation**
  * Updated DRA design and feature-gate docs, including extended resource allocation behavior.

* **Tests**
  * Added unit/Ginkgo coverage for DRA reconciliation and cleanup/sync flows.


doc/design/DRA-integration.md
```diff
@@ -0,0 +1,1348 @@
+---
+title: DRA (Dynamic Resource Allocation) Integration
+authors:
+  - rollandf
+reviewers:
+  - SchSeba
+  - adrianchiris
+creation-date: 11-02-2026
+last-updated: 18-03-2026
```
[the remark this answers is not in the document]
rollandf: We are in 2026

go.mod
```diff
@@ -16,13 +16,14 @@ require (
 	github.com/google/renameio/v2 v2.0.2
 	github.com/google/uuid v1.6.0
 	github.com/hashicorp/go-retryablehttp v0.7.8
-	github.com/jaypipes/ghw v0.22.0
+	github.com/jaypipes/ghw v0.23.0
 	github.com/jaypipes/pcidb v1.1.1
+	github.com/k8snetworkplumbingwg/dra-driver-sriov v0.0.0-20260322160827-52d399b5dbdf
```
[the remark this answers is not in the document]
rollandf: we are in 2026

wizhaoredhat:
Hi @rollandf , I think we need this PR in soon. I would be happy to help along with Seb to get this moving.
I think a future enhancement is to have DRA Standalone mode running along side with Device Plugin. Both managing 2 different pools of devices. But I think this simpler approach first with the feature flag for toggling is better than nothing.

Naturally I will be reviewing this PR at length soon.

controllers/suite_test.go
```diff
@@ -172,6 +173,8 @@ var _ = BeforeSuite(func() {
 	By("registering schemes")
 	err = sriovnetworkv1.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
+	err = sriovdrav1alpha1.AddToScheme(scheme.Scheme)
```
wizhaoredhat, line 176: I think we are missing this above:
`err = os.Setenv("SRIOV_DRA_DRIVER_IMAGE", "mock-image")
	Expect(err).NotTo(HaveOccurred())`

wizhaoredhat (commented):
We also need changes to hack/env.sh for the image substitution.
Also hack/run-e2e-test.sh.

pkg/daemon/daemon.go
```diff
@@ -215,8 +215,21 @@ func (dn *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
-			if vars.FeatureGate.IsEnabled(consts.BlockDevicePluginUntilConfiguredFeatureGate) &&
+			// Unblock device plugin or DRA driver pods when config is already in sync (e.g. after restart).
+			// Both use the same wait-for-config annotation; config daemon removes it so the init container can exit.
+			if vars.FeatureGate.IsEnabled(consts.DynamicResourceAllocationFeatureGate) {
+				draDriverPods, err := dn.getDRADriverPodsForNode(ctx)
+				if err != nil {
+					reqLogger.Error(err, "failed to get DRA driver pods")
+					return ctrl.Result{}, err
+				}
+				if len(desiredNodeState.Spec.Interfaces) > 0 {
```
wizhaoredhat, line 226: We skip annotation removal when `len(desiredNodeState.Spec.Interfaces)` is 0. Wouldn't the init container wait on device-plugin-wait-config indefinitely? Is this by design?

I am worried that we will have many pods stuck in Init which may indicate something unhealthy is going on.

I understand that we run the DRA Driver on all config daemon nodes unlike the device plugin which runs on nodes with SR-IOV. I am wondering why we choose to deviate here and looking at DRA-integration md file, it seems like it was a choice.

controllers/sriovoperatorconfig_controller.go
```diff
@@ -201,8 +201,26 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
...
-		return reconcile.Result{}, err
+	// Deploy either DRA driver or device plugin based on feature gate
+	if r.FeatureGate.IsEnabled(consts.DynamicResourceAllocationFeatureGate) {
+		logger.Info("DRA feature gate enabled, deploying DRA driver instead of device plugin")
+		// Clean up device plugin if it exists
+		if err = cleanupDevicePluginObjs(ctx, r.Client); err != nil {
+			logger.Error(err, "Failed to cleanup device plugin objects")
+			return reconcile.Result{}, err
+		}
+		if err = syncDRADriverObjs(ctx, r.Client, r.Scheme, defaultConfig); err != nil {
```
wizhaoredhat, line 212: As safer pattern would be to sync the desired object first. If successfully remove the old objects. This prevents a situation where we lose both DRA and DP.
E.g.
```
if err = syncDRADriverObjs(ctx, r.Client, r.Scheme, defaultConfig); err != nil {
   if err = cleanupDevicePluginObjs(ctx, r.Client); err != nil {
...
```

controllers/helper.go
```diff
@@ -207,6 +242,143 @@ func syncPluginDaemonObjs(ctx context.Context,
...
+
+func deleteIfNotFound(ctx context.Context, c k8sclient.Client, obj k8sclient.Object) error {
+	if err := c.Delete(ctx, obj); err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	return nil
+}
+
+// cleanupDevicePluginObjs removes device plugin objects when switching to DRA mode
+func cleanupDevicePluginObjs(ctx context.Context, client k8sclient.Client) error {
```
wizhaoredhat, line 362: Shouldn't we do the same as "cleanupDRADriverObjs" and cleanup the RBAC, SAs, etc...?
Any reason we only cleanup the DP's daemonset only?

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -704,3 +732,512 @@ func updateDevicePluginResource(
...
+
+// sriovResourcePolicyNodeSelectorForHostname returns a NodeSelector that matches exactly
+// one node by kubernetes.io/hostname (same intent as the former map nodeSelector).
+func sriovResourcePolicyNodeSelectorForHostname(hostname string) *corev1.NodeSelector {
+	return &corev1.NodeSelector{
+		NodeSelectorTerms: []corev1.NodeSelectorTerm{
+			{
+				MatchExpressions: []corev1.NodeSelectorRequirement{
+					{
+						Key:      corev1.LabelHostname,
```
wizhaoredhat, line 924: Shouldn't we get the hostname from node.Labels[corev1.LabelHostname] ? node.Name is not guaranteed to equal the label I believe

controllers/helper.go
```diff
@@ -220,6 +392,21 @@ func syncDsObject(ctx context.Context, client k8sclient.Client, scheme *runtime.
...
 		}
+	case clusterRoleResourceName, clusterRoleBindingResourceName:
+		// ClusterRole/ClusterRoleBinding are cluster-scoped; do not set controller reference from namespaced dc.
+		if err := apply.ApplyObject(ctx, client, obj); err != nil {
+			logger.Error(err, "Fail to sync", "Kind", kind)
+			return err
+		}
+	case deviceClassResourceName:
+		if err := apply.ApplyObject(ctx, client, obj); err != nil {
+			if apimeta.IsNoMatchError(err) {
```
wizhaoredhat, line 403: Is there a reason why for "IsNoMatchError" where the CRD is not present that we just log Info and return nil? Shouldn't we error the reconcile?

pkg/daemon/daemon.go
```diff
@@ -636,6 +684,25 @@ func (dn *NodeReconciler) getDevicePluginPodsForNode(ctx context.Context) ([]cor
 	return pods.Items, nil
 }
 
+// getDRADriverPodsForNode returns the DRA driver pods running on this node
+func (dn *NodeReconciler) getDRADriverPodsForNode(ctx context.Context) ([]corev1.Pod, error) {
+	funcLog := log.Log.WithName("getDRADriverPodsForNode")
+	pods := &corev1.PodList{}
+	err := dn.client.List(ctx, pods, &client.ListOptions{
+		Namespace: vars.Namespace, Raw: &metav1.ListOptions{
+			LabelSelector: "app=sriov-dra-driver",
```
wizhaoredhat, line 693: Shouldn't we iterate through this list and match on pods not only on app=sriov-dra-driver but also if it makes what we expect for the service account and the owner reference matches with controller? This makes it more tight and we don't (albeit unlikely) hit the wrong pod(s).

pkg/daemon/daemon.go
```diff
@@ -693,6 +760,61 @@ func (dn *NodeReconciler) restartDevicePluginPod(ctx context.Context) error {
...
+		if errors.IsNotFound(err) {
+			funcLog.Info("pod to delete not found")
+			continue
+		}
+		if err != nil {
+			funcLog.Error(err, "Failed to delete DRA driver pod")
+			return err
+		}
+		newPod := &corev1.Pod{}
+		if err := wait.PollUntilContextCancel(ctx, time.Second, true, func(ctx context.Context) (bool, error) {
```
wizhaoredhat, line 792: Shouldn't we use PollUntilContextTimeout to prevent looping indefinitely?

controllers/helper.go
rollandf: I added this to pass the OpenShift pipeline

SchSeba:
closing this one in favor of https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1151 that will continue the great work from @rollandf

PR #1070, opened 2026-04-14
Enable TLS cipher, version, and curve preferences control across operator components

SchSeba:
## Summary by CodeRabbit

* **New Features**
  * Added configurable TLS cipher suites, minimum TLS version, and curve preferences for operator webhooks and the metrics exporter.
  * On OpenShift, TLS settings can be derived from the cluster APIServer TLS security profile when supported.
  * Helm chart and operator deployment now expose TLS configuration values.
* **Bug Fixes**
  * Webhook and metrics proxy TLS flags are now conditionally included only when configured (avoids forcing defaults).
* **Tests**
  * Expanded TLS unit tests and added/extended TLS conformance coverage for correct flag propagation and cipher conversion.
* **Chores**
  * Updated RBAC to allow watching APIServer objects; refreshed e2e test tooling/settings.


e0ne (approved):
LGTM

cmd/webhook/start.go
```diff
@@ -50,6 +52,10 @@ func init() {
 	startCmd.Flags().IntVar(&port, "port", 443,
 		"Secure port that the webhook listens on")
 	startCmd.Flags().BoolVar(&enableHTTP2, "enable-http2", false, "If HTTP/2 should be enabled for the metrics and webhook servers.")
+	startCmd.Flags().StringVar(&tlsCipherSuites, "tls-cipher-suites", "",
+		"Comma-separated list of TLS 1.2 and earlier cipher suites (OpenSSL or IANA names). Insecure cipher suites are rejected. If empty, uses Go defaults.")
```
adrianchiris: should: TLS 1.2 and earlier cipher suites

be: TLS 1.2 and later cipher suites ?

deployment/sriov-network-operator-chart/values.yaml
```diff
@@ -9,6 +9,12 @@ operator:
   nodeSelector: {}
   # custom extra environment variables for the operator container
   extraEnv: {}
+  # TLS configuration for admission webhooks and metrics proxy (Kubernetes only).
+  # On OpenShift, TLS settings are read from the APIServer TLS profile.
```
adrianchiris: can you add the CR object group as well ? so its clear its taken from config.openshift.io

controllers/drain_controller_test.go
```diff
@@ -174,34 +174,37 @@ var _ = Describe("Drain Controller", Ordered, func() {
 			simulateDaemonSetAnnotation(node1, constants.DrainRequired)
 			simulateDaemonSetAnnotation(node2, constants.DrainRequired)
 
-			// Only the first node drains
-			expectNodeStateAnnotation(nodeState1, constants.DrainComplete)
-			expectNodeStateAnnotation(nodeState2, constants.DrainIdle)
+			// Exactly one of the two nodes should drain first.
+			// Which one drains first is non-deterministic due to concurrent reconciles.
+			firstDrainedNode, firstDrainedNodeState, secondDrainedNode, secondDrainedNodeState :=
```
adrianchiris, line 179: fix for general test issue ? or is it related to changes in this PR ?

controllers/sriovoperatorconfig_controller.go
```diff
@@ -175,11 +203,45 @@ func defaultConfigPredicate() predicate.Predicate {
...
+	}
+
+	return builder.Complete(r)
+}
+
+// isAPIServerCRDAvailable checks if the OpenShift APIServer CRD is available in the cluster.
+func (r *SriovOperatorConfigReconciler) isAPIServerCRDAvailable(mgr ctrl.Manager) bool {
+	log.Log.WithName("isAPIServerCRDAvailable").V(2).Info("Checking if OpenShift APIServer CRD is available")
+	gvk := schema.GroupVersionKind{
+		Group:   "config.openshift.io",
```
adrianchiris: should we use the GroupVersion definned in configv1 ?

controllers/sriovoperatorconfig_controller_test.go
```diff
@@ -139,7 +150,7 @@ var _ = Describe("SriovOperatorConfig controller", Ordered, func() {
 				}
 				err = k8sClient.Update(ctx, config)
 				Expect(err).NotTo(HaveOccurred())
-			}, util.APITimeout, util.RetryInterval).Should(Succeed())
+			}, util.APITimeout*3, util.RetryInterval).Should(Succeed())
```
adrianchiris: this change really needed ?

pkg/orchestrator/kubernetes/kubernetes.go
```diff
@@ -42,3 +43,23 @@ func (k *Kubernetes) BeforeDrainNode(_ context.Context, _ *corev1.Node) (bool, e
...
+// Environment variables:
+//   - TLS_CIPHER_SUITES: comma-separated list of cipher suites
+//   - TLS_MIN_VERSION: minimum TLS version (e.g., VersionTLS12, VersionTLS13)
+func (k *Kubernetes) GetTLSConfig(_ context.Context) (*consts.TLSConfig, error) {
+	cipherSuites := os.Getenv("TLS_CIPHER_SUITES")
+	minVersion := os.Getenv("TLS_MIN_VERSION")
+
+	// If no TLS env vars are set, return nil to use component defaults
+	if cipherSuites == "" && minVersion == "" {
+		return nil, nil
```
adrianchiris: if both are empty shouldnt it still return TLSConfig struct with empty values ?

adrianchiris: then when templating we always render the cipherSuite and minTLSVersion flags just with empty values the mean default ?

just a thought. as it will slightly simplify the rendering

pkg/orchestrator/openshift/tlsprofile.go
```diff
@@ -0,0 +1,103 @@
...
+			result.MinTLSVersion = "VersionTLS12"
+		}
+	}
+
+	return result, nil
+}
+
+func openshiftCipherSuitesToIANAAndSecure(cipherSuites []string) []string {
+	insecureCipherIDs := map[uint16]struct{}{}
+	for _, insecureID := range cliflag.InsecureTLSCiphers() {
```
adrianchiris: cliflag implementation just copys the map returned by:

```
import (
  "crypto/tls"
)

tls.InsecureCipherSuites() 
```

adrianchiris: so maybe we can do the same here. i dont think that k8s cliflag should be the source of truth here WDYT ?

pkg/orchestrator/openshift/tlsprofile.go
```diff
@@ -0,0 +1,103 @@
...
+		insecureCipherIDs[insecureID] = struct{}{}
+	}
+
+	result := make([]string, 0, len(cipherSuites))
+	for _, cipherSuite := range cipherSuites {
+		normalized := openshiftCipherSuiteNameToIANA(cipherSuite)
+		ids, err := cliflag.TLSCipherSuites([]string{normalized})
```
adrianchiris: same. 
all ciphers are taken from `tls.CipherSuites()`

pkg/orchestrator/openshift/tlsprofile.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil || len(ids) != 1 {
+			continue
+		}
+		if _, found := insecureCipherIDs[ids[0]]; found {
+			continue
+		}
+		result = append(result, normalized)
```
adrianchiris: so we essentially do filtering over: the supported cipher suites by  crypto/tls minus the insecure ones from crypto/tls.

do we want to apply the same logic in the kubernetes case ?

pkg/orchestrator/openshift/tlsprofile.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/consts"
+)
+
+// GetTLSConfig retrieves the cluster-wide TLS security profile from the OpenShift APIServer resource.
+// Returns nil if no TLSSecurityProfile is configured on the cluster (preserving backward compatibility
+// by using component defaults).
+func (c *OpenshiftOrchestrator) GetTLSConfig(ctx context.Context) (*consts.TLSConfig, error) {
+	apiServer := &configv1.APIServer{}
+	err := c.kubeClient.Get(ctx, types.NamespacedName{Name: "cluster"}, apiServer)
+	if err != nil {
```
adrianchiris, line 39: do we want to log anything ?

adrianchiris: also, should we fallback to the env-vars ? like we do in the k8s orchestrator

pkg/orchestrator/orchestrator.go
```diff
@@ -47,6 +47,14 @@ type Interface interface {
 	//   - bool: true if the post-drain operations completed successfully, false if more time is needed
 	//   - error: an error if the post-drain operations failed
 	AfterCompleteDrainNode(context.Context, *corev1.Node) (bool, error)
+
+	// GetTLSConfig retrieves the cluster-wide TLS security profile configuration.
```
adrianchiris: the description is not correct. for k8s we dont return nil always.

but maybe it makes sense and we should return cluster level tls config, then if its nil, the operator can try to build it from the information passed in env vars ? WDYT ?

also instead of "TLS security profile configuration" just "TLS configuration"

cmd/webhook/start.go
```diff
@@ -154,6 +160,25 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		panic(err)
 	}
 
+	// Parse TLS cipher suites
+	cipherSuites, err := webhook.ParseCipherSuitesFlag(tlsCipherSuites)
```
adrianchiris: the parsing supports both openssl and IANA naming. dont we normalize to IANA in the operator?

why should we support both formats here ? does metrics-exporter and operator-webhook support both ?

cmd/webhook/start.go
```diff
@@ -154,6 +160,25 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		setupLog.Error(err, "invalid tls-cipher-suites flag")
+		panic(err)
+	}
+
+	// Parse minimum TLS version
+	minVersion, err := webhook.TLSVersionToGo(tlsMinVersion)
```
adrianchiris: same comment for this one. on just using one format in the webhook. WDYT ?

cmd/webhook/start.go
SchSeba: right!

    I also removed the insecure rejected as we spoke in the community meeting the validation is done in the operator not the webhook

deployment/sriov-network-operator-chart/values.yaml
SchSeba: done

pkg/orchestrator/orchestrator.go
SchSeba: I like the idea.

I moved the ENV variable to the reconciler and after that we override if the orchestrator function return something that is not nil

pkg/orchestrator/openshift/tlsprofile.go
SchSeba: log added. we don't fallback that can be problematic in OCP we just failed and reconcile again until the user fix the API call

pkg/orchestrator/openshift/tlsprofile.go
SchSeba: please check the function now I am using the openshift list and changed a bit the implementation here

pkg/orchestrator/kubernetes/kubernetes.go
SchSeba: I am passing nill on the kubernetes orchestrator and in the controller we are going to use the env variable

controllers/drain_controller_test.go
SchSeba: not related.
I will open a different PR for this

SchSeba:
failing tests are because we need https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/205 to get merged first

cmd/webhook/start.go
```diff
@@ -154,6 +163,21 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
+		"minVersion", tlsMinVersion)
+
+	var cipherSuites []uint16
+	if tlsCipherSuites != "" {
+		cipherSuites = openshiftcrypto.CipherSuitesOrDie(strings.Split(tlsCipherSuites, ","))
+	}
+
+	var minVersion uint16 = tls.VersionTLS12
+	if tlsMinVersion != "" {
+		minVersion = openshiftcrypto.TLSVersionOrDie(tlsMinVersion)
```
adrianchiris: WDYT about adding this one in pkg/utils/tls.go ? then we can use it from there ? 
that way we dont add openshift import directly here ?

adrianchiris:
minor comment otherwise lgtm.

SchSeba:
we need https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/216 as we added support for curves.

be aware curves configuration is disable in the metrics exporter as kube-rbac-proxy still doesn't support it

pkg/consts/constants.go
```diff
@@ -14,6 +14,22 @@ const (
...
+// TLSConfig holds cluster-wide TLS configuration retrieved from the orchestrator or based on the environment variables.
+// Empty string values indicate that the component should use its default configuration.
+type TLSConfig struct {
+	// CipherSuites is a comma-separated list of TLS cipher suites.
+	// If empty, the component should use its default cipher suites.
+	CipherSuites string
+	// MinTLSVersion is the minimum TLS version (e.g., "VersionTLS12", "VersionTLS13").
+	// If empty, the component should use its default minimum version.
+	MinTLSVersion string
+	// CurvePreferences is a comma-separated list of TLS group/curve names
```
adrianchiris, line 26: the operands accept curve (numeric) IDs.

i wonder if we should stick to numeric IDs here as well.

also, about the operator env var, should it contain the list of numeric IDs directly or you think its better to have the string (and translate.)

in the openshift orchestrator we would then do the translation when building the IDs.

deployment/sriov-network-operator-chart/values.yaml
```diff
@@ -9,6 +9,13 @@ operator:
   nodeSelector: {}
   # custom extra environment variables for the operator container
   extraEnv: {}
+  # TLS configuration for admission webhooks and metrics proxy (Kubernetes only).
+  # On OpenShift, TLS settings are read from the config.openshift.io/v1 APIServer TLS profile.
+  # Leave empty to use Go defaults.
+  tls:
+    cipherSuites: ""
```
adrianchiris, line 18: can you add a comment for each with example input ?

pkg/consts/constants.go
adrianchiris: LMK what you think.

SchSeba: I prefer to leave them consistant with `cipherSuites` and the `MinTLSVersion` they both have names.

deployment/sriov-network-operator-chart/values.yaml
SchSeba: sure done

pkg/consts/constants.go
adrianchiris: ack.

adrianchiris (approved):
lets have CI green and we can merge.

zeeke (approved):
LGTM

SchSeba:
Thanks for the help folks!
merging this one

PR #1075, opened 2026-04-26
fix rdma test

SchSeba:
Fix the rdma test, when the device plugin is restarted, the allocatable resource goes to 0 and then back to the original value that caused the test to fail.

To not have a flaky test we don't use the consistency check, but an eventually check.

### DIFF
--- a/test/conformance/tests/test_networkpool.go
+++ b/test/conformance/tests/test_networkpool.go
@@ -267,14 +267,24 @@ var _ = Describe("[sriov] NetworkPool", Ordered, func() {
 				}, 2*time.Minute, time.Second).Should(BeTrue())
 			}
 
-			By("checking the amount of allocatable devices remains after device plugin reset")
-			Consistently(func() int64 {
+			By("waiting for the new sriov device-plugin pod to be running on the node")
+			Eventually(func(g Gomega) []corev1.Pod {
+				pods, err := clients.Pods(operatorNamespace).List(context.Background(), metav1.ListOptions{
+					LabelSelector: "app=sriov-device-plugin",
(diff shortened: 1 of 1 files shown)


test/conformance/tests/test_networkpool.go
```diff
@@ -267,14 +267,27 @@ var _ = Describe("[sriov] NetworkPool", Ordered, func() {
...
+			By("waiting for the new sriov device-plugin pod to be running on the node")
+			Eventually(func(g Gomega) []corev1.Pod {
+				pods, err := clients.Pods(operatorNamespace).List(context.Background(), metav1.ListOptions{
+					LabelSelector: "app=sriov-device-plugin",
+					FieldSelector: "spec.nodeName=" + testNode,
+				})
+				g.Expect(err).ToNot(HaveOccurred())
+				if err != nil {
+					return nil
+				}
```
zeeke: no need for this. L276 quits the function if err != nil
```suggestion
```

zeeke (commented):
LGTM, only one minor comment

rollandf (approved):

test/conformance/tests/test_networkpool.go
adrianchiris: +1

SchSeba: done

adrianchiris: still here

SchSeba: this time it's done :P

adrianchiris (approved):

SchSeba:
Thanks!

Merging this one

PR #1080, opened 2026-05-10
Dockerfile: add systemd for mstfwreset warm-reboot support

zhiqiangf:
## Summary

`mstfwreset.py:rebootMachine()` calls `reboot` as a direct subprocess ([mstfwreset.py:1663](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/vendor/github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vendors/mellanox/mellanox.go)) on the `WARM_REBOOT` path (`mstfwreset -l 3`). The `reboot` binary is provided by the `systemd` package, which was not installed in the container image.

## Failure

Without `systemd`, the `WARM_REBOOT` path fails with:

```
Return code: 127
stderr: '/bin/sh: line 1: reboot: command not found'
Exception: Failed to reboot machine please reboot machine manually
```

This was reproduced by directly invoking `mstfwreset.py:rebootMachine()` inside the `sriov-network-config-daemon` container on OCP 4.21 with the `mellanoxFirmwareReset` feature gate enabled.

## Context

This path is triggered when:
1. `mellanoxFirmwareReset` feature gate is enabled
2. `mstfwreset` is called with `-l 3` (WARM_REBOOT level)
3. Firmware activation requires a machine reboot

The `kmod` package was previously added to support `lsmod` (also called by `mstfwreset.py`). `systemd` is the remaining missing dependency for the complete mstfwreset warm-reboot code path.

## Change

```diff
-RUN if [ "$(uname -m)" = "s390x" ]; then yum -y install hwdata pciutils; else yum -y install hwdata pciutils kmod "${MSTFLINT}"; fi && yum clean all
+# systemd is required for the reboot binary used by mstfwreset.py:rebootMachine() on the
+# WARM_REBOOT path (mstfwreset -l 3), triggered when mellanoxFirmwareReset feature gate is enabled.
+RUN if [ "$(uname -m)" = "s390x" ]; then yum -y install hwdata pciutils; else yum -y install hwdata pciutils kmod systemd "${MSTFLINT}"; fi && yum clean all
```

## Test plan

- [ ] Build `sriov-network-config-daemon` image and verify `which reboot` returns `/usr/sbin/reboot`
- [ ] Enable `mellanoxFirmwareReset` feature gate and verify `mstfwreset -l 3` can reach the reboot step without `command not found`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Dockerfile.sriov-network-config-daemon
+++ b/Dockerfile.sriov-network-config-daemon
@@ -7,7 +7,9 @@ FROM quay.io/centos/centos:stream9
 ARG MSTFLINT=mstflint
 # We have to ensure that pciutils is installed. This package is needed for mstfwreset to succeed.
 # xref pkg/vendors/mellanox/mellanox.go#L150
-RUN if [ "$(uname -m)" = "s390x" ]; then yum -y install hwdata pciutils; else yum -y install hwdata pciutils kmod "${MSTFLINT}"; fi && yum clean all
+# systemd is required for the reboot binary used by mstfwreset.py:rebootMachine() on the
+# WARM_REBOOT path (mstfwreset -l 3), triggered when mellanoxFirmwareReset feature gate is enabled.
+RUN if [ "$(uname -m)" = "s390x" ]; then yum -y install hwdata pciutils; else yum -y install hwdata pciutils kmod systemd "${MSTFLINT}"; fi && yum clean all
 LABEL io.k8s.display-name="sriov-network-config-daemon" \
       io.k8s.description="This is a daemon that manage and config sriov network devices in Kubernetes cluster"
 COPY --from=builder /go/src/github.com/k8snetworkplumbingwg/sriov-network-operator/build/_output/cmd/sriov-network-config-daemon /usr/bin/


zhiqiangf:
Thank you for the review. After tracing the code more carefully, I agree the `systemd` package addition is not the right fix, but I disagree that a dummy `exit 0` script is the right approach either.

**Why dummy `exit 0` is incorrect:**

`MlxResetFW()` in [mellanox.go:180](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/vendors/mellanox/mellanox.go#L180) calls `mstfwreset -l 3 --skip_driver` and checks the return code:

```go
_, stderr, err := m.utils.RunCommand("mstfwreset", cmdArgs...)
if err != nil {
    errs = append(errs, err)
}
```

There is **no separate** `needReboot` flag set after this call — `mstfwreset -l 3` is the mechanism that performs both the PCIe-level firmware activation AND the OS warm reboot as a single operation. A dummy `exit 0` would make the operator believe the reset succeeded while the firmware change was never activated, leaving the NIC misconfigured.

The daemon's `rebootNode()` (daemon.go:710) is a **separate** code path for configuration changes that require a reboot — it uses `Chroot(consts.Host)` + `systemd-run`. It is not related to `MlxResetFW`.

**You are correct that `reboot` inside a container fails via D-Bus.** The right fix is to move the reboot trigger out of `mstfwreset.py` and into the Go operator code, using the host chroot path the daemon already uses:

**Option A** — Call `mstfwreset` at a level that only does the PCIe reset without triggering an OS reboot (e.g. `-l 1` or `-l 2`), then let the operator trigger the host reboot via its existing `rebootNode()` mechanism.

**Option B** — After `mstfwreset` completes the PCIe reset, have the Go code trigger the host reboot via `Chroot(consts.Host)` + `systemd-run`, exactly as `rebootNode()` does.

I'll update the PR to reflect the correct fix. Dropping the Dockerfile change.

zhiqiangf:
**Update — closing this PR, the fix is not needed here.**

After deeper analysis of `mstfwreset.py`, the reset levels are:

| Flag | Value | What it does |
|------|-------|-------------|
| `-l 0` | `LIVE_PATCH` | No driver/link disruption |
| `-l 1` | `IMMEDIATE_RESET` | ARM-side reset only |
| `-l 3` | `PCI_RESET` | Driver restart + PCIe link reset → calls `resetPciAddr()` |
| `-l 4` | `WARM_REBOOT` | OS warm reboot → calls `rebootMachine()` → calls `reboot` |

**The operator calls `mstfwreset -l 3` (PCI_RESET) — this path calls `resetPciAddr()`, NOT `rebootMachine()`.** The `reboot` binary is only invoked on `-l 4` (WARM_REBOOT), which the operator never uses.

`rebootMachine()` is only triggered if:
1. A PCIe switch device is used (auto-selects `-l 4`), or
2. The tool is called manually with `-l 4`

**Neither condition applies to the SR-IOV operator's `MlxResetFW()` usage.**

Closing this PR — no Dockerfile change is needed. The `reboot` binary absence does not affect the operator's current code path. I'll update the OCPBUGS tickets accordingly.

PR #1082, opened 2026-05-18
Enable MellanoxFirmwareResetFeatureGate by default

SchSeba:
This feature gate is used to enable the Mellanox firmware reset feature.

The reason we enable this feature gate is because on newer mstflint version mstconfig command sends MFRL with bit 6 AND bit 3 ((1<<6)|(1<<3)). some cards like ConnectX-6 Dx firmware does not support the bit 3 (PCIe link toggle) reset trigger for this operation, so the MFRL SET command fails (rc != 0).
The code returns "Please power cycle machine to load new configurations."

The operator does not have access to out of band management system to make a power cycle. so instead we run mstfwreset command to reset the firmware.

### DIFF
--- a/pkg/featuregate/featuregate.go
+++ b/pkg/featuregate/featuregate.go
@@ -16,7 +16,7 @@ var DefaultFeatureStates = map[string]bool{
 	consts.MetricsExporterFeatureGate:                  false,
 	consts.ManageSoftwareBridgesFeatureGate:            false,
 	consts.BlockDevicePluginUntilConfiguredFeatureGate: true,
-	consts.MellanoxFirmwareResetFeatureGate:            false,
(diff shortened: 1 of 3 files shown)


SchSeba:
cc @adrianchiris

adrianchiris (approved):

SchSeba:
Thanks merging this oen

wizhaoredhat:
@SchSeba We might need to revert this. Upgrading the firmware should fix the issue. We don't need the feature gate

wizhaoredhat:
CC: @zhiqiangf

wizhaoredhat:
It took some debugging from @zhiqiangf. We upgraded the firmware, and it then allowed the "soft reboot" option to apply mstconfig configs.

PR #1085, opened 2026-05-20
Test Fix for AWS platform

SchSeba:
Fixes the test for AWS platform to not fail if the config daemon pod is not found.

we can't use `getConfigDaemonPod` as it uses `gomega Expect` and failed if the pod doesn't exist

### DIFF
--- a/test/conformance/tests/test_aws_platform.go
+++ b/test/conformance/tests/test_aws_platform.go
@@ -15,6 +15,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/utils/ptr"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


test/conformance/tests/test_aws_platform.go
```diff
@@ -191,10 +193,19 @@ var _ = Describe("[sriov] aws platform", Ordered, func() {
 			Expect(err).ToNot(HaveOccurred())
 
 			By("waiting for the new config daemon pod to be running")
-			Eventually(func() bool {
-				newPod, err := getConfigDaemonPod(node)
-				// Make sure the replacement pod exists and is different from the deleted one
-				if err != nil || newPod.Name == oldPodName {
+			Eventually(func(g Gomega) bool {
```
adrianchiris, line 196: why not use:

	newPod, err := getConfigDaemonPod(node)
    g.Expect(err).ToNot(HaveOccurred())
    if newPod.Name == oldPodName { 
        return false
    }
    ....

adrianchiris: this basically re-implements getConfigDaemonPod IIUC (unless im reading it wrong)

SchSeba: oh sorry this was already fixed!

check this PR (https://github.com/openshift/sriov-network-operator/pull/1219/changes)
in here it was failing but we already fix that https://github.com/SchSeba/sriov-network-operator/blob/cf8fcc90da6ef1c9456ce0d80f240321401822b2/test/conformance/tests/test_sriov_operator.go#L1967

it was fixed here https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1066

let me close this one

PR #1088, opened 2026-05-31
Rebootloop Limit Feature

yahlifried:
this is the design for a feature to stop bootloops after a certain amount of boots, giving a clear message to the user

## Summary by CodeRabbit

* **New Features**
  * Added a safeguard that limits node reboots to five attempts per configuration generation.
  * Nodes that reach the limit now enter a failed sync state with a clear error message.
  * Reboot attempts are tracked persistently and reset after successful configuration.
  * Counters start fresh when the configuration generation changes.

* **Tests**
  * Added coverage for reboot limits, counter resets, generation changes, and tracker errors.
  * Added validation that successful configuration resets the persisted counter.

* **Documentation**
  * Documented reboot tracking behavior, upgrade and downgrade handling, and recovery scenarios.

### DIFF
--- a/doc/design/node-reboot-loop-limit.md
+++ b/doc/design/node-reboot-loop-limit.md
@@ -1,0 +1,213 @@
+---
+title: Node Reboot Loop Limit
+authors:
+  - yahlifried
+reviewers:
+  - SchSeba
+creation-date: 25-05-2026
+last-updated: 03-06-2026
+---
+
+# Node Reboot Loop Limit
+
+## Summary
+
+Limit the number of reboots allowed for a single `SriovNetworkNodeState` generation
+without manual intervention, preventing infinite reboot loops caused by
+misconfigurations or bugs.
+
+## Motivation
+
+The SR-IOV Network Operator drains and reboots nodes for several reasons during the
+operator's lifecycle, such as updating IOMMU, increasing VFs number for Mellanox or
+Nvidia, switching eSwitch mode, etc.
+
+In some cases, misconfigurations or unknown bugs can cause the applied configuration
+to fail repeatedly after each reboot. Because the daemon re-enters the reconciliation
+loop on startup and sees the same desired state, it triggers another reboot, creating
+an infinite reboot loop. This makes the node unavailable and requires manual
+intervention to recover.
+
+### Use Cases
+
+* As a cluster administrator, I want nodes to stop rebooting automatically after a
+  limited number of failed attempts so that I can investigate and fix the issue
+* As a cluster administrator, I want to see a clear failure status on the
+  `SriovNetworkNodeState` when the reboot limit is reached so that I can identify
+  which nodes are affected
+
+### Goals
+
+* Limit the number of reboots for the same configuration generation and prevent
+  reboot loops
+* Set the `SriovNetworkNodeState` sync status to `Failed` with a descriptive message
+  when the limit is reached
+* Reset the reboot counter when a new generation is applied or when configuration
+  succeeds without requiring a reboot
+
+### Non-Goals
+
+* Exposing an API to the user to configure the maximum number of reboots
+* Automatic recovery or retry after the limit is reached
+
+## Proposal
+
+Track the number of reboots per `SriovNetworkNodeState` generation in a YAML file on
+the host filesystem. Before each reboot, increment the counter. On each reconciliation,
+check if the counter has reached the maximum allowed reboots. If so, set the sync
+status to `Failed` and stop reconciling.
+
+### Workflow Description
+
+#### Normal Operation (reboot count below limit)
+
+1. Daemon reconciles `SriovNetworkNodeState` and determines a reboot is required
+2. Daemon reads the reboot tracker file from disk
+3. If the file does not exist or the stored generation differs from the current
+   generation, the reboot count is treated as 0
+4. Reboot count is below `MaxRebootsPerGeneration` (5), so the daemon proceeds
+5. Daemon reads the current boot ID from `/proc/sys/kernel/random/boot_id` and
+   compares it to the stored boot ID in the tracker file
+6. If the boot ID matches (a reboot was already requested in this boot), the daemon
+   skips the reboot and requeues
+7. If the boot ID differs, the daemon increments the reboot counter, updates the
+   boot ID and generation in the tracker file, and reboots the node
+
+#### Reboot Limit Reached
+
+1. Daemon reconciles `SriovNetworkNodeState` after a reboot
+2. Daemon reads the reboot tracker file and finds the reboot count has reached
+   `MaxRebootsPerGeneration`
+3. Daemon sets `SyncStatus` to `Failed` with message:
+   `"reached maximum number of allowed reboots (5) while trying to configure node"`
+4. Daemon stops reconciling and returns
+
+#### Successful Configuration
+
+1. Daemon applies configuration successfully (no reboot needed, or configuration
+   succeeds after reboot)
+2. Daemon resets the reboot counter to 0 for the current generation
+3. Normal operation continues
+
+#### New Generation
+
+1. A new `SriovNetworkNodeState` generation is created (user updates configuration)
+2. Daemon reads the reboot tracker and sees the stored generation differs from the
+   current generation
+3. Reboot count is treated as 0, allowing fresh reboot attempts for the new
+   configuration
+4. On the first reboot for the new generation, `incrementRebootCounter` writes the
+   new generation, count 1, and the current boot ID to the tracker file, replacing
+   the old generation
+
+```mermaid
+flowchart TD
+    A[apply] --> B{Reboot required?}
+    B -->|Yes| C[Read reboot tracker from disk]
+    C --> D{File exists and\ngeneration matches?}
+    D -->|No| E[Treat count as 0]
+    D -->|Yes| F[Use stored count]
+    E --> G{Count >= MaxReboots?}
+    F --> G
+    G -->|Yes| H[Set SyncStatus=Failed\nwith message]
+    G -->|No| I{Boot ID matches\nstored boot ID?}
+    I -->|Yes| J[Skip reboot, requeue]
+    I -->|No| K[Increment counter\nUpdate boot ID\nWrite to disk]
+    K --> L[Reboot node]
+    B -->|No| M{Config succeeded?}
+    M -->|Yes| N[Reset counter to 0]
+    M -->|No| O[Return error]
+```
+
+### Implementation Details/Notes/Constraints
+
+#### Reboot Tracker File
+
+The reboot tracker is stored as a YAML file on the host at
+`/var/lib/sriov/reboot-tracker.yaml` with the following structure:
+
+```yaml
+generation: 3
+rebootCount: 2
+bootID: "a1b2c3d4-e5f6-7890-abcd-ef1234567890"
+```
+
+The file is lazily initialized: it is created on the first write (when incrementing
+or resetting the counter). Reads return a zero count when the file does not exist.
+
+The `generation` field is updated whenever the counter is written — both
+`incrementRebootCounter` and `resetRebootCounter` write the current generation. When
+a new generation arrives, the first write replaces the old generation in the file.
+
+The `bootID` field stores the value of `/proc/sys/kernel/random/boot_id` at the time
+the reboot was requested. This prevents the reconciler from incrementing the counter
+multiple times within the same boot — the reconciler can be triggered more than once
+between when the reboot command is sent and when the node actually reboots.
+
+#### Host Helpers Interface
+
+A new `RebootTrackerInterface` is added to the host helpers layer with three methods:
+
+* `ReadRebootTracker() (*RebootTrackerFile, error)` — reads and unmarshals the file,
+  returns `nil, nil` if the file does not exist
+* `WriteRebootTracker(tracker *RebootTrackerFile) error` — marshals and writes the
+  file, creating the parent directory if needed
+* `ReadBootID() (string, error)` — reads the current boot ID from
+  `/proc/sys/kernel/random/boot_id`
+
+This follows the same pattern as `ReadSriovResult`/`WriteSriovResult` on
+`SystemdInterface`. The interface is separate from `SystemdInterface` because the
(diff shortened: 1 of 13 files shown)


doc/design/node-reboot-loop-limit.md
```diff
@@ -0,0 +1,213 @@
...
+In some cases, misconfigurations or unknown bugs can cause the applied configuration
+to fail repeatedly after each reboot. Because the daemon re-enters the reconciliation
+loop on startup and sees the same desired state, it triggers another reboot, creating
+an infinite reboot loop. This makes the node unavailable and requires manual
+intervention to recover.
+
+### Use Cases
+
+* As a cluster administrator, I want nodes to stop rebooting automatically after a
+  configurable number of failed attempts so that I can investigate and fix the issue
```
zeeke: The max number of failed reboot attempts does not look configurable, as the `MaxRebootsPerGeneration` is a constant set to 5. 
If you ask me, I'd keep it as a constant, as it's simpler to maintain.

zeeke (commented):
Left one comment. 
Besides that, I'm only wondering if we can get rid of the `boot_id` field and make this process simpler. I don't have any good idea about it as of now

yahlifried:
> Left one comment. Besides that, I'm only wondering if we can get rid of the `boot_id` field and make this process simpler. I don't have any good idea about it as of now

I think it is the best way, I explored some techniques using time or uptime but they are less reliable and more complicated. boot id is pretty reliable and just requiring reading from a file. the only thing is, if the file does not exists for some reason (linuxes before kernel 2.4 from 1999, thin containers, etc...), the daemon will get stuck. I can create a fallback to ignore this and make it less reliable but I think it might be solving a problem that doesn't really exist, since I'm not sure such machines will be able to run `sriov-network-operator` anyways

zeeke:
> > Left one comment. Besides that, I'm only wondering if we can get rid of the `boot_id` field and make this process simpler. I don't have any good idea about it as of now
> 
> I think it is the best way, I explored some techniques using time or uptime but they are less reliable and more complicated. boot id is pretty reliable and just requiring reading from a file. the only thing is, if the file does not exists for some reason (linuxes before kernel 2.4 from 1999, thin containers, etc...), the daemon will get stuck. I can create a fallback to ignore this and make it less reliable but I think it might be solving a problem that doesn't really exist, since I'm not sure such machines will be able to run `sriov-network-operator` anyways

Sounds fair. I'm ok with the boot_id for now. Thanks

zeeke (approved):
LGTM

doc/design/node-reboot-loop-limit.md
```diff
@@ -0,0 +1,213 @@
...
+
+Limit the number of reboots allowed for a single `SriovNetworkNodeState` generation
+without manual intervention, preventing infinite reboot loops caused by
+misconfigurations or bugs.
+
+## Motivation
+
+The SR-IOV Network Operator drains and reboots nodes for several reasons during the
+operator's lifecycle, such as updating IOMMU, increasing VFs number for Mellanox or
+Nvidia, switching eSwitch mode, etc.
```
SchSeba: nit: `NVIDIA Mellanox cards firmware`

doc/design/node-reboot-loop-limit.md
```diff
@@ -0,0 +1,213 @@
...
+
+In some cases, misconfigurations or unknown bugs can cause the applied configuration
+to fail repeatedly after each reboot. Because the daemon re-enters the reconciliation
+loop on startup and sees the same desired state, it triggers another reboot, creating
```
SchSeba: sees the same current state != desired state

doc/design/node-reboot-loop-limit.md
```diff
@@ -0,0 +1,213 @@
...
+an infinite reboot loop. This makes the node unavailable and requires manual
+intervention to recover.
```
SchSeba: add: this is even more critical for single nodes where removing the policy is even challenger as the API server is down while rebooting the node

doc/design/node-reboot-loop-limit.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+The file is lazily initialized: it is created on the first write (when incrementing
+or resetting the counter). Reads return a zero count when the file does not exist.
+
+The `generation` field is updated whenever the counter is written — both
+`incrementRebootCounter` and `resetRebootCounter` write the current generation. When
+a new generation arrives, the first write replaces the old generation in the file.
+
+The `bootID` field stores the value of `/proc/sys/kernel/random/boot_id` at the time
```
SchSeba: I think we can just increate the number just before we exec the reboot command after that gets called I don't think we are going to hit again that step before we reboot the node

pkg/host/internal/reboot/reboot.go
```diff
@@ -0,0 +1,106 @@
+/*
```
SchSeba: maybe change the name to reboot_tracker? reboot is a bit misleading

pkg/daemon/reboot_limit_test.go
```diff
@@ -0,0 +1,190 @@
+package daemon
```
SchSeba: please put the test inside the same already existing daemon_test package. 
also check the file uses the daemon_test package not the daemon package to have better isolation in the test

pkg/daemon/daemon.go
```diff
@@ -405,7 +405,31 @@ func (dn *NodeReconciler) apply(ctx context.Context, desiredNodeState *sriovnetw
 	}
 
 	if reqReboot {
+		rebootsCount, err := dn.readRebootCountFromDisk(desiredNodeState.Generation)
+		if err != nil {
+			reqLogger.Error(err, "failed to read reboot count from disk")
+			return ctrl.Result{}, err
+		}
+		if rebootsCount >= consts.MaxRebootsPerGeneration {
+			reqLogger.Info("maximum reboot retries reached, setting sync status to failed")
```
SchSeba, line 414: lets also send an event in `dn.eventRecorder.SendEvent(ctx, "RebootNode", "Reboot node has been initiated")`

pkg/daemon/daemon.go
```diff
@@ -455,12 +479,64 @@ func (dn *NodeReconciler) apply(ctx context.Context, desiredNodeState *sriovnetw
...
+		return ctrl.Result{}, err
+	}
+
 	// update the lastAppliedGeneration
 	dn.lastAppliedGeneration = desiredNodeState.Generation
 
 	return ctrl.Result{RequeueAfter: consts.DaemonRequeueTime}, nil
 }
 
+func (dn *NodeReconciler) readRebootCountFromDisk(generation int64) (int, error) {
```
SchSeba: this is an implementation detailed if you ask me it should go into the host pacakge

pkg/daemon/daemon.go
```diff
@@ -455,12 +479,64 @@ func (dn *NodeReconciler) apply(ctx context.Context, desiredNodeState *sriovnetw
...
+	if tracker == nil {
+		return 0, nil
+	}
+	if tracker.Generation != generation {
+		return 0, nil
+	}
+	return tracker.RebootCount, nil
+}
+
+func (dn *NodeReconciler) incrementRebootCounter(generation int64) (bool, error) {
```
SchSeba: same this function needs to be in the host package

pkg/daemon/daemon.go
```diff
@@ -455,12 +479,64 @@ func (dn *NodeReconciler) apply(ctx context.Context, desiredNodeState *sriovnetw
...
+			Generation:  generation,
+			RebootCount: 0,
+		}
+	}
+	tracker.RebootCount++
+	tracker.BootID = bootID
+	return true, dn.hostHelpers.WriteRebootTracker(tracker)
+}
+
+func (dn *NodeReconciler) resetRebootCounter(generation int64) error {
```
SchSeba: same :)

SchSeba (commented):
nice work! 

and thanks for the contribution!

I left some small comments, but the general one is that I think we should remove the boot_id.

when we reboot we use `stdOut, StdErr, err := dn.hostHelpers.RunCommand("systemd-run", "--unit", "sriov-network-config-daemon-reboot",
		"--description", "sriov-network-config-daemon reboot node", "/bin/sh", "-c", "systemctl stop kubelet.service; reboot")
`

this is going to imidiate reboot the node I don't think we are going to have time for another reconcile loop after running that command.

in any of your tests did you find a race where the number increase? (another reconcile loop manage to finish?)

also please use the RDMA functional test to check that we write the file and the number os what you expect so we also have coverage in the functional tests

SchSeba:
@coderabbitai review

doc/design/node-reboot-loop-limit.md
yahlifried: Done

doc/design/node-reboot-loop-limit.md
yahlifried: Done

doc/design/node-reboot-loop-limit.md
yahlifried: Done

doc/design/node-reboot-loop-limit.md
yahlifried: Done

pkg/daemon/daemon.go
yahlifried: Done

pkg/host/internal/reboot/reboot.go
yahlifried: Done

pkg/daemon/daemon.go
yahlifried: Done

pkg/daemon/daemon.go
yahlifried: Done

pkg/daemon/daemon.go
yahlifried: Done

pkg/daemon/reboot_limit_test.go
yahlifried: Done

pkg/consts/constants.go
```diff
@@ -123,7 +124,7 @@ const (
 	SysBusPciDrivers      = SysBus + "/pci/drivers"
 	SysBusPciDriversProbe = SysBus + "/pci/drivers_probe"
 	SysClassNet           = "/sys/class/net"
-	ProcKernelCmdLine     = "/proc/cmdline"
+	ProcKernelCmdLine = "/proc/cmdline"
```
SchSeba: please run `go fmt ./...`

pkg/daemon/export_test.go
```diff
@@ -0,0 +1,3 @@
+package daemon
+
+var ApplyForTest = (*NodeReconciler).apply
```
SchSeba: why we use this?

pkg/daemon/reboot_limit_test.go
```diff
@@ -0,0 +1,165 @@
+package daemon_test
```
SchSeba: I think we can add the tests into the `daemon_test.go` there is any special reason to split the tests into multiple files?

SchSeba (approved):
LGTM just please fix the fmt issue

pkg/daemon/reboot_limit_test.go
yahlifried: Done

pkg/consts/constants.go
yahlifried: Done

pkg/daemon/export_test.go
yahlifried: in order to export private `apply` for tests

SchSeba: we have a full reconciler running + mocks so you can driver via changes in the mocks function to the state this is a much realistict test to do then just run the apply function.

check if you can change the test so it will just use the reconciler.
and you check that mocks to reboot + write the files as you expect

yahlifried: Done

pkg/daemon/daemon_test.go
```diff
@@ -774,6 +779,186 @@ var _ = Describe("Daemon CheckSystemdStatus", func() {
 	})
 })
 
+var _ = Describe("Reboot Limit", func() {
+	var (
+		rebootMockCtrl   *gomock.Controller
```
SchSeba: please re-use the vars we have in the top level.

pkg/daemon/daemon_test.go
```diff
@@ -774,6 +779,186 @@ var _ = Describe("Daemon CheckSystemdStatus", func() {
...
+
+	Context("apply with reqReboot=true", func() {
+		It("should proceed to reboot when count is below the limit", func() {
+			mockPlugin.EXPECT().OnNodeStateChange(gomock.Any()).Return(true, true, nil)
+			rebootHostHelper.EXPECT().GetRebootCount(int64(1)).Return(2, nil)
+			rebootHostHelper.EXPECT().IncrementRebootCounter(int64(1)).Return(nil)
+			rebootHostHelper.EXPECT().Chroot(gomock.Any()).Return(func() error { return nil }, nil)
+			rebootHostHelper.EXPECT().RunCommand("systemd-run", gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return("", "", nil)
+
+			result, err := reconciler.Reconcile(ctx, req)
```
SchSeba: do not call the function directly just start a TEST_ENV K8s like we do on the other tests and apply the changes via the regular K8s API

SchSeba (commented):
please check the daemon_tests again.

I was pointing to use the k8s TEST_ENV and not calling the reconcile.

check how we do it in `Describe("Daemon Controller"` and you can replicate the same I think

pkg/daemon/daemon_test.go
yahlifried: Done

pkg/daemon/daemon_test.go
yahlifried: Done

SchSeba (approved):
Great work!

SchSeba:
please fix the fmt issue so we have a clean CI

PR #1089, opened 2026-05-31
Implement persistent logging for sriov-config-daemon based CNF-24281

gavrielg1:
Implement persistent logging for sriov-config-daemon based [CNF-24281](https://redhat.atlassian.net/browse/CNF-24281)

## Summary by CodeRabbit

* **New Features**
  * Added configurable persistent host-file logging for the SR-IOV Network Config Daemon.
  * Supports enabling/disabling logs, custom paths under `/var/log`, rotation limits, retention, and compression.
  * Logs remain available across daemon restarts, with safe default settings when unspecified.
  * Added validation for configuration values and unsafe or invalid log paths.

* **Documentation**
  * Documented `logConfig` settings, defaults, constraints, and configuration examples.

### DIFF
--- a/cmd/sriov-network-config-daemon/start.go
+++ b/cmd/sriov-network-config-daemon/start.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strings"
 
 	ocpconfigapi "github.com/openshift/api/config/v1"
@@ -46,6 +47,7 @@ import (
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/helper"
 	snolog "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/log"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/platform"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
 )
 
@@ -196,6 +198,46 @@ func initLogLevel(defaultConfig *sriovnetworkv1.SriovOperatorConfig) error {
 	return nil
 }
 
+func initLogFile(operatorConfig *sriovnetworkv1.SriovOperatorConfig) {
+	fnLogger := log.Log.WithName("initLogFile")
+
+	cfg := vars.LogCfg
+
+	lc := operatorConfig.Spec.LogConfig
+	if lc != nil {
+		if lc.Enabled != nil && !*lc.Enabled {
+			fnLogger.V(1).Info("persistent file logging disabled via LogConfig")
+			cfg.Enabled = false
+			vars.LogCfg = cfg
+			return
+		}
+		if lc.MaxSizeMB != nil {
+			cfg.MaxSizeMB = *lc.MaxSizeMB
+		}
+		if lc.MaxFiles != nil {
+			cfg.MaxFiles = *lc.MaxFiles
+		}
+		if lc.MaxAgeDays != nil {
+			cfg.MaxAgeDays = *lc.MaxAgeDays
+		}
+		if lc.Compress != nil {
+			cfg.Compress = *lc.Compress
+		}
+		if lc.HostPath != nil && *lc.HostPath != "" {
+			cfg.HostPath = *lc.HostPath
+		}
+	}
+	vars.LogCfg = cfg
+
+	logFilePath := utils.GetHostExtensionPath(filepath.Join(cfg.HostPath, "config-daemon.log"))
+	if err := snolog.InitLogWithFile(logFilePath, cfg.MaxSizeMB, cfg.MaxFiles, cfg.MaxAgeDays, cfg.Compress); err != nil {
+		fnLogger.Error(err, "failed to initialize file logging, continuing with console only")
+		return
+	}
+	fnLogger.Info("persistent file logging enabled (async-buffered, entries are flushed before chroot windows)",
+		"path", logFilePath, "maxSizeMB", cfg.MaxSizeMB, "maxFiles", cfg.MaxFiles, "maxAgeDays", cfg.MaxAgeDays)
+}
+
 func runStartCmd(cmd *cobra.Command, args []string) error {
 	setupLog := log.Log.WithName("sriov-network-config-daemon")
 	stopSignalCh := ctrl.SetupSignalHandler()
@@ -288,6 +330,10 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
 		setupLog.Error(err, "failed to initialize log level")
 		return err
 	}
+
+	// init persistent file logging
+	initLogFile(operatorConfig)
+	defer snolog.CloseFileLogger()
 
 	// init disable drain
 	vars.DisableDrain = operatorConfig.Spec.DisableDrain
--- a/pkg/daemon/config.go
+++ b/pkg/daemon/config.go
@@ -2,7 +2,9 @@ package daemon
 
 import (
 	"context"
+	"path/filepath"
 
+	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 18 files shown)


pkg/log/log.go
```diff
@@ -71,3 +160,386 @@ func zapToOperatorLevel(zapLevel zapcore.Level) int {
...
+}
+
+// safWrite defers the write until we are outside any chroot window, then calls
+// lj.Write which may transparently rotate the log file.
+func (fw *fileWriter) safWrite(data []byte) {
+	if vars.InChroot {
+		waitStart := time.Now()
+		for vars.InChroot && !fw.closed.Load() {
+			time.Sleep(chrootPollInterval)
+		}
```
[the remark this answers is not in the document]
gavrielg1: resolved

SchSeba (commented):
@coderabbitai review

cmd/sriov-network-config-daemon/start.go
```diff
@@ -196,6 +198,46 @@ func initLogLevel(defaultConfig *sriovnetworkv1.SriovOperatorConfig) error {
...
+		if lc.MaxFiles != nil {
+			cfg.MaxFiles = *lc.MaxFiles
+		}
+		if lc.MaxAgeDays != nil {
+			cfg.MaxAgeDays = *lc.MaxAgeDays
+		}
+		if lc.Compress != nil {
+			cfg.Compress = *lc.Compress
+		}
+		if lc.HostPath != nil && *lc.HostPath != "" {
```
SchSeba, line 226: we need to add avalidation here to be sure the user doesn't try something like `/` or `../`

cmd/sriov-network-config-daemon/start.go
```diff
@@ -196,6 +198,46 @@ func initLogLevel(defaultConfig *sriovnetworkv1.SriovOperatorConfig) error {
...
+			cfg.HostPath = *lc.HostPath
+		}
+	}
+	vars.LogCfg = cfg
+
+	logFilePath := utils.GetHostExtensionPath(filepath.Join(cfg.HostPath, "config-daemon.log"))
+	if err := snolog.InitLogWithFile(logFilePath, cfg.MaxSizeMB, cfg.MaxFiles, cfg.MaxAgeDays, cfg.Compress); err != nil {
```
SchSeba, line 233: we have two options here that are better as we do `vars.logCfg` the function can just use it to read the configuration.
I think the same for the all validation in this function it can be done in the initLogWithFile if you ask me

cmd/sriov-network-config-daemon/start.go
```diff
@@ -289,6 +331,10 @@ func runStartCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// init persistent file logging
+	initLogFile(operatorConfig)
```
SchSeba, line 335: I think we should just call `snolog.InitLogWithFile` here and that function should handle all the need verifications and validation

pkg/daemon/config.go
```diff
@@ -63,3 +74,58 @@ func (oc *OperatorConfigNodeReconcile) SetupWithManager(mgr ctrl.Manager) error
...
+	}
+
+	oc.latestLogCfg = newCfg
+	reqLogger.Info("persistent file logging reconfigured (async-buffered, entries are flushed before chroot windows)",
+		"path", logFilePath, "maxSizeMB", newCfg.MaxSizeMB, "maxFiles", newCfg.MaxFiles)
+}
+
+// effectiveLogCfg derives a LogFileSettings from an operator LogConfig pointer,
+// applying defaults for any field that is nil or zero.
+func effectiveLogCfg(lc *sriovnetworkv1.LogConfig) vars.LogFileSettings {
```
SchSeba, line 107: I think this can go to helpers in the API where for the sriovOperatorConfig we have a function of GetEffectiveLogsConfig or something like that

pkg/daemon/config.go
```diff
@@ -63,3 +74,58 @@ func (oc *OperatorConfigNodeReconcile) SetupWithManager(mgr ctrl.Manager) error
 		For(&sriovnetworkv1.SriovOperatorConfig{}).
 		Complete(oc)
 }
+
+// reconcileLogConfig re-initializes or closes the file logger.
+func (oc *OperatorConfigNodeReconcile) reconcileLogConfig(reqLogger logr.Logger, lc *sriovnetworkv1.LogConfig) {
+	newCfg := effectiveLogCfg(lc)
+	if newCfg == oc.latestLogCfg {
```
SchSeba, line 81: does struct equal is going to work here just like that?

pkg/daemon/config.go
```diff
@@ -63,3 +74,58 @@ func (oc *OperatorConfigNodeReconcile) SetupWithManager(mgr ctrl.Manager) error
...
+	vars.LogCfg = newCfg
+
+	if !newCfg.Enabled {
+		oc.latestLogCfg = newCfg
+		reqLogger.V(1).Info("persistent file logging disabled via LogConfig update")
+		snolog.CloseFileLogger()
+		return
+	}
+
+	logFilePath := utils.GetHostExtensionPath(filepath.Join(newCfg.HostPath, "config-daemon.log"))
```
SchSeba, line 94: here I think we need to clean up if the location of the file change? or the size/number is different does the package knows how to handle that and clean up the extra ones?

for example before I had 5 files and now I request 2 it should clean the old 3 files

pkg/daemon/config.go
```diff
@@ -63,3 +74,58 @@ func (oc *OperatorConfigNodeReconcile) SetupWithManager(mgr ctrl.Manager) error
...
+	if err := snolog.InitLogWithFile(logFilePath, newCfg.MaxSizeMB, newCfg.MaxFiles, newCfg.MaxAgeDays, newCfg.Compress); err != nil {
```
SchSeba, line 95: this reconcile runs in a go routine and we have the main reconcile loop on node state we need to be sure there is no race between when we reset the log configuration and other go routines trying to write logs

pkg/daemon/config.go
```diff
@@ -11,18 +13,24 @@ import (
...
 	snolog "github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/log"
+	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/utils"
 	"github.com/k8snetworkplumbingwg/sriov-network-operator/pkg/vars"
 )
 
 // OperatorConfigNodeReconcile represents the reconcile struct for the OperatorConfig.
 type OperatorConfigNodeReconcile struct {
 	client             client.Client
 	latestFeatureGates map[string]bool
+	latestLogCfg       vars.LogFileSettings
```
SchSeba, line 24: I think we don't need this we can just use the vars one no?

pkg/daemon/config_logcfg_test.go
```diff
@@ -0,0 +1,93 @@
+package daemon
```
SchSeba, line 1: please add the tests to the existing config_test file and use the ginkgo framework

SchSeba (commented):
Thanks for the PR!

I left some initial comments. I still need to make a deeper review on the log package changes itself

pkg/webhook/validate.go
```diff
@@ -53,9 +53,24 @@ func validateSriovOperatorConfig(cr *sriovnetworkv1.SriovOperatorConfig, operati
...
 	return true, warnings, nil
 }
 
+// validateSriovOperatorConfigLogConfig validates Spec.LogConfig (incl. HostPath).
+func validateSriovOperatorConfigLogConfig(cr *sriovnetworkv1.SriovOperatorConfig) error {
+	if cr.Spec.LogConfig == nil {
+		return nil
+	}
+	if _, err := cr.GetEffectiveLogConfig(); err != nil {
+		return fmt.Errorf("invalid LogConfig: %w", err)
```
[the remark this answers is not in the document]
gavrielg1: those bounds are already enforced at the API server (OpenAPI min/max + CEL), so a webhook re-check adds cost and no new safety.

Invalid maxSizeMB / maxFiles / maxAgeDays never enter etcd, never hit the webhook, and never trigger reconcile.

PR #1090, opened 2026-06-01
Fix drain escalation race that could reboot node without full drain

SchSeba:
A race condition existed where a node could be rebooted with production workloads still running. The scenario:

1. Daemon requests Drain_Required (partial drain - only SR-IOV pods evicted)
2. Drain controller performs partial drain and sets DrainComplete
3. Meanwhile, a new policy requiring reboot (e.g. RDMA mode change) arrives
4. Daemon sees DrainComplete and proceeds to reboot without a full drain

The root cause is that the daemon could not distinguish whether the completed drain was sufficient for its current (escalated) requirements.

Fix: introduce a new annotation `sriovnetwork.openshift.io/drain-action` on SriovNetworkNodeState, owned by the drain controller, that records the drain type actually performed. The daemon verifies this matches its current need before proceeding to apply configuration or reboot.

The drain controller:
- Sets drain-action when starting a drain (atomically with current-state)
- Re-reads desired-state after drain completes to detect mid-drain escalation
- Resets to Draining and re-drains if desired-state escalated beyond drain-action
- Clears drain-action when returning to Idle

The daemon:
- Checks drain-action satisfies its requirement at DrainComplete
- Escalates desired-state during Draining if reboot becomes needed
- Never proceeds to apply/reboot if drain-action is insufficient

Also adds DrainStateAnnotationPredicate watch on desired-state changes and AnnotateObjectMultiple utility to batch annotation writes in a single API call.

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -1010,6 +1010,27 @@ func OwnerRefToString(cr client.Object) string {
 	return cr.GetObjectKind().GroupVersionKind().GroupKind().String() + "/" + cr.GetNamespace() + "/" + cr.GetName()
 }
 
+// DrainActionSatisfiesDesired returns true if the performed drain action
+// is sufficient for the desired drain type.
+// An empty performedAction means the drain controller hasn't set it yet (backward compat) — treated as satisfied.
+// A desiredAction of Idle means no drain is needed — any performed action satisfies this.
+// Reboot_Required (full drain) satisfies Drain_Required (partial drain).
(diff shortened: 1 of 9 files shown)


SchSeba:
Hi @zeeke @adrianchiris please check this is the other option we talk about in the community meeting yesterday https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1092

let me know which one you prefer please

SchSeba:
closing this one

PR #1010, opened 2026-01-11
Conditions/network crds

SchSeba:
## Summary by CodeRabbit

* **New Features**
  * Added shared condition primitives and embedded condition status across SR-IOV and OVS network CRDs.
  * Added `Ready` and `Age` printer columns and strengthened `status.conditions` validation.
* **Bug Fixes**
  * Controllers now consistently set `Ready` via server-side apply for provisioning errors, missing namespaces, and NAD ownership/name conflicts.
  * Finalizer cleanup treats “already gone” targets as success.
* **Tests**
  * Expanded readiness/reason/`observedGeneration` coverage across all network types and added status transition unit tests.
* **Documentation**
  * Reintroduced the Kubernetes conditions integration design document.
* **Chores**
  * Updated build header generation and refreshed boilerplate year.

PR #1091, opened 2026-06-01
Conditions/operator config v2

SchSeba:
## Summary by CodeRabbit

* **New Features**
  * `SriovOperatorConfig` now shows `Ready` and `Age` in `kubectl get` output via CRD printer columns.
  * Kubernetes-standard `status.conditions` are used consistently across the operator and related network CRDs, including operator-specific reason codes.
* **Bug Fixes**
  * `SriovOperatorConfig` now consistently updates the `Ready` condition after each reconcile, with correct reasons for sync failures and unsupported configurations.
  * `observedGeneration` updates properly after spec changes.
* **Documentation**
  * Updated API/design documentation to reflect the `conditions` model and `Ready` semantics.
* **Tests**
  * Expanded reconcile and failure-case tests for condition setting and error/reason messaging.


SchSeba:
Depends on: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1010
Closes: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/1011

api/v1/sriovoperatorconfig_types.go
```diff
@@ -72,6 +72,7 @@ type SriovOperatorConfigSpec struct {
 
 // SriovOperatorConfigStatus defines the observed state of SriovOperatorConfig
 type SriovOperatorConfigStatus struct {
+	NetworkStatus `json:",inline"`
```
rollandf: Do we want here a different struct? As the doc refers to `Conditions represent the latest available observations of the network's state`
Or maybe make `NetworkStatus` more generic to fit all CRs?

rollandf (approved):
LGTM. Nit about the field doc

api/v1/sriovoperatorconfig_types.go
SchSeba: yep you are right I changed this one to be generic thanks!

api/v1/conditions.go
```diff
@@ -28,18 +28,25 @@ const (
...
+	ReasonOperatorConfigSyncFailed = "OperatorConfigSyncFailed"
+	ReasonUnsupportedConfiguration = "UnsupportedConfiguration"
 )
 
-// NetworkStatus defines the common observed state for network-type CRDs
-type NetworkStatus struct {
-	// Conditions represent the latest available observations of the network's state
+// ConditionStatus defines the common observed state for SR-IOV operator CRDs
+type ConditionStatus struct {
+	// Conditions represent the latest available observations of the resource's state
```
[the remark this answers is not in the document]
SchSeba: we don't need this we still didn't make a release for the project with this new capability

controllers/sriovoperatorconfig_controller.go
```diff
@@ -52,17 +53,54 @@ import (
...
+
 // SriovOperatorConfigReconciler reconciles a SriovOperatorConfig object
 type SriovOperatorConfigReconciler struct {
 	client.Client
 	Scheme            *runtime.Scheme
 	Orchestrator      orchestrator.Interface
 	FeatureGate       featuregate.FeatureGate
+	StatusPatcher     status.Interface
 	UncachedAPIReader client.Reader
+	renderManifestFn  renderManifestFunc
```
adrianchiris: could you add a comment that the fields below are used for testing.

controllers/helper.go
```diff
@@ -207,16 +212,19 @@ func syncPluginDaemonObjs(ctx context.Context,
 	return nil
 }
 
-func syncDsObject(ctx context.Context, client k8sclient.Client, scheme *runtime.Scheme, dc *sriovnetworkv1.SriovOperatorConfig, obj *uns.Unstructured) error {
+func syncDsObject(ctx context.Context, client k8sclient.Client, scheme *runtime.Scheme, dc *sriovnetworkv1.SriovOperatorConfig, obj *uns.Unstructured, applyFn applyManifestFunc) error {
 	logger := log.Log.WithName("syncDsObject")
 	kind := obj.GetKind()
 	logger.V(1).Info("Start to sync Objects", "Kind", kind)
+	if applyFn == nil {
```
adrianchiris: is there a case where this is nil ?
from what i see you always pass the r.applyManifest func

controllers/helper.go
```diff
@@ -236,12 +244,16 @@ func syncDsObject(ctx context.Context, client k8sclient.Client, scheme *runtime.
...
 
-// renderDsForCR returns a busybox pod with the same name/namespace as the cr
-func renderDsForCR(path string, data *render.RenderData) ([]*uns.Unstructured, error) {
+// renderDsForCR renders daemon objects using the provided renderer when one is available.
+func renderDsForCR(path string, data *render.RenderData, renderFn renderManifestFunc) ([]*uns.Unstructured, error) {
 	logger := log.Log.WithName("renderDsForCR")
 	logger.V(1).Info("Start to render objects")
 
-	objs, err := render.RenderDir(path, data)
+	if renderFn == nil {
```
adrianchiris: same as above but with r.renderManifests func

controllers/sriovoperatorconfig_controller.go
```diff
@@ -212,8 +260,9 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 	// For Openshift we need to create the systemd files using a machine config
 	if r.Orchestrator.ClusterType() == consts.ClusterTypeOpenshift {
 		// TODO: add support for hypershift as today there is no MCO on hypershift clusters
-		if r.Orchestrator.Flavor() == consts.ClusterFlavorHypershift {
```
adrianchiris, line 215: i guess we never ran on hypershift until now :)

controllers/sriovoperatorconfig_controller.go
```diff
@@ -225,6 +274,31 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 	return reconcile.Result{RequeueAfter: consts.ResyncPeriod}, nil
 }
 
+// applyReadyCondition updates the shared Ready condition for SriovOperatorConfig
+// using the shared SSA status patcher.
+func (r *SriovOperatorConfigReconciler) applyReadyCondition(ctx context.Context, config *sriovnetworkv1.SriovOperatorConfig, reconcileErr error) error {
+	if r.StatusPatcher == nil || config == nil {
```
adrianchiris: do we need this check ?

controllers/sriovoperatorconfig_controller.go
```diff
@@ -163,6 +203,16 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 		return r.handleSriovOperatorConfigDeletion(ctx, defaultConfig, logger)
```
adrianchiris: do we want to update conditions on the deletion flow ? 
e.g in case of error we dont remove the finalizer so conditions can be updated

controllers/sriovoperatorconfig_controller.go
```diff
@@ -163,6 +203,16 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
...
 	}
 
+	defer func() {
```
adrianchiris, line 200: related to the above comment, if we move this defer above the deletion handling, then we can cover the delete case as well.

the only change is to ignore "object not found" errors when we try to patch (e.g in case the object was deleted.)

adrianchiris (approved):

PR #1102, opened 2026-07-06
Move child rbac to helm

### LINKED ISSUE #1100: Move child-component RBAC resources from runtime creation to Helm chart
The operator currently creates ServiceAccounts, ClusterRoles, ClusterRoleBindings, Roles, and RoleBindings for its child components at runtime by
  rendering Go templates from `bindata/manifests/` and applying them via `ApplyObject`. This requires the operator's own ClusterRole to include `create`
  on `clusterroles` and `clusterrolebindings` (without `resourceNames` restriction), plus `get`, `create`, `update` on `serviceaccounts`.

  These child-component RBAC resources are static — their definitions don't change based on user-provided CR fields. They could instead be defined in the
  Helm chart and created at install time, like the operator's own ServiceAccount and ClusterRole already are.

  ## What is created at runtime today

  From `bindata/manifests/`:

  | Component | File | Resources |
  |---|---|---|
  | network-resources-injector | `webhook/002-rbac.yaml` | SA `network-resources-injector-sa`, ClusterRole `network-resources-injector`,
  ClusterRoleBinding `network-resources-injector-role-binding` |
  | operator-webhook | `operator-webhook/002-rbac.yaml` | SA `operator-webhook-sa`, ClusterRole `operator-webhook`, ClusterRoleBinding
  `operator-webhook-role-binding` |
  | metrics-exporter | `metrics-exporter/metrics-rbac.yaml` | SA `metrics-exporter-sa`, Role `metrics-exporter-role`, RoleBinding `metrics-exporter-rb`,
  ClusterRole `sriov-metrics-kube-rbac-role`, ClusterRoleBinding `sriov-metrics-kube-rbac-rolebinding` |
  | device-plugin | `plugins/002-rbac.yaml` | SA `sriov-device-plugin`, Role `sriov-plugin`, RoleBinding `sriov-device-plugin`, Role
  `sriov-device-plugin-pod-access`, RoleBinding `sriov-device-plugin-pod-access` |

  These are rendered via `pkg/render/render.go` and applied in `controllers/sriovoperatorconfig_controller.go` (`syncWebhookObjs`, `syncMetricsExporter`,
  `syncPluginDaemonObjs`).

  ## Proposed change

  1. Add Helm chart templates for the ServiceAccounts, ClusterRoles, ClusterRoleBindings, Roles, and RoleBindings listed above (similar to how
  `templates/serviceaccount.yaml` already defines the operator and config-daemon SAs).
  2. Remove the RBAC manifests from `bindata/manifests/` (keep the non-RBAC manifests like Deployments and DaemonSets).
  3. Update the controller code to skip applying the RBAC objects that are now managed by Helm.
  4. Remove the `create` verb on `clusterroles`/`clusterrolebindings` from the operator's ClusterRole (lines 37-39 of `deploy/clusterrole.yaml`), and
  narrow `serviceaccounts` verbs to only `get`.

  ## Why

  The operator doesn't need to manage these RBAC resources itself — their content is static and known at build time. Moving them to the Helm chart reduces
  the operator's own RBAC footprint, specifically eliminating the unrestricted `create` on `clusterroles` and `clusterrolebindings`. The install tooling
  (Helm) already has the necessary privileges to create these resources.

  Related: #1097 flagged the operator's ClusterRole permissions as overbroad — this change would address part of that.

  ## Considerations

  - The `{{.Namespace}}` template variable in the current bindata manifests maps directly to `{{ .Release.Namespace }}` in Helm.
  - Lifecycle: the operator currently skips setting owner references on cluster-scoped resources (ClusterRoles, ClusterRoleBindings), so they already
  survive CR deletion. Helm-managed resources follow `helm uninstall` lifecycle, which is the same behavior.
  - Components that can be toggled (webhooks, metrics-exporter) may need the Helm templates gated by the corresponding values (e.g., `{{- if
  .Values.operator.enableWebhooks }}`). The controller should still manage the non-RBAC resources (Deployments, etc.) for toggling at runtime via
  `SriovOperatorConfig`.

yahlifried:
this commit moves child objects RBAC from the operator's runtime creation to helm, and minimize some RBAC allowed actions.

**When upgrading from a previous version via Helm, the first upgrade requires helm upgrade --take-ownership (Helm 3.17+) to adopt the child RBAC resources that were previously created by the operator at runtime. This flag is only needed once and is safe to always include.**

fixes #1100 

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened RBAC permissions across the operator and components by removing unnecessary cluster-wide privileges and restricting verbs/resources to the minimum required.
  * Updated the Helm deployment to consistently create the needed service accounts plus “child” RBAC roles and bindings.
  * Adjusted conformance tests to align with the revised RBAC behavior when related features are disabled.


e0ne:
@yahlifried will it work for operator upgrade?

@SchSeba  please confirm that it doesn't affect OpenShift deployments

deploy/child_role.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
rollandf, line 2: Roles are namespaced awaer

deploy/child_service_account.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: ServiceAccount
```
rollandf, line 2: namespace?

deploy/clusterrole.yaml
```diff
@@ -8,31 +8,34 @@ rules:
...
   resources: ["pods/eviction"]
   verbs: ["create"]
 - apiGroups: ["apps"]
   resources: ["daemonsets"]
   verbs: ["get"]
 - apiGroups: [""]
-  resources: ["namespaces", "serviceaccounts"]
-  verbs: ["*"]
+  resources: ["namespaces"]
+  verbs: ["get", "list", "watch"]
```
rollandf, line 20: Do we need these on `namespaces`?

deploy/clusterrole.yaml
```diff
@@ -8,31 +8,34 @@ rules:
...
+- apiGroups: [""]
+  resources: ["serviceaccounts"]
+  verbs: ["get"]
```
rollandf: Still need `serviceaccounts`?

deploy/child_role.yaml
yahlifried: Added

deploy/child_service_account.yaml
yahlifried: Done

deploy/clusterrole.yaml
yahlifried: removed it

yahlifried:
> @yahlifried will it work for operator upgrade?
> 
> @SchSeba please confirm that it doesn't affect OpenShift deployments

yes, should work correctly, also for openshift with OLM (more changes will be added when synced to openshift)

deploy/clusterrole.yaml
yahlifried: yes, https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/controllers/generic_network_controller.go#L197
watch and list in line 302

e0ne (approved):
LGTM

hack/deploy-operator-helm.sh
```diff
@@ -26,6 +26,11 @@ HELM_VALUES_OPTS="\
 
 PATH=$PATH:${root}/bin
 make helm
+HELM_EXTRA_OPTS=""
+if [ "${HELM_MODE}" = "upgrade" ]; then
+  HELM_EXTRA_OPTS="--take-ownership"
```
SchSeba, line 31: check if we also need to update the documentation on how to install

SchSeba (commented):
can you please check the comment from coderabbit I think we can improve a bit more the RBAC permissions

hack/deploy-operator-helm.sh
yahlifried: done

deployment/sriov-network-operator-chart/templates/role.yaml
```diff
@@ -56,13 +56,23 @@ rules:
...
       - roles
       - rolebindings
     verbs:
-      - '*'
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - delete
```
SchSeba, line 75: I think this is wrong the all point was that the operator will not have permission to create them no?

deployment/sriov-network-operator-chart/templates/child-rbac.yaml
```diff
@@ -0,0 +1,236 @@
...
+kind: Role
+metadata:
+  name: sriov-device-plugin-pod-access
+  namespace: {{ .Release.Namespace }}
+  labels:
+  {{- include "sriov-network-operator.labels" . | nindent 4 }}
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list", "watch", "update", "patch"]
```
SchSeba, line 220: nit: lets be consistant we use lists instead of array on the others so lets follow the same in all the yamls please

SchSeba (commented):
lets also change the `child` to `operands` and the PR should be ready nice work!

deployment/sriov-network-operator-chart/templates/role.yaml
yahlifried: Done

deployment/sriov-network-operator-chart/templates/child-rbac.yaml
yahlifried: Done

deploy/role.yaml
```diff
@@ -3,136 +3,62 @@ kind: Role
...
-  resources:
-    - events
-  verbs:
-    - create
-    - patch
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list", "watch", "delete", "patch"]
+- apiGroups: ["sriovnetwork.openshift.io"]
+  resources: ["*", "sriovnetworknodestates"]
```
SchSeba, line 41: why we need the `*` here? the daemon should only access the sriovNetworkNodeState and only to watch list, patch, update not to delete

deploy/role.yaml
```diff
@@ -3,136 +3,62 @@ kind: Role
...
+  resources: ["infrastructures"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["coordination.k8s.io"]
+  resources: ["leases"]
+  verbs: ["*"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: sriov-network-config-daemon
```
SchSeba, line 35: I think the config-daemon should also be moved to the operands yamls not here

deploy/role.yaml
```diff
@@ -3,136 +3,62 @@ kind: Role
...
-  - list
-  - watch
-- apiGroups:
-  - 'coordination.k8s.io'
-  resources:
-  - 'leases'
-  verbs:
-  - '*'
+- apiGroups: [""]
+  resources: ["pods", "services", "endpoints", "persistentvolumeclaims", "events", "configmaps", "secrets"]
```
SchSeba, line 7: please check I think we can reduce the services here

deploy/role.yaml
```diff
@@ -3,136 +3,62 @@ kind: Role
...
+  verbs: ["*"]
+- apiGroups: ["apps"]
+  resources: ["deployments", "daemonsets", "replicasets", "statefulsets"]
```
SchSeba, line 10: same here

deploy/role.yaml
```diff
@@ -3,136 +3,62 @@ kind: Role
...
-  - get
-  - list
-  - watch
-- apiGroups:
-  - 'coordination.k8s.io'
-  resources:
-  - 'leases'
-  verbs:
-  - '*'
+- apiGroups: [""]
```
SchSeba, line 6: in general please review the file again, againts what the operator really does I think we can improve the RBAC here

deploy/role.yaml
```diff
@@ -3,136 +3,62 @@ kind: Role
...
+  resources: ["configmaps"]
+  verbs: ["get", "update"]
+- apiGroups: ["coordination.k8s.io"]
+  resources: ["leases"]
+  verbs: ["*"]
+- apiGroups: [""]
+  resources: ["events"]
+  verbs: ["create", "patch"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
```
SchSeba, line 57: this is only an operand lets move it to that yaml file

deployment/sriov-network-operator-chart/templates/clusterrole.yaml
```diff
@@ -49,6 +49,6 @@ rules:
   - apiGroups: [""]
```
SchSeba, line 49: lets move this to the operands yaml please

deployment/sriov-network-operator-chart/templates/clusterrole.yaml
```diff
@@ -10,31 +10,31 @@ rules:
     verbs: ["get", "list", "watch", "patch", "update"]
   - apiGroups: [""]
     resources: ["pods"]
-    verbs: ["*"]
+    verbs: ["get", "list", "watch", "delete"]
```
SchSeba, line 13: please check if we need the delete, I think if you have the pods/eviction that is what the operator uses

deployment/sriov-network-operator-chart/templates/clusterrole.yaml
```diff
@@ -10,31 +10,31 @@ rules:
...
   - apiGroups: [""]
     resources: ["pods/eviction"]
     verbs: ["create"]
   - apiGroups: ["apps"]
```
SchSeba, line 17: please check if we really need this one I don't think we do a get outside our namespace to daemonsets

deployment/sriov-network-operator-chart/templates/operands-serviceaccounts.yaml
```diff
@@ -0,0 +1,33 @@
+{{- if .Values.operator.admissionControllers.enabled }}
```
SchSeba, line 1: I don't think this is right, as after we deploy the user can select to deploy the resource injector or the operator webhook. so the SA must always exist

deployment/sriov-network-operator-chart/templates/operands-rbac.yaml
```diff
@@ -0,0 +1,229 @@
+{{- if .Values.operator.admissionControllers.enabled }}
```
SchSeba, line 1: same as the other comment.

also to make it easer for review lets have one file for every operand with it own clusterrole, clusterrolebinding, role and rolebinding please

deployment/sriov-network-operator-chart/templates/role.yaml
```diff
@@ -7,76 +7,31 @@ metadata:
...
-      - get
-      - list
-      - watch
-  - apiGroups:
-      - 'coordination.k8s.io'
-    resources:
-      - 'leases'
-    verbs:
-      - '*'
+  - apiGroups: [""]
```
SchSeba, line 10: check the comments from the deployment folder it should be the same

deployment/sriov-network-operator-chart/templates/rolebinding.yaml
```diff
@@ -29,3 +29,21 @@ roleRef:
   kind: Role
   name: sriov-network-config-daemon
   apiGroup: rbac.authorization.k8s.io
+{{- if .Values.operator.admissionControllers.enabled }}
```
SchSeba, line 32: same as the other comments related to admission

test/conformance/tests/test_sriov_operator.go
```diff
@@ -938,8 +937,6 @@ var _ = Describe("[sriov] operator", Ordered, func() {
 				}, 2*time.Minute, 10*time.Second).Should(BeZero())
 
 				assertObjectIsNotFound("network-resources-injector-service", &corev1.Service{})
-				assertObjectIsNotFound("network-resources-injector", &rbacv1.ClusterRole{})
```
SchSeba, line 941: I think we can leave this validations, for OCP it may contain a different name auto generated by OLM but we can add a label and search by label, let me know what do you think

SchSeba (commented):
nice! 
I left some more comments it's getting ready :)

PR #1107, opened 2026-07-15
Retry VF driver discovery before reporting no driver

almaslennikov:
## Summary

Retry VF driver discovery for a short bounded window before reporting a VF as driverless.

A VF can be briefly unbound during normal lifecycle operations such as driver rebinds. If config-daemon samples `/sys/bus/pci/devices/<vf>/driver` in that window, the current code publishes the VF with an empty driver. `NeedToUpdateSriov()` then treats that empty driver as drift, and the generic plugin can request an unnecessary node drain.

This change keeps the existing behavior for a persistently driverless VF, but avoids turning a single transient driver read miss into node-state drift.

## Changes

- Add bounded VF driver discovery retry in `getVfInfo()` path.
- Return as soon as a non-empty driver is observed.
- Preserve empty-driver reporting after retry exhaustion so real driverless VFs still trigger the existing update logic.
- Add unit coverage for immediate success, transient read failure recovery, retry exhaustion, and the downstream empty-driver `NeedToUpdateSriov()` behavior.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SR-IOV virtual function handling when a VF has no bound driver, ensuring correct update decisions.
  - Added retry-based VF driver discovery to reduce transient failures during device detection.
  - VF information collection is now parallelized for faster SR-IOV discovery.
  - Discovery now stops promptly on cancellation to improve responsiveness during shutdown.
- **Tests**
  - Expanded SR-IOV retry and cancellation coverage to validate success, transient recovery, retry exhaustion, and “driverless” behavior.

### DIFF
--- a/pkg/host/internal/sriov/sriov.go
+++ b/pkg/host/internal/sriov/sriov.go
@@ -32,11 +32,22 @@ const (
 	// sysfsWriteTimeout is the timeout for writing to sysfs files (e.g. sriov_numvfs).
 	// Kernel drivers can block indefinitely on these writes if the device is in a bad state.
 	sysfsWriteTimeout = 2 * time.Minute
+
+	// vfDriverReadAttempts and vfDriverReadInterval bound how long discovery
+	// re-reads a VF's bound driver before reporting it as driverless. VFs can
+	// be briefly unbound during normal lifecycle events such as driver rebinds.
+	vfDriverReadAttempts = 5
+	vfDriverReadInterval = 200 * time.Millisecond
 )
 
 type interfaceToConfigure struct {
 	Iface       sriovnetworkv1.Interface
 	IfaceStatus sriovnetworkv1.InterfaceExt
+}
+
+type vfInfoResult struct {
+	index int
+	vf    sriovnetworkv1.VirtualFunction
 }
 
 type sriov struct {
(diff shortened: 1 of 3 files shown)


almaslennikov:
Addressed the CodeRabbit findings and force-pushed amended commit `92294b33d`:

1. **Error handling:** `getVfDriverNameWithRetry` now returns `(string, error)` and wraps the final driver-read error with device/attempt context. The caller logs the contextual error while preserving existing empty-driver behavior after retry exhaustion.
2. **Context-aware retry wait:** the retry delay now uses `ctx.Done()` instead of an uninterruptible `time.Sleep`, and there is a unit test for canceled context behavior.
3. **Sequential retry cost:** VF driver reads are now retried concurrently per PF via `getVfDriverNames`, so a PF with many transiently/unintentionally driverless VFs waits for one bounded retry window instead of accumulating `N * retryWindow`.
4. **Tests:** updated the helper tests for the new error return and added coverage for persistent no-driver-without-error and context cancellation.

Verification rerun in Linux Docker:

```bash
go test ./pkg/host/internal/sriov/... ./api/v1/... -v
```

Result: `pkg/host/internal/sriov` 56/56 specs passed; `api/v1` passed.

```bash
golangci-lint v2.7.2 run ./pkg/host/internal/sriov/... ./api/v1/...
```

Result: `0 issues.`

almaslennikov:
@SchSeba @zeeke PTAL

pkg/host/internal/sriov/sriov.go
```diff
@@ -138,11 +150,64 @@ func (s *sriov) ResetSriovDevice(ifaceStatus sriovnetworkv1.InterfaceExt) error
...
+	drivers := make(map[string]string, len(vfAddrs))
+	for range vfAddrs {
+		result := <-results
+		if result.err != nil {
+			log.Log.Error(result.err, "getVfDriverNames(): unable to get VF driver", "device", result.vfAddr)
+		}
+		drivers[result.vfAddr] = result.driver
+	}
+	return drivers
+}
```
zeeke: I guess this kind of parallelism is needed because the retry mechanism might increase the discovery time a lot, but what about calling the entire getVfInfo(...) in a per vf go routine?
I think it would become more efficient and readable

e0ne (approved):
LGTM

almaslennikov:
Addressed the latest feedback and force-pushed amended commit `baf812596`:

1. **Parallelize full VF discovery:** replaced the driver-only fan-out with `getVfInfos`, which runs the full `getVfInfo(...)` path per VF in a goroutine and reassembles results in the original VF-list order. This keeps the bounded retry behavior while making the discovery path simpler and avoiding serial latency across VFs.
2. **Test assertion context:** added descriptive messages to the new Gomega assertions from the retry tests.

Verification rerun in Linux Docker:

```bash
go test ./pkg/host/internal/sriov/... ./api/v1/... -v
```

Result: `pkg/host/internal/sriov` 56/56 specs passed; `api/v1` passed.

```bash
golangci-lint v2.7.2 run ./pkg/host/internal/sriov/... ./api/v1/...
```

Result: `0 issues.`

pkg/host/internal/sriov/sriov.go
almaslennikov: @zeeke addressed

zeeke (approved):
LGTM

PR #1115, opened 2026-07-27
Disable metrics-bind-address by default and add Helm chart configuration support

schsebabot:
## Summary

This PR disables the metrics-bind-address by default in the sriov-network-operator and adds Helm chart configuration support to allow users to re-enable and customize the metrics endpoint when needed. Setting the default to `"0"` follows the controller-runtime convention for disabling metrics serving.

## Key Changes

- **`main.go`**: Changed the default value of the `--metrics-bind-address` flag from `":8080"` to `"0"`, disabling metrics serving by default.
- **`values.yaml`**: Added a new Helm value `operator.metricsBindAddress` defaulting to `"0"` (disabled).
- **Operator Deployment template** (`deployment/sriov-network-operator/templates/operator.yaml`): Wired the new Helm value through as the `--metrics-bind-address` flag in the operator container args, so users can override it (e.g., set to `":8080"` to re-enable metrics).
- **Kustomize deployment** (`config/manager/manager.yaml`): Updated the manager container args to explicitly pass `--metrics-bind-address=0`, keeping kustomize-based deployments consistent with the new default.

## Notable Implementation Decisions

- The default of `"0"` was chosen per the [controller-runtime documentation](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager#Options), which specifies that setting the metrics bind address to `"0"` disables metrics serving entirely.
- Users who need metrics can re-enable them by setting `operator.metricsBindAddress` to a valid bind address (e.g., `":8080"`) in their Helm values.
- Both Helm and kustomize deployment paths are updated to ensure consistent behavior regardless of the deployment method.

## Summary by CodeRabbit

* **New Features**
  * Added configurable metrics endpoint binding through deployment settings.
  * Metrics can be enabled by specifying a bind address, such as `:8080`.

* **Bug Fixes**
  * Metrics serving is now disabled by default, avoiding unintended endpoint exposure.

### DIFF
--- a/deployment/sriov-network-operator-chart/templates/operator.yaml
+++ b/deployment/sriov-network-operator-chart/templates/operator.yaml
@@ -43,6 +43,10 @@ spec:
           image: {{ .Values.images.operator }}
           command:
             - sriov-network-operator
+          {{- if .Values.operator.metricsBindAddress }}
+          args:
+            - --metrics-bind-address={{ .Values.operator.metricsBindAddress }}
+          {{- end }}
           securityContext:
             readOnlyRootFilesystem: true
             allowPrivilegeEscalation: false
--- a/deployment/sriov-network-operator-chart/values.yaml
+++ b/deployment/sriov-network-operator-chart/values.yaml
@@ -1,4 +1,8 @@
 operator:
+  # The address the operator metrics endpoint binds to.
+  # Leave empty to use the default (metrics disabled).
+  # Set to e.g. ":8080" to enable metrics on that address.
+  metricsBindAddress: ""
   tolerations:
     - key: "node-role.kubernetes.io/master"
       operator: "Exists"
--- a/main.go
+++ b/main.go
@@ -77,7 +77,7 @@ func main() {
 	var metricsAddr string
 	var probeAddr string
 
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metric endpoint binds to. Use \"0\" to disable metrics serving.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 
 	snolog.BindFlags(flag.CommandLine)


deploy/operator.yaml
```diff
@@ -49,6 +49,8 @@ spec:
           image: $SRIOV_NETWORK_OPERATOR_IMAGE
           command:
           - sriov-network-operator
+          args:
+          - --metrics-bind-address=0
```
SchSeba: we don't need this as this is the default value

deployment/sriov-network-operator-chart/templates/operator.yaml
```diff
@@ -43,6 +43,8 @@ spec:
           image: {{ .Values.images.operator }}
           command:
             - sriov-network-operator
+          args:
```
SchSeba, line 47: lets have this in a if state if the `.Values.operator.metricsBindAddress` is not empty

deployment/sriov-network-operator-chart/values.yaml
```diff
@@ -1,4 +1,8 @@
 operator:
+  # The address the operator metrics endpoint binds to.
+  # Set to "0" to disable metrics serving (default).
+  # Set to e.g. ":8080" to enable metrics on that address.
+  metricsBindAddress: "0"
```
SchSeba: lets have this as empty and update the comment

schsebabot: Agreed — since the Go default handles the disabled case, the Helm value should default to empty. I'll change to `metricsBindAddress: ""` and update the comment to explain that leaving it empty uses the code default (metrics disabled).

deploy/operator.yaml
schsebabot: Good point — since the default in `main.go` is now `"0"`, explicitly passing it here is redundant. I'll remove the `args` block from `deploy/operator.yaml`.

deployment/sriov-network-operator-chart/templates/operator.yaml
schsebabot: Makes sense — I'll wrap the `args` block in `{{- if .Values.operator.metricsBindAddress }}` so it only renders when the user explicitly sets a non-empty value.

SchSeba (approved):

SchSeba:
Hi @adrianchiris please take a look on this

adrianchiris (approved):

SchSeba:
The failing test is not related to this change.
there is a problem with a machine config that makes the master node to reboot in the middle of the tests I am investigating that.

merging this one

PR #1131, opened 2026-08-05
Make mstfwreset best-effort in Mellanox plugin — log errors but don't block apply

schsebabot:
This PR changes the `mstfwreset` error handling in the Mellanox plugin from fatal (blocking the apply flow) to best-effort (log and continue). Since the sriov-network-operator will trigger a node reboot when FW config changes are needed, a transient `mstfwreset` failure is non-fatal — the FW configuration will still be applied correctly after the reboot.

## Key Changes

- **`pkg/vendors/mellanox/mellanox.go`**: In `MlxResetFW()`, removed the `errs = append(errs, err)` call when `mstfwreset` fails. The error is now logged as a warning with the PCI address for context, but no longer propagates to the caller.
- **`pkg/vendors/mellanox/mellanox_test.go`**: Updated test cases for single-NIC and multi-NIC `mstfwreset` failure scenarios to expect `nil` error instead of an error return, reflecting the new best-effort behavior.

## Notable Implementation Decisions

- The log message was updated to explicitly communicate the best-effort intent: `"mstfwreset failed, continuing as best-effort since config will be applied after reboot"`, making it clear to operators that this is intentional behavior and not a silent failure.
- The `pciAddress` is included in the log output to help with debugging which device encountered the transient failure.
- `mstfwreset` is treated as an optimization (avoid requiring reboot for FW config activation). If it fails, the node reboot path remains the fallback, so blocking the entire apply flow on this transient failure would be strictly worse than proceeding.

## Summary by CodeRabbit

* **Bug Fixes**
  * Firmware reset operations now continue if a device reset fails, allowing remaining devices to be processed.
  * Reset failures are handled more gracefully during single- and multi-device operations, avoiding unnecessary interruption.
  * Virtual function reset failures continue to return an error, preserving existing error reporting behavior.

### DIFF
--- a/pkg/vendors/mellanox/mellanox.go
+++ b/pkg/vendors/mellanox/mellanox.go
@@ -171,8 +171,7 @@ func (m *mellanoxHelper) MlxResetFW(pciAddresses []string, mellanoxNicsStatus ma
 		// We have to ensure that pciutils is installed into the container image Dockerfile.sriov-network-config-daemon
 		_, stderr, err := m.utils.RunCommand("mstfwreset", cmdArgs...)
 		if err != nil {
-			log.Log.Error(err, "mellanox-plugin resetFW(): failed", "stderr", stderr)
-			errs = append(errs, err)
+			log.Log.Error(err, "mellanox-plugin resetFW(): mstfwreset failed, continuing as best-effort", "stderr", stderr, "pciAddress", pciAddress)
 		}
 	}
 
--- a/pkg/vendors/mellanox/mellanox_test.go
+++ b/pkg/vendors/mellanox/mellanox_test.go
@@ -120,7 +120,7 @@ var _ = Describe("SRIOV", func() {
 			Expect(err.Error()).To(ContainSubstring("failed to reset VFs on secondary port"))
 		})
 
-		It("should return error if firmware reset fails", func() {
+		It("should not return error if firmware reset fails (best-effort)", func() {
 			mellanoxNicsStatus["0000:d8:00."] = map[string]sriovnetworkv1.InterfaceExt{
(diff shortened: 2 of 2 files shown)


almaslennikov (approved):

zeeke (approved):

rollandf (approved):
LGTM

pkg/vendors/mellanox/mellanox.go
```diff
@@ -171,8 +171,7 @@ func (m *mellanoxHelper) MlxResetFW(pciAddresses []string, mellanoxNicsStatus ma
 		// We have to ensure that pciutils is installed into the container image Dockerfile.sriov-network-config-daemon
 		_, stderr, err := m.utils.RunCommand("mstfwreset", cmdArgs...)
 		if err != nil {
-			log.Log.Error(err, "mellanox-plugin resetFW(): failed", "stderr", stderr)
-			errs = append(errs, err)
+			log.Log.Error(err, "mellanox-plugin resetFW(): mstfwreset failed, continuing as best-effort since config will be applied after reboot", "stderr", stderr, "pciAddress", pciAddress)
```
SchSeba: remove `since config will be applied after reboot` from the log

SchSeba:
fix also the commit message

pkg/vendors/mellanox/mellanox.go
schsebabot: Good catch, removed that part. The log message now reads: `"mellanox-plugin resetFW(): mstfwreset failed, continuing as best-effort"` without the reboot rationale.

SchSeba (approved):

PR #1141, opened 2026-08-31
fix(daemon): do not mark a concurrent spec update as already applied

SchSeba:
updateSyncState copies live ObjectMeta onto the in-memory NodeState, including Generation, while Reconcile still holds the spec from Get(). Capturing generation at the start of apply() is too late: the InProgress status update already ran, so lastAppliedGeneration could jump to a newer spec that OnNodeStateChange never saw.

Requeue when generation changes during reconcile, record the generation fetched at Get() as lastAppliedGeneration, and treat spec vs status RDMA mismatch as drift so a reboot request is not skipped after a partial drain completes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented stale node configuration from being applied when its specification changes during reconciliation.
  - Added detection for mismatched requested and reported RDMA modes so updates are correctly triggered.
  - Avoided unnecessary RDMA updates when the requested and current modes already match.

- **Tests**
  - Added coverage for RDMA mode changes and matching-mode scenarios.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -240,6 +240,16 @@ func (dn *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		return ctrl.Result{}, err
 	}
 
+	// updateSyncState copies live ObjectMeta onto desiredNodeState. If a newer spec
+	// arrived during this reconcile, Generation no longer matches the Spec we hold.
+	// Requeue so the next loop fetches the latest spec instead of applying a stale one.
+	if desiredNodeState.GetGeneration() != latest {
+		reqLogger.Info("nodeState generation changed during reconcile, requeue",
+			"reconcile-generation", latest,
+			"latest-generation", desiredNodeState.GetGeneration())
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+
 	reqReboot, reqDrain, err := dn.checkOnNodeStateChange(desiredNodeState)
 	if err != nil {
 		return ctrl.Result{}, err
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -240,6 +240,16 @@ func (dn *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
 	}
 
+	// updateSyncState copies live ObjectMeta onto desiredNodeState. If a newer spec
+	// arrived during this reconcile, Generation no longer matches the Spec we hold.
+	// Requeue so the next loop fetches the latest spec instead of applying a stale one.
+	if desiredNodeState.GetGeneration() != latest {
+		reqLogger.Info("nodeState generation changed during reconcile, requeue",
+			"reconcile-generation", latest,
+			"latest-generation", desiredNodeState.GetGeneration())
+		return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
SchSeba: this is fine as this is the first time we update the CR in the reconcile flow. before that we don't need as it's just checks

zeeke: should we set RequeueAfter ?

```suggestion
	return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```

zeeke (commented):
only one minor comment

pkg/daemon/daemon.go
SchSeba: yeah you are right if not we will not reconcile again

zeeke (approved):

rollandf (approved):

PR #1151, opened 2026-09-11
feat: Integrate dra-driver-sriov as device-plugin alternative follow-up

wizhaoredhat:
Based on the great work @rollandf did in PR #1061.

Adds an optional Dynamic Resource Allocation (DRA) path for SR-IOV via [dra-driver-sriov](https://github.com/k8snetworkplumbingwg/dra-driver-sriov), controlled by `featureGates.dynamicResourceAllocation` on `SriovOperatorConfig`. When enabled, the operator deploys the DRA driver (instead of the device plugin) and syncs DRA CRs from `SriovNetworkNodePolicy`; when disabled, behavior stays on the existing device-plugin / ConfigMap path. Mode selection is **cluster-wide only** (no per-node mixed mode).

## Feature: DRA as device-plugin alternative

**Deployment & RBAC**
- Bindata under `bindata/manifests/dra-driver/`: base DeviceClass, ServiceAccount, ClusterRole/Role (pod access + OpenShift SCC), ClusterRoleBinding, DaemonSet with wait-for-config init container (`SRIOV_DRA_DRIVER_IMAGE` / `SRIOV_NETWORK_CONFIG_DAEMON_IMAGE`).
- Operator and Helm chart ClusterRoles extended for DRA driver APIs and `resource.k8s.io` DeviceClass management.
- Chart: `images.sriovDraDriver` value and operator env wiring; e2e/hack scripts pass `SRIOV_DRA_DRIVER_IMAGE`.

**CRDs & dependencies**
- Vendor `SriovResourcePolicy` and `DeviceAttributes` CRDs (chart + test util) aligned with `dra-driver-sriov`.
- Add `github.com/k8snetworkplumbingwg/dra-driver-sriov` module and register types in the manager scheme.

**Operator controllers**
- `SriovOperatorConfig`: sync DRA driver objects and skip the device-plugin DaemonSet when DRA is on; tear down DRA and restore the device plugin when off.
- `SriovNetworkNodePolicy`: when DRA is on, sync `DeviceAttributes` (per resource pool), per-node `SriovResourcePolicy` (`kubernetes.io/hostname` selector), and extended-resource `DeviceClass` objects; cleanup on policy or mode change. Device-plugin ConfigMap path unchanged when DRA is off.
- Helpers: `syncDRADriverObjs`, `cleanupDRADriverObjs`, shared DRA label/name helpers; DRA reconcile logic in `sriovnetworknodepolicy_dra.go`.

**Config daemon**
- Discover and coordinate with `sriov-dra-driver` pods alongside existing device-plugin logic (bounded restarts, wait-for-config annotation handling).

**Docs**
- `doc/design/DRA-integration.md` and updates in `doc/advanced-features.md`.

## Hardening & correctness (follow-ups in this PR)

**Provider mode switch**
- Sync the replacement DaemonSet and wait for rollout before removing the old provider, so a failed switch cannot leave neither provider present; wait only while the previous provider still exists so sync/cleanup keep progressing.
- Device-plugin cleanup removes RBAC as well as the DaemonSet.

**Ownership & cleanup**
- Adopt operator-managed DRA objects after managed-label drift; persist controller owner refs; refuse DeviceClass adoption without an ownership marker.
- Tag base DeviceClass and DRA cluster RBAC on sync; delete them on mode switch only when the management marker matches.
- Watch `SriovOperatorConfig` so DRA CRs are removed promptly when the feature gate is disabled (no ResyncPeriod wait).

**Validation & resource naming**
- Admission rejects colliding `resourceName`s after DRA normalization (always, not only when the gate is on); controller applies the same retained resource set when rendering DRA objects.
- Use typed `resource/v1` DeviceClass objects; fail DRA driver sync when the DeviceClass API is missing (policy sync still skips safely if unavailable).
- Omit empty DRA resource filters; skip NetFilter policies during `SriovResourcePolicy` conversion (filters cannot express `netFilter` yet).

**Runtime / platform**
- Tighten DRA driver RBAC (namespace-scoped pod patch for init; nodes are read-only at cluster scope).
- Complete DaemonSet runtime config (requests/limits; MULTUS mode, no NRI socket).
- Grant DRA driver SA `privileged` SCC use on OpenShift.
- Clear wait-for-config when a node has no interfaces so DRA driver pods are not stuck in Init.

**Tests & CI**
- Unit coverage for DRA sync/cleanup, helpers, and webhook collisions.
- Operator-level DRA conformance: feature-gate provider swap and policy-driven DRA CR sync/cleanup on virtual CI lanes (including OpenShift).
- OCP e2e: install DRA CRDs and set `CNI_BIN_PATH` so the DRA driver can become Ready.

## Requirements

- Kubernetes with DRA / `resource.k8s.io` (design targets 1.34+).
- DRA driver image via `images.sriovDraDriver` / `SRIOV_DRA_DRIVER_IMAGE` (chart default `:latest` during active development).

## Summary by CodeRabbit

* **New Features**
  * Added opt-in Kubernetes Dynamic Resource Allocation (DRA) support for SR-IOV devices.
  * Added DRA driver deployment, configuration, permissions, and Helm settings.
  * Added DeviceAttributes, SriovResourcePolicy, and DeviceClass resources for DRA allocation.
  * Legacy device-plugin allocation remains the default.

* **Bug Fixes**
  * Improved cleanup and synchronization when switching allocation modes.
  * Prevented DRA driver pods from remaining blocked when no interfaces are configured.
  * Improved validation for conflicting and out-of-scope resource names.

* **Documentation**
  * Documented DRA configuration, requirements, behavior, and examples.


SchSeba:
Hi @wizhaoredhat, please check the CI and the rabbitAI comments

bindata/manifests/dra-driver/003a-role-pod-access.yaml
```diff
@@ -0,0 +1,26 @@
...
+# Same pattern as sriov-device-plugin-pod-access in bindata/manifests/plugins/002-rbac.yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: sriov-dra-driver-pod-access
+  namespace: {{.Namespace}}
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list", "watch", "patch"]
```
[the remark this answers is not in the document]
wizhaoredhat: DRA intentionally matches the existing device-plugin wait-for-config RBAC. If we want to tighten this (to not give such permission to modify any other pods in the operator namespace) then we need to use something other than annotations.

bindata/manifests/dra-driver/004-daemonset.yaml
```diff
@@ -0,0 +1,159 @@
...
+        - name: KUBELET_PLUGINS_DIRECTORY_PATH
+          value: /var/lib/kubelet/plugins
+        - name: V
+          value: "5"
+        - name: NRI_PLUGIN_NAME
+          value: "dra-driver-sriov"
+        - name: NRI_PLUGIN_IDX
+          value: "00"
+        - name: HEALTHCHECK_PORT
+          value: "-1"
```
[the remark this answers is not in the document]
wizhaoredhat: The DRA driver DaemonSet uses hostNetwork: true (same as the SR-IOV device plugin). With HEALTHCHECK_PORT > 0, the driver’s gRPC health server listens in the node’s network namespace, not an isolated pod network. A fixed port (e.g. 51515) can conflict with other hostNetwork pods on that node (config daemon, device plugin, other infra) or with host services. This operator does not assign per-node health ports or use admission/policy to prevent overlap, so enabling health checks and kubelet probes would be fragile. The device-plugin DaemonSet also runs hostNetwork without gRPC/HTTP probes. Keeping HEALTHCHECK_PORT=-1 matches that operational model until there is a conflict-safe approach (non–hostNetwork health path, dynamic port + probe wiring, or shared convention across all hostNetwork SR-IOV components).

deployment/sriov-network-operator-chart/crds/sriovnetwork.k8snetworkplumbingwg.io_deviceattributes.yaml
```diff
@@ -0,0 +1,111 @@
...
+              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: DeviceAttributesSpec holds the attributes to apply to devices.
+            properties:
+              attributes:
+                additionalProperties:
+                  description: DeviceAttribute must have exactly one field set.
```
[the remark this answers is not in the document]
wizhaoredhat: This is part of the vendored in CRD from sriov-dra-driver. We should fix it there first. This is out of scope for this PR.

deployment/sriov-network-operator-chart/values.yaml
```diff
@@ -147,10 +147,18 @@ images:
   ovsCni: ghcr.io/k8snetworkplumbingwg/ovs-cni-plugin
   rdmaCni: ghcr.io/k8snetworkplumbingwg/rdma-cni
   sriovDevicePlugin: ghcr.io/k8snetworkplumbingwg/sriov-network-device-plugin
+  sriovDraDriver: ghcr.io/k8snetworkplumbingwg/dra-driver-sriov:latest
```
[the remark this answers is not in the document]
wizhaoredhat: We intentionally keep images.sriovDraDriver at :latest (and the same default in hack/env.sh) so upstream CI and lab runs exercise the current dra-driver-sriov build together with the other operands. Pinning to a digest is expected at distribution layer (e.g. OpenShift image references), not in the community operator chart defaults.

bindata/manifests/dra-driver/003a-role-pod-access.yaml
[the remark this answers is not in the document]
wizhaoredhat: We’re keeping namespace-scoped `pods` patch on the DRA driver SA for this PR so the wait-for-config init can set its annotation, matching the existing device-plugin RBAC and flow in `bindata/manifests/plugins/002-rbac.yaml`.

We understand this does not eliminate token exposure within the namespace. Tightening that is a separate hardening effort (alternate coordination and/or admission that limits patch targets) and should apply consistently to both device-plugin and DRA driver rather than only changing DRA here.

bindata/manifests/dra-driver/003-clusterrole.yaml
```diff
@@ -0,0 +1,84 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
danwinship, line 3: So this appears to match the ClusterRole helm chart from `dra-driver-sriov`, but is it really correct? Does it actually need all of the write access that it requests? Does it really need to write to (non-status fields of) resourceclaims, pods, and nodes? Does it really need to be able to delete resourceclaims?

EDIT: ah, you change some of this in the next commit...

bindata/manifests/dra-driver/004-daemonset.yaml
```diff
@@ -0,0 +1,151 @@
...
+        - name: CONFIGURATION_MODE
+          value: "MULTUS"
```
danwinship, line 106: It would be more normal to configure all of this stuff via command-line args...

controllers/helper.go
```diff
@@ -62,12 +65,23 @@ var (
 )
 
 const (
+	// DRA bindata (constants.DRADriverPath) — keep resource names in sync with manifests.
```
danwinship: `consts.DRADriverPath`, right? Though also, it's not obvious why you're referencing that here

controllers/helper.go
```diff
@@ -172,6 +196,17 @@ func GetNodeSelectorForDevicePlugin(dc *sriovnetworkv1.SriovOperatorConfig) map[
 	return tmp.ConfigDaemonNodeSelector
 }
 
+// GetNodeSelectorForDRADriver returns the same node selector as the config daemon (workers).
+// Unlike the device plugin, the DRA driver does not use the device-plugin=Enabled label
+// because that label is only set when in device-plugin mode; in DRA mode we need the
+// driver to run on all config-daemon nodes.
+func GetNodeSelectorForDRADriver(dc *sriovnetworkv1.SriovOperatorConfig) map[string]string {
```
danwinship, line 203: This comment won't age well; it assumes that the reader already understands device-plugin mode but needs to have DRA mode explained to them. (ie, it assumes that the reader is an sriov-network-operator maintainer from around September 2026.)

controllers/helper.go
```diff
@@ -211,6 +246,145 @@ func syncPluginDaemonObjs(ctx context.Context,
...
+			return err
+		}
+	}
+
+	logger.Info("Cleaned up DRA driver DaemonSet, RBAC, ServiceAccount, and base DeviceClass")
+	return nil
+}
+
+func deleteIfNotFound(ctx context.Context, c k8sclient.Client, obj k8sclient.Object) error {
+	if err := c.Delete(ctx, obj); err != nil && !errors.IsNotFound(err) {
```
danwinship: more like `deleteUnlessNotFound`?

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -131,9 +138,30 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(ctx context.Context, req ct
 	if err = r.syncAllSriovNetworkNodeStates(ctx, defaultOpConf, policyList, nodeList); err != nil {
 		return reconcile.Result{}, err
 	}
-	// Sync Sriov device plugin ConfigMap object
-	if err = r.syncDevicePluginConfigMap(ctx, defaultOpConf, policyList, nodeList); err != nil {
-		return reconcile.Result{}, err
+
+	// Sync either device plugin ConfigMap or DRA resources (DeviceAttributes + SriovResourcePolicy) based on feature gate
+	if r.FeatureGate.IsEnabled(constants.DynamicResourceAllocationFeatureGate) {
```
danwinship, line 145: In standard k8s/OCP usage of the term "feature gates", this would be incorrect; feature gates don't directly enable new features, they just make the features be _available_ to be enabled/configured.

It seems like sriov-network-operator may be using the term differently, but `api/v1/sriovoperatorconfig_types.go` still describes `FeatureGates` as being `to enable experimental features`, and this won't always be an experimental feature, so it seems like it ought to be an ordinary config option, not a feature gate.

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -704,3 +732,512 @@ func updateDevicePluginResource(
...
+			continue
+		}
+		if p.Spec.ResourceName != "" {
+			desiredResourceNames[p.Spec.ResourceName] = struct{}{}
+		}
+	}
+	gvk := schema.GroupVersionKind{Group: "resource.k8s.io", Version: "v1", Kind: "DeviceClassList"}
+	list := &unstructured.UnstructuredList{}
+	list.SetGroupVersionKind(gvk)
+	if err := r.List(ctx, list, client.MatchingLabels{deviceClassGeneratedByLabel: deviceClassOperatorLabelVal}); err != nil {
```
danwinship: Why is this using unstructured rather than using the proper types directly? (I suppose the answer might be "because we're importing a too-old version of `k8s.io/api` and `k8s.io/client-go`, but in that case, fix that?)

controllers/sriovoperatorconfig_controller_test.go
```diff
@@ -155,7 +155,7 @@ var _ = Describe("SriovOperatorConfig controller", Ordered, func() {
 					FeatureGates:          map[string]bool{},
 				}
 				err = k8sClient.Update(ctx, config)
-				Expect(err).NotTo(HaveOccurred())
+				g.Expect(err).NotTo(HaveOccurred())
```
danwinship: (unrelated drive-by fix? maybe should be a separate commit?)

controllers/sriovoperatorconfig_controller.go
```diff
@@ -251,25 +251,25 @@ func (r *SriovOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.
 		return reconcile.Result{}, err
 	}
 
-	// Deploy either DRA driver or device plugin based on feature gate
+	// Deploy either DRA driver or device plugin based on feature gate.
+	// Sync the replacement provider before cleaning up the old one so a failed
+	// switch does not leave the cluster without either provider.
```
danwinship, line 265: But doesn't this mean a failed switch will leave it with both providers? Does the operator cope with that?

danwinship (commented):
"One wrong commit followed by 20 fixups" isn't great for reviewability. Either you review the wrong commit first and then have to interpret all of the fixups in terms of that, or else you ignore the individual commits and just review the final diff.

Anyway, I tried to review this a little, but I don't really know this code at all, so...

bindata/manifests/dra-driver/003-clusterrole.yaml
wizhaoredhat: Yes sorry about the 20+ commits. I wanted to keep the original author's commit unchanged and append changes on top.

I reviewed it again, the accesses should be correct for dra-driver-sriov.

controllers/sriovnetworknodepolicy_controller.go
wizhaoredhat: Right now we plan for this feature to be behind a feature gate. The proper place is to make device plugin and DRA run side by side. However this will require a big change and the change to the SriovNetworkNodePolicy CRD. We aren't ready for this and it is low priority.

@SchSeba It maybe a conversation we want to have now about the Migration Path.

controllers/sriovoperatorconfig_controller.go
wizhaoredhat: For this particular case, the expectation is that somehow the dra-driver-sriov cannot start at all. Hence the Device Plugin will be left around. Having both providers being operational would be a problem.

Do you think we should just remove the Device Plugin / DRA blindly and then add the vice versa also blindly?

@SchSeba / @rollandf Your input here would be great!

bindata/manifests/dra-driver/004-daemonset.yaml
wizhaoredhat: IMHO, Env is better if we want to make these configuration modifiable in the feature. Mostly on the basis that it is more readable when we template it.

controllers/helper.go
wizhaoredhat: Revised the comment a bit better. PTAL

controllers/helper.go
wizhaoredhat: I changed it to be more appropriate.

controllers/sriovnetworknodepolicy_controller.go
wizhaoredhat: Thank you, I fixed it.

wizhaoredhat:
@coderabbitai resume

wizhaoredhat:
@coderabbitai review

wizhaoredhat:
@coderabbitai review

wizhaoredhat:
@coderabbitai full review

wizhaoredhat:
@coderabbitai resume

controllers/sriovnetworknodepolicy_controller.go
wizhaoredhat: Discussed offline.

SchSeba:
@coderabbitai full review

